PR #2, opened 2020-04-03
Repo setup

jbaxleyiii:
Initial basic setup of the repo with testing and release of artifiacts


graphql/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
```
queerviolet, line 1: Maybe let's leave it out till we figure out how we want to pull in the existing graphql-parser? From my limited experimenting with it, it was easiest to treat it as an external crate even if we fork it into this repo

queerviolet:
Ah, one other comment: could we include getting started instructions in `README.md`? Maybe it's just bc I'm new to Rust and could use the help, but I find it odd that every js project includes the same `npm install && npm start` instructions, but in Rustland, where the build process is not necessarily as obvious (to me?), we don't seem to do that?

graphql/Cargo.toml
jbaxleyiii: @queerviolet yep that sounds good to me. I mainly wanted to show that workspaces / multi crates was easy to setup. Happy to remove!

queerviolet: It is good to see, I hadn't known about this feature, actually!

README.md
```diff
@@ -1,2 +1,51 @@
...
+
+```sh
+cargo 1.42.0 (86334295e 2020-01-31)
+```
+
+### Running the project
+To run the project locally, you can use `cargo` to build and exexecute the bin of the `apollo` crate.
+
+```
+cargo run --bin apollo
```
queerviolet, line 26: Let's put `cargo run` right up top, akin to the `npm install && npm run` part of so many JS readmes.

Also, if we get rid of the `graphql` crate—or make it a library crate, I think?—we won't have to specify the `--bin apollo`, we can just do `cargo run`

jbaxleyiii: @queerviolet we can get rid of the `--bin apollo` but I suspect we will want another crate pretty quickly into this build out. I'll add a section about running the CLI as well. I suspect there is a way to make only one of the packages a "bin" command to ease this as well

jbaxleyiii: @queerviolet yep! That worked great :)

jbaxleyiii:
@queerviolet @JakeDawkins are you okay with us merging this?

JakeDawkins:
@jbaxleyiii Yep! This looks like a fantastic start :)

PR #3, opened 2020-04-06
Import graphql-parser and add a print command.

queerviolet:
Import graphql-rust's [graphql-parser](https://github.com/graphql-rust/graphql-parser) as a subtree, add it as a dependent crate, and add an `apollo print schema.gql` command which uses the parser.

## subtrees

This PR pulls in all of [graphql-parser](https://github.com/graphql-rust/graphql-parser) using [git subtree](https://www.atlassian.com/git/tutorials/git-subtree).

Subtrees aren't submodules. As you can see, the files are actually in this repo, along with all their history. Here's how I brought them in:

```sh
git remote add graphql-parser git@github.com:graphql-rust/graphql-parser
git subtree add --prefix graphql-parser graphql-parser master
```

This is functionally equivalent to using `git read-tree` to copy over the file with their history. AFAIK, the only thing that's special about the `subtree` command is that it adds markers to the commit message (such as you can see in b6ae66d) which help it pull, push, and merge the subtree in the future 

In the future, to pull in updates from upstream, we can do this:

```sh
git subtree pull --prefix graphql-parser graphql-parser master
```

And similarly, we can `git subtree push` to contribute a branch upstream.


jbaxleyiii:
@queerviolet should we try bringing in the external package as a git submodule or something so we could open upstream changes directly from this repo?

queerviolet:
Yeah, so we can, and the last time I did this, that's what I did. The trouble with that approach is that it means that some commits span multiple repositories, which is a pain. Like if we add a parser feature and add a flag to switch it on, that's two PRs, two branches, updating the submodule, doing it in the right order so the tests don't fail…

Maybe there's some way to make it less painful? But AFAIK there isn't, and I don't really want to live like that.

What I _do_ need to do is add this code via while preserving the source history, which I'll hack on now.

jbaxleyiii:
@queerviolet ah that is a good point. Okay! Thanks for the background!

queerviolet:
Maybe I should have squashed.

queerviolet:
I changed this branch to pull in `graphql-parser` as a [git subtree](https://www.atlassian.com/git/tutorials/git-subtree). Adding more to the PR description.

queerviolet:
Okay, a question about history.

`git subtree` gives us an option to squash the remote repository history down to a single commit when importing it. I didn't use it, and this is the result—we can see the whole history of `graphql-parser` being imported into this repo.

Do we want that? On the one hand, I think it can be nice to have the whole history, particularly if we're trying to learn how things have evolved in the way that we have. On the other hand, it can be a bit noisy, and it means that there will be unfamiliar names popping up all over inline blame (if you use something like GitLens, which I do). Back on the first hand, we are _using the code written by these folks_ so maybe this is a nice way to get acquainted with them.

Thoughts?

abernix:
Fwiw, I agree with your first hand's approach!

If this repository's history was particularly long, I might consider it more extensively, but I don't see this as being too much of a departure than the more common "fork" pattern that GitHub rallies around.  The history will prove to be valuable, and I'm sure there's some great persona behind those lines of `git blame`.

jbaxleyiii:
@abernix @queerviolet I'm good with what y'all think here.

PR #11, opened 2020-04-16
Some suggestions / experiments for the command structure

queerviolet:
# quick guide to adding a command in this structure

1. Add a struct for it in [`commands/mod.rs`](apollo/src/commands/mod.rs). You can copy-pasta one of the existing commands, like `print` or `login`.
2. Add it to the `Apollo` enum for it in [`commands/mod.rs`](apollo/src/commands/mod.rs).
3. Add a module for it in [`commands/`](apollo/src/commands). If your command is big and needs multiple files, feel free to give it a directory inside commands. The module will `impl Command for <your-command>`
4. Add a case for it to the `impl Command for Apollo`, again in [`commands/mod.rs`](apollo/src/commands/mod.rs).
5. Enjoy!

# pr notes

I started commenting on #10 and then realized I needed to know Rust better to make good suggestions. Anyway, I did some exploring, I now know Rust better, so here are some thoughts / suggestions.

In b2198aa, we're introducing a named struct as the enum variant data for each command. I like this and think we should do it! It means that each subcommand can receive a struct with all its inputs which feels exactly right—like, that seems like the big win of using `structopt` to begin with. It's also a bit more like receiving a graphql input type if we decide to go that route.

In 6c50816, I'm trying out declaring a `Command` trait which each of these structs implements. The command trait has one function, `run`. Somewhat annoyingly, we still need to enumerate all the commands with a `match` in `main.rs`… and unconditionally call `run` on whichever one we match. But we don't have to explicitly `use` each implementation—the trait takes care of that for us. Also, I strongly suspect there's a way to write a macro that does this for us. Also-also, maybe there's a better way of unpacking enum variant data when every case satisfies some trait. I just don't know what that is.

I think the command trait is kindof nice, but it also feels like maybe using language features for the sake of it? Would love thoughts / input.


queerviolet:
Update: the command trait is growing on me

PR #12, opened 2020-04-19
Setup test suite for shell installer

jbaxleyiii:
The `install.sh` command is pretty critical for the sucessful distribution of the CLI so we really should ensure that it is as well tested as we can. This PR setups the `bats` testing framework to test the `install.sh` command


cdn/public/cli/install.sh
```diff
@@ -1,59 +1,93 @@
 #!/usr/bin/env bash
```
benjamn, line 1: Since the install script is BASH, should we use `bash` instead of `sh` for the install command?
```
curl -sSL https://install.apollographql.workers.dev/cli | bash
```

PR #14, opened 2020-04-20
don't error / log to sentry on unsupported methods

jbaxleyiii:
This limits the worker to supporting `HEAD` (for caching) and `GET` requests when people use the CLI installer

### DIFF
--- a/cdn/src/handler.ts
+++ b/cdn/src/handler.ts
@@ -19,8 +19,15 @@ type CLIArgs = {
 
 type PossiblePaths = [undefined, Product?, CLIArgs["platform"]?, string?];
 
+const SUPPORTED_METHODS = ["GET", "HEAD"]
 export async function handleRequest(event: FetchEvent): Promise<Response> {
   try {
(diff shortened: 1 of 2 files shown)


cdn/src/handler.ts
```diff
@@ -19,8 +19,15 @@ type CLIArgs = {
 
 type PossiblePaths = [undefined, Product?, CLIArgs["platform"]?, string?];
 
+const SUPPORTED_METHODS = ["GET", "HEAD"]
 export async function handleRequest(event: FetchEvent): Promise<Response> {
   try {
+    if (!SUPPORTED_METHODS.includes(event.request.method)) {
+      return Promise.resolve(
+        new Response(`This proxy only supports the following request types: ${SUPPORTED_METHODS.join(", ")}`, { status: 500 })
+      )
```
abernix, line 28: Is the wrapping in `Promise.resolve` necessary since we're already an `async` function that will return a `Promise`?

PR #18, opened 2020-04-24
literally turn index.html into the install script

queerviolet:
This pr turns index.html into a shell script that installs the cli tooling. That is, it is a shell script, which begins with the shebang, which nevertheless looks pretty in the browser.

When displayed in a browser, the user will see the full code for the install script, along with a stylish informational message. When displayed in a text editor / with `cat`, the message will still be readable, just a bit less stylish.

When viewed in a browser, the code of the install script is styled and highlighted, because why not? We're using highlight.js delivered via cloudflare's js cdn (we may alternately prefer to serve hljs ourselves).

The one caveat here is that browsers will interpret the file in quirks mode, which is (1) a generous description of a shell script massaged into html file form (2) fine.

Screen Shot 2020-04-23 at 7 07 55 PM

one last thing: clicking the command copies it now.


designatednerd:
a) 🤯🤯🤯 this is bonkers and awesome at the same time
b) Any particular reason why there's both an inline stylesheet and a `style.css`?

queerviolet:
@designatednerd oh i think i just screwed up

jbaxleyiii:
Merged this into #19. This idea was just amazing

PR #20, opened 2020-04-27
Return an exit code

jsegaran:
Commands will return an int, that will then be used for the exit code. Also removed the login command since we aren't going to be implementing for the moment.

This is from the hackday project @zionts and I worked on.


jbaxleyiii:
@jsegaran Thank you so much for opening this! We are exploring a more full error story with https://github.com/apollographql/apollo-cli/pull/21 and I'd love your thoughts on it

PR #21, opened 2020-04-28
Move installation of CLI to be in $HOME/.apollo

jbaxleyiii:
This commit builds some core infrastructure of the CLI (errors, layout, logging), cleans up some prototypes, and creates a new `setup` command which adds the path to the CLI bins for the Apollo toolchain (currently `ap` but will grow to include the `apollo-language-server`) to common startup scripts (i.e. `.profile`).

This work builds off on the excellent Volta CLI which I think has some really great practices and structures.

@queerviolet @JakeDawkins this takes some opinions on structure / crates / errors that I really like but am completely open to changing. I've opened this as a draft as I expect quite a lot of back and forth

Needs test btw. Will test once we agree on the structural changes


cli/src/cli.rs
```diff
@@ -0,0 +1,62 @@
...
+    )]
+    pub quiet: bool,
+}
+
+impl Apollo {
+    pub fn run(self) -> Fallible<ExitCode> {
+        if let Some(command) = self.command {
+            command.run()
+        } else {
+            Apollo::from_iter(["ap", "help"].iter()).run()
```
queerviolet, line 40:
```suggestion
            Apollo::from_iter([std::env::args()[0], "help"].iter()).run()
```

jbaxleyiii:
@queerviolet my basic thought has been if it has to do with the filesystem / registry then it belongs in Atlas. The CLI / Language Server / Editors will need common functionality for 1) looking up the user 2) using the atlas cache 3) sharing common error codes across tooling.

Very much happy to change this it isn't what you were thinking about for atlas!

queerviolet:
mmm… in my thinking, the atlas is built around a graphql schema that describes a bunch of graphql schemas. it's like the introspection schema, but it's able to handle namespaces, assignments, and so on.

that schema is the spiritual heart of the atlas module. i'm thinking of the `atlas` module's purpose as basically being:

  1. define the metagraph / enhanced introspection schema with rust bindings
  2. implement resolvers

i don't think the atlas will handle general filesystem operations. or, if it does, we'll first define them in terms of that schema and write resolvers for them. i'd rather identify and move those bits later than put everything there now.

all that said, this all looks like _super great_ cli tooling and i think it should live in the `cli` package.

jbaxleyiii:
Closing in favor of #23, #24, and #25

PR #22, opened 2020-04-29
Add snapshot testing and interface implements interface

queerviolet:
Add snapshot testing with [insta](https://crates.io/crates/insta), and then add parser support for [interfaces implementing interfaces](https://github.com/graphql/graphql-spec/pull/373), new in GraphQL 15.

This also, incidentally, adds a `-s` flag to the `print` command to print the syntax tree.

insta lets us simplify our parser tests somewhat. We check snapshots of the `Result` of `parse_{query,schema}`, which means we don't have to special case failing and succeeding tests, and on the whole the process feels less fragile.


queerviolet:
Ugh, we renamed stuff so the merge is a nightmare. It's also a bit messy to lump these; I'm going to break this into a few PRs

queerviolet:
Snapshot testing is now in #29

PR #23, opened 2020-05-01
Implement CLI wide logging

jbaxleyiii:
This commit allows useage of `info!`, `warn!`, and `debug!` within the CLI command space which will follow the instructions of using `--debug`/`--verbose` or `--quiet`/`--silent`.

### Todo
- [ ] tests
- [x] documentation


cli/src/cli.rs
```diff
@@ -0,0 +1,72 @@
...
+    global_setting = AppSettings::ColoredHelp,
+    global_setting = AppSettings::ColorAuto,
+    global_setting = AppSettings::DeriveDisplayOrder,
+    global_setting = AppSettings::DontCollapseArgsInUsage,
+    global_setting = AppSettings::VersionlessSubcommands,
+)]
+/// The [Experimental] Apollo CLI, for supporting all your graphql needs :)
+pub struct Apollo {
+    #[structopt(subcommand)]
+    pub command: Option<Subcommand>,
```
JakeDawkins, line 21: Oo this is an interesting approach! I was gonna suggest we name this `Command` instead of `Subcommand` but that's already taken 🙃

cli/src/cli.rs
```diff
@@ -0,0 +1,72 @@
...
+                .expect("Called help without a path to the binary");
+
+            Apollo::from_iter([&command_name, "help"].iter()).run();
+        }
+    }
+}
+
+#[derive(StructOpt)]
+pub enum Subcommand {
+    #[structopt(name = "login")]
```
JakeDawkins, line 61: is this not handled by the `kebab-case` setting?

cli/src/cli.rs
queerviolet: We could also rename the trait if we like

cli/src/cli.rs
queerviolet: I think we have to flag this Subcommand struct as kebab as well?

cli/src/commands/print.rs
```diff
@@ -1,33 +0,0 @@
-use crate::commands::Command;
```
queerviolet, line 1: can we not lose this command? we can rename it or put it under a `debug` or `experimental` or something, but i've been using it a lot to look at how the parser is parsing things

PR #25, opened 2020-05-01
Move installation to $HOME/.apollo/bin

jbaxleyiii:
This PR changes the install location of the CLI to be $HOME/.apollo/bin and provides a setup command for windows and unix environments to allow for that path to be added to the $PATH of their profile

### DIFF
--- a/cli/src/log.rs
+++ b/cli/src/log.rs
@@ -25,6 +25,7 @@ pub fn init_logger(verbose: bool, quiet: bool, env_log_level: Result<String, Var
         };
         env_logger::builder()
             .filter_level(flag_log_level)
+            .format_level(false)
             // only show timestamps on verbose and trace levels
             // only show module path on verbose and trace levels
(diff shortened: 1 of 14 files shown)


cli/src/log.rs
```diff
@@ -25,6 +25,7 @@ pub fn init_logger(verbose: bool, quiet: bool, env_log_level: Result<String, Var
         };
         env_logger::builder()
             .filter_level(flag_log_level)
+            .format_level(false)
```
JakeDawkins, line 28: Why do we not want to print the level?

jbaxleyiii: The `info` outputs get a `[Info]` bracket even though we want to use it as a normal log where the output shouldn't go to stdout

jbaxleyiii: I'm going to write a custom formatter for this to better match what we want

PR #26, opened 2020-05-05
Atlas design and example structure

queerviolet:
Design and project structures exemplifying cross-schema name resolution, atlas file structures, and documentation of implementation details (all v. in progress)


queerviolet:
this pr is failing tests which tells me that maybe the tests are a little too finicky because it has no code? @jbaxleyiii

jbaxleyiii:
@queerviolet the CDN tests are broken on master (but fixed on #21) and since we can't get coverage reliably I'm going to turn it off 😭

atlas/atlas.star
```diff
@@ -0,0 +1,175 @@
+"""
+A URL is an absolute location within the atlas.
+"""
+scalar URL
```
jbaxleyiii, line 4: @queerviolet do you have any specification on this URL?

atlas/atlas.star
```diff
@@ -0,0 +1,175 @@
...
+and `String`.
+"""
+scalar Name
+
+type Query {
+  at(id: URL): Point
+  edge(name: Name, from: Name, to: Name): Edge
+  type(id: URL): Type
+  interface(id: URL): Interface
+  edge(id: URL): Edge
```
jbaxleyiii, line 17: Since we have a `Point` interface, instead can we have a `point(id: URL)` and `points(ids: [URL!]!)` here instead of one off for each type of point in the graph?

atlas/atlas.star
```diff
@@ -0,0 +1,175 @@
...
+"""
+scalar Name
+
+type Query {
+  at(id: URL): Point
+  edge(name: Name, from: Name, to: Name): Edge
+  type(id: URL): Type
+  interface(id: URL): Interface
+  edge(id: URL): Edge    
+  points: [Point]
```
jbaxleyiii, line 18: We should probably think about pagination early on in this design as it always bites us trying to add it later. Especially as we deal with massive graphs

atlas/atlas.star
```diff
@@ -0,0 +1,175 @@
...
+}
+
+## This would be cool:
+# interface Point @key(fields: "id", via: "Query.at(id)") {
```
jbaxleyiii, line 22: What does the `via` do here? Is that meant to represent its canonical fetch point in the graph? How does this differ from the current `_entities` entry point design?

atlas/atlas.star
```diff
@@ -0,0 +1,175 @@
...
+  id: URL
+  name: Name
+}
+
+type Enum implements Point @key(fields: "id", via: "Query.at(id)") {
+  id: URL
+  name: Name
+}
+
+type List @key(fields: "of") {
```
jbaxleyiii, line 58: I find this name to be confusing / to abstract? What is this a list of?

atlas/atlas.star
```diff
@@ -0,0 +1,175 @@
...
+  of: Point
+}
+
+type Must @key(fields: "of") {
```
jbaxleyiii, line 62: Same comment with the List type, what does this do?

atlas/atlas.star
```diff
@@ -0,0 +1,175 @@
...
+  SCALAR
+  OBJECT
+  FIELD_DEFINITION
+  ARGUMENT_DEFINITION
+  INTERFACE
+  UNION
+  ENUM
+  ENUM_VALUE
+  INPUT_OBJECT
+  INPUT_FIELD_DEFINITION @abc @1 @2(x: y) @2
```
jbaxleyiii, line 170: What are these directives?

atlas/atlas.star
```diff
@@ -0,0 +1,175 @@
...
+}
+
+interface Foo implements Bar  & Baz & X @0 {
```
jbaxleyiii, line 173: Should this be here?

atlas/fixtures/basic/.atlas/atlas.toml
```diff
@@ -0,0 +1,15 @@
+# should atlas lookup proceed up the directory structure?
+#   default: false
+child = false
```
jbaxleyiii, line 3: child seems to arbitrary in this context. Can we make it something like `recurse_ancestor_atlases = false`

atlas/fixtures/basic/@.gql
```diff
@@ -0,0 +1,9 @@
+schema @id(name: "lyrical") {
```
jbaxleyiii, line 1: I'm confused by the naming of this file as `@`?

atlas/fixtures/basic/@.gql
```diff
@@ -0,0 +1,9 @@
...
+  query: Query
+  mutation: Mutation
+}
+
+type Query {
+  findUser(byId: ID): User @from(field: "users.Query.findUser")
```
jbaxleyiii, line 7: Is this meant to be a composite example? of users + admin

atlas/fixtures/basic/@.gql
```diff
@@ -0,0 +1,9 @@
...
+  me: User @from(field: "auth.Query.me")
```
jbaxleyiii, line 8: with the other file being named `admin`, how does atlas know to call this `auth`? Does each service need to name its graph with an id?

atlas/readme.md
```diff
@@ -0,0 +1,325 @@
...
+- schemas
+
+### `atlas:` urls uniquely address points
+
+Technically atlas URLs are URIs or maybe URNs but [WHATWG basically thinks the distinction is unnecessary and confusing](https://url.spec.whatwg.org/#goals) so we'll just follow their lead and call them URLs.
+
+They take this form:
+
+```
+atlas:specifier:path?layer
```
jbaxleyiii, line 102: Down below you use `::` as a separator but here only `:`? Is that intentional?

jbaxleyiii: What is `atlas` here?  Is it literally the string `atlas` or is it a reference to something?

atlas/readme.md
```diff
@@ -0,0 +1,325 @@
...
+
+Every part except the initial `atlas:` is optional.
+
+Some examples:
+
+```
+# these are all in the main atlas
+
+## a schema
+atlas::spotify
```
jbaxleyiii, line 113: How do schema variants come into play here? Are they defined or referenced elsewhere?

atlas/readme.md
```diff
@@ -0,0 +1,325 @@
...
+
+## a point within the schema
+atlas::spotify.User
```
jbaxleyiii, line 116: I've grown really fond of rust's import syntax and am wondering how close we can keep to that:

`atlas::spotify::{User,Node}`

atlas/readme.md
```diff
@@ -0,0 +1,325 @@
...
+
+## a field
+### note: still thinking about the syntax here
+atlas::spotify.id/spotify.User=spotify.ID
```
jbaxleyiii, line 120: What is this trying to do?

atlas/readme.md
```diff
@@ -0,0 +1,325 @@
...
+
+## identified by atlas id
+atlas:[3deff29adabade98f3234acb9320]:svc.secrets/User=String
+```
+
+#### Atlas URLs start off by specifying the atlas
+
+Atlases can be specified either by *id* or *alias*. IDs are specified in `[brackets]`, aliases are bare words:
+
+Specified by id:
```
jbaxleyiii, line 140: What is the use case for specifying by id? That isn't something we would expect people to author?

atlas/readme.md
```diff
@@ -0,0 +1,325 @@
...
+atlas::genius.User
+atlas::nyt.Article
+```
+
+#### Edges are located within their declaring schema
+
+All `id` fields declared on any type *by* the `genius` schema:
+
+```
+atlas::genius.id
```
jbaxleyiii, line 185: what does this do? Since `id` could have any arbitrary number of return types I'm not clear what this does?

atlas/readme.md
```diff
@@ -0,0 +1,325 @@
...
+```
+
+A field declared in `genius` named `id` on the type `genius.User`:
+
+```
+atlas::genius.id/genius.User
```
jbaxleyiii, line 191: Why wouldn't you write this `atlas::genius::User::id`?

atlas/readme.md
```diff
@@ -0,0 +1,325 @@
...
+- inputs
+- interfaces
+- fields
+- unions
+- enums
+- scalars
+- directives
+- schemas
+
+### `atlas:` urls uniquely address points
```
jbaxleyiii, line 95: Who writes these urls and when?

atlas/readme.md
```diff
@@ -0,0 +1,325 @@
...
+Layers are described in more detail in the [next section](#atlas-layers).
+
+### Layers stack to form the atlas
+
+When accessed from some particular client, atlas data is likely strewn across various places. We call these *layers*.
+
+For example, when compiling types against the public atlas, the layers might include:
+  - schema files in the workspace
+  - the user's global cache in ~/.atlas
+  - the public atlas in heaven
```
jbaxleyiii, line 224: Probably worth clarifying this line ;)

atlas/readme.md
```diff
@@ -0,0 +1,325 @@
...
+
+In another instance, we might be using the atlas within a language server. The LSP provides access to unsaved content in the user's editor, which we can represent as another layer within the atlas.
```
jbaxleyiii, line 226: oh hell yes

atlas/readme.md
```diff
@@ -0,0 +1,325 @@
...
+```
+
+In either case, to complete the type `User`, the compiler needs to resolve `users.User`. This happens in two steps:
+
+1. first, we look up the first part of the dotted path—`user`, by scanning up the [scope chain](#scopes-mirror-the-file-tree) until a scope is found in which the name `user` has been defined.
+2. then, we descend into that definition, looking for a `User` element.
+
+### Scopes mirror the file tree
+
+The scope chain consists of:
```
jbaxleyiii, line 267: Do you have a preference on if tooling defaults to local install (i.e. `./.atlas/cache`) or global (`~/.atlas/cache`)?

atlas/readme.md
```diff
@@ -0,0 +1,325 @@
...
+    - all names introduced with `let`
+3. For each ancestor directory up to the [atlas root](#atlas-root) or filesystem root, the namespace of that directory, containing:
+    - the base names (extension stripped) of all files with recognized schema extensions: `.star`, `.gql`, and `.graphql`. It is an error for two sibling schemas to share a base name (i.e. you cannot have `users.star` and `users.gql` as siblings in the same directory); the compiler will warn about this situation and refuse to resolve the name entirely.
+    - the names of any directories containing, in themselves or their 
+4. If the root is a child, the scope chain continues up the directory tree to the next atlas root.
+    
+    
+### atlas root
+
+A `.atlas` directory marks the root of a constellation analagously to how a `.git` directory marks the root of a git repository.
```
jbaxleyiii, line 283: What is the difference between constellation and graph here?

atlas/readme.md
```diff
@@ -0,0 +1,325 @@
...
+
+TK, but generally we use cacache holding the current state for each atlas under its proper uri as a key. The state is a lemma which references others, also held in the cache under their own hash.
+
+## Distributed data format
+
+We can represent an atlas as a collection of signed structs. These are sketched in [data.star](./data.star).
+
+The first such struct an *axiom*. It's nothing more than a message describing the atlas and signed by some identity. The hash of the entire axiom becomes the atlas' id.
+
+From there, we can change the atlas by creating *lemmas*. A lemma is a signed message describing some change to the atlas. For example, lemmas  can `Grant` other identities write access to namespaces within the atlas, or `Revoke` the same. We can also `Announce` endpoints for schemas, create lemmas for directive metadata, and so on.
```
jbaxleyiii, line 323: I think we need to understand better the signing process here. Can you explain the why and how of that?

atlas/atlas.star
```diff
@@ -0,0 +1,175 @@
+"""
+A URL is an absolute location within the atlas.
+"""
+scalar URL
+
+"""
+Names are dotted paths of GraphQL identifiers, like `spotify.User`
```
AlexanderMann, line 7: Is there a length constraint on this? ie, "we only allow 4 levels of nesting/paths"?

Additionally, when you say "identifier" is that in reference to [`ID`](https://spec.graphql.org/draft/#sec-ID) or maybe [`Name`](https://spec.graphql.org/draft/#Name)? (`Name` would make more sense to me since that would limit the characters possible)

It seems like we'll

AlexanderMann: (also, I think a _maybe_ better name here is `Path` so as to not overshadow the graphql spec but I'm not wed to that opinion in the least)

dariuszkuc:
Closing old stale PR which is no longer applicable to the new codebase.

PR #29, opened 2020-05-11
Add snapshot testing for graphql-parser

queerviolet:
Change the graphql-parser tests to use snapshot tests with [insta](https://crates.io/crates/insta).

Previously, all the tests were round trip tests—we'd parse a query or schema, then print it, then ensure that the input and output are identical. That's… probably fine in almost all cases, but does open the possibility of subtle bugs in both the parser and printer conspiring to make a broken system pass the tests anyway.

In the new tests, we're still testing the round trip, but we're also verifying a snapsnot of the `Result` of `parse_schema`/`parse_query`. Verifying the `Result` is nice because it also allows us to eliminate the special-case tests for failure—failure tests are just snapshots of errors. This simplification pleases me.


queerviolet:
goddamnit, windows.

queerviolet:
let me guess that we need to do the `\r\n` to `\n` conversion before parsing so the positions don't get jacked up

graphql-parser/tests/schemas.rs
```diff
@@ -0,0 +1,69 @@
+// extern crate graphql_parser;
```
jbaxleyiii, line 1: Should we just delete this file?

graphql-parser/tests/test.rs
```diff
@@ -0,0 +1,129 @@
+extern crate graphql_parser;
+#[cfg(test)] #[macro_use] extern crate pretty_assertions;
+use insta::assert_debug_snapshot;
```
jbaxleyiii, line 3: This is so great

jbaxleyiii:
@queerviolet I think you missed deleting a file but after that and windows is working, this looks great to me!

queerviolet:
all tests passing 🎉 PTAL 

One thing in particular I'd like to draw our attention to: we're currently running all samples in the corpus against both parsers. That is, we don't distinguish between "schema" samples and "query" samples anymore. We just generate snapshot tests with both `parse_query` and `parse_schema` for everything.

I'm doing this because it's easier, but also because we're about to introduce the a new, joint grammar, which can parse both queries and schemas. That grammar should be able to parse all the files.

Currently, about half the snapshots are of errors. I actually don't think that's such a bad thing. Since we're planning on eventually producing a joint grammar, I think we do actually want to notice if e.g. something which wasn't parsing as a query now does parse as a query, even if it's "only" part of the schema grammar. That is there's a conceptual interdependence there that's a bit subtle which I'd like to capture.

Also, there's a tendency to have lots of succeeding tests and not so many failing ones; this somewhat neatly avoids that.

Open to thoughts, but it would also be a pain to go back and separate them and work that distinction into the test macros, so keep that in mind ;)

queerviolet:
interpreting ❤️ as 👍

PR #31, opened 2020-05-12
Improved output of installer

jbaxleyiii:
At some point I'll stop messing with this installer buttttttttt...

installer_one
installer_with_prompt
installer_with_prompt_and_force

### DIFF
--- a/cdn/public/install.sh
+++ b/cdn/public/install.sh
@@ -2,31 +2,143 @@
 # <body><div class="f"><pre id="flip"><code class="bash">
 # Welcome to the Apollo CLI Install script!
 
+# Options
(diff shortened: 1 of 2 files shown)


cdn/public/install.sh
```diff
@@ -2,63 +2,148 @@
...
-  fi
+prepare_installation() {
+  local sudo_command
+  local msg
+  local install="$(command -v ap)"
 
-  return
+  
+  if [ -n "$install" ]; then
+    warn "An existing version of 'ap' is already installed at $install.\n"
```
JakeDawkins, line 116: This is lovely ❤️

PR #32, opened 2020-05-12
setup error hooks for unexpected panics

jbaxleyiii:
One of the biggest mistakes of the last CLI was not making it very clear when there was a user error vs a CLI error (panic/throw). The new CLI already has a special output for errors but this PR introduces a panic handler using [human panic](https://crates.io/crates/human-panic)

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -2,6 +2,7 @@
 name = "apollo-cli"
 version = "0.0.3"
 authors = ["Apollo <opensource@apollographql.com>"]
+homepage = "https://github.com/apollographql/apollo-cli"
 edition = "2018"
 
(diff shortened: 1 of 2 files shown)


JakeDawkins:
`We take privacy seriously, and do not perform any automated error collection.` Is part of the output for this. Should we actually say that we don't do this? In case we want to add telemetry or error reporting in the future?

jbaxleyiii:
@JakeDawkins its part of the output from the package that we can't change! But also I don't think we will want to collect errors. If we do end up changing that we will need to revist this

JakeDawkins:
@jbaxleyiii Yeah, I saw that we can't do anything about it. I'm fine with leaving it for now. I assume we'll be looking at a pretty rapid release cycle, so if we want to revise this later, I'm fine with that.

PR #33, opened 2020-05-12
inital scaffolding of nested subcommands for discussion

jbaxleyiii:
@queerviolet @JakeDawkins I'd love your thoughts on the setup to build out nested commands (i.e: `apollo schema push`)

### DIFF
--- a/cli/src/cli.rs
+++ b/cli/src/cli.rs
@@ -44,7 +44,15 @@ impl Apollo {
         if let Some(command) = self.command {
             command.run()
         } else {
-            run_command(&mut vec!["help".to_string()])
+            // per the docs on std::env::arg
+            // The first element is traditionally the path of the executable, but it can be set to
+            // arbitrary text, and may not even exist. This means this property should not be
+            // relied upon for security purposes.
+            let command_name = args()
+                .next()
+                .expect("Called help without a path to the binary");
+
+            Apollo::from_iter([&command_name, "help"].iter()).run()
         }
     }
 }
@@ -57,10 +65,7 @@ pub enum Subcommand {
 
     #[structopt(name = "schema")]
     ///  ⚛   manage your schema
-    Schema {
-        #[structopt(subcommand)]
-        command: Option<commands::schema::SchemaCommand>,
-    },
+    Schema(commands::Schema),
 
     #[structopt(name = "print", setting = AppSettings::Hidden)]
     ///  🖨   parse and pretty print schemas to stdout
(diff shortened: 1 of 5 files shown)


cli/src/cli.rs
```diff
@@ -76,6 +76,24 @@ impl Subcommand {
             Subcommand::Login(login) => login.run(),
             Subcommand::Print(print) => print.run(),
             Subcommand::Setup(setup) => setup.run(),
+            Subcommand::Schema { command } => match command {
```
queerviolet, line 79: it seems kindof unfriendly to expect every subcommand implementer to do this?

cli/src/cli.rs
```diff
@@ -62,6 +54,14 @@ pub enum Subcommand {
     #[structopt(name = "login")]
     ///  🔓  log in to apollo
     Login(commands::Login),
+
+    #[structopt(name = "schema")]
+    ///  ⚛   manage your schema
+    Schema {
+        #[structopt(subcommand)]
+        command: Option<commands::schema::SchemaCommand>,
```
queerviolet, line 62: what happens if this isn't optional?

queerviolet:
Something feels a little off about us having to manually invoke help. What happens if we don't make the command an option. Does that just give us the behavior we want (i.e. if it's not there, structopt prints help?)

queerviolet:
Particularly as we look at adding nested subcommands, it seems like it would be nice not to have to replicate that check and re-invoke pattern at every subcommand and every level of the tree.

jbaxleyiii:
@queerviolet yeah I agree with your comments which is where my general 😒 about this comes from. I'll try some other ideas but maybe we can pair on Friday on it?

queerviolet:
Definitely down to pair!

cli/src/cli.rs
```diff
@@ -76,6 +81,10 @@ impl Subcommand {
             Subcommand::Login(login) => login.run(),
             Subcommand::Print(print) => print.run(),
             Subcommand::Setup(setup) => setup.run(),
+            Subcommand::Schema(command) => match command {
```
queerviolet, line 84: maybe move this into `schema/mod.rs`?

cli/src/commands/schema/push.rs
```diff
@@ -0,0 +1,13 @@
+use structopt::StructOpt;
+
+use crate::commands::Command;
+use crate::errors::{ExitCode, Fallible};
+
+#[derive(StructOpt)]
+pub struct Push {}
+
+impl Command for Push {
```
queerviolet, line 9: is the `Command` trait actually doing anything at this point? i introduced it so we wouldn't have to directly reference the implementation files from the command schema, but we're doing that now, so we may as well get rid of the trait

jbaxleyiii:
I think this is a good enough direction for now. Going to close this but when we start building out subcommands its a good template to use

PR #35, opened 2020-05-12
Embed queries and fragments within the schema grammar.

queerviolet:
Enable dropping named queries and fragments into the SDL. The typescript parser allows this, and even if it didn't, we need to be able to point tooling at a graphql file and parse whatever's inside of it to e.g. do codegen on stored queries.

The parser here is a bit more permissive than I think we'll want the actual language to be. For instance, it's okay with you putting a bare, unnamed `query` into a schema file:

```graphql
query { some { stuff } }
```

I don't think we actually want that—I don't know what it means, really, or what we'd want to do with it (suggestions welcome tho!) That said, I think it's probably better that the parser doesn't barf on it, and instead that we treat it as a semantic error at some higher level of atlas or forge.

### DIFF
--- a/graphql-parser/src/query/ast.rs
+++ b/graphql-parser/src/query/ast.rs
@@ -42,6 +42,7 @@ pub enum Definition<'a, T: Text<'a>> {
 #[derive(Debug, Clone, PartialEq)]
 pub struct FragmentDefinition<'a, T: Text<'a>> {
     pub position: Pos,
+    pub description: Option<String>,
     pub name: T::Value,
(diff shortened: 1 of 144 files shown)


queerviolet:
"144 files changed" and most of them are just adding `description: None`. This is basically The Case Against Snapshot Testing, but idk. I'd rather have more robust tests that are a little noisy on PRs that update the AST / grammar. We're doing it a lot now, but that should taper off relatively quickly.

queerviolet:
I want to do a bit more refactoring around the AST but will do that in another PR

PR #40, opened 2020-05-19
Add IntelliJ related configuration files

Enrico2:
These files were auto-created by Intellij when following the instructions on in https://intellij-rust.github.io/

Seems to set up things pretty well.

Some global settings I had to tweak to make the experience better:
1. Set `Preferences -> Rust -> Expand declarative macros` to `Expand with experimental engine`
image
2. check `Preferences -> Rust -> Cargo -> Run external linter to analyze code on the fly`
image

This is a good article explaining which IntelliJ files to put in VCS: https://intellij-support.jetbrains.com/hc/en-us/articles/206544839-How-to-manage-projects-under-Version-Control-Systems

Note that we're not a maven/gradle based project, and so we **do** add the .iml files, because they're not automatically created.

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,0 +1,8 @@
+# Default ignored files
+/shelf/
+/workspace.xml
+# Datasource local storage ignored files
+/dataSources/
+/dataSources.local.xml
+# Editor-based HTTP Client requests
+/httpRequests/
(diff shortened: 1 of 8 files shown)


Enrico2:
Heads up, this set up wasn't quite right. I'm working on fixing it, but it may be conflicting with a bug in the Rust plugin for IntelliJ.

PR #41, opened 2020-05-20
Slightly flatten the AST

queerviolet:
Flatten some of the most egregious unnecessary abstractions within the AST. Specifically:

* flatten `enum OperationDefinition { Query(...), Mutation(...), Subscription(...) }` into `struct OperationDefinition { kind: Operation }`
* remove the `Number` abstraction. Spec says they should be `i32`, we're using `i64`. "to be safe"? But maybe we should actually use the specified 32 bits?
* remove the `TypeCondition` `enum` which had only one variant(!!)


graphql-parser/src/query/format.rs
```diff
@@ -249,17 +250,17 @@ impl<'a> Displayable for InlineFragment<'a>
...
+// {
+//     fn display(&self, f: &mut Formatter) {
+//         match *self {
+//             TypeCondition::On(ref name) => {
+//                 f.write("on ");
+//                 f.write(name.as_ref());
+//             }
+//         }
+//     }
+// }
```
jbaxleyiii, line 263: Do you want to delete these bits all together?

queerviolet: Yes, ty!

PR #44, opened 2020-05-26
Simplify error construction and usage

jbaxleyiii:
This PR cleans up the custom ApolloError to be a simple wrapper
around the standard error. It adds a `From` impl as well as builds
out a context for results to make it easier to cast into an ApolloError

I also removed the login comman noop and added a CTA for bug reports 🎉


cli/src/errors/error.rs
```diff
@@ -1,49 +1,92 @@
...
     }
 }
 
-/// The failure trait for all Apollo errors.
-pub trait ApolloFail: std::error::Error {
-    /// Returns the process exit code that should be returned if the process exits with this error.
-    fn exit_code(&self) -> ExitCode;
+impl fmt::Display for ApolloError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.inner.kind.fmt(f)
```
Enrico2, line 48: I [opened a bug](https://github.com/intellij-rust/intellij-rust/issues/5457) for this marking issue with the IJ Rust plugin, but if you wanna make my IntelliJ happy, this is not shown as red:
```rust
(&self.inner.kind).fmt(f)
```

image

vs.

image

cli/src/errors/error.rs
```diff
@@ -1,49 +1,92 @@
...
+    }
 }
 
-impl<T: ApolloFail> From<T> for ApolloError {
-    fn from(error: T) -> Self {
-        let exit_code = error.exit_code();
+impl Error for ApolloError {
+    fn source(&self) -> Option<&(dyn Error + 'static)> {
```
Enrico2, line 53: Q: What's the purpose of using the `'static` lifetime for the response?

cli/src/errors/error.rs
```diff
@@ -1,49 +1,92 @@
...
+    fn with_context<F>(self, f: F) -> Fallible<T>
+    where
+        F: FnOnce() -> ErrorDetails;
+}
+
+impl<T, E> Context<T> for Result<T, E>
+where
+    E: Error + 'static,
+{
+    fn with_context<F>(self, f: F) -> Fallible<T>
```
Enrico2, line 76: Q: When does one use `self` vs `&self`?

cli/src/errors/error.rs
```diff
@@ -1,49 +1,92 @@
...
         ApolloError {
-            msg: error.to_string(),
-            exit_code,
+            inner: Box::new(Inner { kind, source: None }),
         }
     }
 }
 
+pub trait Context<T> {
+    fn with_context<F>(self, f: F) -> Fallible<T>
```
Enrico2, line 67: I don't see anything calling this, would prefer to avoid dead code, unless there's a concrete use case coming up? Is this used in a later PR?

(I suspect there is no warning because the trait is public.)

cli/src/errors/error.rs
jbaxleyiii: The `'static` (as I understand) supports downcasting the trait object back to a concrete type. [A few](https://rust-lang.github.io/rfcs/2504-fix-error.html#why-cause---source) [reference](https://users.rust-lang.org/t/what-does-it-mean-to-return-dyn-error-static/37619/12) links.

Tbh I'm not totally sure this is right either!

cli/src/errors/error.rs
jbaxleyiii: Happy to remove this. I mainly took it from https://docs.rs/anyhow/1.0.31/anyhow/trait.Context.html#tymethod.with_context. I've used `with_context` a few times but have always ended up refactoring to not need it or use `map_err`. I *think* this is better since its lazy but again, I'm mainly trying to replicate what others have done 😅

cli/src/errors/error.rs
jbaxleyiii: `self` takes ownership over the object which we want to do since we are changing the underlying error. 
image

cli/src/errors/error.rs
jbaxleyiii: Sure happy to make that change!

cli/src/errors/error.rs
jbaxleyiii: `with_context` means you don't have to handle the source of the error, instead we can cast it for you.

So `some_result_fn().with_context(|| ErrorDetails::CustomError)?` over `some_result_fn().map_error(|e| ErrorDetails::CustomError)?` 

I *think* the latter looses the source information of the error

Enrico2: _Generally_ I prefer adding something the first time it's used, or if I have concrete plans to use it soon. While it might seem like a good idea, it's easy to add it when we need it.

PR #45, opened 2020-05-26
Support updating the CLI from the CLI

jbaxleyiii:
This commit supports a dedicated update command for the CLI. It does this by making a HEAD request to the CDN proxy which should have the latest version information cached. It allows a user to manually call `ap update` which will present them with a version if they don't have the latest. The user can then confirm the update or deny it.

This PR also lands a background task which checks, once a day when used, if the CLI could be updated. If it does, then it prints to stderr update instructions.

I also cleaned up the setup command here to be less agressive on writing to shell configs both in the installer and in the setup command itself.

Depends on #44


jbaxleyiii:
Depends on #44

PR #50, opened 2020-05-28
Gravity/auth setup

AlexanderMann:
# Motivation

Gravity team is looking to add a number of features to the CLI, namely around schema uploading/download and checks. To start all of this, we want to be able to set the appropriate `API_KEY` in config to allow users to issue many authenticated requests.

## Team

- @AlexanderMann 
- @jsegaran 

## Questions

- [x] for testing, do we want to:
  - mock user input
  - decouple file and user input from the actual config modification
  - etc.
- [x] for env var do we want to:
  - overwrite global config when env var present
  - merge config based on env var prefix, ie:
    - AP_CLI__FOO_BAR__BAZ -> foo-bar {baz: ...} being a merged config
  - similar to K8s and GCP use a different config file path based on an env var
- [x] for configs do we want to allow for a global and local config?
  - if yes, does `auth setup` then result in modifying local?

## Functional Testing

- [x] no config present
  - [x] run `cargo run auth setup` to test user input
    - [x] exit code 0, new key shows up in `cat ~/.apollo/config.toml`
  - [x] run `printf '...' | cargo run auth setup` to test stdin input
    - [x] exit code 0, new key shows up in `cat ~/.apollo/config.toml`
- [x] config present
  - [x] run `cargo run auth setup` to test user input
    - [x] When I hit and confirm `y` to overwrite, exit code 0, new key shows up in `cat ~/.apollo/config.toml`
    - [x] When I hit and confirm `n` to cancel overwrite, exit code 0, old key shows up in `cat ~/.apollo/config.toml`
  - [x] run `printf 'y\n...' | cargo run auth setup` to test stdin input
    - [x] exit code 0, new key shows up in `cat ~/.apollo/config.toml`
  - [x] run `printf 'n\n...' | cargo run auth setup` to test stdin input
    - [x] exit code 0, old key shows up in `cat ~/.apollo/config.toml`

## Suggested Musical Pairing

https://soundcloud.com/megan-thee-stallion/savageremix-feat-beyonce

### DIFF
--- a/cli/src/cli.rs
+++ b/cli/src/cli.rs
@@ -75,7 +75,7 @@ impl Subcommand {
             Subcommand::Setup(setup) => setup.run(session),
             Subcommand::Auth(auth) => match auth {
                 commands::Auth::Setup(setup) => setup.run(session),
-            }
+            },
         }
     }
 }
--- a/cli/src/cli.rs
+++ b/cli/src/cli.rs
@@ -62,6 +62,9 @@ pub enum Subcommand {
     #[structopt(name = "setup", setting = AppSettings::Hidden)]
     ///  🚜  setup the Apollo toolchain in your environment
     Setup(commands::Setup),
+    #[structopt(name = "auth")]
+    ///  🔐 Manage authentication
+    Auth(commands::Auth),
 }
 
 impl Subcommand {
@@ -70,6 +73,9 @@ impl Subcommand {
             Subcommand::Update(update) => update.run(session),
             Subcommand::Print(print) => print.run(session),
             Subcommand::Setup(setup) => setup.run(session),
+            Subcommand::Auth(auth) => match auth {
+                commands::Auth::Setup(setup) => setup.run(session),
+            }
         }
     }
 }
--- a/cli/src/commands/auth.rs
+++ b/cli/src/commands/auth.rs
@@ -2,7 +2,7 @@ use crate::commands::Command;
 use crate::config::CliConfig;
 use crate::errors::{ExitCode, Fallible};
 use crate::telemetry::Session;
-use crate::terminal::input;
+use crate::terminal::{confirm, input};
 use log::warn;
 use structopt::StructOpt;
 
@@ -20,6 +20,14 @@ pub struct Setup {}
 impl Command for Setup {
     fn run(&self, _session: &mut Session) -> Fallible<ExitCode> {
         let mut config = CliConfig::load().unwrap();
+
+        if config.api_key.is_some() {
+            warn!("Config auth already setup.");
+
+            if !confirm("Proceed?")? {
+                return Ok(ExitCode::Success);
+            }
+        }
 
         let key = input("Please paste key:")?;
 
--- a/cli/src/commands/auth.rs
+++ b/cli/src/commands/auth.rs
@@ -38,11 +38,10 @@ impl Command for Setup {
         }
 
         debug!("Setting new key...");
-        let mut config = session.config.clone();
-        config.api_key = Some(key);
+        session.config.api_key = Some(key.to_string());
 
         debug!("Saving new key...");
-        CliConfig::save(&session.config_path, &config)?;
+        CliConfig::save(&session.config_path, &session.config)?;
 
         info!("{} Your personal API key was successfuly set!", KEY);
         Ok(ExitCode::Success)
--- a/cli/src/commands/auth.rs
+++ b/cli/src/commands/auth.rs
@@ -3,7 +3,7 @@ use crate::config::CliConfig;
 use crate::errors::{ExitCode, Fallible};
 use crate::style::KEY;
 use crate::telemetry::Session;
-use crate::terminal::{input};
+use crate::terminal::input;
 use console::style;
 use log::{info, warn};
 use structopt::StructOpt;
@@ -22,9 +22,8 @@ pub struct Setup {}
 impl Command for Setup {
     fn run(&self, session: &mut Session) -> Fallible<ExitCode> {
         session.log_command("auth setup");
-        let mut config = CliConfig::load().unwrap();
 
-        if config.api_key.is_some() {
+        if session.config.api_key.is_some() {
             warn!("Authentication already configured.");
         }
 
@@ -37,6 +36,7 @@ impl Command for Setup {
             return Ok(ExitCode::ConfigurationError);
         }
 
+        let mut config = session.config.clone();
         config.api_key = Some(key);
         CliConfig::write(&config).unwrap();
         info!("{} Your personal API key was successfuly set!", KEY);
--- a/cli/src/commands/auth.rs
+++ b/cli/src/commands/auth.rs
@@ -38,7 +38,7 @@ impl Command for Setup {
 
         let mut config = session.config.clone();
         config.api_key = Some(key);
-        CliConfig::write(&config).unwrap();
+        CliConfig::save(&session.config_path, &config).unwrap();
         info!("{} Your personal API key was successfuly set!", KEY);
         Ok(ExitCode::Success)
     }
--- a/cli/src/commands/auth.rs
+++ b/cli/src/commands/auth.rs
@@ -5,13 +5,12 @@ use crate::style::KEY;
 use crate::telemetry::Session;
 use crate::terminal::sensitive;
 use console::style;
-use log::{info, warn};
+use log::{debug, error, info, warn};
 use structopt::StructOpt;
 
 #[derive(StructOpt)]
 #[structopt(rename_all = "kebab-case")]
 pub enum Auth {
-    /// Setup your auth stuff
     /// Link the CLI to your Apollo account using your Personal API Key.
     Setup(Setup),
 }
@@ -23,6 +22,7 @@ impl Command for Setup {
     fn run(&self, session: &mut Session) -> Fallible<ExitCode> {
         session.log_command("auth setup");
 
+        debug!("Checking is config's api_key is already set...");
         if session.config.api_key.is_some() {
             warn!("Authentication already configured.");
         }
@@ -31,14 +31,19 @@ impl Command for Setup {
             style("https://engine.apollographql.com/user-settings").cyan());
         let key = sensitive("User key:")?;
 
+        debug!("Checking user input...");
         if key.is_empty() {
-            warn!("No key was inputed, quitting without changes.");
+            error!("No key was inputed, quitting without changes.");
             return Ok(ExitCode::ConfigurationError);
         }
 
+        debug!("Setting new key...");
         let mut config = session.config.clone();
         config.api_key = Some(key);
-        CliConfig::save(&session.config_path, &config).unwrap();
+
+        debug!("Saving new key...");
+        CliConfig::save(&session.config_path, &config)?;
+
         info!("{} Your personal API key was successfuly set!", KEY);
         Ok(ExitCode::Success)
     }
--- a/cli/src/commands/auth.rs
+++ b/cli/src/commands/auth.rs
@@ -1,0 +1,25 @@
+use crate::commands::Command;
+use crate::errors::{Fallible, ExitCode};
+use crate::telemetry::Session;
+use structopt::StructOpt;
+use crate::config::CliConfig;
+
+#[derive(StructOpt)]
+#[structopt(rename_all = "kebab-case")]
+pub enum Auth {
+    /// Setup your auth stuff
+    Setup(Setup)
+}
+
+#[derive(StructOpt)]
+#[structopt(rename_all = "kebab-case")]
+pub struct Setup {}
+
+impl Command for Setup {
+    fn run(&self, session: &mut Session) -> Fallible<ExitCode> {
+        let mut config = CliConfig::load().unwrap();
+        config.api_key = Some(String::from("test"));
+        CliConfig::write(&config).unwrap();
+        Ok(ExitCode::Success)
+    }
+}
--- a/cli/src/commands/auth.rs
+++ b/cli/src/commands/auth.rs
@@ -1,8 +1,10 @@
 use crate::commands::Command;
 use crate::config::CliConfig;
 use crate::errors::{ExitCode, Fallible};
+use crate::style::KEY;
 use crate::telemetry::Session;
 use crate::terminal::{confirm, input};
+use console::style;
 use log::{info, warn};
 use structopt::StructOpt;
 
@@ -10,37 +12,38 @@ use structopt::StructOpt;
 #[structopt(rename_all = "kebab-case")]
 pub enum Auth {
     /// Setup your auth stuff
-    /// Requires using an User key which can be found here:
-    /// https://engine.apollographql.com/user-settings
+    /// Link the CLI to your Apollo account using your Personal API Key.
     Setup(Setup),
 }
 
 #[derive(StructOpt)]
-#[structopt(rename_all = "kebab-case")]
 pub struct Setup {}
 
 impl Command for Setup {
-    fn run(&self, _session: &mut Session) -> Fallible<ExitCode> {
+    fn run(&self, session: &mut Session) -> Fallible<ExitCode> {
+        session.log_command("auth setup");
         let mut config = CliConfig::load().unwrap();
 
         if config.api_key.is_some() {
-            warn!("Config auth already setup.");
+            warn!("Authentication already configured.");
 
             if !confirm("Proceed?")? {
                 return Ok(ExitCode::Success);
             }
         }
 
-        info!("Please input a User key which can be found here: https://engine.apollographql.com/user-settings");
+        info!("To link your CLI to your Apollo account go to {} and create a new Personal API Key. Once you've done that, copy the key and paste it into the prompt below.",
+            style("https://engine.apollographql.com/user-settings").cyan());
         let key = input("User key:", true)?;
 
         if key.is_empty() {
-            warn!("Did not update the Apollo CLI Config!");
-            return Ok(ExitCode::Success);
+            warn!("No key was inputed, quitting without changes.");
+            return Ok(ExitCode::ConfigurationError);
         }
 
         config.api_key = Some(key);
         CliConfig::write(&config).unwrap();
+        info!("{} Your personal API key was successfuly set!", KEY);
         Ok(ExitCode::Success)
     }
 }
--- a/cli/src/commands/auth.rs
+++ b/cli/src/commands/auth.rs
@@ -29,7 +29,7 @@ impl Command for Setup {
 
         info!("To link your CLI to your Apollo account go to {} and create a new Personal API Key. Once you've done that, copy the key and paste it into the prompt below.",
             style("https://engine.apollographql.com/user-settings").cyan());
-        let key = sensitive("User key:")?;
+        let key = sensitive("Personal API Key:")?;
 
         debug!("Checking user input...");
         if key.is_empty() {
--- a/cli/src/commands/auth.rs
+++ b/cli/src/commands/auth.rs
@@ -3,13 +3,15 @@ use crate::config::CliConfig;
 use crate::errors::{ExitCode, Fallible};
 use crate::telemetry::Session;
 use crate::terminal::{confirm, input};
-use log::warn;
+use log::{info, warn};
 use structopt::StructOpt;
 
 #[derive(StructOpt)]
 #[structopt(rename_all = "kebab-case")]
 pub enum Auth {
     /// Setup your auth stuff
+    /// Requires using an User key which can be found here:
+    /// https://engine.apollographql.com/user-settings
     Setup(Setup),
 }
 
@@ -29,7 +31,8 @@ impl Command for Setup {
             }
         }
 
-        let key = input("Please paste key:")?;
+        info!("Please input a User key which can be found here: https://engine.apollographql.com/user-settings");
+        let key = input("User key:", true)?;
 
         if key.is_empty() {
             warn!("Did not update the Apollo CLI Config!");
--- a/cli/src/commands/auth.rs
+++ b/cli/src/commands/auth.rs
@@ -3,7 +3,7 @@ use crate::config::CliConfig;
 use crate::errors::{ExitCode, Fallible};
 use crate::style::KEY;
 use crate::telemetry::Session;
-use crate::terminal::input;
+use crate::terminal::sensitive;
 use console::style;
 use log::{info, warn};
 use structopt::StructOpt;
@@ -29,7 +29,7 @@ impl Command for Setup {
 
         info!("To link your CLI to your Apollo account go to {} and create a new Personal API Key. Once you've done that, copy the key and paste it into the prompt below.",
             style("https://engine.apollographql.com/user-settings").cyan());
-        let key = input("User key:", true)?;
+        let key = sensitive("User key:")?;
 
         if key.is_empty() {
             warn!("No key was inputed, quitting without changes.");
--- a/cli/src/commands/auth.rs
+++ b/cli/src/commands/auth.rs
@@ -3,7 +3,7 @@ use crate::config::CliConfig;
 use crate::errors::{ExitCode, Fallible};
 use crate::style::KEY;
 use crate::telemetry::Session;
-use crate::terminal::{confirm, input};
+use crate::terminal::{input};
 use console::style;
 use log::{info, warn};
 use structopt::StructOpt;
@@ -26,10 +26,6 @@ impl Command for Setup {
 
         if config.api_key.is_some() {
             warn!("Authentication already configured.");
-
-            if !confirm("Proceed?")? {
-                return Ok(ExitCode::Success);
-            }
         }
 
         info!("To link your CLI to your Apollo account go to {} and create a new Personal API Key. Once you've done that, copy the key and paste it into the prompt below.",
--- a/cli/src/commands/auth.rs
+++ b/cli/src/commands/auth.rs
@@ -1,14 +1,16 @@
 use crate::commands::Command;
-use crate::errors::{Fallible, ExitCode};
-use crate::telemetry::Session;
-use structopt::StructOpt;
 use crate::config::CliConfig;
+use crate::errors::{ExitCode, Fallible};
+use crate::telemetry::Session;
+use crate::terminal::input;
+use log::warn;
+use structopt::StructOpt;
 
 #[derive(StructOpt)]
 #[structopt(rename_all = "kebab-case")]
 pub enum Auth {
     /// Setup your auth stuff
-    Setup(Setup)
+    Setup(Setup),
 }
 
 #[derive(StructOpt)]
@@ -16,9 +18,17 @@ pub enum Auth {
 pub struct Setup {}
 
 impl Command for Setup {
-    fn run(&self, session: &mut Session) -> Fallible<ExitCode> {
+    fn run(&self, _session: &mut Session) -> Fallible<ExitCode> {
         let mut config = CliConfig::load().unwrap();
-        config.api_key = Some(String::from("test"));
+
+        let key = input("Please paste key:")?;
+
+        if key.is_empty() {
+            warn!("Did not update the Apollo CLI Config!");
+            return Ok(ExitCode::Success);
+        }
+
+        config.api_key = Some(key);
         CliConfig::write(&config).unwrap();
         Ok(ExitCode::Success)
     }
--- a/cli/src/config.rs
+++ b/cli/src/config.rs
@@ -1,8 +1,9 @@
 use std::error::Error;
 use std::fs;
 
+use crate::errors::{ApolloError, ErrorDetails, Fallible};
 use config::{Config, Environment};
-use log::{info, debug};
+use log::{debug, info};
 use serde::{Deserialize, Serialize};
 use uuid::Uuid;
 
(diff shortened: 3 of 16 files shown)


cli/src/errors/details.rs
```diff
@@ -48,6 +48,9 @@ Please ensure you have permissions to edit your environment variables."
 
     #[error("Could not install CLI. {}", .msg)]
     CLIInstallError { msg: String },
+
+    #[error("Could not read cli configuration CLI. {}", .msg)]
```
Enrico2, line 52:
```suggestion
    #[error("Could not read CLI configuration. {}", .msg)]
```

cli/src/errors/details.rs
```diff
@@ -48,6 +48,9 @@ Please ensure you have permissions to edit your environment variables."
...
+    CLIConfigError { msg: String },
```
Enrico2, line 53:
```suggestion
    CliConfigError { msg: String },
```

And while you're here if you can rename the one above too (https://meteor.slack.com/archives/C013EMYJ48J/p1590692356033300)

cli/src/main.rs
```diff
@@ -37,7 +37,16 @@ fn main() {
...
+    let session_result = Session::init()
+        .map_err(|e| ApolloError::from(ErrorDetails::CLIConfigError { msg: e.to_string() }));
+
+    if let Err(err) = session_result {
+        report(&err);
+        let code = err.exit_code();
+        code.exit();
+    };
+
+    let mut session = session_result.unwrap();
```
Enrico2, line 49:
```suggestion
    let mut session = Session::init()
        .map_err(|e| ApolloError::from(ErrorDetails::CLIConfigError { msg: e.to_string() }))
        .unwrap_or_else(|err| {
            report(&err);
            err.exit_code().exit();
        });
```

cli/src/cli.rs
```diff
@@ -3,8 +3,8 @@ use structopt::StructOpt;
 
 use crate::commands::{self, Command};
 use crate::errors::{ExitCode, Fallible};
-use crate::version::command_name;
 use crate::telemetry::Session;
+use crate::version::command_name;
```
jbaxleyiii, line 7: @Enrico2 I think this is another example to your point of us setting up linting

cli/src/commands/auth.rs
```diff
@@ -0,0 +1,46 @@
...
+use crate::telemetry::Session;
+use crate::terminal::{confirm, input};
+use log::{info, warn};
+use structopt::StructOpt;
+
+#[derive(StructOpt)]
+#[structopt(rename_all = "kebab-case")]
+pub enum Auth {
+    /// Setup your auth stuff
+    /// Requires using an User key which can be found here:
```
jbaxleyiii, line 13: Since we use `Personal API Key` in the UI, lets change this to:

```rs
/// Link the CLI to your Apollo account using your personal API key.
```

We can leave the url for the actual output since the `///` shows up in the `help` command

cli/src/commands/auth.rs
```diff
@@ -0,0 +1,46 @@
...
+    /// https://engine.apollographql.com/user-settings
+    Setup(Setup),
+}
+
+#[derive(StructOpt)]
+#[structopt(rename_all = "kebab-case")]
```
jbaxleyiii, line 19: I don't think we need this rename_all here since we don't have any subscommands. We many not need the `[structopt]` at all

cli/src/commands/auth.rs
```diff
@@ -0,0 +1,46 @@
...
+pub struct Setup {}
+
+impl Command for Setup {
+    fn run(&self, _session: &mut Session) -> Fallible<ExitCode> {
+        let mut config = CliConfig::load().unwrap();
```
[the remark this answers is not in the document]
jbaxleyiii: ^^ the logging of the session gives us a path for telemetry. So I'd log the event as `auth setup` so we can understand usage information

jbaxleyiii, line 24: I don't think we should `.unwrap()` here and instead we should handle the error with some dedicated output of what went wrong

cli/src/commands/auth.rs
```diff
@@ -0,0 +1,46 @@
...
+
+        if config.api_key.is_some() {
```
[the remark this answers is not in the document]
jbaxleyiii: I wonder if we need this confirmation at all to be honest. I think you could accomplish the same action by printing with `warn!` that there is an existing key in place at the config path before you ask for their new key

cli/src/commands/auth.rs
```diff
@@ -0,0 +1,46 @@
...
+            warn!("Config auth already setup.");
+
+            if !confirm("Proceed?")? {
+                return Ok(ExitCode::Success);
+            }
+        }
+
+        info!("Please input a User key which can be found here: https://engine.apollographql.com/user-settings");
```
jbaxleyiii, line 34: I'd make this more verbose:

```
To link your CLI to your Apollo account go to
  <in another color>https://engine.apollographql.com/user-settings</color>
and create a new key. Once you've done that, copy the key and paste it into the prompt below.

Enter personal API key: 
```

jbaxleyiii: It also may be worth asking @StephenBarlow his thoughts on language here!

cli/src/commands/auth.rs
```diff
@@ -0,0 +1,46 @@
...
+        let key = input("User key:", true)?;
+
+        if key.is_empty() {
+            warn!("Did not update the Apollo CLI Config!");
```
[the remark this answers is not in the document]
jbaxleyiii: I would also favor erroring out here with a non-zero exit code

cli/src/commands/auth.rs
```diff
@@ -0,0 +1,46 @@
...
+            return Ok(ExitCode::Success);
+        }
+
+        config.api_key = Some(key);
+        CliConfig::write(&config).unwrap();
+        Ok(ExitCode::Success)
```
jbaxleyiii, line 44: I think we should include an `info!("{} Your personal API key was successfuly set!", style::Key)`

cli/src/commands/auth.rs
```diff
@@ -0,0 +1,46 @@
...
+    }
+}
```
jbaxleyiii, line 46: As a general comment for CLI comands, I'd love to have a pretty overt amount of `debug!` style logging for when people run in `--verbose` mode to help figure out issues or help us diagnose bugs. I'd like to see a `debug!` for each logical progression of actions here

cli/src/config.rs
```diff
@@ -17,9 +17,18 @@ To learn more, checkout https://apollo.dev/cli/telemetry\n";
 #[derive(Debug, Serialize, Deserialize)]
 pub struct CliConfig {
     pub machine_id: String,
+    pub api_key: Option<String>,
```
jbaxleyiii, line 20: instead of `api_key`, how about `personal_api_key` since we do have service tokens as well?

cli/src/config.rs
```diff
@@ -17,9 +17,18 @@ To learn more, checkout https://apollo.dev/cli/telemetry\n";
...
 }
 
 impl CliConfig {
+    pub fn write(cli_config: &CliConfig) -> Result<(), Box<dyn Error + 'static>> {
```
jbaxleyiii, line 24: I think we could split this into two functions:
`pub fn save() -> Result`
`pub fn create_config_file()` -> Result`

Save should check for the file to exist before any write and, if needed, create the dir structure and file.

cli/src/config.rs
```diff
@@ -17,9 +17,18 @@ To learn more, checkout https://apollo.dev/cli/telemetry\n";
...
+        let config_path = apollo_config().unwrap();
+        let toml = toml::to_string(cli_config).unwrap();
+
+        fs::create_dir_all(&config_path.parent().unwrap())?;
+        fs::write(&config_path, toml).map_err(|e| From::from(e))
```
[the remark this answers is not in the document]
jbaxleyiii: This is probably and error we want to handle more cleanly. Can we create a custom error that prints out more verbose "we had an issue when trying to write to your global config at {location}. The error was {msg}"

cli/src/main.rs
```diff
@@ -37,7 +37,16 @@ fn main() {
 
     setup_panic_hooks();
 
-    let mut session = Session::init().unwrap();
+    let session_result = Session::init()
+        .map_err(|e| ApolloError::from(ErrorDetails::CLIConfigError { msg: e.to_string() }));
```
jbaxleyiii, line 41: Since this will prevent anyone from using the CLI, we should be as verbose as we can on how they should fix whatever problem happened. Even a `You may need to change permissions or delete the config at ~/.apollo/config.toml` or something to help people recover would be great

cli/src/terminal.rs
```diff
@@ -1,4 +1,20 @@
 use crate::errors::ErrorDetails;
+use console::Term;
+
+// For interactively handling user input
+pub fn input(msg: &str, sensitive: bool) -> Result<String, ErrorDetails> {
```
jbaxleyiii, line 5: Instead of a boolean for sensitive, I think this could be clearer with an enum for ReadMode

```rs
enum ReadMode {
  Secure
  Normal
}
```

Or something like that. Mainly to make the call sites more explict since `input` is likely to be used a lot!

cli/src/commands/auth.rs
```diff
@@ -0,0 +1,46 @@
...
+#[derive(StructOpt)]
+#[structopt(rename_all = "kebab-case")]
+pub struct Setup {}
+
+impl Command for Setup {
+    fn run(&self, _session: &mut Session) -> Fallible<ExitCode> {
+        let mut config = CliConfig::load().unwrap();
+
+        if config.api_key.is_some() {
+            warn!("Config auth already setup.");
```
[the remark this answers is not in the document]
AlexanderMann: Yup, we left this intentionally not-great in hopes of getting feedback. Should have made it "please give us good feedback here" and then called that out in the description.

cli/src/cli.rs
AlexanderMann: ++ @jsegaran and I have noticed there are distinct differences between what VSCode and IJ pick up for formatting. Would be nice to have a source of truth we can simply base our editors on.

cli/src/config.rs
jsegaran: In ci people would use service tokens to push/interact with the cli. I don't think we should differentiate here whether it is a personal or service token, since we don't have a separate workflow for authenticating as a service and authenticating as a user.

cli/src/commands/auth.rs
```diff
@@ -0,0 +1,45 @@
...
+use crate::telemetry::Session;
+use crate::terminal::sensitive;
+use console::style;
+use log::{info, warn};
+use structopt::StructOpt;
+
+#[derive(StructOpt)]
+#[structopt(rename_all = "kebab-case")]
+pub enum Auth {
+    /// Setup your auth stuff
```
jbaxleyiii, line 14: does this part print out when you run `help`? If so, can you remove it? `auth stuff` doesn't give much help in what the command does 😅

cli/src/commands/auth.rs
```diff
@@ -0,0 +1,45 @@
...
+
+impl Command for Setup {
+    fn run(&self, session: &mut Session) -> Fallible<ExitCode> {
+        session.log_command("auth setup");
+
+        if session.config.api_key.is_some() {
+            warn!("Authentication already configured.");
+        }
+
+        info!("To link your CLI to your Apollo account go to {} and create a new Personal API Key. Once you've done that, copy the key and paste it into the prompt below.",
```
jbaxleyiii, line 30: How does this wrap on terminals? Should we truncate it to 80 lines and add a new line?

cli/src/commands/auth.rs
```diff
@@ -0,0 +1,45 @@
...
+            style("https://engine.apollographql.com/user-settings").cyan());
+        let key = sensitive("User key:")?;
+
+        if key.is_empty() {
+            warn!("No key was inputed, quitting without changes.");
```
jbaxleyiii, line 35: This should probably be error! since we are exiting with a non-zero

cli/src/commands/auth.rs
```diff
@@ -0,0 +1,45 @@
...
+            return Ok(ExitCode::ConfigurationError);
+        }
+
+        let mut config = session.config.clone();
+        config.api_key = Some(key);
+        CliConfig::save(&session.config_path, &config).unwrap();
```
jbaxleyiii, line 41: Let's handle failing to save the config better where with a better error description.

cli/src/commands/auth.rs
jbaxleyiii: @AlexanderMann @jsegaran can y'all still add some `debug!` statements in this command?

cli/src/config.rs
```diff
@@ -17,36 +17,60 @@ To learn more, checkout https://apollo.dev/cli/telemetry\n";
...
 impl CliConfig {
-    pub fn load() -> Result<Self, Box<dyn Error + 'static>> {
-        let mut s = Config::new();
+    pub fn save(path: &PathBuf, cli_config: &CliConfig) -> Result<(), Box<dyn Error + 'static>> {
+        let toml = toml::to_string(cli_config).unwrap();
 
-        let config_path = apollo_config().unwrap();
+        fs::create_dir_all(&path.parent().unwrap())?;
+        debug!("Wrote cli config to path {}", path.to_str().unwrap());
+        fs::write(&path, toml).map_err(|e| From::from(e))
```
jbaxleyiii, line 31:
```suggestion
        fs::write(&path, toml).map_err(From::from)
```

cli/src/main.rs
```diff
@@ -38,7 +38,15 @@ fn main() {
 
     setup_panic_hooks();
 
-    let mut session = Session::init().unwrap();
+    let session_result = Session::init();
+
+    if let Err(err) = session_result {
+        report(&err);
+        let code = err.exit_code();
+        code.exit();
```
jbaxleyiii, line 46: Though this will exit the program, should you still put a `return code.exit();` here to signal that this is an early termination of the progression?

cli/src/commands/auth.rs
AlexanderMann: Ha, good catch, missed that.

cli/src/commands/auth.rs
AlexanderMann: So we were debating that. We assumed most folks will have wrapping setup on terminals and that trying to do "smart" wrapping is something _tougher_ which might be best left for another pr.

We can throw newlines in, or try to truncate, but I do think there'll be a usability issue with logging if we require every call site to manage making the log lines pretty for users. Might be worth looking into various packages which can auto-wrap nicely etc.?

cli/src/commands/auth.rs
AlexanderMann: Ah, yes, since we changed the exit code.

cli/src/commands/auth.rs
jbaxleyiii: That's fair! Well reasoned and fine by me!

cli/src/commands/auth.rs
jbaxleyiii: stuff

cli/src/main.rs
AlexanderMann: @jbaxleyiii if you throw the `return` in front of that, I get warnings in the build about `unreachable` code. However, if you _remove_ the semi-colon, you do not get those warnings...

cli/src/commands/auth.rs
AlexanderMann: Did this with the update to `CliConfig`'s error handling

cli/src/config.rs
```diff
@@ -17,33 +18,79 @@ To learn more, checkout https://apollo.dev/cli/telemetry\n";
...
+    // This can be used for a service api key, or a personal api key.
+    // The CLI doesn't differentiate at the moment.
+    pub api_key: Option<String>,
+}
+
+fn save(path: &PathBuf, cli_config: &CliConfig) -> Result<(), Box<dyn Error + 'static>> {
+    let toml = toml::to_string(cli_config).unwrap();
+
+    fs::create_dir_all(&path.parent().unwrap())?;
+    debug!("Wrote cli config to path {}", path.to_str().unwrap());
```
Enrico2, line 30: this line is before we actually write, which will be misleading if we fail writing. Maybe phrase as "Writing CLI config ..." ?

cli/src/config.rs
```diff
@@ -17,33 +18,79 @@ To learn more, checkout https://apollo.dev/cli/telemetry\n";
...
+    if !path.exists() {
+        CliConfig::create(path)?;
+    }
+
+    s.merge(config::File::with_name(path.to_str().unwrap()))?;
+
+    s.merge(Environment::with_prefix("APOLLO_").separator("__"))
+        .unwrap();
+
+    debug!("Loaded cli config from path {}", path.to_str().unwrap());
```
Enrico2, line 46: same here

cli/src/config.rs
```diff
@@ -17,33 +18,79 @@ To learn more, checkout https://apollo.dev/cli/telemetry\n";
...
-            let toml = toml::to_string(&generated.unwrap()).unwrap();
+        // log initial telemetry warning
+        info!("{}", POST_INSTALL_MESSAGE);
+        Ok(config)
+    }
 
-            fs::create_dir_all(&config_path.parent().unwrap())?;
-            fs::write(&config_path, toml)?;
+    pub fn load(path: &PathBuf) -> Fallible<Self> {
+        load(path).map_err(|e| {
```
Enrico2, line 79: this closure is repeated, consider extracting to a function for DRYer code?

cli/src/commands/auth.rs
```diff
@@ -0,0 +1,50 @@
...
+        if session.config.api_key.is_some() {
+            warn!("Authentication already configured.");
+        }
+
+        info!("To link your CLI to your Apollo account go to {} and create a new Personal API Key. Once you've done that, copy the key and paste it into the prompt below.",
+            style("https://engine.apollographql.com/user-settings").cyan());
+        let key = sensitive("User key:")?;
+
+        debug!("Checking user input...");
+        if key.is_empty() {
```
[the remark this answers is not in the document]
AlexanderMann: Will do, and add tests for this

cli/src/commands/auth.rs
```diff
@@ -0,0 +1,50 @@
...
+            error!("No key was inputed, quitting without changes.");
+            return Ok(ExitCode::ConfigurationError);
+        }
+
+        debug!("Setting new key...");
+        let mut config = session.config.clone();
+        config.api_key = Some(key);
```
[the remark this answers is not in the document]
AlexanderMann: So we thought about doing that but also felt weird/gross doing that as we'd be mutating state that potentially others are relying upon. Maybe a better way to communicate that is by making the `session`'s `config` immutable?

Mind you, the call stack here should almost always be painfully short, but I think there's value in making the side-effecty stuff limited.

cli/src/main.rs
AlexanderMann: So I just tried doing this and our the standard Rust Formatter for the file keeps changing it to `return code.exit();`...and then the warnings start popping up. Not saying this to discourage doing this (I committed things and pushed) but simply want a call out that we might need to define our rust fmt config to prevent folks' tools from overwriting this etc.

cli/src/commands/auth.rs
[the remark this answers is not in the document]
AlexanderMann: ...do we want to also `trim` the key before writing it to the config?

cli/src/config.rs
AlexanderMann: I think we can actually make this much more concise in a number of ways, depending on whether we feel that the separate details for writing and reading are valuable? ie, if we could refactor the `CliConfigWriteError` and `CliConfigLoadError` into `CliConfigError` where we say something like: 

```
Configuration error for file {}: {}
```

We could also have an ENUM of "actions", but I'm not sure how valuable it really is specifying whether it's a read/write issue etc. (especially because I suspect this will frequently be a permissions or not found issue)

cli/src/commands/auth.rs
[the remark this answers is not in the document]
Enrico2: Yeah I think that makes sense.

cli/src/commands/auth.rs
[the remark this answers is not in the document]
Enrico2: Unfortunately (or fortunately, depends how you look at it :)) Rust doesn't have an easy way to define mutability at the field level.

> mutating state that potentially others are relying upon

Can you elaborate here? What use case did you have in mind?
I'm as new to this language as you, but it's my understanding that Rust has compiler level support to ensure no 2 pieces of code have access to the same mutable instance at the same time, and having (aliases to the same underlying data is very bad, and not easy to code)

cli/src/config.rs
Enrico2: Hm, I don't know how to answer that; Maybe that's worthwhile but in the spirit of keeping changes small, I just mean that this can be something like
```rust
load(path).map_err(|e| read_error(e, path))
```

cli/src/main.rs
```diff
@@ -38,7 +38,15 @@ fn main() {
 
     setup_panic_hooks();
 
-    let mut session = Session::init().unwrap();
+    let session_result = Session::init();
+
+    if let Err(err) = session_result {
+        report(&err);
+        let code = err.exit_code();
+        return code.exit()
```
Enrico2, line 46: Welp, this sucks 🤷 

```
warning: unreachable expression
  --> cli/src/main.rs:46:9
   |
46 |         return code.exit()
   |         ^^^^^^^-----------
   |         |      |
   |         |      any code following this expression is unreachable
   |         unreachable expression
   |
   = note: `#[warn(unreachable_code)]` on by default
```

Sorry about the back and forth but maybe we have to live with `code.exit()` and no visual indication. Maybe add an obvious comment?
```rust
// XXX: THIS LINE QUITS THE PROGRAM
```

cli/src/telemetry/session.rs
```diff
@@ -58,13 +63,20 @@ impl Session {
             ci_name: ci_info::get().name,
         };
 
-        let release_version = get_installed_version()?.to_string();
+        let release_version = get_installed_version()
+            .map_err(|e| ApolloError::from(ErrorDetails::CliInstallError { msg: e.to_string() }))?
```
Enrico2, line 67: hm, `ApolloError` has `from_source` that keeps track of the original error, but the ergonomics could be improved there. This (and anywhere else we `map_err`) would then be:
```
map_err(|e| ApolloError::from_source(e, ErrorDetails::CliInstallError { msg: e.to_string() }))
```

but then there's also `with_context` on `Result` that we implement in `errors.rs` that should solve this, but that FnOnce doesn't have access to the `E` from `Result<T, E>` -- I tried changing that locally and it's a PITA, so I guess that `from_source` is what we can go with for now.

Enrico2: or maybe that doesn't compile? Eh, never mind, bike shed.

cli/src/telemetry/session.rs
```diff
@@ -79,7 +91,14 @@ impl Session {
         }
 
         let url = format!("{}/telemetry", domain());
-        let body = serde_json::to_string(&self).unwrap();
+        let telemetry_session = serde_json::json!({
```
Enrico2, line 94: This seems like an anti-pattern, we have a Telemetry struct, we should be able to serialize it with `serde_json`.

(I found this: https://serde.rs/attr-skip-serializing.html)

The thing that pops out first is that we have to manually write this out because `machine_id` is out of session in this PR, why is that? 

duplicating strings in maps and variable names that should match is a common source of bugs, and there is tooling to avoid it. 

Another options that is somewhat cumbersome but hard codes less strings is using `serde_json::to_value` which returns a [Option<Value>](https://docs.serde.rs/serde_json/value/enum.Value.html)

cli/src/main.rs
AlexanderMann: @Enrico2 another solution we might like to adopt is would be this:

https://stackoverflow.com/a/30285110

```rs
fn main() {
    let exit_code = real_main();
    std::process::exit(exit_code);
}

fn real_main() -> i32 {
    // the real program here
    if let Err(err) = session_result {
        report(&err);
        return err.exit_code();
    }
    //...
}
```

I think this potentially sets us up for better unit testing as well, and gets us all the visual diffing we're after. I'd be happy to go this route as it's a rather quick change.

cli/src/telemetry/session.rs
AlexanderMann: Yeah, so this is funky and probably a sign that we should refactor a little more heavily. The reason we removed `machine_id` is because we felt that adding the `config` and `config_path` to the `session` (which is passed down to all commands) was a useful change which could be welcomed by all commands in the future.

The notion that `session` was _purely_ for `telemetry` didn't come across as it's something passed down everywhere. As such, it seemed like the natural place to add the new _session_ values.

Once we added these however, we had duplicate data in the `machine_id` (since it's really just stored in the `config`), and we didn't need to serialize `config` or `config_path` over to the Typescript `telemetry` logic.

I agree that a common source of bugs is mapping names in this manner. Maybe a better solution would be to introduce the notion of a `Session` which has less to do with `telemetry` (lives higher up in the file tree) and can send `Telemetry` messages (re-introducing the original struct in `Telemetry`) which it constructs in some well typed code?

cli/src/config.rs
AlexanderMann: I think I'm confused now, wouldn't we then need a `read_error` and a `write_error`? Since the two closures which do the error mapping are building different errors themselves (one a `CliConfigWriteError` the other a `CliConfigReadError`) we'd need have the shared code be able to build one or the other dynamically no?

cli/src/commands/auth.rs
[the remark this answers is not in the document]
AlexanderMann: 1. `derive` for `clone` is much cleaner, many thanks for the tidbit!
2. That sounds good. My original reason for reaching for the hard `clone` was that it signals that you _shouldn't_ do this and as a way to avoid side effects (ie, for those new to things, it's potentially clearer). I also didn't know that about the compiler. I'm game for refactoring this and we can always go back as it's just bits and bytes.

jbaxleyiii:
> I had to stop mid-review, will continue later today. Posting what I have in the meanwhile (`auth.rs` test and `session.rs` left to look at)
> 
> General comment:
> integration tests are great, and I appreciate the fact that this PR contains some!
> 
> I think there is also a lot of value in unit tests, where applicable.
> 
> Rust distinguishes between these two by files in the `tests` folder vs a
> 
> ```
> mod tests {
> 
> }
> ```
> 
> in the file tested.

@Enrico2 I advised them to mainly test with integration test since the core is so new and will be changing quite a lot (as shown in just this PR!). I do like unit tests a lot but I think the main functionality here is best tested with integration test since we are rapidly itterating on the actual internal structure

cli/src/commands/auth.rs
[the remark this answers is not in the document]
jsegaran: We trim the response directly when we get it from the terminal in `terminal.rs`. I think it is redundant to trim it here?

cli/src/config.rs
Enrico2: Oh sorry, maybe I didn't notice they're different error types!

cli/src/telemetry/session.rs
Enrico2: I'm new to this team and OSS in general, so as a result I feel like it's hard for me to advise on the best course of action here.
On the one hand, that refactor is not too related to this PR, and for the sake of progress I'd like to say it's ok to defer to a quick follow up that will focus only on refactoring as per your suggestion.
On the other hand, there's no clear specific owner for this currently, so I don't know who's priority it's going to be to do it? 
If Gravity is going to be adding more commands, it'd be great if we can settle the underlying structures to be clean and easy to build on top of. As early contributors, you're hitting these issues :)

So, from my perspective, I think I'm ok with a TODO FIXME (we don't fail the build on those) left in the code to refactor the serialization. We can discuss later who will own that. Sound good?

cli/src/commands/auth.rs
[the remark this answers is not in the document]
Enrico2: @jsegaran is right we already trim in `input_or_sensitive`

cli/src/commands/auth.rs
```diff
@@ -0,0 +1,50 @@
...
+
+        debug!("Checking is config's api_key is already set...");
+        if session.config.api_key.is_some() {
+            warn!("Authentication already configured.");
+        }
+
+        info!("To link your CLI to your Apollo account go to {} and create a new Personal API Key. Once you've done that, copy the key and paste it into the prompt below.",
+            style("https://engine.apollographql.com/user-settings").cyan());
+        let key_input = sensitive("Personal API Key:")?;
+        let key = key_input.trim();
```
Enrico2, line 33: FYI: Did you know that in rust you can repurpose a variable name?
```rust
let key = sensitive("Personal API Key:")?;
let key = key.trim();
```

cli/src/telemetry/session.rs
AlexanderMann: That sounds like an excellent idea, and yeah, I also want to say that I haven't taken any of your comments to be grumbling or expecting us to magically know the direction to go in, but rather, "this ish is new...so probably this?"

cli/src/commands/auth.rs
AlexanderMann: 🤯 I did not. I think VSCode must have been lagging when I accidentally typed that the first time and it highlighted red.

cli/src/main.rs
```diff
@@ -39,7 +39,16 @@ fn main() {
 
     setup_panic_hooks();
 
-    let mut session = Session::init().unwrap();
+    let session_result = Session::init();
+
+    if let Err(err) = session_result {
+        report(&err);
+        let code = err.exit_code();
+        // TODO: FIXME
```
jbaxleyiii, line 47: @AlexanderMann Can you add more information about this TODO?

cli/src/main.rs
```diff
@@ -74,7 +83,8 @@ fn main() {
         Err(Error::Apollo(err)) => {
             report(&err);
             let code = err.exit_code();
-            code.exit();
+            // TODO: FIXME
```
jbaxleyiii, line 86: Same with this one

PR #51, opened 2020-05-28
Update build.rs to sort test files before generating tests.

queerviolet:
This was just causing a spurious diff to pop up in `test/tests.rs`, depending on the various ways our different machines ordered the sample files.


queerviolet:
ah

PR #52, opened 2020-05-29
Introduce code generation from .graphql files for making queries

Enrico2:
Josh and Alex want to make the CLI make some GraphQL requests.

We have a plan to use https://github.com/graphql-rust/graphql-client with its code generation features along with a schema that we'll keep up to date as a build phase.

To unblock them, this PR adds 2 example queries and tests to show how this works.

I verified this works locally by using my personal api key and removing the `should_panic`.

Their existing code saves the api key on the file system after authenticating.
We should therefore create a `Client` instance that's instantiated with the api key `auth` creates.

See the documentation for the library we're using and see examples here: https://github.com/graphql-rust/graphql-client/tree/master/graphql_client/tests


cli/src/client/mod.rs
```diff
@@ -0,0 +1,44 @@
...
+use http::Uri;
+use reqwest::blocking;
+use std::error::Error;
+
+mod queries;
+
+static PROD_GQL_API_URL: &str = "https://engine-graphql.apollographql.com/api/graphql";
+
+pub struct Client {
+    personal_api_key: String,
```
queerviolet, line 11: any reason not to call this `api_key`?

cli/src/client/queries.rs
```diff
@@ -0,0 +1,73 @@
...
+pub struct MeQuery;
+
+pub fn me_query(client: &Client) -> Result<Option<me_query::MeQueryMe>, Box<dyn Error>> {
+    let response_body: Response<me_query::ResponseData> =
+        client.send::<MeQuery>(me_query::Variables {})?;
+
+    Ok(response_body.data.and_then(|data| data.me))
+}
+
+mod tests {
```
queerviolet, line 48: Maybe these would be better as [documentation tests](https://doc.rust-lang.org/rustdoc/documentation-tests.html)? Then we get documentation!

queerviolet: Looks like we could also [mark them no_run](https://doc.rust-lang.org/rustdoc/documentation-tests.html#attributes)

cli/src/client/mod.rs
Enrico2: I was thinking that somewhere else in the code we're going to ask the user to use their personal api key obtained from https://engine.apollographql.com/user-settings but actually this code specifically can take any type of API key so I'll rename.

cli/src/client/queries.rs
Enrico2: Good suggestion! They really are just there for documentation :)

Enrico2: Something is not working as expected with the doctests. I tried moving these over and my first hint that they're not working is that IntelliJ doesn't render them as rust code, but as simple comments. So I tried adding a compilation error in them, and everything still passes.
I went to read some docs and couldn't for the life of me figure out how to do this.
I took this example 

```
/// First line is a short summary describing function.
///
/// The next lines present detailed documentation. Code blocks start with
/// triple backquotes and have implicit `fn main()` inside
/// and `extern crate <cratename>`. Assume we're testing `doccomments` crate:
///
/// ```
/// let result = doccomments::add(2, 3);
/// assert_eq!(res, 5);
/// ```
pub fn add(a: i32, b: i32) -> i32 {
    a + b
}
```
and changed the last line from `result` to `res` in the assert to make it not compile, and everything still passes. I tried putting this block in `main.rs` too just to make sure.

I then tried putting that `add` block in a different crate (something I set up locally, the minigrep project from The Book) and IT RENDERS THERE. I seriously am lost on this.

Enrico2: Wow.. Figured this out. So I tried running this:
```
$ cargo test --doc
```
And got the response:
```
error: no library targets found in package `apollo-cli`
```

So, we have as `main.rs` but no `lib.rs` -- I created that file and moved all of the `mod ___;` from `main.rs` to `lib.rs` and NOW DOCTESTS WORK.

🤦

Enrico2: https://github.com/rust-lang/rust/issues/50784

Enrico2: Thinking about this, I'd rather not introduce a `lib.rs` just yet for the sole purpose of doctests. It forces us to mark some modules as public just that `main` could use them. So, I think I'll keep the code as is so that tests will compile.

Enrico2:
No clue how Github is counting lines.. 
image

jbaxleyiii:
@Enrico2 I have some thoughts on this that I'll leave today

.github/workflows/test.yml
```diff
@@ -82,8 +82,8 @@ jobs:
         uses: actions-rs/tarpaulin@v0.1
         with:
           version: '0.9.0'
-          timeout: 300 # 5 mins for test coverage
-          args: '--ignore-tests -- --test-threads 1'
+          timeout: 600 # 10 mins for test coverage
```
jbaxleyiii, line 85: @Enrico2 let's go ahead and comment out the tarpaulin job all together and turn back on the normal linux one since code coverage isn't working well right now anyway

cli/.graphqlconfig
```diff
@@ -0,0 +1,16 @@
+{
```
jbaxleyiii, line 1: Is this used instead of a build.rs script?

cli/src/client/mod.rs
```diff
@@ -0,0 +1,44 @@
+use graphql_client::{GraphQLQuery, Response};
+use http::Uri;
+use reqwest::blocking;
+use std::error::Error;
+
+mod queries;
+
+static PROD_GQL_API_URL: &str = "https://engine-graphql.apollographql.com/api/graphql";
```
jbaxleyiii, line 8: I tend to think that urls should be backed by a possible env variable override both for test suites and for testing against our own stacks. What do you think?

cli/src/client/mod.rs
```diff
@@ -0,0 +1,44 @@
...
+        &self,
+        variables: Q::Variables,
+    ) -> Result<Response<Q::ResponseData>, Box<dyn Error>> {
+        let request_body = Q::build_query(variables);
+
+        let res = self
+            .reqwest
+            .post(&self.uri.to_string())
+            .header("Content-Type", "application/json")
+            .header("apollographql-client-name", "internal-apollo-cli")
```
jbaxleyiii, line 35: Do we want to treat this as a new version of the [Apollo CLI](https://github.com/apollographql/apollo-tooling/blob/master/packages/apollo/src/Command.ts#L229-L233) or a new product all together?

cli/src/client/mod.rs
```diff
@@ -0,0 +1,44 @@
...
+            .header("apollographql-client-version", "0.1")
+            .header("x-api-key", &self.api_key)
+            .json(&request_body)
+            .send()?;
+
+        let response_body: Response<Q::ResponseData> = res.json()?;
+        Ok(response_body)
```
jbaxleyiii, line 42: One thing this doesn't do is handle GraphQL errors (i.e. `{ errors }`). What is the type of `ResponseData` here?

cli/src/client/queries.rs
```diff
@@ -0,0 +1,73 @@
...
+) -> Result<Option<GraphQLDocument>, Box<dyn Error>> {
+    let response_body: Response<get_schema_query::ResponseData> =
+        client.send::<GetSchemaQuery>(variables)?;
+
+    let hash: Option<GraphQLDocument> = response_body.data.and_then(|data| {
+        data.service
+            .and_then(|service| service.schema.map(|schema| schema.hash))
+    });
+
+    Ok(hash)
```
jbaxleyiii, line 29: is the hash really a `GraphQLDocument`?

jbaxleyiii: oh I see, this is just an example

cli/.graphqlconfig
Enrico2: Oh sorry, I should have clarified. This is specifically for IntelliJ's graphql plugin integration. It allows for code highlighting when writing queries and some other features.

cli/src/client/mod.rs
Enrico2: Yup, I can do that, and default to this value. It just means using a function because you can't have a static string get a value from the environment.

cli/src/client/mod.rs
Enrico2: No preference, whatever you think is best..

cli/src/client/queries.rs
Enrico2: Actually you found an issue, I initially tested with getting the entire schema which is a `GraphQLDocument` scalar, which is typedef'd into a String, hash happens to be a `String` too so this compiled. I'll fix that.

cli/src/client/mod.rs
Enrico2: `response_body` is of type:
```rust
pub struct Response<Data> {
    /// The absent, partial or complete response data.
    pub data: Option<Data>,
    /// The top-level errors returned by the server.
    pub errors: Option<Vec<Error>>,
}
```

`ResponseData` is codegen'd based on the query, in our case it's a struct with a single `service` field.

.github/workflows/test.yml
Enrico2: sgtm!

Enrico2:
> You and I caught up in our call today but my biggest point of feedback is I'd like the schema to be gitignored to reduce change noise in the repo and to reduce the amount of uneeded information (i.e we only use a very small portion of the schema).

Yes, although for this PR to unblock @AlexanderMann and @jsegaran, let's check it in so that they're not blocked on the build script we discussed, once I do that one, I'll remove the schema from the repo and .gitignore it.

cli/src/client/mod.rs
Enrico2: These are queries are here for example purposes, we should definitely error check once we have real queries (and then we can delete these)

cli/src/client/mod.rs
jsegaran: Do we want to bake this into the config? We could then piggy back of env_var overrides we use there, and we would have a consistent environment variable naming scheme.

Enrico2: @jsegaran I don't follow what you mean. What is the naming scheme? See the update to the PR and lmk if that works for you?

cli/src/client/mod.rs
jbaxleyiii: 🤔 I think it would be cool to keep it the same and watch the stats roll over in Graph Manager!

Enrico2:
Since this is not approved yet, I might actually change it a bit if https://github.com/apollographql/apollo-cli/pull/55 lands first.

Enrico2:
Much better :)
image

PR #53, opened 2020-05-29
Fix `update` command for macos (darwin)

Enrico2:
Github releases the cli with the darwin name, the code expects macos.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -19,6 +19,7 @@ use std::panic;
 use console::style;
 use structopt::StructOpt;
 
+use crate::cli::{Apollo, Subcommand};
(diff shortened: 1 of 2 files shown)


cli/src/main.rs
```diff
@@ -39,9 +40,10 @@ fn main() {
 
     let mut session = Session::init().unwrap();
 
-    let should_check_for_updates = match env::args().nth(1) {
-        Some(cmd) => !cmd.eq("update"),
-        _ => false,
+    let should_check_for_updates = if let Some(Subcommand::Update(_)) = cli.command {
```
jbaxleyiii, line 43: oh this is so much better!

PR #55, opened 2020-06-02
Set up a build script for cli to download the most recent schema.

Enrico2:
This PR adds a build script that allows developers to work with the most recent schema for AGM.

The logic here is that
1. A developer should be able to work offline after the initial build (which requires network anyway for the cargo dependencies). By default we only download the schema if there's isn't an existing one.
1. If we want to explicitly download the schema, we can do so using the `UPDATE_SCHEMA` environment variable.
1. Only do the hard work if we need to, meaning we check the etags before downloading the schema.

This will allow us to use rust code generation from GraphQL without checking in the schema to source control, and introducing noisy PRs.


cli/build.rs
```diff
@@ -0,0 +1,103 @@
...
+use std::path::PathBuf;
+use uuid::Uuid;
+
+use reqwest::blocking::Client;
+use std::env;
+use std::fs;
+
+/// This script downloads the schema if it's not in the file system (in a path that's .gitignored)
+/// If it exists, the script does nothing.
+/// If UPDATE_SCHEMA environment variable is set, we download the schema if the etag is different on the
```
jbaxleyiii, line 11: Can you also add this to the contributing guide?

Enrico2: Yup!

Enrico2:
(got verbal[slack] approval -- merging)

PR #56, opened 2020-06-03
Atlas graph representation and early work on edges and selections

queerviolet:
This is (early, sketchy) work on how we can represent graphs internally within the compiler / other tools that work with graphs. Key features:
- we're flattening all schema elements into a `Point` struct, which has a `name` and `edges` and a `kind` with maybe more data
- `Edge` is an enum which represents the different relationships between `Point`s. Currently, the one variant is `Name(String)`, representing a child known by that name. We'll probably add some additional information to this relationship, including whether it's a lexical child or from elsewhere.
- currently, all edges are created as we walk the AST. But the intention is that we'll be able to define edges which are not pre-computed—like queries, representing e.g. "all descendant fields which are resolved by a particular service"—but are described as an `Edge` variant and can be cached dynamically on the `Point`
- an edge leads to a `Sel`ection of points. This is currently just a `Vec<Point>`, but I think we may make it a struct which can similarly cache edges

There's some TODOs here:

- [ ] organize & document selection mechanics
- [ ] capture remaining AST data into owned selection nodes
- [ ] sketch a use case for query planning
  - [ ] query edges
  - [ ] add support for materializing lazy edges


benweatherman:
Gonna close this "rusty" PR

PR #59, opened 2020-06-04
Schema get

jsegaran:
Create a cli function `ap schema get` that can get a schema from the registry. Currently we hard code acephi production as the graph, so this command doesn't take any arguments.


cli/src/cli.rs
```diff
@@ -57,14 +57,16 @@ pub enum Subcommand {
...
+    ///  🖨  parse and pretty print schemas to stdout
     Print(commands::Print),
     #[structopt(name = "setup", setting = AppSettings::Hidden)]
     ///  🚜  setup the Apollo toolchain in your environment
     Setup(commands::Setup),
     #[structopt(name = "auth")]
-    ///  🔐 Manage authentication
+    ///  🔐  Manage authentication
     Auth(commands::Auth),
+    ///
```
Enrico2, line 68: 📜 <- emoji suggestion

`📜  Get schema for Graph and Variant` ?

cli/src/client/mod.rs
```diff
@@ -21,19 +21,21 @@ pub struct Client {
     reqwest: blocking::Client,
 }
 
+pub type GraphQLDocument = String;
```
Enrico2, line 24: Can you put this in the schema module? Also add a comment explaining we need a typedef for scalars we declare in our schema.

cli/src/commands/schema/get.rs
```diff
@@ -0,0 +1,53 @@
...
+    response_derives = "PartialEq, Debug",
+    deprecated = "warn"
+)]
+struct GetSchemaQuery;
+
+impl Command for Get {
+    fn run(&self, session: &mut Session) -> Fallible<ExitCode> {
+        session.require_api_key()?;
+
+        let client = Client::from((session.config.api_key.as_ref().unwrap()).to_string());
```
Enrico2, line 25: Future note (not this PR): might be nicer to pass in the client in to the run command, if we ever want to enable running multiple commands in the same invocation.

cli/src/telemetry/session.rs
```diff
@@ -53,6 +53,13 @@ pub struct Session {
 }
 
 impl Session {
+    /// This function will return an error if the api key is not defined
+    pub fn require_api_key(&self) -> Fallible<()> {
```
Enrico2, line 57: Why not return `Fallible<String>` (or `&str` or whatever) ? It'd make the rest of the code simpler, no?

cli/src/commands/schema/get.rs
```diff
@@ -0,0 +1,53 @@
...
+
+        let variables = get_schema_query::Variables {
+            graph_id: "acephei".into(),
+            variant: Some("production".into()),
```
Enrico2, line 29: probably wanted to get these from args?

cli/src/client/mod.rs
```diff
@@ -43,13 +45,34 @@ impl Client {
...
+                })
+            })?;
 
-        let response_body: Response<Q::ResponseData> = res.json()?;
-        Ok(response_body)
+        match response_body.errors {
+            Some(err) => Err(ErrorDetails::GraphQLError {
+                msg: err
+                    .into_iter()
+                    .map(|err| err.message.clone())
```
Enrico2, line 70: no reason to clone here

cli/src/commands/schema/get.rs
```diff
@@ -0,0 +1,53 @@
...
+
+        let data: Option<get_schema_query::ResponseData> =
+            client.send::<GetSchemaQuery>(variables)?;
+
+        let maybe_document: Option<String> = data.and_then(|data| {
+            data.service
+                .and_then(|service| service.schema.map(|schema| schema.document))
+        });
+
+        if maybe_document.is_none() {
```
Enrico2, line 40: design style: consider a `match` instead. kind of bad practice to use `is_none` and `is_some` and returns if can easily be avoided.

cli/src/commands/schema/get.rs
jsegaran: Yea. I wasn't sure if we wanted to do this via session, but since we have a fair number of commands that won't need a gql client I didn't think we wanted to have to build/pass it to all commands. Don't have a strong preference.

cli/src/commands/schema/get.rs
jsegaran: Yea. For now we are going to hard code it as we are making tests.

cli/src/commands/schema/get.rs
Enrico2: I didn't mean via session, just an argument, but really not an issue right now :)

cli/src/client/mod.rs
```diff
@@ -1,39 +1,30 @@
...
 use reqwest::blocking;
 use std::env;
-use std::error::Error;
 
 mod queries;
 
-static PROD_GQL_API_URL: &str = "https://engine-graphql.apollographql.com/api/graphql";
-
-fn api_uri() -> Uri {
-    env::var("APOLLO_API_URL")
```
Enrico2, line 12: This line was lost -- there's no longer a way to set the URL.

This makes me wonder how the integration tests actually passed?

cli/src/client/mod.rs
```diff
@@ -1,39 +1,30 @@
...
     reqwest: blocking::Client,
 }
 
 impl Client {
-    fn from(api_key: String) -> Client {
+    pub(crate) fn from(api_key: String, uri: String) -> Client {
         Client {
             api_key,
-            uri: api_uri(),
+            uri: uri.parse::<Uri>().unwrap(),
```
Enrico2, line 19: I think we should be avoiding `unwrap` unless we know we 100% can't fail. This can fail if this function is invoked with an invalid `Uri` (e.g. from an env variable, assuming you'll restore functionality), so at the very least, use `expect` with a descriptive message?

cli/src/client/mod.rs
```diff
@@ -43,13 +34,29 @@ impl Client {
             .header("apollographql-client-name", "experimental-apollo-cli")
             .header(
                 "apollographql-client-version",
-                env::var("CARGO_PKG_VERSION").unwrap(),
+                env::var("CARGO_PKG_VERSION").unwrap_or("testing".into()),
```
Enrico2, line 37: Could we have a better default than `"testing"` ?

cli/tests/commands/schema/get.rs
```diff
@@ -0,0 +1,79 @@
...
+    async fn error_non_200() {
+        let mut cli = get_cli();
+        let response = ResponseTemplate::new(500);
+        add_mock_graphql(&mut cli, response).await.unwrap();
+
+        cli.command
+            .arg("schema")
+            .arg("get")
+            .assert()
+            .code(predicate::ne(0));
```
Enrico2, line 37: We go through the trouble of having a custom exit code based on different errors, then let's verify we get the expected error code instead of just `ne(0)`

same in other tests.

cli/src/config.rs
```diff
@@ -38,9 +41,14 @@ fn load(path: &PathBuf) -> Result<CliConfig, Box<dyn Error + 'static>> {
         CliConfig::create(path)?;
     }
 
+    let mut s_defaults = Config::new();
```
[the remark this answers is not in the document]
jsegaran: It solves the problem with environment variable override nicely. Allowing us to declare a namespace `APOLLO_` and automatically generating the rest of the environment variables from there instead of having to create new ones.
Having to hardcode the `api_url` string isn't ideal but I think using the `Config` abstraction does provide benefits.

cli/src/client/mod.rs
jsegaran: We allow over-writing in the environment variable merge of config in `cli/src/config.rs:51`

cli/src/client/mod.rs
jsegaran: Switched it to `dev`

cli/src/errors/details.rs
```diff
@@ -54,6 +54,18 @@ Please ensure you have permissions to edit your environment variables."
 
     #[error("Error writing config file {}: {}", .path, .msg)]
     CliConfigWriteError { msg: String, path: String },
+
+    #[error("No api key found in config file: Please run `ap auth setup` first")]
+    NoApiKeyError,
+
+    #[error("Received invalid graphql response from Apollo Studio: {}", .msg)]
+    RegistryNetworkError { msg: String },
```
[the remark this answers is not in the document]
jsegaran: Changed it to `InvalidGraphQLError`

cli/src/commands/schema/mod.rs
```diff
@@ -4,10 +4,46 @@ pub mod store;
...
+    if hash_capture.is_some() {
+        let cap = hash_capture.unwrap();
+        return Ok(SchemaRef::SchemaHashRef {
+            graph_id: cap[1].to_string(),
+            hash: cap[2].to_string(),
+        });
+    }
+
+    Err(ErrorDetails::InputError {
+        msg: format!("Schemref {} is not a valid schemaref", src),
```
Enrico2, line 40: typo and .. is `schemaref` a term we publicly use somewhere?

cli/src/commands/schema/mod.rs
```diff
@@ -4,10 +4,46 @@ pub mod store;
...
 pub use get::Get;
+use regex::Regex;
 pub use store::Store;
 
+enum SchemaRef {
+    SchemaVariantRef { graph_id: String, variant: String },
+    SchemaHashRef { graph_id: String, hash: String },
+}
+
+fn parse_schema_ref(src: &str) -> Fallible<SchemaRef> {
```
Enrico2, line 18: This is ideal code for unit testing instead of capturing these in integration tests.

cli/src/commands/schema/get.rs
```diff
@@ -24,10 +27,21 @@ impl Command for Get {
...
             session.config.api_url.as_ref().unwrap().clone(),
-        );
-        let variables = get_schema_query::Variables {
-            graph_id: "acephei".into(),
-            variant: Some("production".into()),
+        )?;
+
+        let schema_ref = parse_schema_ref(&self.schema_ref)?;
+
+        let variables = match schema_ref {
```
Enrico2, line 36: nice and clean 👍

cli/src/commands/schema/mod.rs
Enrico2: this is really the only "request changes" comment.

cli/src/commands/schema/mod.rs
```diff
@@ -0,0 +1,100 @@
...
+#[derive(StructOpt)]
+#[structopt(rename_all = "kebab-case")]
+pub enum Schema {
+    ///  Get schema from a graph variant or schema hash
+    Get(get::Get),
+    Store(store::Store),
+    Check(check::Check),
+}
+
+#[test]
```
Enrico2, line 57: Unit tests go inside a `mod tests` in the same file and add `#[cfg(test)]` on the mod. There are examples in other files in the code base

Enrico2: Good page to read: https://doc.rust-lang.org/book/ch11-01-writing-tests.html

Enrico2: I pushed some changes, please take a look, for next time :)
I want some of these changes in my branch, so I'll merge once tests pass.

PR #65, opened 2020-06-17
Fix flaky tests, refactor some code, clean clippy, and more.

### LINKED ISSUE #63: Find and fix flaky apollo-cli test that causes checks to fail
Reproducible locally:

```
cargo install flaky-finder
flaky-finder -j8 -r20 "cargo test -p apollo-cli"
```

This might contain one of the reasons:
https://github.com/apollographql/rust/pull/60/checks?check_run_id=773610533

```
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: NetworkError', cli/src/version.rs:94:16
stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /Users/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.46/src/backtrace/libunwind.rs:86
   1: backtrace::backtrace::trace_unsynchronized
             at /Users/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.46/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print_fmt
             at src/libstd/sys_common/backtrace.rs:78
   3: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
             at src/libstd/sys_common/backtrace.rs:59
   4: core::fmt::write
             at src/libcore/fmt/mod.rs:1069
   5: std::io::Write::write_fmt
             at src/libstd/io/mod.rs:1504
   6: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:62
   7: std::sys_common::backtrace::print
             at src/libstd/sys_common/backtrace.rs:49
   8: std::panicking::default_hook::{{closure}}
             at src/libstd/panicking.rs:198
   9: std::panicking::default_hook
             at src/libstd/panicking.rs:218
  10: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:511
  11: rust_begin_unwind
             at src/libstd/panicking.rs:419
  12: core::panicking::panic_fmt
             at src/libcore/panicking.rs:111
  13: core::option::expect_none_failed
             at src/libcore/option.rs:1268
  14: core::result::Result<T,E>::unwrap
             at /rustc/49cae55760da0a43428eba73abcb659bb70cf2e4/src/libcore/result.rs:1005
  15: apollo_cli::version::get_latest_release
             at cli/src/version.rs:94
  16: apollo_cli::version::needs_updating
             at cli/src/version.rs:173
  17: apollo_cli::version::background_check_for_updates::{{closure}}
             at cli/src/version.rs:201
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
thread panicked while panicking. aborting.

```

Enrico2:
This PR fixes https://github.com/apollographql/rust/issues/63

The issue contains a long comment on the path I went through to find the root cause, I _really_ took the long way around :)

The TL;DR: is 
* overriding `APOLLO_CDN_URL` in multiple tests caused a race condition for some tests to try and talk to the mock server set up by another test, because they find the mock server through the value in that environment variable.
* Along the way of finding the problem, I went through refactoring some of the code I saw along the way. Looking at individual commits should make it easier to understand.
* I deleted code that was commented out, it didn't seem valuable.
* Finally, I ran `cargo clippy`, found a lot of issues, cleaned them, and committed that too.

### DIFF
--- a/cli/src/telemetry/session.rs
+++ b/cli/src/telemetry/session.rs
@@ -8,7 +8,6 @@ use serde::{Deserialize, Serialize};
 use uuid::Uuid;
 
 use crate::config::CliConfig;
-use crate::domain;
 use crate::errors::{ApolloError, ErrorDetails, Fallible};
 use crate::layout::apollo_config;
 use crate::version::get_installed_version;
@@ -50,6 +49,10 @@ pub struct Session {
 
     /// The current instantiation of CliConfig
     pub config_path: PathBuf,
+
+    /// The current host for the Apollo CDN, normally retreived from the APOLLO_CDN_URL
+    /// environment variable, and defaults to https://install.apollographql.com
+    pub cdn_host: String,
 }
 
 impl Session {
@@ -78,6 +81,9 @@ impl Session {
 
         let config = CliConfig::load(&config_path)?;
 
+        let cdn_host = env::var("APOLLO_CDN_URL")
+            .unwrap_or_else(|_| "https://install.apollographql.com".to_string());
+
         Ok(Session {
             command,
             session_id,
@@ -85,6 +91,7 @@ impl Session {
             release_version,
             config,
             config_path,
+            cdn_host,
         })
     }
 
@@ -98,7 +105,7 @@ impl Session {
             return Ok(false);
         }
 
-        let url = format!("{}/telemetry", domain());
+        let url = format!("{}/telemetry", &self.cdn_host);
         // TODO: FIXME:: https://github.com/apollographql/rust/pull/50#discussion_r434231100
         // Requiring effectively copy paste renaming for
         // data is a _high_ chance for bugs and no something
@@ -130,7 +137,10 @@ impl Session {
 
         debug!("Telemetry request done");
         match resp {
-            Ok(res) => Ok(res.status().is_success()),
+            Ok(res) => {
+                debug!("response status is {}, response is {:?}", res.status(), res);
+                Ok(res.status().is_success())
+            }
             Err(_e) => Ok(false),
         }
     }
@@ -140,18 +150,20 @@ impl Session {
 mod tests {
     use std::env::set_var;
 
-    use wiremock::matchers::{method, PathExactMatcher};
+    use wiremock::matchers::{method, path};
     use wiremock::{Mock, MockServer, Request, ResponseTemplate};
 
     use super::Session;
 
     #[async_std::test]
     async fn reports_session() -> Result<(), Box<dyn std::error::Error>> {
+        let proxy = MockServer::start().await;
+
         // create a session
         let mut session = Session::init()?;
         session.log_command("test");
 
-        let proxy = MockServer::start().await;
+        session.cdn_host = proxy.uri();
 
         let payload_matcher = move |request: &Request| {
             let body: serde_json::Value =
@@ -163,14 +175,13 @@ mod tests {
         };
 
         Mock::given(method("POST"))
-            .and(PathExactMatcher::new("/telemetry"))
+            .and(path("/telemetry"))
             .and(payload_matcher)
             .respond_with(ResponseTemplate::new(200))
             .expect(1)
             .mount(&proxy)
             .await;
 
-        set_var("APOLLO_CDN_URL", &proxy.uri());
         assert_eq!(session.report()?, true);
 
         // the mock will panic if it this report is sent
--- a/cli/src/version.rs
+++ b/cli/src/version.rs
@@ -1,3 +1,10 @@
+use crate::domain;
+use crate::errors::{ErrorDetails, Fallible};
+use crate::layout::apollo_home;
+
+use log::debug;
+use semver::Version;
+use serde::{Deserialize, Serialize};
 use std::env;
 use std::env::consts::OS;
 use std::error::Error;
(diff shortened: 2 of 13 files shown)


Enrico2:
Oh hey look at that all checks passed on the first time ;)

cli/src/version.rs
```diff
@@ -189,23 +190,23 @@ pub fn needs_updating() -> Result<CLIVersionDiff, Box<dyn Error + 'static>> {
...
 }
 
-pub(crate) fn background_check_for_updates() -> mpsc::Receiver<Version> {
+pub(crate) fn background_check_for_updates(cdn_host: &str) -> mpsc::Receiver<Version> {
     debug!("Checking to see if there is a latest version");
     let (sender, receiver) = mpsc::channel();
 
     // don't check for updates if APOLLO_UPDATE_CHECK_DISABLED is set
-    if env::var("APOLLO_UPDATE_CHECK_DISABLED").is_ok() {
-        return receiver;
```
jbaxleyiii, line 198: oh so retunring this early was a source of the bugs as well? Interesting!

cli/tests/cli-runtime.rs
```diff
@@ -24,64 +18,4 @@ mod unix {
...
-    //         .mount(&proxy)
-    //         .await;
-
-    //     Ok(proxy)
-    // }
-
-    // this test only verifies that the request is sent to the CDN as the printing
-    // is flaky given its background thread status
-    // #[async_std::test]
-    // async fn prints_if_update_is_found() -> Result<(), Box<dyn std::error::Error>> {
```
jbaxleyiii, line 50: I would still like to find a way to test this if possible?

cli/src/telemetry/session.rs
```diff
@@ -78,13 +81,17 @@ impl Session {
 
         let config = CliConfig::load(&config_path)?;
 
+        let cdn_host = env::var("APOLLO_CDN_URL")
```
AlexanderMann, line 84: 👍

cli/src/telemetry/session.rs
```diff
@@ -130,7 +137,10 @@ impl Session {
 
         debug!("Telemetry request done");
         match resp {
-            Ok(res) => Ok(res.status().is_success()),
+            Ok(res) => {
+                debug!("response status is {}, response is {:?}", res.status(), res);
```
AlexanderMann, line 141: This is part of what originally tipped you off to the conflicting servers yeah?

cli/src/telemetry/session.rs
```diff
@@ -163,14 +175,13 @@ mod tests {
         };
 
         Mock::given(method("POST"))
-            .and(PathExactMatcher::new("/telemetry"))
+            .and(path("/telemetry"))
```
AlexanderMann, line 178: 👍

cli/src/telemetry/session.rs
```diff
@@ -163,14 +175,13 @@ mod tests {
...
             .and(payload_matcher)
             .respond_with(ResponseTemplate::new(200))
             .expect(1)
             .mount(&proxy)
             .await;
 
-        set_var("APOLLO_CDN_URL", &proxy.uri());
```
AlexanderMann, line 173: 👍

cli/src/version.rs
```diff
@@ -79,40 +76,43 @@ fn get_version_disk(
...
     let resp = reqwest::blocking::Client::new()
         .head(&url)
         .header("User-Agent", "Apollo CLI")
         .send()
-        .map_err(|_err| ExitCode::NetworkError)
-        .unwrap();
+        .map_err(|_err| ErrorDetails::ReleaseFetchError)?;
 
     if !resp.status().is_success() {
         debug!("Request to load the latest release");
```
AlexanderMann, line 96: Why keep this around at all and not wrap it into the new `debug` statement below?

cli/src/telemetry/session.rs
Enrico2: Yes, and I figured this is a good thing as a debug message. 
fwiw I didn't see `debug!` statements printed at all, not sure why, I think I didn't configure it right with how env_logger wants it. This was `eprintln!` while I was debugging.

cli/src/version.rs
Enrico2: hm yeah I'll fix this. That line doesn't make much sense to me anyway.

cli/src/version.rs
Enrico2: Ah, no, this was a design style choice. Generally speaking I try to avoid `return` statements, especially in this case where the returned value is the same in both branches.

I adopted this approach when I started coding Scala (and fwiw was an explicit style guide at Twitter) where it was the first language I wrote that didn't require the `return` keyword explicitly. Rust has the same thing.

The backing idea here, is that returning early hurts readability as a function's code grows, and creates many "exit points" that one needs to keep in their head. Instead, branching as needed and having a single exit point is usually easier to parse. I realize this may be subjective and I'm not phrasing it well :)

Also, I will say that in cases where we want to exit very early and the rest of the function is long, it's actually better to use return, for example:
```rust
fn doSomethingOnVector(v: Vec<u32>) {
  if v.is_empty() return;

  // .. do something long and complex on v
}
```

cli/tests/cli-runtime.rs
Enrico2: I'll give it a shot! (after I merge this PR though)

generally speaking, I think it might be more successful in the long run, in cases like this, where instead of commenting out (no more compilation), use `[#ignore]` on the test and a comment explaining why it's being ignored, ideally with a TODO and a link to an issue?

PR #79, opened 2020-06-22
cargo lock + fix tests

ashleygwilliams:
fixes #78 #74

PR #80, opened 2020-06-22
Fix failing unit tests -- issue with upstream dependencies

### LINKED ISSUE #74: tests panic when run locally, *sometimes*
on a clean `git clone`, running `cargo test` produces:

```
running 6 tests
test commands::print::does_not_fail_with_no_files ... ok
test commands::schema::tests::parse_variant_ref ... ok
test commands::schema::tests::parse_fail ... ok
test commands::schema::tests::parse_hash_ref ... ok
thread panicked while panicking. aborting.
error: test failed, to rerun pass '-p apollo-cli --bin apollo-cli'

Caused by:
  process didn't exit successfully: `/Users/ag_dubs/apollo/rust/target/debug/deps/apollo_cli-d9a8dc7a305f7794` (signal: 4, SIGILL: illegal instruction)
```

Enrico2:
This PR fixes #74 -- but not in a good way.

Turns out it was only reproducible in CI because we didn't check in our Cargo.lock file. I was able to reproduce locally by deleting my local copy of Cargo.lock, tests started failing.

I then compared the passing vs. failing Cargo.lock, and specifically focused on differences that related to this test.

So, here's what I found: While we specifically depend on `async-std = "1.6.1"`, wiremock has that dependency as `version = "1"`, which brings in `1.6.2` (that's what the failing `Cargo.lock` had)

So, in my `Cargo.lock`, I forced `1.6.1`. That change on its own was not enough, I had to also revert `smol` from `0.1.18` to `0.1.11`.

THAT was the thing that enabled tests to pass.


Enrico2:
Changed to draft, the approach here to manually change `Cargo.lock` will not work as its generated by Cargo. 
However, we still have to iterate on fixing #74 , I'll use this branch for that.

Enrico2:
Closing in favor of #79

Issue #71, opened 2020-06-19
Github Workflows change proposal


Enrico2:
Hi,

I spent a good amount of time yesterday thinking about our CI/CD process in this repository and realized we might want to do things a little differently. 

I had a few motivators, the merit of which may or may not be subjective :)
1. Make PR checks finish as quick as possible without significantly affecting the signal we get from them.
1. Fail fast if there's a build failure.
1. Heavily cache so that we can benefit from incremental compilation and avoid duplicate work.

The first thing I noticed was that for each PR commit, we're running the same build 6 times (3 for separate OS targets, one in build.yml and one in test.yml) and tests 3 times.
In rust, running `cargo test` also builds, and because these run in parallel workflows, we wait a long time before tests are actually run. This often means a long time before we get a good signal from the check.

Then I saw that we build a full release binary on every PR commit, something that I'm guessing we don't need 99% of the time.

I started to play around with different variations, these were not attempts at changing the workflows, but rather exploration. You can see the different commits in [this draft](https://github.com/apollographql/rust/pull/70).

At some point, and this became a distraction, a unit test was consistently failing on something very unexpected; a _timeout_ from the wiremock server we spin up, I started digging into that, but because it was only reproducible on CI, my iteration speed was slow af because I noticed the cache is not doing what it's supposed to and we download and compile all of the cargo stuff on each build. The cache always misses. Upon further investigation, this is probably something I BROKE in this branch, but anyway, it got me to think about what we cache, and made me notice another thing, we build the cache key by hashing
```
${{ hashFiles('**/Cargo.lock') }}
```
But, we don't actually check in `Cargo.lock` so that part of the key is empty, which means we may change dependencies' versions but load older dependencies from cache. **This is real bug no. 1**

Along the way, I saw that while we build the entire repo on mac and windows, we only build the CLI on Linux. **This is real bug no. 2**

Back to the main thing, I asked @jbaxleyiii why we've set up the workflow to build a full release CLI on each PR commit, and he replied:

[...]

PR #84, opened 2020-06-23
Github Workflows revamping

Enrico2:
This PR pretty much reworks what we do in our Github workflows. 

I had a few goals in mind:

1. Do not lose functionality.
1. Do not do redundant work.
1. Do less on each PR commit when it's not needed.

The result of this exploration led me to this state:

1. Deploying the CDN is not changed, and happens when we push to `main` but only if anything in the `'cdn/**'` path changed.

1. I separated out building/testing rust from non-rust.

1. Testing the CDN happens only when there are changes to the `'cdn/**'` path on PR commits.

1. For rust code, we have a single workflow that runs on every PR commit. It lints, compiles and tests the entire workspace. Linting doesn't fail the workflow, but will show a red X if it fails. Linting will also annotate the build result, this is visible on the `/files` view and on the action output.

1. We no longer build a `--release` binary on each commit, that took a long time, and we rarely need it. Instead, we can manually trigger the workflow thanks to `workflow_dispatch` ([see this new feature!](https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/)). We also run this workflow on every merge to main. The workflow tests and builds a `--release` binary.

1. Use [actions-rs](https://github.com/actions-rs) where possible which does the boilerplate of installing rust. Also when using [their cargo](https://github.com/actions-rs/cargo) we see cargo errors as PR annotations. Cool!

1. The cache keys have been cleaned up, we use the selected toolchain `rustc_hash` as part of the key. Not that it matters but FYI this is different for stable vs. nightly.

1. Skip compilation step on a cache hit; in the workflow that runs on PR commits (`test-rust.yml`) we have 3 jobs: lint, compile including tests, run tests. While technically running tests already compiles, it helps to compile as a separate job so that we cache the compilation results regardless of test failures. This means that if a commit in a PR fails tests, subsequent commits to fix the tests will run checks faster.

1. DRY DRY DRY and simplify:
	1. Heavily using `matrix` values instead of `env` values per `if: matrix.build == ..`
	1. using actions/cache@v2 which allows for multiple paths instead of a cache action per folder.
	1. always use stable rust, since we did anyway, this meant less lines of code.
	1. the workflows that build a release binary have been consolidated to a single workflow with multiple triggers.

---------

Actual stuff this PR fixes/improves:
1. On each PR commit, we were compiling twice (once for build and once for test)
1. On each PR commit, we were building the entire workspace with `--release` for mac and windows, but only the CLI for linux (the latter is what we _should_ be doing), and making a binary, that's rarely used.
1. On PRs that don't touch rust/cdn, we were building cdn stuff/rust stuff respectively.
1. repeated code, as much as reasonable (we still repeat across workflows, but until Github fixes that, not much we can do. See https://github.com/actions/runner/issues/438)

---------

I iterated on this a lot, I kept the commit log for posterity, but I'll definitely squash this one before merging :)


jbaxleyiii:
@Enrico2 a few things about this PR then a more general comment about the issue / pr.

It looks like tests are either silently failing **or** the verbosity setting of the logs look like they are failing in multiple places

image

image

I'm not sure if they are actually failling of not but the verbosity setting here makes it hard to see what is going wrong or right with the test suite. I'd love to better understand why debug mode is set to on as the default state for the tests as I think it will make for a really noisy set of logs that are hard to grep and understand.

The `build-binary` label doesn't feel quite right to me either. I think we should either totally turn of binary building on PR's or turn it back on to its own job that always runs. I prefer the latter combined with some of you changes here because it gets the benefits of the fail fast setup you were looking for while ensuring that we still know our release style builds will function with every set of changes and not be suprised by anything come release time. We could, alternatively, do a nightly build which I think would be just as sufficient but if it isn't automated I strongly suspect we will forget to do it and miss something accidently breaking. I think this is a great thing for futher discussion with the larger internal teams using the early build feature to understand how they would want to use it (gravity, tooling, server, and client) and we can even go ahead and disable the trigger for it now as a first step.

My overall feedback is that improving the CI / CD pipeline here is a **great** thing to do with real benefits for internal teams and external contributors. However, I don't think that we had enough discussion / consensus from different parties on #71 to move forward quite as quickly with a build out here. There is definitely merit to trying out options and a prototype / design by doing style way of working to inform a better understand of what we can and cant do but let's make sure that everyone has the chance to speak up on their opinions, understanding of the problem space, and hopes and dreams for the change before we go too too far on an implementation. I really like a lot of these changes and suspect we should land most of them but part of this repo being shared across many teams and, eventually, the larger Apollo OSS community, means we have to do somethings that we can individually do quickly, more slowly to build consensus. 

Ultimately what I'd ask is to let this sit for a few days for others to review it and #71. Ask questions, give feedback, etc. Use this as a base for your PRs on the wasm bridge stuff so you get the benefits of it and can try it out in practice, and then as a wider team we can make a decision on how we want to move forward (again I think most of this looks really great!).

Enrico2:
@jbaxleyiii yeah sorry, I didn't clarify my intent. I suspect we're definitely going to do _something_ about our workflows, but I didn't intend to land anything until we've agreed what is it that we want to get done. This PR is me starting things off. Reviews and iteration is definitely expected.

The use of the `build-binary` label was kind of a prototype to show that we can get binary builds on PRs on demand instead of on every commit. Take a look at [the actions tab for this branch](https://github.com/apollographql/rust/actions?query=branch%3Aran%2Fworkflow_revamp), the workflow to build a binary took 16 minutes last time. I believe doing that on every commit does more harm than it's worth, this was an attempt at a happy medium. I'm also good with the scheduled nightly option you're suggesting on top of an "on demand" option. (btw there are other events we can use, not just labeling, this was just one option.)

As for the verbosity: This was the lesson I got from `thread panicked when panicking` that was only happening on CI that didn't show any information. I had to enable that verbosity to see what failed and where. My thought here was 99% of the times we'll care about logs, is when there's a failure, and without verbose logging, we won't have anything actionable to work off of.
These tests that you pointed to are not failing as far as I can tell. I believe what we're seeing is the result of not capturing stdout, and running tests that expect a failure/panic.
I'm totally fine with tweaking this (e.g. trying with `RUST_BACKTRACE=1` only, still capturing, using `info` level debug, or a combination of those) but the previous setting gave me no visibility when there's a problem.

Enrico2:
As I woke up today at 5:30 and couldn't get back to sleep, I thought about this a little more. Specifically about the binary build thing.

So, a nightly job is not necessary, since we can instead set it up to build it when we push to `main`. Any errors resulting from that will both email us (I think?) and show up on the main repo's badges.

@jbaxleyiii When you'd said
> we still know our release style builds will function with every set of changes and not be suprised by anything come release time.

As far as I can tell, we don't do any functionality verifications on the resulting binary, and using the `--release` flag, as far as I know, is only applying further optimizations that would not matter to the core functionality, just the performance.
 
I was talking to @ashleygwilliams about this yesterday (although our conversation got cut short) and she was suggesting that we can expose the artifacts resulting from a regular build (the `debug` release) if anyone wants to test out a binary from a PR. She also mentioned that the bulk of time spent by the CLI would be on network calls, and she doubts we'll see a big improvement at all from a release build (sorry if I'm misquoting or misunderstood here!)

FWIW, I'm also ok making these changes in multiple steps; first just DRY, simplify and separate rust from cdn stuff, but keep the same functionality of workflows, then apply whatever we decide in #71 

But now for real, I'm leaving this here (and maybe basing off of it) until we know what we want.

Enrico2:
Closing in favor of #118

PR #85, opened 2020-06-24
Suggestions on top of qv-visitor #83

Enrico2:
- Fix `build.rs` to sort `use` statements
- Rename `Mapping` to `Folding`
- put tests in `mod tests`
- `new()` for `Folding`
- Rename stuff for readability
- return `Option<F::Output>` instead of `Folding<F>`

### DIFF
--- a/graphql-parser/src/lib.rs
+++ b/graphql-parser/src/lib.rs
@@ -135,7 +135,7 @@
 //! You can also `map` an AST into another form.
 //!
 //! ```rust
-//! use graphql_parser::{parse_query, Map, query, query::{Node, Document, Definition, SelectionSet, Selection}};
+//! use graphql_parser::{parse_query, Fold, query, query::{Node, Document, Definition, SelectionSet, Selection}};
 //! let ast = graphql_parser::parse_query(r#"
 //! query {
 //!     someField
@@ -143,8 +143,8 @@
 //! }
 //! "#)?;
 //! struct ToIndentedNodeTypes {}
-//! impl Map for ToIndentedNodeTypes {
-//!     // We're mapping the query AST into a string
+//! impl Fold for ToIndentedNodeTypes {
+//!     // We're folding the query AST into a string
 //!     type Output = String;
 //!     // The *merge* function controls how we merge child output data up the tree
 //!     // when the map of the child is complete. Here we join parent and child
@@ -153,7 +153,7 @@
 //!         format!("{}\n{}", parent, child)
 //!     }
 //! }
-//! impl query::Map for ToIndentedNodeTypes {
+//! impl query::Fold for ToIndentedNodeTypes {
 //!     fn query<'a>(&mut self, _: &Document<'a>, stack: &[Self::Output]) -> Self::Output {
 //!         format!("{}query", "  ".repeat(stack.len()))
 //!     }
@@ -168,7 +168,7 @@
 //!     }
 //! }
 //!
-//! let tx = ast.map(ToIndentedNodeTypes{});
+//! let tx = ast.fold(ToIndentedNodeTypes{});
 //! pretty_assertions::assert_eq!(tx.output, Some(String::from("query
 //!   query_def
 //!     sel_set
@@ -203,4 +203,4 @@ pub use crate::schema::parse_schema;
 
 mod name;
 pub use crate::name::*;
-pub use visit::Map;
+pub use visit::Fold;
--- a/graphql-parser/src/lib.rs
+++ b/graphql-parser/src/lib.rs
@@ -114,11 +114,11 @@
 //! };
 //!
 //! /// Schemas can contain queries, so all visitors must at least implement query::Visitor.
-//! impl query::Visitor for Fields {}
+//! impl query::QueryVisitor for Fields {}
 //!
 //! /// To collect field definitions, we'll also want to implement the appropriate method of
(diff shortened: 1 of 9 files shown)


graphql-parser/src/lib.rs
```diff
@@ -114,11 +114,11 @@
 //! };
 //!
 //! /// Schemas can contain queries, so all visitors must at least implement query::Visitor.
-//! impl query::Visitor for Fields {}
```
ashleygwilliams, line 117: i'm curious why you would choose to name this this way? it feels redundant to me (the word query is already there!)

ashleygwilliams: fwiw there's a clippy lint for this: https://rust-lang.github.io/rust-clippy/master/index.html#module_name_repetitions. not saying we should necessarily be dogmatic about it or turn on the lint, but i would say it is generally idiomatic rust to not have the module name in the type name (see Rust std library).

Enrico2: There is a seemingly identical `Node` trait in both query and schema. It took me a while to realize why, because they use different types of `Visitor`. I had four files open in front of me, all using the same names for structs but referring to different types, some of them extending extending each other. It was a lot harder to keep track of before renaming.

ashleygwilliams: Seems like we could do some improvement with naming, but I don't think making this naming redundant is the right call.

Enrico2: I can rename it back if you feel strongly about this, but I'm doing while thinking this hurts readability, as the reader of code I've not written (this PR is change _suggestions_ on another PR, #83 so if you have naming ideas, that one would be the place for it..)

This is one of the few Rust things I really don't like; The convention to not include the module before types makes a lot of sense, but at the same time there's no quick way (i.e. "with a glance") way to answer the question "what module does this class belong to?", and it seems to be a common pattern to use the same name in the same crate under different modules.
e.g. in a JVM language there's a package declaration at the top, so that question is easily answerable. With Rust it depends on the file name, but not really because a trait in `foo/goo.rs` can be made public directly on `foo` with 
```rust
pub use self::goo::*;
```
in `foo/mod.rs`

So we end up with lines like this:
```rust
pub trait Visitor: query::Visitor {
```
but that `query::Visitor` is actually in `query/visit.rs`
or this:
image
2 traits that are verbatim identical, but then the reader is supposed to notice that the `Visitor` in `V: Visitor` is referring to different visitors, and again there's no quick way to see that without further navigation.

Conventions are meant to ease readability, by aligning everyone's expectations, but in this case I don't see how it helps. (trust me, I'm a sucker for aligned consistency and sticking with conventions and I'll be the main person to comment on that stuff..)

ashleygwilliams: i didn't necessarily suggest that you return it to what it was, i meant to say that i think we can think more broadly and potentially come up with a better solution than simply repeating the module name. when things get difficult to read because of naming it often indicates a larger question. i will need to read the PR more in order to make a suggestion myself, but the rust naming convention exists for a reason and i dont think adding the module name is the only possible way to improve this. indeed, far from it.

in a more general sense, i think it is important that you lead with these *problems* and not simply lead with a solution presented in a PR, and i am personally frustrated. i think bringing up that you found the 2 visitors confusing would be a good *comment* on the original PR and that this entire PR should be closed in favor of conversation on the original PR. none of the changes in this PR are significant enough or certain enough to warrant a separate, uncolicited, PR. i'm sure that @queerviolet has reasons she named these the way she did (or has thought about the question of naming and has questions/concerns of her own) and I would like to hear them!

queerviolet: I gave them parallel names because that's how the parser is already architected. There's a `schema::Document` and a `query::Document`, for example. Am I super thrilled with that in every way? No, for the reasons you describe. But it's what's there.

The only reason to have two grammars at all is because the query grammar is much smaller, and thus, presumably faster. But we could benchmark it and decide that the confusion isn't worth it.

Other approaches: I tried to genericize the `Node` trait more, but higher-kinded types aren't currently possible in Rust, and that's what a straightforward translation would require (specifically, the `Node` trait needs to be generic over some visitor trait `V` and some map trait `M`). We could avoid duplicating the code with a small macro, but I don't deduplicate code until I get to three copies, because the abstraction cost is IMO never worth it before that.

Finally, I thought about departing from the design of the rest of the parser and just having a single Visitor which can visit any query or schema node. That would be okay for `Visitor`, because you only have to implement the methods for nodes you care about. But the `Map` trait makes you implement all node types, and I thought it would be weird to have to provide empty implementations of, say, schema node mappings when you know you're parsing a query. Also, it was already written the other way. I figure if we do benchmark it and decide it's not worth it to have two parsers (really, two parser entry points—the schema parser depends on parts of the query parser), we can fold everything together then.

queerviolet:
Okay, I really like `Fold` as a word. We'll use that as "the thing `Map` returns". We do actually want to return the whole `Fold` tho, because it owns the `map`, and the map can hold state we'll want to recover. Specifically for us, it'll have graph layer data.

Updated the original PR.

graphql-parser/src/lib.rs
Enrico2: Yeah I tried that Node abstraction too, it got hairy :)
Am also down with the rule of 3 👍👍

PR #88, opened 2020-06-30
JSON De/Serialization for `QueryPlan` (compatible with JS)

Enrico2:
This PR adds the ability to serialize and deserialize a `QueryPlan` into JSON in the same format as it is on the JS side.
We will use then be able to use the same cucumber tests that we have in apollo-gateway to test the Rust implementation of the query planner.

The PR heavily uses the `serde` and `serde_json` crates, and their provided annotations to to achieve the format compatibility. I had to do some minor changes to the model to support this, nothing that really affects ergonomics (for context, the model code was implemented very recently and when I wrote it then I could have just as easily decided to model it like in this PR, but only happened to not do that).

The only real workaround that I had to add was to create a private alternative, single variant enum (`QueryPlanSerde`) for QueryPlan that will serialize with a `kind` field.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -85,6 +85,10 @@ dependencies = [
 [[package]]
 name = "apollo-query-planner"
 version = "0.0.1"
+dependencies = [
+ "serde 1.0.114",
+ "serde_json",
+]
 
 [[package]]
 name = "arrayref"
@@ -2231,9 +2235,9 @@ dependencies = [
 
 [[package]]
 name = "serde_json"
-version = "1.0.55"
+version = "1.0.56"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ec2c5d7e739bc07a3e73381a39d61fdb5f671c60c1df26a130690665803d8226"
+checksum = "3433e879a558dde8b5e8feb2a04899cf34fdde1fafb894687e52105fc1162ac3"
 dependencies = [
  "itoa",
  "ryu",
--- a/query-planner/Cargo.toml
+++ b/query-planner/Cargo.toml
@@ -8,3 +8,5 @@ edition = "2018"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+serde = { version = "1.0.114", features = ["derive"] }
+serde_json = "1.0.56"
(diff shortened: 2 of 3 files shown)


ashleygwilliams:
@Enrico2 could you elaborate on "The only real workaround that I had to add was to create a private alternative, single variant enum (QueryPlanSerde) for QueryPlan that will serialize with a kind field."? i don't know the domain well enough but i'm curious to know more

Enrico2:
@ashleygwilliams Sure! So, the JSON format of a query plan starts off like this:
```
{
  "kind": "QueryPlan",
  "node": ... // a single `PlanNode` enum, it has 4 enum variants. This can also be `undefined`
}
```

Our current Rust version of the `QueryPlan` model is 
```rust
struct QueryPlan(Option<PlanNode>)
```
I couldn't see a way to tell `serde_json` to serialize the latter into the former without manually implementing ser/de myself.

But, this hypothetical struct:
```
enum QueryPlan {
  QueryPlan { node: Option<PlanNode> }
}
```
can be annotated to tell `serde_json` to look like that JSON on top:
```
#[serde(rename_all = "PascalCase", tag = "kind")]
```
this tells `serde_json` to add a "tag" field, name that field "kind" and the value for that field is the name of the (only) variant. The rest of the fields in the resulting json are the fields from variant struct.

So my options were to either
1. Change the top level QueryPlan struct to a single-variant enum: We'd like our top level `QueryPlan` to be easy to handle in Rust code, so creating a single-variant enum that we'll use everywhere seems like the wrong call.
2. Implement Ser/De manually for QueryPlan: As far as I could tell, more complex than the alternative.
3. Use the workaround that I did in this PR where going to/from `serde_json::Value` from/to `QueryPlan` goes through an intermediate, private (to the module) struct that's compatible with the serialization format.

Writing this whole thing out made me realize 2 things:
1. `QueryPlan(Option<PlanNode>)` will not be as nice or ergonomic to use compared to `QueryPlan { node: Option<PlanNode> }`, maybe we should change that? I could add it on here as this PR already changes some parts of the model.

2. _when we need_ JSON compatibility with the JS side, we'll have to go through `into_json` or `from_json` that I added to the `QueryPlan` impl if we need to, and can't use `serde_json::to_value` directly because that'll result in a slightly different looking JSON.

FWIW I don't think we'll need compatibility outside of unit tests that use a shared serialized view.

PR #102, opened 2020-07-08
Add dependabot.yml for automatic updates. 

Enrico2:
See https://docs.github.com/en/github/administering-a-repository/enabling-and-disabling-version-updates

Config: https://docs.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+version: 2
+updates:
+  - package-ecosystem: cargo
+    directory: "/"
+    schedule:
+      interval: "daily"
+    reviewers:
+      - "apollographql/rustronauts"


abernix:
I enjoy automated dependency management and think this is a worthy consideration.

In my experience, I've found that its success largely depends on the extensiveness of the repository's testing suite to exercise the stability and expectations of our usage of those dependencies (to be clear: not re-testing the package, but testing the expectations of the integration with the package).  It does, to varying degrees, result in what appears to be successful upgrades which end up breaking something unexpectedly and undetected by the tests in place.

As this is a relatively new repository, and we're a relatively new team to Rust, I would lean on punting this for a few months.  I think when the time arrives that the components from this repository are actually in a deliverable-ish form where they are ready to be consumed by external, that it's prudent to have dependency management to make sure bug fixes and security fixes are applied.  (Though I think GitHub will still provide Security advisories regardless!)

Enrico2:
> As this is a relatively new repository, and we're a relatively new team to Rust, I would lean on punting this for a few months.

I don't quite understand that rationale, to me that's a reason to do the opposite and adopt a version updater sooner.

When we bring in new dependencies right now we just default to the most recent version _because_ we have just a little bit of code and less dependencies and use less APIs that might have issues when upgrading.
As I see it, the more we punt, the more likely it is that we lose track of version updates and end up encountering larger breaking changes.

Adding this bot is saving me from doing it manually, are you suggesting that we don't apply version updates at all?

ashleygwilliams:
i would generally agree with @abernix - unless we really trust our test suite (i'll be the first to say i don't currently), a tool like this can be more of a foot gun than a help. it is true that we currently have to manually update dependencies, but it is also not terribly urgent that we do so. if we were to hold off updating dependencies for a month or so it wouldn't be a huge deal. i think working on building out the code and our confidence in our test suite is a better priority, and i think we can eventually add a tool like this later on!

jbaxleyiii:
> As this is a relatively new repository, and we're a relatively new team to Rust, I would lean on punting this for a few months.

I'm like @Enrico2 and don't really understand this point but would love to better! I'm largely in favor of this getting merged through both as way to keep our deps up to date which I largely view as best practice and as way to help us find gaps in our young test suite!

Enrico2:
How do you see your concerns manifesting in a way that we'd not otherwise discover, except later? 
In other words, how would having a bot that does a manual thing for us (with our approval), going to cause an issue that will have otherwise be prevented? 

I generally prefer doing something, and undoing if it causes problems compared with doing nothing;
Learning involves experimentation, and that is about taking action, and learning from consequences (be them positive or negative). In our case (early repository, no published crates, very little code) an "bad" version update is going to have a tiny blast radius, if any, that we'll learn more from, and we can easily take back the decision and remove the bot.

I'd rather discover _early_ that a version update is breaking CI; it's true that our test coverage is not amazing, but it's not bad either. We have some tests, and we're going to add more as we add code (I'm assuming good test coverage is an ideal we all share for new PRs)

jbaxleyiii:
@Enrico2 I think you summed up my feelings on this well!

abernix:
I don't understand how updating our dependencies more often in an automated fashion would do anything to find gaps in a test suite that doesn't test those dependencies nuances.

I do not believe that automated dependency management, or even bumping our dependencies manually, is a critical part of this project at a stage like this and is more like to create subtle problems that we are more likely to get hung up on.  Put another way, if we're building something today I don't see the need to bump dependencies ruthlessly as new versions come out and surprise us during our progress, particularly as we're just learning — and less equipped — how to debug subtle surprises.

abernix:
In my experience, the automatic bumping of transitive dependencies by Renovate has only allowed surprise changes that were not documented in CHANGELOGs or were semantically larger breaking changes than claimed to slide in.  Not automatically updating dependencies can actually maintain stability.

This repository should already get security updates through other means and running a dependency once every six months is, I think, not unreasonable.

The automatic updates also create extra PR noise.  I'm not saying we should not turn this on eventually, I just don't see why this is a priority this month.  We can turn this bot on in the future with the same number of lines of code.

Enrico2:
> I don't see the need to bump dependencies ruthlessly as new versions come out and surprise us during our progress, particularly as we're just learning — and less equipped — how to debug subtle surprises.

Great question, this is exactly what I _want_ to happen; The best way for me to get more equipped to debug subtle surprises, and gain more familiarity with the crates we depend on, is to face these issues, and creating a situation in which I can face them as they happen makes them more sparse. Postponing version updates increases the likelihood of facing too many subtle surprises all at once, to a point where it's more likely that priorities will cause me to force an older version and accrue tech debt.

Version updates, especially the kind that the bot updates (minor updates) are more often than not bug fixes, ones that we may have not encountered, or ones that we may not even know we suffer from. Updating also means _decreasing_ the likelihood of subtle surprises.

Enrico2:
Closing in favor of #118

abernix:
Glad we landed this in #118.  Thanks for pushing it through when the timing was more mutually agreed upon.

PR #105, opened 2020-07-13
some nits

ashleygwilliams:
nit PRs can be a pain! it's my goal that these are seen as simplifying and clarifying and not matters of personal style. i intend to file several separate so that folks can see the general patterns.

### DIFF
--- a/cli/build.rs
+++ b/cli/build.rs
@@ -1,10 +1,9 @@
-use std::fs::{read, write};
-use std::path::PathBuf;
-use uuid::Uuid;
-
-use reqwest::blocking::Client;
 use std::env;
 use std::fs;
+use std::path::PathBuf;
+
+use reqwest::blocking::Client;
+use uuid::Uuid;
 
 /// This script downloads the schema if it's not in the file system (in a path that's .gitignored)
 /// If it exists, the script does nothing.
@@ -26,7 +25,8 @@ fn main() -> std::io::Result<()> {
     // and most of the time it doesn't do anything.
     println!("cargo:rerun-if-changed=.schema/last_run.uuid");
     fs::create_dir_all(".schema").expect("failed creating the .schema directory");
-    write(".schema/last_run.uuid", Uuid::new_v4().to_string()).expect("Failed writing uuid file");
+    fs::write(".schema/last_run.uuid", Uuid::new_v4().to_string())
+        .expect("Failed writing uuid file");
 
     let schema_url = env::var("APOLLO_GPAPHQL_SCHEMA_URL")
         .unwrap_or_else(|_| "https://engine-graphql.apollographql.com/api/schema".to_owned());
@@ -51,7 +51,7 @@ fn main() -> std::io::Result<()> {
             update_schema(&client, schema_url_str)
         } else {
             println!(".schema/etag.id exists");
-            let curr_etag = String::from_utf8(read(etag_path).unwrap()).unwrap();
+            let curr_etag = String::from_utf8(fs::read(etag_path).unwrap()).unwrap();
             println!("curr etag: {}", curr_etag);
 
             let response = client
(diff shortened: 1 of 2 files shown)


cli/build.rs
```diff
@@ -26,10 +25,11 @@ fn main() -> std::io::Result<()> {
     // and most of the time it doesn't do anything.
     println!("cargo:rerun-if-changed=.schema/last_run.uuid");
     fs::create_dir_all(".schema").expect("failed creating the .schema directory");
-    write(".schema/last_run.uuid", Uuid::new_v4().to_string()).expect("Failed writing uuid file");
+    fs::write(".schema/last_run.uuid", Uuid::new_v4().to_string())
+        .expect("Failed writing uuid file");
 
     let schema_url = env::var("APOLLO_GPAPHQL_SCHEMA_URL")
-        .unwrap_or_else(|_| "https://engine-graphql.apollographql.com/api/schema".to_owned());
+        .unwrap_or_else(|_| String::from("https://engine-graphql.apollographql.com/api/schema"));
```
Enrico2, line 32: FWIW also scatter `to_string()` on `&str` here and there, which literally calls `String::from`. Personally I don't love that the common way is so many characters, and that it puts the actual string (what the reader is probably interested in) later, but no strong objections or anything.

Can you elaborate on this preference?

cli/build.rs
```diff
@@ -40,9 +40,7 @@ fn main() -> std::io::Result<()> {
 
     println!(
         "UPDATE_SCHEMA = {}",
-        env::var("UPDATE_SCHEMA")
-            .unwrap_or_else(|_| String::from("null"))
-            .as_str()
+        &env::var("UPDATE_SCHEMA").unwrap_or_else(|_| String::from("null"))
```
Enrico2, line 43: I'm still confused on `&String` vs. `&str` -- I know the latter is a slice, and I understand the difference between the two, but I'm not yet sure what I should be using one or the other when _I_ declare the type.

cli/build.rs
```diff
@@ -1,10 +1,9 @@
-use std::fs::{read, write};
+use std::env;
```
Enrico2, line 1: Is there some convention for import _line grouping_ [*]?
`cargo fmt` sorts, which is great, but I'm not sure if there's prior opinion or work on choosing to group lines together (like here).

Same Q for using `foo.{x, y}` vs `foo.*` vs.
```rust
use foo.x;
use foo.y;
```

Ideally if there's a config file we can write to tell `cargo fmt` what to do about these cases, that'd be great.

cli/build.rs
ashleygwilliams: so this is not a pref so much as when considering the file as a whole, there was already a pattern of calling `String::from`

ashleygwilliams: i don't really have a personal pref, i kinda dig having the type information up front, don't really care about character count? but as stated, i chose `String::from` here because the doc already had a predominant pattern of using it

i do have a pref for `to_string` over `to_owned` ... preferring `to_owned` in truly generic situations, and `to_string` when we know we want a string

cli/build.rs
```diff
@@ -61,7 +61,7 @@ fn main() -> std::io::Result<()> {
             println!("remote etag: {}", remote_etag.unwrap_or("None"));
 
             match remote_etag {
-                Some(etag) if etag == &curr_etag => {
+                etag if etag == Some(&curr_etag) => {
```
Enrico2: At this point just use an `if`, there's no de-structuring anymore.

cli/build.rs
ashleygwilliams: the main changes i've made here are to group the std imports and to not import `read` and `write` into the namespace, preferring to have to reference them with the `fs` prefix because it improves readability because `read` and `write` are common function names

it may be possible to configure but whenever possible i like to keep the cargo fmt defaults and not grow them further, opting to only enforce so much as it impacts readability

cli/build.rs
ashleygwilliams: in many cases `&String` can act as `&str` because it impls `Deref<Target = str>` - in these situations, the `&` feels less obtrusive to what is happening in the logic

PR #106, opened 2020-07-14
Implement query planner tests

### LINKED ISSUE #103: Implement query-planner tests (i.e., Cradle) into this repository
With the landing of https://github.com/apollographql/apollo-server/pull/4297 in the `apollo-server` repository, we now have the "`.feature` files" which are the beginnings of what we are calling _Cradle_. 

These test files now act as compatibility acceptance tests for the current JavaScript query planner.  The `.feature` format is meant to be language-agnostic and introducing them to this repository should act as the basis for implementing the port of [our current `buildQueryPlan`](https://github.com/apollographql/apollo-server/blob/5a7fd856f6cbb1efe0ea91797cb81b9492bfe380/packages/apollo-gateway/src/buildQueryPlan.ts) into (I think) the `query-planner` module in this repository.

We imagine _Cradle_ growing into something more complete in the future (along with finding an agnostic home for them to match our intention for them to stand on their own), but that work will be tracked in another issue.  This issue is meant to track bringing those tests into this repository!

I believe the deliverables here are:

1. Be able to iteratively develop on implementation of a Rust `build_query_plan` using the `.feature` files introduced in https://github.com/apollographql/apollo-server/pull/4297.
2. Ensure that the same tests are also executed within CI.
3. All tests to be demonstrated as being executed, but failing, as there is no implementation yet! (And the implementation is out of scope for this issue)
4. Wholesale disabling of these tests in order to allow them to be incrementally disabled as the implementation grows to support their use-cases.  This disabling might just be demonstrated as a separate commit.

Just to repeat: the very large caveat here is that all the tests will be failing. 😄 

Some (possible!) considerations worth discussing, either along the way to or during exploration to implementing it might be:

* Determine _where_ the `.feature` files and the tests should live in this repository

  _Likely no need research extensively — as I think moves would be somewhat effortless — but some parity with best-practices or patterns of other repos utilizing Cucumber tests might be worth considering._

* Confirming that it's _okay_ to merely *copy* (yes, roughly, `cp there/*.feature here/.`) the `.feature` files into this repository.

  _**Copy**, you say?!  My claim would be that, while I DO think we need to find a more permanent home for them, that they are not currently changing at a rapid-enough pace that copying them will be a problematic ergonomic.  With this query-plan-porting-work in-progress, the JS query planner is effectively feature-frozen in its current state.  I think the source of truth could remain in the `apollo-server` repository until either the point the Rust query planner is shipped OR we answer the question: "where is the permanent home for Cradle tests?" and migrate them there.  (Also, I will open a separate issue for this!)_

* How to ignore `.feature` files.

  _I don't know enough about the implementation that will be necessary to understand what that would look like, but I think it will be important to be able to _skip_ them or focus specific tests._

Possibly more! Feel free to discuss!

(cc @JakeDawkins who has been thinking about this)

JakeDawkins:
Resolves #103 

This PR takes the existing query planner test suite from the apollo-server repo (cradle) and sets up this project to run them whenever there is a runnable query planner. 

The signature of the query planner that I've stubbed out here is from this discussion https://github.com/apollographql/rust/issues/101#issuecomment-656356447

Right now, you can run this suite by running `cargo test` or `cargo test -p apollo-query-planner --test cucumber`.

Deliverables from #103:

- [x] Be able to iteratively develop on implementation of a Rust build_query_plan using the .feature files introduced in apollographql/apollo-server#4297.
- [x] Ensure that the same tests are also executed within CI.
- [x] All tests to be demonstrated as being executed, but failing, as there is no implementation yet! (And the implementation is out of scope for this issue)
- [x] Wholesale disabling of these tests in order to allow them to be incrementally disabled as the implementation grows to support their use-cases. This disabling might just be demonstrated as a separate commit.
  > this is possible by commenting out the scenarios in the `.feature` file, but I haven't found a way to skip integration tests yet like you can with regular unit tests using `#[ignore]`

### DIFF
--- a/query-planner/Cargo.toml
+++ b/query-planner/Cargo.toml
@@ -15,6 +15,7 @@ serde_json = "1.0.56"
 [dev-dependencies]
 cucumber = { package = "cucumber_rust", version = "^0.6.0" } 
 
+# Having this commented out is preventing the cucumber tests from running, but still compiles them. Uncomment to run them.
 # [[test]]
 # name = "cucumber"
 # harness = false # Allows Cucumber to print output instead of libtest
--- a/query-planner/Cargo.toml
+++ b/query-planner/Cargo.toml
@@ -15,6 +15,6 @@ serde_json = "1.0.56"
 [dev-dependencies]
 cucumber = { package = "cucumber_rust", version = "^0.6.0" } 
 
-[[test]]
-name = "cucumber"
-harness = false # Allows Cucumber to print output instead of libtest
+# [[test]]
+# name = "cucumber"
+# harness = false # Allows Cucumber to print output instead of libtest
--- a/query-planner/Cargo.toml
+++ b/query-planner/Cargo.toml
@@ -10,3 +10,10 @@ edition = "2018"
 [dependencies]
 serde = { version = "1.0.114", features = ["derive"] }
 serde_json = "1.0.56"
+
+[dev-dependencies]
+cucumber = { package = "cucumber_rust", version = "^0.6.0" } 
+
+[[test]]
+name = "cucumber"
+harness = false # Allows Cucumber to print output instead of libtest
--- a/query-planner/Cargo.toml
+++ b/query-planner/Cargo.toml
@@ -8,6 +8,7 @@ edition = "2018"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+graphql-parser = { path = "../graphql-parser" }
 serde = { version = "1.0.114", features = ["derive"] }
 serde_json = "1.0.56"
 
--- a/query-planner/tests/cucumber.rs
+++ b/query-planner/tests/cucumber.rs
@@ -67,7 +67,7 @@ mod query_planner_tests {
                 Some(query_string) => {
                     // store the query on context so other steps can acces it
                     context.query = Some(query_string.clone());
-                }, 
+                },
                 None => panic!("no argument to query step")
             }
         };
@@ -77,12 +77,12 @@ mod query_planner_tests {
                 Some(query) => {
                     // todo use a schema here rather than this stub
                     let built_plan = build_query_plan_from_str("type Query { hello: String }", query.as_ref());
-                    
+
                     match &step.docstring {
                         // todo: deserialize from string to query plan
                         Some(_expected_plan_string) => {
                             assert_eq!(built_plan, QueryPlan { foo: "okay then".to_string() });
-                        }, 
+                        },
                         None => panic!("no argument to query plan step")
                     }
                 },
--- a/query-planner/tests/cucumber.rs
+++ b/query-planner/tests/cucumber.rs
@@ -92,18 +92,11 @@ mod query_planner_tests {
     });
 }
 
+// TODO: how to ignore
 cucumber! {
-    // path relative to cargo.lock
     features: "./", // Path to our feature files
     world: crate::MyWorld, // The world needs to be the same for steps and the main cucumber call
     steps: &[
         query_planner_tests::steps // the `steps!` macro creates a `steps` function in a module
     ]
-    // setup: setup, // Optional; called once before everything
-    // before: &[
-    //     a_before_fn // Optional; called before each scenario
-    // ],
-    // after: &[
-    //     an_after_fn // Optional; called after each scenario
-    // ]
 }
--- a/query-planner/tests/cucumber.rs
+++ b/query-planner/tests/cucumber.rs
@@ -1,46 +1,93 @@
 use cucumber::cucumber;
+use std::fmt;
+
+type Result<T> = std::result::Result<T, QueryPlanError>;
+
+// stub out a custom QueryPlanError
+#[derive(Debug, Clone)]
+struct QueryPlanError;
+impl fmt::Display for QueryPlanError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "error in building query plan")
+    }
+}
+
+struct Schema {}
+fn parse_schema(_schema: &str) -> Result<Schema> {
+    Ok(Schema {})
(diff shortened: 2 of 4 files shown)


query-planner/tests/cucumber.rs
```diff
@@ -0,0 +1,102 @@
...
+// stub out a custom QueryPlanError
+#[derive(Debug, Clone)]
+struct QueryPlanError;
+impl fmt::Display for QueryPlanError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "error in building query plan")
+    }
+}
+
+struct Schema {}
```
Enrico2, line 15: `graphql_parser::schema::Document`

query-planner/tests/cucumber.rs
```diff
@@ -0,0 +1,102 @@
...
+fn parse_schema(_schema: &str) -> Result<Schema> {
+    Ok(Schema {})
+}
+
+struct Query {}
+fn parse_query(_query: &str) -> Result<Query> {
+    Ok(Query {})
+}
+#[derive(Debug)]
```
Enrico2, line 24: `graphql_parser::query::Document`

query-planner/tests/cucumber.rs
```diff
@@ -0,0 +1,102 @@
...
+struct QueryPlan {
```
Enrico2, line 25: `crate::model::QueryPlan` (or maybe it has to be `apollo-query-planner::`, not sure)

query-planner/tests/cucumber.rs
```diff
@@ -0,0 +1,102 @@
...
+    foo: String,
+}
+impl std::cmp::PartialEq for QueryPlan {
```
Enrico2, line 28: `QueryPlan` already implements this trait, and fwiw it's derivable

query-planner/tests/cucumber.rs
```diff
@@ -0,0 +1,102 @@
...
+#[derive(Debug, Clone)]
+struct QueryPlanError;
+impl fmt::Display for QueryPlanError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "error in building query plan")
+    }
+}
+
+struct Schema {}
+fn parse_schema(_schema: &str) -> Result<Schema> {
```
Enrico2, line 16: `use graphql_parser::{parse_query, parse_schema};`

query-planner/tests/cucumber.rs
```diff
@@ -0,0 +1,102 @@
...
+}
+impl cucumber::World for QueryPlannerTestContext {}
+impl std::default::Default for QueryPlannerTestContext {
+    fn default() -> QueryPlannerTestContext {
+        // This function is called every time a new scenario is started
+        QueryPlannerTestContext { query: None }
+    }
+}
+
+mod query_planner_tests {
```
Enrico2, line 59: Is this `mod` necessary for cucumber? This code is already in an integration test module named `cucumber`.

query-planner/tests/cucumber.rs
```diff
@@ -0,0 +1,102 @@
...
+                        None => panic!("no argument to query plan step")
+                    }
+                },
+                None => std::unreachable!()
+            }
+        };
+    });
+}
+
+// TODO: how to ignore
```
Enrico2, line 95: One of the options is to conditionally compile the entire module:
```rust
#[cfg(feature = "with_qp_tests")]
mod cucumber
```

in `mod.rs`

(idk if mod.rs is a thing in a tests folder, if not we can put it on `mod query_planner_tests`

query-planner/tests/cucumber.rs
```diff
@@ -0,0 +1,102 @@
+use cucumber::cucumber;
+use std::fmt;
+
+type Result<T> = std::result::Result<T, QueryPlanError>;
+
+// stub out a custom QueryPlanError
+#[derive(Debug, Clone)]
+struct QueryPlanError;
```
Enrico2, line 8: This will come from my current branch, It's not actually used anywhere right now, right?

Should I put up a PR for some no-op code that only contains the signatures and this error struct so that you could reference them directly?

query-planner/tests/cucumber.rs
Enrico2: oops this was on the wrong line, I meant it on `struct Schema`

query-planner/tests/cucumber.rs
JakeDawkins: I'm okay with leaving this for now, since we don't use that error really. we can just use the one on your branch when it's ready

query-planner/tests/cucumber.rs
JakeDawkins: As we saw yesterday, removing the `[[test]]` section of the cargo.toml will still compile, but not run the cucumber tests, which seems like a decent middle-ground for now

query-planner/tests/cucumber.rs
JakeDawkins: I think so, yes 🙃

JakeDawkins:
I opened https://github.com/bbqsrc/cucumber-rust/issues/56 to see if there's any way to trim down or remove the full printed output from these tests, since I'd rather not print the _full_ content of the test for every scenario 😞

query-planner/Cargo.toml
```diff
@@ -8,5 +8,13 @@ edition = "2018"
...
 
 [dependencies]
+graphql-parser = { path = "../graphql-parser" }
 serde = { version = "1.0.114", features = ["derive"] }
 serde_json = "1.0.56"
+
+[dev-dependencies]
+cucumber = { package = "cucumber_rust", version = "^0.6.0" } 
+
+# [[test]]
```
Enrico2, line 18:
```suggestion
# Having this commented out is preventing the cucumber tests from running, but still compiles them. Uncomment to run them.
# [[test]]
```

Issue #112, opened 2020-08-05
Implement a minified printer for `[query|schema]::ast`


Enrico2:
When serializing a query plan into a json string, the operation field for each fetch node is the graphql document to run on the service, we print the minified operation there, but our current printer doesn't support it. There's a janky beginning of an implementation of this in https://github.com/apollographql/rust/blob/ran/build_query_plan/graphql-parser/src/format.rs

Enrico2:
We've done this for `query::ast` in #113 but we should re-implement it from scratch I think. Let's keep this issue around for that re-implementation.

PR #125, opened 2020-08-28
Implement a minified printer for `query::ast`

Enrico2:
This PR implements query minified string from scratch instead of the temporary solution of hacking it into the existing graphql-parser printer.

The PR also restores the code in the various `format.rs` prior to #113

This PR doesn't quite cl_ose #112 because it only implements it for queries.


Enrico2:
In terms of how to read this: `minified.rs` is all of the new stuff, the rest of the changes are almost a 1:1 restoration of the code prior to #113 + the `*Ref` implementations for `Displayable`.

PR #126, opened 2020-08-28
Fix query with `__schema` introspection case

Enrico2:
## Bugfix

It should be allowed to invoke such a query against a query planner. The planner's behavior is to skip those fields.
This was not working prior because the .ts implementation's graphql schema can fetch the field definition for introspection fields.

## Tests improvement
This PR sets up an easier way to add test cases that have different schemas; instead of a single `features` directory with `csdl.graphql` in it, we now have a `basic` subdirectory with the previous set of tests in it. If anyone wants to add test cases with a different schema, all one has to do is create another subdirectory and name the schema in it `csdl.graphql`.


Enrico2:
Reviewed by @trevor-scheer during pairing, merging.

PR #135, opened 2020-09-08
[Spike] Query Executor in Rust

jbaxleyiii:
spike

## What this is
This is a spike of a porting the JavaScript [query executor](https://github.com/apollographql/apollo-server/blob/60d974240769aac3e1c75ff968657b2d3e60e9b6/packages/apollo-gateway/src/executeQueryPlan.ts) which takes the query plan generated by the new Rust query planner and does a simple http fetch to the federated service and merges the data back together. This is **not something we should intend to merge**, rather it was a spike to see where things will be tricky as we work to do the port genuinely and where we could slice things for smaller deliverables.

## What do I do with this?
What I would to get out of this exercise is a shared understanding of some of the footguns which may lie ahead, what a common data structure needs to look like for the executor, and some of the challenges (and opportunities) in moving from a single threaded environment to a multi-threaded one. So

1. Don't review this for code quality / idiomatic Rust code. Its not that I'm not looking to improve my Rust, but rather this is a sketch of an idea not a real thing.
2.  Do read through the [main executor](http/src/main.rs). Below is a more detailed guide of what is interesting
3. Do ask questions about the data structures, general shape of the problem.
4. Don't focus on the http parts (tide, parsing of the response, etc). This was just for local testing.

## Detailed Review Notes

- The JS query executor relies heavily on two things to be successful. First, the single threaded nature of JavaScript and second, the ease of interior mutability within that thread. In particular, the `FlattenNode` and the `FetchNode` both mutate the result sink in a way that works if there is only one thread and a shared reference to the sink. For example, [flattenResultsAtPath](https://github.com/apollographql/apollo-server/blob/60d974240769aac3e1c75ff968657b2d3e60e9b6/packages/apollo-gateway/src/executeQueryPlan.ts#L165) mutates the results while creating a zip for subsequent nodes to work on. This is really hard / not what we should do in a world where you have multiple threads and interior mutability isn't really the way to go.
- `RwLock` is incredible. In the executor we need to be able to read from a shared reference at multiple times without block access to the data but block on writes. This is important because prior to sending a fetch we need to read the results to generate variables for downstream requests and we don't want to block on this for every request that is being formed. However, we do need to block on writes so that reads get the updated data correctly. A `Mutex` blocks on both but `RwLock` gets us just what we wanted. You can see where we [create the lock](https://github.com/apollographql/federation/compare/main...jbaxleyiii/gateway-spike#diff-1d121b833f22f1fd43615c8db9fa578bR78) and an example where we [write back to it](https://github.com/apollographql/federation/compare/main...jbaxleyiii/gateway-spike#diff-1d121b833f22f1fd43615c8db9fa578bR293)
- deep merging of json isn't easy. I found a library but am not 100% sure it actually does what we want. It may or may not be why  parallel flatten nodes [don't work](https://github.com/apollographql/federation/compare/main...jbaxleyiii/gateway-spike#diff-1d121b833f22f1fd43615c8db9fa578bR170-R175)
- Instead of manually forking threads, I am relying on [futures to manage promises](https://github.com/apollographql/federation/compare/main...jbaxleyiii/gateway-spike#diff-1d121b833f22f1fd43615c8db9fa578bR109-R117) for me. I'm not sure if this is actually doing what I want it to do but it behaves like `Promise.allSettled` which is similar to the existing one. Maybe we should go more all in on threads here, not sure.
- We are going to need to do [some cloning](https://github.com/apollographql/federation/compare/main...jbaxleyiii/gateway-spike#diff-1d121b833f22f1fd43615c8db9fa578bR215-R227) when it comes time to send requests. Mainly of the variables because these could be used across multiple requests to the executeNode can't take ownership of them. I don't think this is a problem but wanted to flag it here for discussion.
- [Some idioms in JS are just not possible in Rust.](https://github.com/apollographql/apollo-server/blob/main/packages/apollo-gateway/src/executeQueryPlan.ts#L206). This is on the whole a good thing and it has already helped to improve this implementation but don't count on a direct port being the same shape in each part. 
- We will need to figure out lifetimes / ownership of the schema within threads so it can be created outside of a request. I [didn't worry about that right now](https://github.com/apollographql/federation/compare/main...jbaxleyiii/gateway-spike#diff-1d121b833f22f1fd43615c8db9fa578bR24-R50) but that is definitely something we will need to do

## Suggested next steps
I think after an initial read through, a pairing session could be a helpful way to talk through this. Overall I'm quite thrilled at how quickly this came together and it makes me even more excited about when we tackle this for real. Let me know if you have any questions! 👍 

## Additional changes
#186 was merged into this one. Its description:

This PR was developed on top of #135

First and foremost, any changes to the existing _executor.rs_ and related code written in #135 are mostly there as **suggestions**. @jbaxleyiii lmk if you want to pair on these or feel free to take / reject parts. I didn't dig too deep into the implementation of the executor yet, I'm going to start doing that this week.

Second: This PR separates out the Stargate library from the binary, for a few reasons:
1. Separating out the business logic from the wrapping server makes for easier testing.
2. The library doesn't choose an async runtime, nor does it assume any specific implementation of an HTTP server.
3. Use actix-web instead of tide; we've discussed these options and feel more comfortable going with a more mature server implementation, as well as one that uses tokio.

@jbaxleyiii If these changes look good, we can merge them into #135 and we can continue from there. If you want to do it differently, lmk!

- [ ] Address `todo!` / `unimplemented!` / `FIXME` in the code
- [ ] Tests for the executor (ideally these should be unit tests on the Stargate library, for coverage)
- [ ] Integration tests. I was thinking we can set up acephei services locally and then the gateway would have `localhost` references to those on different ports, integration tests would be entirely localhost.
- [ ] Add tracing/metrics middleware -- I believe there are some existing crates.
- [ ] Write more log lines so that they're integrated with the logger of the http server (use `env_logger` ?)


http/src/main.rs
```diff
@@ -0,0 +1,515 @@
...
+use graphql_parser::schema;
+
+#[async_std::main]
+async fn main() -> std::result::Result<(), Box<dyn std::error::Error + Send + Sync>> {
+    tide::log::start();
+    let mut app = tide::new();
+
+    // Start server
+    app.at("/").post(|req: Request<()>| async move {
+        let request_context = req.body_graphql().await?;
```
EverlastingBugstopper, line 21: :O

Enrico2:
@trevor-scheer @abernix @jbaxleyiii That last commit should bring this branch to a mergeable state without breaking the release process. I've created an issue for follow ups (#198)

Enrico2 (approved):

stargate/Dockerfile
```diff
@@ -5,8 +5,7 @@
 ARG BASE_IMAGE=ekidd/rust-musl-builder:latest
 FROM alpine/git:latest AS cloner
 
-RUN git clone https://github.com/apollographql/federation.git /src && \
-  git -C /src checkout jbaxleyiii/gateway-spike
+RUN git clone https://github.com/apollographql/federation.git /src
 
 # Our first FROM statement declares the build environment.
 FROM ${BASE_IMAGE} AS builder
```
abernix, line 11: This only existed as a workaround for the time-being while we were based on this branch.   This is meant to be replaced with a [`COPY ../ /src`](https://docs.docker.com/engine/reference/builder/#copy) (inside the `builder` stage) so that this can be built from CI _from the current commit_.

stargate/Dockerfile
```diff
@@ -23,9 +22,9 @@ FROM alpine:latest
...
 RUN apk --no-cache add ca-certificates
 COPY --from=builder \
-  /home/rust/src/target/x86_64-unknown-linux-musl/release/stargate \
+  /home/rust/src/target/x86_64-unknown-linux-musl/release/apollo-stargate \
   /usr/local/bin/
 
 # User can provide an alternate port via "-e 'PORT=<your port>'" when running the container.
 # PORT defaults to 8080 otherwise.
-CMD /usr/local/bin/stargate --address=http://0.0.0.0:${PORT:-8080} --manifest=/graph
+CMD /usr/local/bin/apollo-stargate --port=${PORT:-8080} --manifest=${CSDL:-"/graph.graphql"}
```
abernix, line 30: There's a specific reason `PORT` is a variable here and that the internal CSDL path to the container was not, since it's nearly the de-facto standard for configuring the listening/bind port for many Docker containers (common practice).

I don't think that customizing the path where the CSDL is mounted in the container is a necessary feature; the user shouldn't need to change the expected locations of things inside this runtime image, and they can configure the volume mount with `-v /local/path/to/their/csdl.graphql:/graph` when running the image (or the appropriate instruction in their container manifest/specification).

I don't particularly care where the graph is mounted _inside_ (including whether it has the `.graphql` extension or not), but I think it should be our implementation detail.

abernix (commented):
This feedback is meant for @Enrico2, who has taken over this PR, and only about one small part that I had my attention called to in https://github.com/apollographql/federation/pull/135#issuecomment-701706929

stargate/Dockerfile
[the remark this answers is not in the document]
Enrico2: Well, so this is an interesting bit. Are you asking that we change this so that the user facing binary we release is just `stargate`?
This is getting the name from the crate name, and at some point we said that we want to prefix those with `apollo-`.

If we release Stargate as a binary, we can rename the file to whatever we want as part of the download / extraction code.

I think that we might want to, at some point, publish the binary to crates.io, at which point the crate name will be the "user facing" name, and rust folks could install it with `cargo install stargate` or `cargo install apollo-stargate` depending on what we end up with.

I see that `stargate` is available, so maybe we should actually go ahead and squat on that name..

Gonna change it!

stargate/Dockerfile
Enrico2: My Dockerfile foo isn't great to make sure I'm doing the right thing, I see what you're saying though, I hope it's ok but I'm going to merge this and maybe you can fix this later / we can do it tomorrow together?

stargate/Dockerfile
Enrico2: Makes sense! I'll fix.

stargate/Dockerfile
Enrico2: Hey @abernix , @trevor-scheer and I tried to address this comment, but couldn't quite figure it out quickly, we can't `COPY ../ .` (`/src` would be wrong) because it's a parent directory.
We tried some variations but failed at various stages. I had to stop for lunch and then do other more pressing stuff :)

abernix: @Enrico2 @trevor-scheer  All good.  I can handle this!

http/src/main.rs
avidal: The quality contributions Avery is known for. Keep up the good work!

PR #157, opened 2020-09-15
Auto Fragmentization in the Rust Query Planner

Enrico2:
This PR implements the Auto Fragmentization feature in the Rust Query Planner.

Be sure to "ignore whitespace" when reviewing.

## Process

Prior to this PR, we have copied over the scaffolding from the .ts implementation of the query planner to implement this feature later. In the .ts implementation, we discover and create "internal fragments" as we work through the query planner. However, as I started working at this problem, I decided that separating auto-frag to a processing phase that happens separately makes for simpler, more extensible code, that will allow us to iterate on this specific feature in a decoupled way from the rest of the query planner.

The main query planning process creates `FetchNode`s, each of which has an `operation` field that contains the literal string that will be sent to the downstream process. This is where we want to use queries that contain fragments we've automatically discovered.
When a query is planned with auto-frag enabled (a new flag to the `plan` function in `QueryPlanner`), right before we serialize the `operation` for the `FetchNode`, we _move_ the selection set through an auto fragmentization pass.

The process recursively iterates through the selection set, where for each selection set that has more than 2 fields (see parity section below) we first check if we've seen this selection set before (see hashing section below). If we create a fragment if there is none, and replace the selection set with one that only has one selection, the fragment spread for the fragment that we just found/created.

### Parity, not optimization

This implementation is not very intelligent, and the resulting query highly depends on the incoming query. For example, we may create many fragments that are only used once or we may not create fragments for actually really complex selection sets, e.g. selection sets with 2 selections whose sub-trees are deeply nested selection sets that repeat often.

The goal for this initial implementation is to reach parity with the .ts implementation so that we could completely replace the .ts implementation with an invocation of this one via wasm, without affecting users who are relying on the existing implementation.

Down the line, we can work on various optimizations for this code, as the need arises, or accept contributions to it from the community. (e.g. detecting fragments in subsets of selection sets? :)).

### Hashing

In order to identify if we've seen a selection set before, to use as a key in a map, we need the ability to hash a selection set based on _content_. In order to get this hash value we had to do 2 things:

1. Limit the `f64` values to `NotNan` [using this crate](https://docs.rs/ordered-float/2.0.0/ordered_float/struct.NotNan.html). Rust cannot derive a `Hash` for `f64` because of NaN values. This is a whole beast you can read about [here](https://internals.rust-lang.org/t/f32-f64-should-implement-hash/5436). Looking at the GraphQL Spec, a Float literal is defined as
```
FloatValue :: 
  IntegerPart FractionalPart
  IntegerPart ExponentPart
  IntegerPart FractionalPart ExponentPart
```
which as far as I can tell, from a _parsing_ perspective, any literal value cannot be a `NaN`, and thus, hashable. graphql-js doesn't allow this as well. [See this issue](https://github.com/graphql/graphql-spec/issues/778) by @glasser 

2. In order for the hash values to be based on _content_, the `Hash` implementations have to skip any field that references a _position_ for the element in the original query. I used the `derivative` crate for this (good find @trevor-scheer !!) to derive `Hash` and skip `position` and `span` fields.

## Important note about testing

This code prints out the fragments in each operation in predictable order; sorted by fragment name.
The tests we have in the `.feature` files have different ordering. We only have test cases with 1 or 2 fragments, so I can't really say if this is always "reverse" order, in any case it most likely depends on insertion order in the .ts implementation. But because this implementation is different, we can't resolve that by using a LinkedHashMap like in other cases before.

And so I changed the relevant expected result in the .feature files to use the lexicographic ordering. This is the 2nd case where we'll have to adapt if we want to use the same feature files for different code bases. My recommendation would be to either make the other implementations' output have predictable ordering, or do the comparison in the test in a way that doesn't rely on ordering where it's not important.
In this PR's case, if we wanted to make the test agnostic to fragment ordering, it'd mean parsing the `operation` in a FetchNode and comparing the parsed object, while _ignoring_ any position/span fields (which means we can't use the derived `PartialEq` unless we change that too), given the complexity of that, I opted to change the .feature file instead. Lmk if this is a problem.

My one simple test case passed, but when enabling testing using the `.feature` files that have auto-frag, I saw some failures but had to go afk, I will debug these before this PR is opened as not a draft.

## Integration with Javascript

This PR doesn't change the API used by the wasm module just yet. I figured we'd do that in a separate PR that involves a minor version update to the wasm crate.

## Separate question on versions

This PR changes the `.plan` function signature, which is public. Do we want to update the version in the same PR? And if so, do we not want all 3 versions (`query-planner`, `query-planner-wasm` and `query-planner-wasm`'s package.json) to use the same version? Right now they do not.

TODO:
- [x] fix tests
- [x] macroify similar code blocks in the auto-frag

### DIFF
--- a/query-planner/src/autofrag.rs
+++ b/query-planner/src/autofrag.rs
@@ -29,6 +29,20 @@ macro_rules! autofrag_field {
     }};
 }
 
+macro_rules! autofrag_inline {
+    ($inline:ident, $context:ident, $frags:ident, $counter:ident, $ss:expr, $else:expr) => {{
+        if let Some(tc) = $inline.type_condition {
+            let new_parent = $context.names_to_types[tc];
+            SelectionRef::InlineFragmentRef(inline_fragment_ref!(
+                $inline,
+                auto_frag_selection_set($context, $frags, $counter, new_parent, $ss)
+            ))
+        } else {
+            SelectionRef::InlineFragmentRef($else)
+        }
+    }};
+}
+
 pub(crate) fn auto_fragmentization<'q>(
     context: &'q QueryPlanningContext<'q>,
     selection_set: SelectionSetRef<'q>,
@@ -94,23 +108,14 @@ pub(crate) fn auto_fragmentization<'q>(
                     SelectionSetRef::from(&field.selection_set),
                     field_ref!(field)
                 ),
-                Selection::InlineFragment(inline) => {
-                    if let Some(tc) = inline.type_condition {
-                        let new_parent = context.names_to_types[tc];
-                        SelectionRef::InlineFragmentRef(inline_fragment_ref!(
-                            inline,
-                            auto_frag_selection_set(
-                                context,
-                                frags,
-                                counter,
-                                new_parent,
-                                SelectionSetRef::from(&inline.selection_set)
-                            )
-                        ))
-                    } else {
-                        SelectionRef::InlineFragmentRef(inline_fragment_ref!(inline))
-                    }
-                }
+                Selection::InlineFragment(inline) => autofrag_inline!(
+                    inline,
+                    context,
+                    frags,
+                    counter,
+                    SelectionSetRef::from(&inline.selection_set),
+                    inline_fragment_ref!(inline)
+                ),
                 Selection::FragmentSpread(_) => {
                     unreachable!("Fragment spreads is only used at the end of query planning")
                 }
@@ -133,23 +138,14 @@ pub(crate) fn auto_fragmentization<'q>(
                 field.selection_set,
                 field
             ),
-            SelectionRef::InlineFragmentRef(inline) => {
-                if let Some(tc) = inline.type_condition {
-                    let new_parent = context.names_to_types[tc];
-                    SelectionRef::InlineFragmentRef(inline_fragment_ref!(
-                        inline,
-                        auto_frag_selection_set(
-                            context,
-                            frags,
-                            counter,
-                            new_parent,
-                            inline.selection_set
-                        )
-                    ))
-                } else {
-                    SelectionRef::InlineFragmentRef(inline)
-                }
-            }
+            SelectionRef::InlineFragmentRef(inline) => autofrag_inline!(
+                inline,
+                context,
+                frags,
+                counter,
+                inline.selection_set,
+                inline
+            ),
             SelectionRef::FragmentSpreadRef(_) => {
                 unreachable!("Fragment spreads is only used at the end of query planning")
             }
--- a/query-planner/src/autofrag.rs
+++ b/query-planner/src/autofrag.rs
@@ -1,11 +1,15 @@
 use crate::builder::get_field_def_from_type;
 use crate::consts::{MUTATION_TYPE_NAME, QUERY_TYPE_NAME};
 use crate::context::QueryPlanningContext;
-use graphql_parser::query::refs::{FragmentDefinitionRef, SelectionRef, SelectionSetRef};
+use graphql_parser::query::refs::{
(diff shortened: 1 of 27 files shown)


query-planner/src/builder.rs
```diff
@@ -777,3 +765,19 @@ fn flat_wrap(kind: NodeCollectionKind, mut nodes: Vec<PlanNode>) -> PlanNode {
         }
     }
 }
+
+fn maybe_auto_fragmentization<'a, 'q: 'a>(
+    context: &'q QueryPlanningContext<'q>,
+    selection_set: SelectionSetRef<'q>,
+) -> (String, SelectionSetRef<'a>) {
```
trevor-scheer, line 772: Nit: just for consistency, can we do the printing/minification of both frags and selection set in the same place? So we'd either return `(String, String)` or (not sure if syntax is right, but...) `([FragmentDefinitionRef<'a>], SelectionSetRef<'a>)`

query-planner/src/lib.rs
```diff
@@ -35,9 +36,9 @@ impl<'s> QueryPlanner<'s> {
         QueryPlanner { schema }
     }
 
-    pub fn plan(&self, query: &str) -> Result<QueryPlan> {
+    pub fn plan(&self, query: &str, auto_fragmentization: bool) -> Result<QueryPlan> {
         let query = parse_query(query).expect("failed parsing query");
-        build_query_plan(&self.schema, &query)
+        build_query_plan(&self.schema, &query, auto_fragmentization)
```
trevor-scheer, line 41: I feel pretty ok about query and schema as positional parameters, but we may have other options / tweaks for query planning in the future. This might only be an appropriate JS suggestion (so please correct me if it is) but I'd recommend this 3rd arg be an options struct.

query-planner/src/macros.rs
```diff
@@ -1,4 +1,3 @@
-#[macro_export]
```
trevor-scheer, line 1: No issue with this (as I don't know any better), just would like to understand why we don't need it anymore.

trevor-scheer (approved):
LGTM! Going to take a 2nd pass over the core autofrag stuff so possibly a few more comments but nice work 🙂

query-planner/src/autofrag.rs
```diff
@@ -0,0 +1,305 @@
...
+use crate::context::QueryPlanningContext;
+use graphql_parser::query::refs::{
+    FragmentDefinitionRef, FragmentSpreadRef, SelectionRef, SelectionSetRef,
+};
+use graphql_parser::query::{Operation, Selection};
+use graphql_parser::schema::TypeDefinition;
+use graphql_parser::Name;
+use std::collections::hash_map::DefaultHasher;
+use std::collections::HashMap;
+use std::hash::{Hash, Hasher};
```
trevor-scheer, line 12: `Hasher` unused?
```suggestion
use std::hash::Hash;
```

query-planner/src/lib.rs
Enrico2: Oh sure, good call. We should probably make its construction a builder pattern with default value so that future changes are backwards compatible.

query-planner/src/macros.rs
Enrico2: IIUC it makes the macro public outside of the crate. I don't think we ever needed it. Either I misunderstood what it does, or maybe we had integration tests that use these macros? (integration tests are "external" to the crate)

query-planner/src/autofrag.rs
Enrico2: Importing `Hasher` allows us to use `s.finish()` in line 176.
For future reference, one of the lint checks we do with clippy in the github action will warn about unused imports.

trevor-scheer (approved):

Issue #143, opened 2020-09-10
[js-repo-migration] Migrate federation specification


abernix:
Relates to #142 and follows up on #134. 

Because the federation specification currently lives within the Apollo Server documentation, this task is _very likely_ to be covered by the work in #142, but just for completeness, we also want to move the federated specification over to this repository.

abernix:
Done!

Issue #142, opened 2020-09-10
[js-repo-migration] Migrate federation documentation


abernix:
Follows up #134 

The federation documentation lives on the web at https://www.apollographql.com/docs/apollo-server/federation/introduction/ and in GitHub at https://github.com/apollographql/apollo-server/tree/main/docs/source/federation.

Once the `@apollo/federation` and `@apollo/gateway` packages have moved here (which #134 accomplishes), we need to also move their documentation over here to keep the code and documentation relatively lock-step.

We should probably do a brief review of the existing information architecture during that move, but it might be as simple as moving the `/federation` subdirectory over here along with the appropriate `gatsby-config.js` and corresponding deploy changes.

### Next step

* Coordinate with our wonderful DX team to see if they have any input on this.

### Known action items

- [x] Move documentation here
- [x] Setup Netlify deployments
- [ ] Re-sync changes from `apollo-server`'s `docs/source/federation` directory
- [x] Mount Netlify deployment to `/docs/federation` using our [`website-router` (private)](https://github.com/apollographql/website-router) repository.
- [x] Setup redirects for existing `/docs/apollo-server/federation/` URLs using `apollo-server`'s [`docs/static/_redirects`](https://github.com/apollographql/apollo-server/blob/HEAD/docs/static/_redirects) file.
- [x] Algolia Docsearch adjustments (Branch started here: https://github.com/algolia/docsearch-configs/compare/master...abernix:patch-1)

abernix:
This was completed!

PR #158, opened 2020-09-16
docs: Bring/Setup documentation deployment for Federation from Server repo.

abernix:
This brings the existing documentation and its supporting infrastructure (in
this case, modeled after/copied from the Apollo Server repository, but this
is standard "Apollo infrastructure" in the general sense) to the Federation
repository where it can live alongside the code that was migrated over in
apollographql/federation#134.

### Next Steps 

> ⚠️ Each of these blocks the next, but neither of them blocks this PR from merging now as merging it would not result in the documentation becoming available in any linked-sense anywhere.

- [ ] Short-term solution to doc-set switcher (@StephenBarlow) (Tracked in https://github.com/apollographql/gatsby-theme-apollo/issues/147)
- [ ] Setup redirects from Apollo Server (@abernix) (Tracked in https://github.com/apollographql/apollo-server/issues/4565

Related: apollographql/federation#142
Related: apollographql/federation#143

(Both of these will close when the TODOs are completed).


abernix:
@trevorblades Although `apollo-server` [sets its Netlify `base` as `docs/` and its `publish` directory as `docs/public`](https://github.com/apollographql/apollo-server/blob/7d5cf189921fef29afe7a07a15cf7835a2c2161c/netlify.toml#L3), that precise configuration over here (with what I believe is 100% the same structure) resulted in an Error cannot find `/docs/docs/public`.

Therefore, I had to introduce https://github.com/apollographql/federation/pull/158/commits/f69888a9e625d19adbfc75a47003ae64ca616b36 to this PR to make it build. 

To be honest, having `publish` be relative to `base` makes a lot of sense to me — [and that's exactly what the Netlify documentation suggests is the case](https://docs.netlify.com/configure-builds/file-based-configuration/#sample-file) — so I'm actually more surprised that `apollo-server` is requiring that treatment and _happy_ that this makes sense, but it's odd that they're different.  Perhaps Apollo Server's setup is grandfathered into a mode that requires relativity to the `netlify.toml`?  Anyhow, just worth mentioning, and soliciting your perspective, not looking for anything to be changed necessarily (unless I'm doing something wrong!)

trevorblades:
@abernix looks good to me! Regarding the relativity of the `publish` property in `netlify.toml`, I think you're right:

> Perhaps Apollo Server's setup is grandfathered into a mode that requires relativity to the netlify.toml?

Tbh, reading the Netlify docs on this subject make me question how the other docs sites are working with their current configs, but I remember looking into this before, and coming to the same conclusion then. 🤷‍♂️

I'll coordinate with @StephenBarlow to get an update to the docset switcher into the docs theme 👍

trevorblades (approved):

abernix:
> Tbh, reading the Netlify docs on this subject make me question how the other docs sites are working with their current configs, but I remember looking into this before, and coming to the same conclusion then. 🤷‍♂️

@trevorblades Same.

> I'll coordinate with @StephenBarlow to get an update to the docset switcher into the docs theme 👍

Thank you so much, to you both, in advance!

PR #171, opened 2020-09-21
feat(gateway): support alternate query plan cache implementations

trevor-scheer:
This commit allows the user to specify their own cache to be used by the gateway for query plan caching. One intended use case, for example, is a distributed cache to be shared among a number of gateways.

For distributed cache implementations, you may find these packages useful!
* [Redis](https://github.com/apollographql/apollo-server/tree/master/packages/apollo-server-cache-redis)
* [Memcached](https://github.com/apollographql/apollo-server/tree/master/packages/apollo-server-cache-memcached)

A user provided cache must implement the `QueryPlanCache` interface, which can be seen [here](https://github.com/apollographql/apollo-server/pull/4026/files#diff-7b5e66976a637efc823826a985fa8e65R163-R167) within the PR for reference.

---
Originally at: apollographql/apollo-server#4026


trevor-scheer:
Note: no previous discussion to copy over.

PR #172, opened 2020-09-21
fix(query-planner): Build correct query plans when fragments are reused

trevor-scheer:
TODO: implement the fix and update PR description. This is currently only a reproduction.

Referenced issue: https://github.com/apollographql/apollo-server/issues/4063
cc @benjaminjkraft

---
Originally at: apollographql/apollo-server#4281


pcmanus:
This test has been included with the fix in PR #652, hence the closing.

PR #173, opened 2020-09-21
fix(query-planner): Gateway builds different query plan depending on fragmentation of incoming query

trevor-scheer:
TODO: implement the fix and update PR description. This PR currently only provides a reproduction for the bug as a jumping off point.

---
Originally at: apollographql/apollo-server#3830


dariuszkuc:
Closing old stale PR.

PR #174, opened 2020-09-21
FIX: Minifying a String argument should escape quotes and slashes

### LINKED ISSUE #170: Syntax Error: Expected \":\", found String \":\"." when sending as string with latest @apollo/gateway package
**Description:**
We are seeing the following error with respect to the latest @apollo/gateway npm package version `0.20.1` when sending the query input as string for example.

**Query**
```

query {

   someEvents(input: "{\"name\":\"john\",\"age\":\"22\",\"class\":\"mca\"}")

}

```
**ERROR**
We see following error:

```

Syntax Error: Expected \":\", found String \":\"." when sending as string

```

But the same works fine with the @apollo/gateway gateway version `0.19.1`. Not sure if something has been updated with the latest version.

Enrico2:
fixes #170
    
This was a regression with the new Rust query planner; the minified implementation has a bug where if a String argument contains either `\` character or a `"` character, they should be printed with escaping so that the eventual operation string is valid.
    
Test cases were added to ensure we don't regress. I've verified they fail without the fix.

Unrelated: Update license in Cargo.toml, this was wrong initially.

### DIFF
--- a/gateway-js/CHANGELOG.md
+++ b/gateway-js/CHANGELOG.md
@@ -4,7 +4,12 @@
 
 > The changes noted within this `vNEXT` section have not been released yet.  New PRs and commits which introduce changes should include an entry in this `vNEXT` section as part of their development.  When a release is being prepared, a new header will be (manually) created below and the appropriate changes within that release will be moved into the new section.
 
-- _Nothing yet! Stay tuned!_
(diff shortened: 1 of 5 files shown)


trevor-scheer (approved):
CHANGELOG entry and LGTM!

PR #175, opened 2020-09-21
Update version for query-planner-wasm

### LINKED ISSUE #170: Syntax Error: Expected \":\", found String \":\"." when sending as string with latest @apollo/gateway package

Enrico2:
This version update and subsequent release fixes #170 , a regression that was introduced with `0.20.1`

### DIFF
--- a/gateway-js/package.json
+++ b/gateway-js/package.json
@@ -26,7 +26,7 @@
   },
   "dependencies": {
     "@apollo/federation": "file:../federation-js",
-    "@apollo/query-planner-wasm": "0.0.3",
+    "@apollo/query-planner-wasm": "0.0.4",
     "@types/node-fetch": "2.5.4",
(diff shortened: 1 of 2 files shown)


Enrico2:
@trevor-scheer all those `package-lock.json` updates were a result of deleting it before running `npm install`, is that ok?

trevor-scheer (approved):

PR #178, opened 2020-09-22
Lost directives on inline fragment

### LINKED ISSUE #177: Directives which are located on INLINE FRAGMENT are skipped and are not sent to the service
I'm using gateway v0.20.1 and here is my schema:

```
directive @fragmentDirective on INLINE_FRAGMENT
directive @fieldDirective on FIELD

interface Human {
    name: String!
}

type User implements Human {
    name: String!
    email: String!
}

type Query {
    selectHuman:[Human]
}
```

After sending the following query, @fragmentDirective will be skipped and will not come to the service, the @fieldDirective is NOT skipped
```
{
 selectHuman{
    name
    ... on User @fragmentDirective {
      email @fieldDirective
    }
  }
}
```

After debugging I found that after parsing the request, the directive is present in the operation context, please see
directive is present in operation context
directive is present in seletion set

But it was lost after building query plan
directive is absent in query plan

I will appreciate any help or suggestions. 
Thank you!

Enrico2:
As reported in #177, we lose any directives on inline fragments from the original query

## Problem

We create new scopes either when we drill into object fields, inline fragments, or fragment spreads.

When we go into object fields, we collect the entire field node, so we don't lose directives.
We do however, lose directives when we create a new scope for inline fragments and fragment spreads.

## Solution for inline fragments

This PR fixes #177 by adding in any directives we find when collecting inline fragments and adding them to the scope we create for any fields in the fragment.

When we rebuild the operation that goes to downstream services, we wrap them in inline fragments if needed, if the fields we're wrapping have optional directives on the scope they're in, we add it to the inline fragment we're creating.

## Fragment spreads

As far as I can tell, we have the same problem with fragment spreads and directives on fragment definitions because the planner unfurls any fragments. I've opened #179 for this case.

### DIFF
--- a/query-planner/Cargo.toml
+++ b/query-planner/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "apollo-query-planner"
-version = "0.0.3"
+version = "0.0.4"
 authors = ["Apollo <opensource@apollographql.com>"]
 homepage = "https://github.com/apollographql/federation"
 edition = "2018"
--- a/query-planner/tests/features/basic/fragments.feature
+++ b/query-planner/tests/features/basic/fragments.feature
@@ -327,6 +327,11 @@ Scenario: supports directives on inline fragments (https://github.com/apollograp
     vehicle(id:"rav4") {
       ... on Car @fragmentDirective {
         price
+        thing {
+          ... on Ikea {
+            asile
+          }
+        }
       }
       ... on Van {
         price @fieldDirective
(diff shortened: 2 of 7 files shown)


query-planner/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "apollo-query-planner"
-version = "0.0.3"
+version = "0.0.4"
```
trevor-scheer, line 3: I think version changes should wait for their actual release. I wasn't sure about this until I thought about it this way: multiple feature additions still end up on the same minor, we only bump the minor once on release. The same could be said for bug fixes, though we'll likely release a patch for most bug fixes individually.

query-planner/tests/features/basic/fragments.feature
```diff
@@ -319,3 +319,27 @@ Scenario: supports root fragments
     }
   }
   """
+
+Scenario: supports directives on inline fragments (https://github.com/apollographql/federation/issues/177)
```
trevor-scheer, line 323: I really like tagging the test case with the related issue 👍

trevor-scheer (approved):

query-planner/Cargo.toml
Enrico2: I agree, in this case though, this is because I forgot updating this one yesterday (Assuming we want to keep query-planner-wasm and query-planner the same.

PR #181, opened 2020-09-23
Setup CircleCI/Lerna/npm to compile and publish `@apollo/query-planner-wasm`

abernix:
This allows us to compile and depend upon the local version of the WASM query planner package and get the tarballs that are packed with the right version.  Additionally, this allows Lerna to handle the version bumping and publishing for `@apollo/query-planner-wasm`, since it is uniquely positioned to do that best in a monorepo orchestration that involves multiple npm packages with relative `file:` dependencies on each other and coupled with our existing needs for Lerna and publishing workflows we have built.

Technically speaking, this relieves the `Cargo.toml` from its responsibility of managing the `version` and shifts that responsibility to `lerna` and other npm-specific tooling we have connected to that.  In fact, this change completely eliminates the need for the `wasm-pack` generated `package.json` that is derived from the Cargo.toml (and rendered into the "out-dir", previously `pkg/`).

While I would believe that there are some advanced `wasm-pack` cases where this above change might be undesirable, for now, the constraint incurred by that transposition is unblocking in other ways.  For example, rather than necessitating parallel metadata support in the `Cargo.toml` for every `package.json` property (think about npm specific properties like "labels", "bugs", "private"), we can simply change our `package.json` as we see fit.

I am claiming that this side-stepping of version control defined within `Cargo.toml` is acceptable since we do not have any intention of publishing the `query-planner-wasm` package to crates.io on its own.  Instead, we will continue to publish the `wasm-pack`'d-from-this-crate `@apollo/query-planner-wasm` package directly to npm's registry.  To reflect that desire, I've marked the `query-planner-wasm` package as "private" using the `private = true` property in its `Cargo.toml`.

In theory, this change could de-stabilize the work that `wasm-pack` does and that's worth calling out.  For example, by not allowing it to generate the `files` property (which indicates which artifacts are emitted into the `npm pack`'d bundle), we might be not allowing it to add other (Future?  Unexpected?) important emitted files.  I've instead put those `files` directly into the new `package.json` source of truth, along with other appropriate properties, like `types` and `main`.  The largest risk here - based on our intended use of the package - seems to be relevant only if we were to rename the host Crate.  To demitigate that risk, I've explicitly wired up the `package.json` with `--out-dir` and `--out-file` flags to ensure that it always emits `index`-prefixed files.  Furthermore, they are now emitted into the `dist` directory, to be parallel with all our other npm package patterns which do the same.

I will note that, `wasm-pack` may address some of the work-arounds here in the future, but best I can tell, they will be roughly compatible with what we're doing here.  I am basing this outstanding issues, PRs, RFCs and documentation on the `wasm-pack` project, referenced below.

Ref: https://rustwasm.github.io/rfcs/008-npm-dependencies.html#unresolved-questions (See last section)
Ref: https://rustwasm.github.io/docs/wasm-pack/commands/build.html#extra-options (See footnote)
Ref: https://github.com/rustwasm/wasm-pack/issues/606
Ref: https://github.com/rustwasm/wasm-pack/issues/840

### DIFF
--- a/query-planner-wasm/Cargo.toml
+++ b/query-planner-wasm/Cargo.toml
@@ -1,6 +1,7 @@
 [package]
 name = "query-planner-wasm" # this ends up being the name in the published npm package.
-version = "0.0.4"
+version = "0.0.0-DO.NOT.CHANGE" # The version is managed in the package.json, by Lerna.
+private = true
 authors = ["Apollo <opensource@apollographql.com>"]
 homepage = "https://github.com/apollographql/federation"
 description = "Bridge code written in Rust to Javascript/Typescript, to be internally used by Apollo Gateway. This package is not meant to be independently consumed."
--- a/query-planner-wasm/package.json
+++ b/query-planner-wasm/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@apollo/query-planner-wasm",
-  "version": "0.0.3",
+  "version": "0.0.4",
   "description": "Bridge code written in Rust to Javascript/Typescript, to be internally used by Apollo Gateway. This package is not meant to be independently consumed.",
   "scripts": {
     "wasm-pack": "wasm-pack build --target nodejs --out-dir dist --out-name index --scope apollo",
--- a/query-planner-wasm/package.json
+++ b/query-planner-wasm/package.json
@@ -1,0 +1,34 @@
+{
+  "name": "@apollo/query-planner-wasm",
+  "version": "0.0.3",
+  "description": "Bridge code written in Rust to Javascript/Typescript, to be internally used by Apollo Gateway. This package is not meant to be independently consumed.",
+  "scripts": {
+    "wasm-pack": "wasm-pack build --target nodejs --out-dir dist --out-name index --scope apollo",
+    "preinstall": "npm run wasm-pack"
+  },
+  "author": "opensource@apollographql.com",
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/apollographql/federation.git",
+    "directory": "query-planner-wasm/"
+  },
+  "bugs": {
+    "url": "https://github.com/apollographql/federation/issues"
+  },
+  "homepage": "https://github.com/apollographql/federation#readme",
+  "keywords": [
+    "GraphQL",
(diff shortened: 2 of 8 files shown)


query-planner-wasm/package.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "@apollo/query-planner-wasm",
+  "version": "0.0.3",
+  "description": "Bridge code written in Rust to Javascript/Typescript, to be internally used by Apollo Gateway. This package is not meant to be independently consumed.",
+  "scripts": {
+    "wasm-pack": "wasm-pack build --target nodejs --out-dir dist --out-name index --scope apollo",
```
trevor-scheer, line 6: This is great.

query-planner-wasm/package.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "@apollo/query-planner-wasm",
+  "version": "0.0.3",
```
trevor-scheer, line 3: Should this be...
```suggestion
  "version": "0.0.4",
```

trevor-scheer: Or maybe you've already done a release elsewhere that will merge into this branch.

trevor-scheer (approved):
🎉

query-planner-wasm/package.json
abernix: Nope, you're right!

query-planner-wasm/Cargo.toml
```diff
@@ -1,6 +1,7 @@
 [package]
 name = "query-planner-wasm" # this ends up being the name in the published npm package.
-version = "0.0.4"
+version = "0.0.0-DO.NOT.CHANGE" # The version is managed in the package.json, by Lerna.
```
Enrico2, line 3: (apologies for the late reply, I didn't get to look at this yesterday, it was a short day for me that was packed with meetings in the morning)

Is there any way we can keep these in sync instead of forcing this version to be a stub? 
If possible, I don't want to prevent us from being able to publish this crate as a standalone to crates.io

abernix: This doesn't prevent us from doing it permanently, only in the current iteration of the natural growth of this project.

PR #186, opened 2020-09-25
Stargate suggestions to #135 + actix-web support

Enrico2:
This PR was developed on top of #135

First and foremost, any changes to the existing _executor.rs_ and related code written in #135 are mostly there as **suggestions**. @jbaxleyiii lmk if you want to pair on these or feel free to take / reject parts. I didn't dig too deep into the implementation of the executor yet, I'm going to start doing that this week.

Second: This PR separates out the Stargate library from the binary, for a few reasons:
1. Separating out the business logic from the wrapping server makes for easier testing.
2. The library doesn't choose an async runtime, nor does it assume any specific implementation of an HTTP server.
3. Use actix-web instead of tide; we've discussed these options and feel more comfortable going with a more mature server implementation, as well as one that uses tokio.

@jbaxleyiii If these changes look good, we can merge them into #135 and we can continue from there. If you want to do it differently, lmk!

- [ ] Address `todo!` / `unimplemented!` / `FIXME` in the code
- [ ] Tests for the executor (ideally these should be unit tests on the Stargate library, for coverage)
- [ ] Integration tests. I was thinking we can set up acephei services locally and then the gateway would have `localhost` references to those on different ports, integration tests would be entirely localhost.
- [ ] Add tracing/metrics middleware -- I believe there are some existing crates.
- [ ] Write more log lines so that they're integrated with the logger of the http server (use `env_logger` ?)


query-planner-wasm/Cargo.toml
```diff
@@ -1,7 +1,6 @@
 [package]
 name = "query-planner-wasm" # this ends up being the name in the published npm package.
 version = "0.0.0-DO.NOT.CHANGE" # The version is managed in the package.json, by Lerna.
-private = true
```
trevor-scheer, line 4: Is this intentional? I suspect @abernix could speak to its existence. This seems likely to be a relic of `package.json`'s `private` key - maybe it's reflected in a generated `package.json`?

stargate/crates/stargate-lib/src/common.rs
```diff
@@ -0,0 +1,23 @@
...
+#[structopt(
+    name = "stargate",
+    about = "A production ready federation server from Apollo"
+)]
+pub struct Opt {
+    /// Manifest CSDL
+    #[structopt(long, parse(from_os_str))]
+    pub manifest: PathBuf,
+
+    /// Where to write the output: to `stdout` or `file`
```
trevor-scheer, line 14: Copypasta?

stargate/crates/stargate-lib/src/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+impl<'app> Stargate<'app> {
+    pub fn new(schema: &'app str) -> Stargate<'app> {
+        // TODO(ran) FIXME: gql validation on schema
+        let planner = QueryPlanner::new(schema);
+        let service_list = get_service_list(&planner.schema);
+        Stargate {
+            planner,
+            service_list,
+        }
+    }
```
trevor-scheer, line 29: Can `get_service_list` be a fn that lives on a `QueryPlanner` instance? Then `Stargate` would only need a planner passed in to its `new` fn, which could then get the `service_list` itself.

trevor-scheer (approved):
I don't think it's problematic or blocking, but it seems strange to have `graphql-parser` "belong" to `stargate`. I'd expect we'll be using this crate in all kinds of tooling and it seems pretty standalone.

query-planner-wasm/Cargo.toml
Enrico2: afaict this had no affect, and in fact caused a warning that said there's an unknown key in the file.
Also, for future reference, the only keys that get copied to the `package.json` that gets generated by `wasm-pack` are [here](https://github.com/rustwasm/wasm-pack/blob/master/src/manifest/npm/commonjs.rs)

stargate/crates/stargate-lib/src/common.rs
Enrico2: Yes thanks!

stargate/crates/stargate-lib/src/lib.rs
Enrico2: It can, I tried that at some point. I think I landed on this because it makes sense for the service list to be a property of Stargate. It felt a little awkward to get that from the planner in terms of ownership.

Enrico2: Oh I remember, in order to do that, query-planner (which is a different crate) would need to either be aware of `ServiceDefinition` which comes from stargate-lib, and so that definition would have to move, and that has cascading effects.

query-planner-wasm/Cargo.toml
abernix: Yeah, all good to remove this; this doesn't appear to do what I had thought it would.  I was trying to prevent this package from getting accidentally published to Crates.io.  Though I think that what I had meant to do was probably to use the [`publish = false`](https://doc.rust-lang.org/cargo/reference/manifest.html#the-publish-field) option.

PR #189, opened 2020-09-28
Create stargate Dockerfile and GH Publish action

trevor-scheer:
This commit introduces a Dockerfile for building a stargate image.
Currently this is building from an in-flight PR (#135), which we
can update as soon as it lands on `main`.

The additional GH action publishes the container to the new
GitHub Container Registry. For now, this requires a PAT (as noted
in the comments). We should remove this in favor of GITHUB_TOKEN
as soon as it's supported.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,27 @@
+name: Publish on Release
+
+on:
+  release:
+    types: created
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  publish:
+    name: Publish
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Code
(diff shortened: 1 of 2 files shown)


Enrico2 (commented):
This won't work anymore if #186 is merged to #135 

Do you want to base this on #186 and assume stargate-actix-web will be the one we use? The `Dockerfile` location can remain the same, just the build should build the binary in `integrations/stargate-actix-web`

abernix (approved):
> This won't work anymore if (once?) #186 is merged to #135 
> Do you want to base this on #186 and assume stargate-actix-web will be the one we use?

While true, that makes #186 a blocker factor of this PR unnecessarily.

Instead, let's have #186 make the appropriate change — when finished/ratified — to this file so this can merge right away.

PR #197, opened 2020-09-30
Publish stargate with appropriately versioned tags

trevor-scheer:
This PR updates Stargate's publishing workflow to include versioned tags as well as `latest` when publishing to GHCR. These additional versions will allow users to pin to major, minor, or patch versions of the container, like so (for `v0.1.2`):

`apollographql/federation/stargate:0`
`apollographql/federation/stargate:0.1`
`apollographql/federation/stargate:0.1.2`
`apollographql/federation/stargate:latest`

The patch tag will never be overwritten, but the `latest`, minor, and major tags will be updated as appropriate.

A couple examples:
- On minor release (`1.2.1` -> `1.3.0`), the  `.../stargate/latest` and `.../stargate:1` tags would be updated, while the `.../stargate:1.2` and `.../stargate:1.2.1` tags would remain the same. New `.../stargate:1.3.0` and `.../stargate:1.3` tags would be created.

- On major release (`1.2.1` -> `2.0.0`), only the `latest` tag would be updated. New `.../stargate:2`, `.../stargate:2.0` and `.../stargate:2.0.0` tags would be created.

This publishing behavior is meant to replicate the capabilities of the `^`, `~`, and `=` operators that semver uses to specify pinning the major/minor/patch version of a package.

Associated test release:
https://github.com/apollographql/federation/releases/tag/v0.1.2

Associated successful job run to validate:
https://github.com/apollographql/federation/actions/runs/282569418

Associated image SHA:
sha256:18b0afb05feb02ff582483272a35b518cfc0a6b6b4c395f0358fde6d162e32a3
Can be found [here](https://github.com/orgs/apollographql/packages/container/package/federation%2Fstargate)


.github/workflows/publish.yml
```diff
@@ -14,7 +14,9 @@ jobs:
     steps:
       - name: Checkout Code
         uses: actions/checkout@v2
-
+      - name: Get release version
+        id: get_version
+        run: echo ::set-env name=RELEASE_VERSIONS::$(echo "${GITHUB_REF}" | sed -e "s/refs\/tags\///g" | sed -E "s/v?([0-9]+)\.([0-9]+)\.([0-9]+)(-[a-zA-Z]+(\.[0-9]+)?)?/\1.\2.\3\4,\1.\2\4,\1\4/g")
```
Enrico2, line 19: You can use `${{ github.event.release.tag_name }}` to get the tag instead of parsing the ref.

Enrico2, line 19: I'm not 100% sure about this `sed`, it's the sort of stuff that if I were to write in regular code, I'd write a lot of unit tests for, and the consequence of making a mistake here is publishing something to the wrong version and potentially affecting people's actual running machines.

E.g.
```
$ echo "v1.2.3-alpha.3" | sed -E "s/v?([0-9]+)\.([0-9]+)\.([0-9]+)(-[a-zA-Z]+(\.[0-9]+)?)?/\1.\2.\3\4,\1.\2\4,\1\4/g"
1.2.3-alpha.3,1.2-alpha.3,1-alpha.3
```
But then also
```
$ echo "v1.2.4-alpha.3" | sed -E "s/v?([0-9]+)\.([0-9]+)\.([0-9]+)(-[a-zA-Z]+(\.[0-9]+)?)?/\1.\2.\3\4,\1.\2\4,\1\4/g"
1.2.4-alpha.3,1.2-alpha.3,1-alpha.3
```
notice that `1.2-alpha.3,1-alpha.3` is repeated for basically a path version and potentially multiple alphas behind.

Generally though (this is not criticism!) this kind of regex parsing feels a bit too brittle, and is a hotbed for bugs that might cause us to actually release something to a version we don't want to. (I wouldn't trust myself to not make mistakes here, even reviewing this..)

Have you looked into existing Github actions out there that do similar stuff?

Or maybe we can use [this script](https://github.com/cloudflare/semver_bash/blob/master/semver.sh)?

Enrico2 (changes requested):
I just noticed we're using `types: created`, can you speak to that? I saw on this [SO post](https://stackoverflow.com/questions/59518658/how-to-use-github-release-version-number-in-github-action) that maybe we should be using `published`? 
from [here](https://docs.github.com/en/free-pro-team@latest/developers/webhooks-and-events/webhook-events-and-payloads#release):

> * **created**: a draft is saved, or a release or pre-release is published without previously being saved as a draft
> * **published**: a release, pre-release, or draft of a release is published

stargate/Dockerfile
```diff
@@ -23,9 +23,9 @@ FROM alpine:latest
...
 RUN apk --no-cache add ca-certificates
 COPY --from=builder \
-  /home/rust/src/target/x86_64-unknown-linux-musl/release/stargate \
+  /home/rust/src/target/x86_64-unknown-linux-musl/release/apollo-stargate \
   /usr/local/bin/
 
 # User can provide an alternate port via "-e 'PORT=<your port>'" when running the container.
 # PORT defaults to 8080 otherwise.
-CMD /usr/local/bin/stargate --address=http://0.0.0.0:${PORT:-8080} --manifest=/graph
+CMD /usr/local/bin/apollo-stargate --address=http://0.0.0.0:${PORT:-8080} --manifest=/graph
```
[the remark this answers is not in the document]
trevor-scheer: Resolved in 92aae4d

.github/workflows/publish.yml
trevor-scheer: Outdated as your fork now removes the need for any of this

stargate/Dockerfile
```diff
@@ -22,7 +22,7 @@ FROM alpine:latest
 # TODO: Unsure if needed
 RUN apk --no-cache add ca-certificates
 COPY --from=builder \
-  /home/rust/src/target/x86_64-unknown-linux-musl/release/stargate \
+  /home/rust/src/target/x86_64-unknown-linux-musl/release/apollo-stargate \
```
trevor-scheer, line 25:
```suggestion
  /home/rust/src/target/x86_64-unknown-linux-musl/release/stargate \
```

Enrico2 (approved):

.github/workflows/publish.yml
Enrico2: We fixed this concern by forking the action repo and addressing it in code, see [Enrico2/Publish-Docker-Github-Action](http://github.com/Enrico2/Publish-Docker-Github-Action)

abernix (approved):
LGTM!  Are we aiming to contribute the feature up stream?  (I hope we do / are.  This is, I believe, one of the more popular Actions and was well tested and maintained which seemed to make it a good "reach for" option.  Maintaining a fork will require the fork be updated.)

In the future I suspect we may want to consider not relying on a GitHub Action for this important publishing step since I don't believe we can easily invoke it locally in an (urgent) pinch.  Put another way, having this as a script in the repository seems appealing to me.

The GitHub Action at that point would just be:

```yaml
  - run: ./scripts/publish-this-image $version_to_publish
```

Which, on CircleCI, works, well, with _exactly_ the same YAML since they share that common form. :grin:

Let's also remember that this will make every version the latest, even if we're marking it as a pre-release right now in the _Releases_ interface, or even if we publish a patch fix to a no-longer active release line (i.e., we'll have to be careful when posting a bug fix on 1.5.2 when we're already on 2.0.0).

(For now we can just run `docker build . -t ghcr.io/apollographql/stargate:latest ghcr.io/apollographql/stargate:0 ghcr.io/apollographql/stargate:0.1 ghcr.io/apollographql/stargate:0.1.2` and `docker push` on each, accordingly.).

Enrico2:
@abernix I'm not sure they'd be open to a PR for this change, given the conversation here: https://github.com/elgohr/Publish-Docker-Github-Action/issues/70

But I'll clean up the code and PR it, we'll see what the feedback is like. I agree about the maintenance part.

trevor-scheer:
Landed this in the interest of time but will circle back next week to take action items from @abernix's comments.

PR #208, opened 2020-10-05
Lock down fn/struct visibility and a round of version upgrades

Enrico2:
This PR does 2 things:

## Visibility lock down

By default we wrote a lot of functions and structs as `pub` where they are really only used internally in the crate. I've changed most of them to `pub(crate)`. I did a "find & replace", compiled, fixed errors, and eventually ran tests.

## Version upgrades

Dependabot opened a few different PRs, this collates them into one.
closes #202 #203 #204 #205 #206

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -37,9 +37,9 @@ dependencies = [
 
 [[package]]
 name = "actix-cors"
-version = "0.3.0"
+version = "0.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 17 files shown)


Enrico2:
This PR does 2 things:

## Visibility lock down

By default we wrote a lot of functions and structs as `pub` where they are really only used internally in the crate. I've changed most of them to `pub(crate)`. I did a "find & replace", compiled, fixed errors, and eventually ran tests.

## Version upgrades

Dependabot opened a few different PRs, this collates them into one.
closes #202 #203 #204 #205 #206

PR #218, opened 2020-10-09
Add serviceFetcher option to change fetcher used to connect to downstream services.

jsegaran:


trevor-scheer:
@jsegaran is this still an API you're interested in adding? (I'm going to preemptively close but happy to revisit)

PR #226, opened 2020-10-12
Add support for multiple wasm QueryPlanner objects

### LINKED ISSUE #210: Support for multiple wasm query planner objects?
In our application, we instantiate multiple different Apollo Gateway objects, each with slightly different schemas. We do this because we're currently migrating code from one service to another and we want to do side-by-side testing of the old and new implementations. Having two gateway objects allows us to compare the query plans that each produces and to make a side-by-side test whenever we notice a difference between those plans.

After the wasm query planner was introduced `@apollo/gateway` in v0.20.1, we found that we could no longer do this - it seemed that we always got query plans for whichever gateway was initialized last (see reproduction code below).

I am not a rust expert, but I suspect the issue is here in [this file](https://github.com/apollographql/federation/blob/88cb9164341363d908c8a414f879ecb143017688/query-planner-wasm/src/lib.rs#L7-L23) where there currently is just a global SCHEMA and DATA objects that are always reused whenever a gateway object updates its schema. 

Given that these variables are setup right now as vectors rather than just global pointers, I'm hoping that y'all are planning to support multiple query planner objects in the future. Is that the case, or is that something you'd be open to accepting a PR for? We'd like to update to the latest query planner since it seems to fix some bugs we're experiencing, but this issue will block us from doing so.

<details>
<summary> To reproduce the issue here:</summary>

```
const { ApolloGateway } = require("@apollo/gateway");
const { parse } = require("graphql");

const gatewayA = new ApolloGateway({
  localServiceList: [
    { name: "accounts", url: "http://localhost:4001/graphql", typeDefs: parse("type Query { foo: String, bar: String}")},
    { name: "reviews", url: "http://localhost:4002/graphql", typeDefs: parse("type Query { baz: String }")},
  ],
});

const gatewayB = new ApolloGateway({
  localServiceList: [
    { name: "accounts", url: "http://localhost:4001/graphql", typeDefs: parse("type Query { foo: String }")},
    { name: "reviews", url: "http://localhost:4002/graphql", typeDefs: parse("type Query { baz: String, bar: String }")},
  ],
});

// The two schemas are different
console.log(JSON.stringify(gatewayA.schema) == JSON.stringify(gatewayB.schema)) // expected: 'false', actual: 'false'

// However the two gateway objects share a pointer to the same query
// planner wasm object, which means they will always return the same
// query plans (those of gatewayB)
console.log(gatewayA.queryPlannerPointer == gatewayB.queryPlannerPointer) // expected: 'false', actual: 'true'
```
</details>

jaredly:
Fixes #210

I considered using a HashMap, but it's a little more complicated, and I'm assuming we won't have people creating many hundreds of query plans 🙃 if that's a use case we want to support, we could move to HashMaps.

One nice thing about passing around an index instead of a pointer is that it's much more robust :) rust doesn't really make guarantees about raw pointers remaining valid in the presence of allocations.

cc @Enrico2 looks like you did the first impl of this stuff?

### DIFF
--- a/query-planner-wasm/src/lib.rs
+++ b/query-planner-wasm/src/lib.rs
@@ -38,10 +38,8 @@ pub fn get_query_planner(schema: JsString) -> usize {
 }
 
 /// Drop a query planner (and associated Schema string) to free up memory.
-/// Most applications will have a single query planner that they use
-/// for the duration of the app's lifetime, but if you are working
-/// with multiple QueryPlanners, you'll want to call this when you
-/// are done with one.
+/// This can happen when an ApolloGateway updates & replaces its QueryPlanner,
+/// or when an ApolloGateway is no longer needed.
 #[wasm_bindgen(js_name = dropQueryPlanner)]
 pub fn drop_query_planner(planner_idx: usize) {
     unsafe {
--- a/query-planner-wasm/src/lib.rs
+++ b/query-planner-wasm/src/lib.rs
@@ -13,7 +13,7 @@ static mut DATA: Vec<Option<QueryPlanner>> = vec![];
 /// getQueryPlanner creates a QueryPlanner if needed, and returns an "id"
 /// for later use with `getQueryPlan`. Calling this multiple times with
 /// the same schema will only result in the schema being parsed once, and the
-/// QueryPlanner is reused.
+/// QueryPlanner is cloned.
 #[wasm_bindgen(js_name = getQueryPlanner)]
 pub fn get_query_planner(schema: JsString) -> usize {
     let schema = String::from(schema);
@@ -21,7 +21,11 @@ pub fn get_query_planner(schema: JsString) -> usize {
         for i in 0..SCHEMA.len() {
             match &SCHEMA[i] {
                 Some(x) if x == &schema => {
-                    return i;
+                    let id = SCHEMA.len();
+                    SCHEMA.push(Some(schema));
+                    // No need to re-parse, we can just clone!
+                    DATA.push(DATA[i].clone());
+                    return id;
                 }
                 _ => (),
             }
--- a/query-planner-wasm/src/lib.rs
+++ b/query-planner-wasm/src/lib.rs
@@ -4,31 +4,61 @@ use apollo_query_planner::{QueryPlanner, QueryPlanningOptions};
 use js_sys::JsString;
 use wasm_bindgen::prelude::*;
 
-static mut SCHEMA: Vec<String> = vec![];
-static mut DATA: Vec<QueryPlanner> = vec![];
+/// These vectors serve as repositories for the Schema string and QueryPlanner
+/// so that they don't get garbage collected. The JavaScript side gets the index
+/// into the vector, which it can use in `getQueryPlan`.
+static mut SCHEMA: Vec<Option<String>> = vec![];
+static mut DATA: Vec<Option<QueryPlanner>> = vec![];
 
+/// getQueryPlanner creates a QueryPlanner if needed, and returns an "id"
+/// for later use with `getQueryPlan`. Calling this multiple times with
+/// the same schema will only result in the schema being parsed once, and the
+/// QueryPlanner is reused.
 #[wasm_bindgen(js_name = getQueryPlanner)]
 pub fn get_query_planner(schema: JsString) -> usize {
+    let schema = String::from(schema);
     unsafe {
-        if SCHEMA.is_empty() {
-            SCHEMA.push(String::from(schema));
-            DATA.push(QueryPlanner::new(&SCHEMA[0]));
-        } else {
-            SCHEMA[0] = String::from(schema);
-            DATA[0] = QueryPlanner::new(&SCHEMA[0]);
+        for i in 0..SCHEMA.len() {
+            match &SCHEMA[i] {
+                Some(x) if x == &schema => {
+                    return i;
+                }
+                _ => (),
+            }
         }
-        let data = &DATA[0];
-        data as *const QueryPlanner as usize
+        let id = SCHEMA.len();
+        SCHEMA.push(Some(schema));
+        DATA.push(Some(QueryPlanner::new(&SCHEMA[id].as_ref().unwrap())));
+        return id;
+    }
+}
+
+/// Drop a query planner (and associated Schema string) to free up memory.
+/// Most applications will have a single query planner that they use
+/// for the duration of the app's lifetime, but if you are working
+/// with multiple QueryPlanners, you'll want to call this when you
+/// are done with one.
+#[wasm_bindgen(js_name = dropQueryPlanner)]
+pub fn drop_query_planner(planner_idx: usize) {
+    unsafe {
+        if planner_idx < DATA.len() {
+            // Setting the index to None will allow what was there to
+            // be freed. We keep them there so that all indices will be
+            // preserved.
+            SCHEMA[planner_idx] = None;
+            DATA[planner_idx] = None;
+        }
     }
 }
 
 #[wasm_bindgen(js_name = getQueryPlan)]
-pub fn get_query_plan(planner_ptr: usize, query: &str, options: &JsValue) -> JsValue {
-    let options: QueryPlanningOptions = options.into_serde().unwrap();
+pub fn get_query_plan(planner_idx: usize, query: &str, options: &JsValue) -> JsValue {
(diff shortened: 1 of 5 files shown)


jaredly:
image
here's the test failure 🙃

query-planner-wasm/src/lib.rs
```diff
@@ -4,31 +4,61 @@ use apollo_query_planner::{QueryPlanner, QueryPlanningOptions};
...
+        let id = SCHEMA.len();
+        SCHEMA.push(Some(schema));
+        DATA.push(Some(QueryPlanner::new(&SCHEMA[id].as_ref().unwrap())));
+        return id;
+    }
+}
+
+/// Drop a query planner (and associated Schema string) to free up memory.
+/// Most applications will have a single query planner that they use
+/// for the duration of the app's lifetime, but if you are working
```
Enrico2, line 38: This sentence is a bit misleading; One of the ways in which Apollo Gateway can run is by polling for downstream schema changes, composing a new federated schema (csdl) and updating the runtime schema on the fly.

That fact brings to question if this is the right solution, because some people do in fact create a lot of new schemas during the runtime of the server by design.
I like that you've added the function to drop a planner, if we want to go with this approach, we'd have to make sure and drop the old planner whenever a schema updates (i.e. identify schema changes vs. additions).

At the root of the problem though, is that there is one global wasm module loaded, but there can be multiple ApolloGateway instances. We might want to solve this at the root of the problem, if it's possible, rather than adapting the global model to a local one.

Let me stew on this for a bit.

Generally though, the use of an index is indeed better than a pointer!

Enrico2 (commented):
As for the test failure, I'm fixing that wasm-pack installation issue separately, rebase your branch once #228 is merged.

query-planner-wasm/src/lib.rs
```diff
@@ -4,31 +4,63 @@ use apollo_query_planner::{QueryPlanner, QueryPlanningOptions};
...
-        } else {
-            SCHEMA[0] = String::from(schema);
-            DATA[0] = QueryPlanner::new(&SCHEMA[0]);
+        for i in 0..SCHEMA.len() {
+            match &SCHEMA[i] {
+                Some(x) if x == &schema => {
+                    let id = SCHEMA.len();
+                    SCHEMA.push(Some(schema));
+                    // No need to re-parse, we can just clone!
+                    DATA.push(DATA[i].clone());
```
Enrico2, line 27: I see why you went with this approach, 2 calls using the same index and one of them dropping the planner would cause a dangling reference to an index.

### Why not clone?

We've been intentional about not implementing `Clone` for structs that can potentially be very large. For the `QueryPlanner` case, its data is the parsed schema ast, which can be a very large deeply nested tree. One of the design tenets for this gateway is to be able to operate over very large schemas (e.g. imagine an entire enterprise's data graph, composed from 100+ different services) and cloning that would create a noticeable performance degradation.

Cloning here creates a state where the gateway will be susceptible to undue memory pressure, if a user mis-uses the js library or even in the normal case, if there is a schema composition event but no change in the resulting composed schema, and we keep cloning the same planner over and over.

Generally speaking, using an `Rc` would solve that, but there's a bug here....

### Bug: `QueryPlanner` references the schema `String`.

This is the definition:
```rust
pub struct QueryPlanner<'s> {
    pub schema: schema::Document<'s>,
}
```

that `'s` is the lifetime referenced by `Document`, which, if you look at where that ends up coming from, is the lifetime of the original `&str` that contains the schema. So when we put the schema in `SCHEMA`, we then use a reference to it to pass to the parser. By cloning (even if this was an `Rc`), the cloned planner still references the `String` that was used to create the original planner. When dropping that planner and its schema, any cloned planner will have dangling references.

### How should we fix the root issue?

We've not had a chance to discuss this, and we appreciate the contribution and ideas, I think they may inform an eventual solution. Ideally I'd like to figure out if we can fix the problem outside of the surface area of these objects somehow.

### Idea?

Just had a thought, maybe we should be adding a static `GATEWAYS` vector that will contain a schema and planner, which would map to a js `ApolloGateway` instance?
Meaning, if you have 2 `ApolloGateway` instances, each of which is constantly replacing its schema, you'd have 2 items in a `GATEWAYS` vector, and for each gateway we'd still just have one schema and planner that's being replaced?
In that situation we'd not have to implement any `drop` (which really adds a lot of complexity)

That might be a better design. Doing that would mean introducing some breaking API changes, the names would match what we're returning:

```rust
#[wasm_bindgen(js_name = getGatewayBridge)]
pub fn get_gateway_bridge(schema: JsString) -> usize { .. }

#[wasm_bindgen(js_name = getQueryPlan)]
pub fn get_query_plan(gateway_idx: usize, query: &str, options: &JsValue) -> JsValue { .. }
```

(Noting that this would be a breaking API change, which will be relevant when releasing)

jaredly:
@Enrico2 thanks for the great review! I've reworked my solution to fix the bugs, and I think I've got a pretty neat solution that doesn't involve any breaking changes! Memory access is localized, and cloning is done behind an `Rc` which makes it very cheap.

Enrico2:
@jaredly Thanks for the update! We're discussing how to proceed with this. This adds some level of complexity that ideally I'd like to avoid, and so I'm researching if there are other ways to solve the problem. Stay tuned!

Enrico2:
Take a look at #261 as a proposed simpler alternative.

jaredly:
hmmm @Enrico2 it's simpler because it doesn't allow you to free memory if you're done with an `ApolloGateway` object?

Enrico2:
> hmmm @Enrico2 it's simpler because it doesn't allow you to free memory if you're done with an `ApolloGateway` object?

It's simpler because it removes the need for memory management from JS while still allowing multiple Gateways to run in the same process, as well as update a gateway's schema on the fly without leaking memory.

ApolloGateway is not meant to "be done with", under what circumstances would you be done with it while keeping the node process around?

dkapadia:
(@jaredly is a colleague of mine, working on the same project - he's also far more fluent in rust/wasm than I am).

>Take a look at #261 as a proposed simpler alternative.

From an API level, it looks like our tests pass now with #261 as well as this PR, so that's a definite improvement over main (where our tests fail due to the overwriting of the schema). I haven't had a chance to profile the memory consumption of the rust query planner yet, but if it's comparable to the JS query plan code we'll definitely need to worry about freeing unused gateway objects and so would greatly prefer this PR over #261.

> ApolloGateway is not meant to "be done with", under what circumstances would you be done with it while keeping the node process around?

We currently create ApolloGateway objects as part of our deploy testing flow to allow devs to test out and validate schema changes in production before those changes are rolled out to 'real' users. For example, if dev visits <test-schema-version>.example.com, we create (and cache) an ApolloGateway object for `test-schema-version`, then execute the request using that newly created gateway, instead of the gateway used for requests sent to `example.com` (which uses the production schema for end users). Once they are happy testing against the test-schema, we make the test-schema the default production schema, by updating the production ApolloGateway's schema.

Each ApolloGateway we create uses up a fair bit of memory, so we have a fairly small LRU cache setup that evicts unused ApolloGateway objects when they are no longer used. This is fine for our use case since we only need a max of ~5 test versions at any given time, but there might be ~50 such versions created over the course of a day, so we can't really have the memory usage grow unbounded.

Happy to go into more detail if any of that is unclear, and very much appreciate your time looking into this issue!

abernix:
Copying and pasting my comment from https://github.com/apollographql/federation/pull/261#issuecomment-821091620:

> This should no longer be necessary with the landing of https://github.com/apollographql/federation/pull/622 (specifically, the changing of the query planner back to TypeScript).  Put another way, the code this changes is no longer present and the query planners are now scoped within the Gateways that generate and execute them.

Thanks very much for opening this PR in the first place; very much appreciated contribution, and hopefully this simplifies your configuration in your deploy testing.

jaredly:
oh neat! Does this mean y'all are moving away from rust/wasm? (I see reducing the number of languages in a project as a positive thing 😅)

abernix:
@jaredly We're still excited about Rust (and hiring for it!).  For `@apollo/gateway` in particular though, we are focused on iterating on the existing codebase in TypeScript (read: reducing the number of languages in this repository).  We didn't necessarily want to be maintaining multiple languages within a project for long, though that was part of what unfolded in this experiment.  Happy to get back to a single language here and simplify the packaging though, for sure.

Overall, this will allow us to iterate, improve and learn on the Federation model _as a whole_, bringing new features and bug fixes + adding value for many current users.

PR #228, opened 2020-10-13
GH Action: Install wasm-pack only if not available.

Enrico2:
Installing wasm-pack [fails](https://github.com/rustwasm/wasm-pack/blob/173a05bd829bc499533c3967a2b980dc6069ce19/src/installer.rs#L96-L103) if it's already installed. This change checks if `wasm-pack` exists already, and installs it otherwise.


EverlastingBugstopper (approved):
lgtm!

PR #233, opened 2020-10-15
feat(stargate): Propagate headers to downstream services

trevor-scheer:
This PR introduces request header propagation to the gateway in its simplest form. For authentication or out-of-graphql-band context purposes, there's usually a need for the gateway to pass along headers to the implementing services. This can now be accomplished by setting the flag `--propagate-request-headers` when running the stargate binary. The value is a space delimited list of header names which the gateway should "forward" to the outgoing requests being sent during execution.

TODO: address likely case-sensitivity of this implementation


stargate/crates/stargate-lib/src/request_pipeline/service_definition.rs
```diff
@@ -35,13 +35,31 @@ impl Service for ServiceDefinition {
...
             .await?;
 
         data.ok_or_else(|| unimplemented!("Handle error cases in send_operation"))
     }
 }
+
+// TODO: Add tests
+
+#[cfg(test)]
+mod tests {}
```
Enrico2, line 65: I like to add a failing test to remind myself to add real tests :)

stargate/crates/stargate-lib/src/request_pipeline/service_definition.rs
```diff
@@ -35,13 +35,31 @@ impl Service for ServiceDefinition {
             variables: Some(Map::from_iter(variables.into_iter()).into()),
         };
 
+        let headers = &context.request_context.header_map;
+
+        let mut request_builder = surf::post(&self.url).header("userId", "1");
+        for (&name, &value_bytes) in headers.into_iter() {
+            match std::str::from_utf8(value_bytes) {
+                Ok(value) => {
+                    request_builder = request_builder.header(name, value);
```
Enrico2, line 44: Looks like `.header`'s 2nd argument's type is `impl ToHeaderValues` which means it can accept any struct that implements that trait.
One of the implementors is `HeaderValue` from `http_types`. That struct has a `from_bytes` method, which is the value in the HashMap we have in RequestContext. That seems like the most appropriate way to pass that in.

stargate/src/main.rs
```diff
@@ -1,26 +1,44 @@
+use std::collections::HashMap;
```
Enrico2, line 1: Nit: IJ's regular format moves around imports, I have not yet found where to configure that. OTOH Rust's `cargo fmt` sorts them if they're not grouped (i.e. no line spaces between them) but IJ adds those empty lines.
So, IJ has 2 ways to format Rust, it's own formatter and applying the rust formatter:
image

I've conditioned myself the Rustfmt one on rust files.
tldr: remove the empty lines between use statements..

trevor-scheer: Done! Workflow tips always appreciated

stargate/crates/stargate-lib/src/common.rs
```diff
@@ -27,6 +27,10 @@ pub struct Opt {
     /// Accepts [http|udp]://host:port/path (path is optional)
     #[structopt(short, long)]
     pub tracing_endpoint: Option<TracingConfig>,
+
+    /// A space separated list of header names which Stargate should propagate to downstream services.
```
Enrico2, line 31: Heh [this is my one of my most popular tweets](https://twitter.com/rmgn/status/650069074769149952), and sort of shows that the use of downstream/upstream is not ideal. Let's use `implementing services` if that's our federation nomenclature

stargate/crates/stargate-lib/src/common.rs
```diff
@@ -121,10 +140,12 @@ impl Default for Opt {
 
 #[cfg(test)]
 mod tests {
-    use crate::common::{Opt, TracingConfig, TracingProtocol};
     use std::path::PathBuf;
+
     use structopt::StructOpt;
```
Enrico2, line 146: nit: no spaces

stargate/crates/stargate-lib/src/common.rs
```diff
@@ -27,6 +27,10 @@ pub struct Opt {
     /// Accepts [http|udp]://host:port/path (path is optional)
     #[structopt(short, long)]
     pub tracing_endpoint: Option<TracingConfig>,
+
+    /// A space separated list of header names which Stargate should propagate to downstream services.
+    #[structopt(short, long)]
+    pub propagate_headers: Option<Vec<String>>,
```
Enrico2, line 33: naming: We know we're going to want different lists for incoming and outgoing

stargate/crates/stargate-lib/src/lib.rs
```diff
@@ -16,16 +17,23 @@ mod utilities;
 pub struct Stargate<'app> {
     service_list: HashMap<String, ServiceDefinition>,
     pub planner: QueryPlanner<'app>,
+    pub propagate_headers: Option<Vec<String>>,
+}
+
+pub struct StargateOpts<'app> {
+    pub schema: &'app str,
+    pub propagate_headers: Option<Vec<String>>,
 }
```
Enrico2, line 26: Any specific reason you chose to model things like this?

I had imagined something like
```rust
struct Stargate<'app> {
  service_list: HashMap<String, ServiceDefinition>,
  pub planner: QueryPlanner<'app>,
  options: StargateOptions
}

struct StargateOptions {
  pub propagate_headers:Vec<String>
}

impl Default for StargateOptions {
  pub fn default() -> Self {
    Self {
      propagate_headers: vec![]
    }
  }
}
```

I don't think we need the schema in StargateOptions, it's already in the planner.
i.e. `StargateOptions` is not `Opts`

Also, in cases where the behavior for `None` is the same as `Some(vec![])`, I'd prefer making the type just a vector that can be empty.

stargate/src/main.rs
```diff
@@ -76,3 +103,16 @@ async fn main() -> std::io::Result<()> {
     .run()
     .await
 }
+
+#[cfg(test)]
```
Enrico2, line 107: remove

Enrico2:
Oh this is nice GH actions don't run when the branch has conflicts

stargate/crates/stargate-lib/src/common.rs
abernix: I agree we should pick one nomenclature, but it's worth pointing out that we call it [`downstream service(s)` in a lot of places](https://github.com/apollographql/federation/search?q=%22downstream+service%22+OR+%22downstream+services%22)! ([vs. `implementing service(s)`](https://github.com/apollographql/federation/search?q=%22implementing+service%22+OR+%22implementing+services%22)).

Since this is just a comment, this might be a better suggestion for a wholesale name re-alignment.  We should probably have that conversation with product and do it the change at once more broadly though and some of us have recently noted that _implementing services_ is a name that might benefit from revisiting.

trevor-scheer: Implementing is a good suggestion for now. I hope we do have that conversation in the future, I'm rooting for `subgraph` :)

stargate/crates/stargate-lib/src/common.rs
```diff
@@ -27,6 +27,10 @@ pub struct Opt {
     /// Accepts [http|udp]://host:port/path (path is optional)
     #[structopt(short, long)]
     pub tracing_endpoint: Option<TracingConfig>,
+
+    /// A space separated list of header names which Stargate should propagate to implementing services.
+    #[structopt(short, long)]
```
ashleygwilliams, line 32: this make the short flag `-p` which may be surprising to some users!

stargate/crates/stargate-lib/src/common.rs
```diff
@@ -50,6 +54,21 @@ impl Opt {
         } else {
             buf.push_str("tracing: disabled");
         }
+        buf.push('\n');
```
ashleygwilliams, line 57: the `with_capacity` here is surprising, i don't think we need it. i also think refactoring this into a format string would improve readability!

Enrico2:
@trevor-scheer As discussed offline, I gave this a pass and pushed some additions. The main change, that cascaded to some other stuff is to avoid redundant allocations that happen on the critical path (i.e. every request), that led to some references and lifetimes but nothing too complex.

I'm stuck on other bugs with the executor so I'm in no rush for this to merge today, feel free to decide to merge on Wednesday if you like these changes. Thanks!

Enrico2 (approved):
approved having addressed own concerns (got consent from author!)

Issue #156, opened 2020-09-15
Unable to resolve external fields on an interface


jgnieuwhof:
## Summary
It seems as though I've uncovered a bit of odd behaviour while implementing Relay's global node resolver spec in a federated graph. I _think_ the issue more broadly applies to the resolution strategy for fields that return an interface.

I'm happy to fully generalize the description below, but I feel as though it may be helpful to work through this specific example at the same time. I've [added a node service to a forked version of federation-demo](https://github.com/jgnieuwhof/federation-demo/pull/1), with the following schema:

```graphql
# node service
interface Node {
  id: ID!
  createdAt: String! @external
}

extend type User implements Node {
  id: ID! @external
  createdAt: String! @external
}

type Query {
  node(id: ID!): Node
}

# accounts service
interface Node {
  id: ID!
  createdAt: String!
}

type User implements Node @key(fields: "id") {
  id: ID!
  createdAt: String!
  name: String
  username: String
}
```

## Expected Behaviour
Generally, I would expect fields marked as `@external` on an interface to be resolved by the service that defines the base concrete entity being resolved. This would entail using the resolved object's `__typename` to determine where to fetch the external fields from.

Specifically, given the schema above, when running a `node` query, which returns a `User`, I would expect:
- `Node.id` to be resolved by the node service
- `Node.createdAt` to be resolved by the "accounts" service

## Actual Behaviour
Generally:
- If the interface is specified in a query fragment, the gateway tries to resolve all fields from the service that defines the field which returns the interface.
- If the concrete entity is specified in a query fragment, the gateway reaches out to the appropriate service for the external fields.

Specifically:
- If the interface is specified in a query fragment, the gateway tries to resolve both `Node.id` and `Node.createdAt` from the "node" service:
```graphql
# query
{
  node(id: "User:1") {
    ... on Node {
      __typename
      id
      createdAt
    }
  }
}

# query plan
QueryPlan {
  Fetch(service: "node") {
    {
      node(id: "User:1") {
        __typename
        id
        createdAt
      }
    }
  },
}
```

- If the concrete entity is specified in a query fragment, the gateway resolves `Node.id` from the "node" service, and `Node.createdAt` from the "accounts" service:

```graphql
# query
{
  node(id: "User:1") {
    ... on User {
      __typename
      id
      createdAt
    }
  }
}

# query plan
QueryPlan {
  Sequence {
    Fetch(service: "node") {
      {
        node(id: "User:1") {
          __typename
          ... on User {
[...]

PR #243, opened 2020-10-19
External interface resolution across services

MarcelCutts:
## Motivation

Following Jordan's issue described [here](https://github.com/apollographql/federation/issues/156) in which certain kinds of interface queries are not properly resolved across services, we seek to understand why. We start here at the lowest level, the query planner.

## Approach
We added a test case to the query planner Rust crate which [fails for us when using the federation package](https://github.com/apollographql/federation/issues/156), translating our SDL into the domain understood by the planner. Two cases were added, one concrete type which should work as expected, and one abstract interface type which we expect to fail

## Outcome
Both tests pass with enough annotation on the `csdl.graphql` file. This implies that the issue could be higher up somewhere - perhaps at the layer that generates the annotations.

tenor (38)

### DIFF
--- a/stargate/crates/query-planner/tests/features/external-field/csdl.graphql
+++ b/stargate/crates/query-planner/tests/features/external-field/csdl.graphql
@@ -1,0 +1,48 @@
+schema
+@graph(name: "nodes", url: "https://node.api.com")
+@graph(nameL: "accounts", url: "https://accounts.com")
+@composedGraph(version: 1)
+{
+    query: Query
+
+}
+
+# Nodes
+
+type Query {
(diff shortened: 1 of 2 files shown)


trevor-scheer:
@MarcelCutts did you mean to close this? I see the branch was restored but the PR wasn't reopened.

MarcelCutts:
Hey @trevor-scheer - Yes! I apologise for not checking my remote up front. We were encountering some bugs with a previous version of the query planner that I strove to write a failnig test for before attempting to dig at the bug proper. This was going to be for my own help until I or someone else on my team figured out a fix.

However, both actions turned out to be unnecessary out the issue has been resolved within the last month!

PR #264, opened 2020-10-29
Clean `unwrap()`s from `query-planner`

Enrico2:
Started from the query-planner crate.

Only remaining `unwrap()`s are in unit tests.

(also cleaned a clippy thing)

### DIFF
--- a/stargate/crates/stargate-lib/src/request_pipeline/executor.rs
+++ b/stargate/crates/stargate-lib/src/request_pipeline/executor.rs
@@ -34,7 +34,7 @@ pub async fn execute_query_plan<'req>(
     let data_lock: RwLock<Value> = RwLock::new(json!({}));
 
     if let Some(ref node) = query_plan.node {
-        execute_node(&context, node, &data_lock, &vec![]).await;
+        execute_node(&context, node, &data_lock, &[]).await;
     } else {
         unimplemented!("Introspection not supported yet");
     };
@@ -47,7 +47,7 @@ fn execute_node<'schema, 'request>(
     context: &'request ExecutionContext<'schema, 'request>,
     node: &'request PlanNode,
     results: &'request RwLock<Value>,
-    path: &'request ResponsePath,
+    path: &'request [String],
 ) -> BoxFuture<'request, ()> {
     async move {
         match node {
@@ -71,7 +71,7 @@ fn execute_node<'schema, 'request>(
                 //   }
             }
             PlanNode::Flatten(flatten_node) => {
-                let mut flattend_path = Vec::from(path.as_slice());
+                let mut flattend_path = Vec::from(path);
                 flattend_path.extend_from_slice(flatten_node.path.as_slice());
 
                 let inner_lock: RwLock<Value> = RwLock::new(json!({}));
@@ -130,7 +130,7 @@ fn execute_node<'schema, 'request>(
     .boxed()
 }
 
-fn merge_flattend_results(parent_data: &mut Value, child_data: &Value, path: &ResponsePath) {
+fn merge_flattend_results(parent_data: &mut Value, child_data: &Value, path: &[String]) {
     if path.is_empty() || child_data.is_null() {
         merge(&mut *parent_data, &child_data);
         return;
@@ -249,7 +249,7 @@ async fn execute_fetch<'schema, 'request>(
 
 fn flatten_results_at_path<'request>(
     value: &'request mut Value,
-    path: &ResponsePath,
+    path: &[String],
 ) -> &'request Value {
     if path.is_empty() || value.is_null() {
         return value;
@@ -283,7 +283,7 @@ fn flatten_results_at_path<'request>(
     }
 }
 
-fn execute_selection_set(source: &Value, selections: &SelectionSet) -> Value {
+fn execute_selection_set(source: &Value, selections: &[Selection]) -> Value {
     if source.is_null() {
         return Value::default();
     }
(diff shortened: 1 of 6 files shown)


stargate/crates/query-planner/src/builder.rs
```diff
@@ -364,11 +364,11 @@ fn complete_field<'a, 'q: 'a>(
         // the type_name could be a primitive type which is not in our names_to_types map.
         let return_type = context.names_to_types.get(type_name);
 
-        if return_type.is_none() || !return_type.unwrap().is_composite_type() {
+        if return_type.is_none() || !return_type.expect("checked not None").is_composite_type() {
```
ashleygwilliams, line 367: i think this would be better as an `if let`

stargate/crates/query-planner/src/context.rs
```diff
@@ -214,35 +217,35 @@ impl<'q> QueryPlanningContext<'q> {
         }
 
         let return_type = self.names_to_types.get(field_def.field_type.as_name());
-        let field_type_is_not_composite =
-            return_type.is_none() || !return_type.unwrap().is_composite_type();
 
-        if field_type_is_not_composite {
-            return vec![];
-        }
+        if let Some(return_type) = return_type {
```
ashleygwilliams, line 221: the logic here is hard to read and would benefit from being broken into some smaller functions

stargate/crates/query-planner/src/context.rs
```diff
@@ -214,35 +217,35 @@ impl<'q> QueryPlanningContext<'q> {
...
-                    &mut res,
-                    &mut visited_fragment_names,
-                );
-                res
-            };
-            provided_fields.chain(fields).collect()
+            if let Some(provides) = self.federation.provides(field_def) {
+                let fields = {
+                    let mut visited_fragment_names: HashSet<&str> = HashSet::new();
+                    let mut res = vec![];
```
ashleygwilliams, line 234: this smells like it would be better as a struct w/impl

stargate/crates/query-planner/src/context.rs
```diff
@@ -214,35 +217,35 @@ impl<'q> QueryPlanningContext<'q> {
...
+                    );
+                    res
+                };
+                provided_fields.chain(fields).collect()
+            } else {
+                provided_fields.collect()
+            }
         } else {
-            provided_fields.collect()
+            vec![]
```
ashleygwilliams, line 248: is this potentially better as a default value?

stargate/crates/stargate-lib/src/request_pipeline/executor.rs
```diff
@@ -47,7 +47,7 @@ fn execute_node<'schema, 'request>(
     context: &'request ExecutionContext<'schema, 'request>,
     node: &'request PlanNode,
     results: &'request RwLock<Value>,
-    path: &'request ResponsePath,
+    path: &'request [String],
```
ashleygwilliams, line 50: not clear why this type change? seems unrelated to removing unwraps

stargate/crates/stargate-lib/src/request_pipeline/executor.rs
```diff
@@ -283,7 +283,7 @@ fn flatten_results_at_path<'request>(
     }
 }
 
-fn execute_selection_set(source: &Value, selections: &SelectionSet) -> Value {
+fn execute_selection_set(source: &Value, selections: &[Selection]) -> Value {
```
ashleygwilliams, line 286: same here, not sure why type change is in this pr

ashleygwilliams (changes requested):
there are a lot of expects here that suggest that a vec has only one item- why are we using a vec to model this data

replacing unwraps with expects is great but i would still encourage us to use proper error handling. even if we reason it is impossible, code changes, and those assumptions change. using errors and ? is much better than panicking, even if the panic has a messge.

stargate/crates/query-planner/src/builder.rs
Enrico2: How? afaik I can't add a condition with an `||` to an `if let`

stargate/crates/stargate-lib/src/request_pipeline/executor.rs
Enrico2: Recommendation (that I agree with) from Clippy.
`ResponsePath` is a type alias for `Vec<String>`, using a slice `&[String]` is better than expecting a `&Vec<String>` -- it saves an allocation and allows the function to be called with an array.

Enrico2:
I don't have the bandwidth right now to improve this code in general and refactor things out. The only purpose was to replace `unwrap`s with `expect` (and fixing that Clippy warning).

I appreciate the other suggestions but they should not block merging this.

stargate/crates/query-planner/src/context.rs
abernix: Tracking this in https://github.com/apollographql/federation/issues/477

stargate/crates/query-planner/src/context.rs
abernix: Tracking this in https://github.com/apollographql/federation/issues/477

stargate/crates/query-planner/src/context.rs
abernix: Tracking this in https://github.com/apollographql/federation/issues/477

abernix:
At a high level, tracking the need to consider revisiting the `vec` in https://github.com/apollographql/federation/issues/476

PR #273, opened 2020-11-05
Very sketchy draft of CSDL spec.

queerviolet:
Here's about 30% of a first pass of the CSDL spec. It's mostly taken from @JakeDawkins's [notion page](https://www.notion.so/Introduction-to-CSDL-121682c03e5d43c5b1a891c59ac4aefd) about the same.

feel free to comment / push changes to the branch / whatever as we develop this

cc: @JakeDawkins @trevor-scheer @sachindshinde @zionts


docs/source/csdl-spec.mdx
```diff
@@ -0,0 +1,319 @@
...
+In addition to these fields, all fields on the root `Query` or `Mutation` types also need a `@resolve` directive, since those types aren't owned by any service
+
+```graphql
+type Query {
+  rockets: [Rocket]! @resolve(graph: "rockets")
+  astronauts: [Astronaut]! @resolve(graph: "astronauts")
+}
+```
+
+### `@csdl` and `@graph`
```
queerviolet, line 274: I want to change `@composedGraph(version:)` to `@csdl(version:)` bc it feels like it more clearly indicates that it's the version of CSDL this schema requires, rather than the version of this composed graph.

docs/source/csdl-spec.mdx
```diff
@@ -0,0 +1,319 @@
...
+  @graph(name: "astronauts", url: "https://astronauts.api.com")
+  @csdl(version: "^1") {
+  query: Query
+}
+```
+
+## Divergence from GraphQL SDL
+
+CSDL is a subset of GraphQL. In particular, CSDL differs from GraphQL SDL in the following ways:
+- **Extensions are forbidden.** You cannot `extend type` or `extend interface` within a CSDL document.
```
queerviolet, line 296: I believe we never need them and this makes the csdl easier to use?

queerviolet: easier to consume, that is.

docs/source/csdl-spec.mdx
```diff
@@ -0,0 +1,319 @@
...
+
+`directive @resolve(graph: String!) on FIELD_DEFINITION`
+
+Once we know about owning services for types, we have to represent what service is responsible for resolving fields that are extended by other services. We do this with the `resolve` directive. For the `rocket` field on our `Astronaut` type and the `astronaut` field on our `Rocket` type, that would look like this:
+
+```graphql
+type Astronaut 
+	@key(fields: "{ id }", graph: "astronauts")
+	@key(fields: "{ id }", graph: "rockets")
+	@owner(graph: "astronauts")
```
queerviolet, line 245: what do we think about changing `@owner(graph:)` to `@default(graph:)`? i'm thinking about the discussions we've been having about something in-between value types and owned types. In principle, the only thing that's special about a type's owner is that it's the default resolver for fields (and also that you can only `@key` off fields from the owner type, but I don't think that's a requirement within this spec, but rather within federation)

queerviolet: @trevor-scheer @JakeDawkins @zionts ☝️

queerviolet:
Let's try out removing `@csdl_owner`. I don't believe we need the concept of type ownership represented in the csdl. Instead:
  - the default assumption is that any field without `@csdl_resolve` is "free". that is, it can be resolved by any subgraph which can return the type (so we can just tack it on to the query)
  - `@csdl_resolve` binds a field to a particular subgraph, optionally with `requires` and `provides`

This does mean that we now have to provide `@csdl_resolve` for every field of an entity, but that seems… fine? The explicitness is nice, and csdl is designed to be convenient to consume more than to produce (because it's expected to be mostly machine-generated).

docs/source/csdl-spec.md
```diff
@@ -0,0 +1,630 @@
...
+  - Jake Dawkins <jake@apollographql.com>
+  - Trevor Scheer <trevor@apollographql.com>
+  - Ashi Krishnan <ashi@apollographql.com>
+---
+
+## Introduction: What is CSDL?
+
+The Composed Schema Definition Language (CSDL) provides a way to describe a GraphQL Schema which is composed from one or more other GraphQL schemas.
+
+The CSDL is intended to be a single artifact which can power a graph router. For example, the CSDL replaces the service list configuration of Apollo Gateway.
```
zionts, line 17: nitpick: As a reader, this feels like a detail that I could take in two ways. One of those is, I think, intended -> "Currently, I see the configuration to my gateway as `serviceList` and now I should see configuration to my gateway as CSDL". The other (and where my head first went) is confusing -> "Currently, the `serviceList` tells my gateway how to compose. Now CSDL will replace that".

I think that, even more concretely, we may very well have an intermediate period where the `serviceList` configuration of Apollo gateway leads to a new version of composition running at startup of that server, and thus producing CSDL for the Apollo gateway to use as configuration of its query planner.

All to say, I think that maybe there's a small disagreement — I think the CSDL is intended to be a single artifact which can power a query planner. A graph router (aside: currently we don't call Apollo gateway that publicly; should we introduce that term here?) may contain more things as configuration, such as plugins to lifecycle, and it may even take an input to _create_ or _fetch_ a CSDL to pass to its component, a query planner.

docs/source/csdl-spec.md
```diff
@@ -0,0 +1,630 @@
...
+
+CSDL is a subset of the [GraphQL Schema Definition Language](https://spec.graphql.org/). It removes certain SDL features (for instance, type extensions are not permitted) while also requiring the definition of certain types and providing a suite of directives to describe graph topology. In particular, a CSDL can:
+- define [subgraphs](#subgraph) and bind them to [endpoints](#endpoint) (with the required [`csdl_Graph`](#csdl_Graph) type and [`@csdl_endpoint`](#csdl_endpoint) directive)
+- assign fields to subgraphs (with [`@csdl_resolve`](#resolve))
+- declare additional data required and provided by subgraph field resolvers (with [`@csdl_key`](#key), and [`@csdl_resolve`](#csdl_resolve))
```
zionts, line 22: I'll leave this comment when I get there below, but I think that `@csdl_key` is misleading as a term that bundles both the `@key` for a field's requirements and a field's `@requires` selection set.

docs/source/csdl-spec.md
```diff
@@ -0,0 +1,630 @@
...
+```graphql
+# Rockets Service
+type Query {
+  rockets: [Rocket]!
+}
+
+type Rocket @key(fields: "id") {
+  id: String! 
+  name: String!
+  captain: Astronaut @provides(fields: "tripId")
```
zionts, line 51: Assuming we're using the `@provides` to illustrate the composite nature of the (currently named) `@csdl_key` directive — should we include an `@requires` as well? I know that the overwhelming majority of _our_ federation usage relies on `@requires` and we don't have a single usage of `@provides` (our priorities are not query planner performance tricks), so I think if we're going to include things beyond the essential, `@requires` would probably be more universally understood and informative, if other teams are anything like ours.

docs/source/csdl-spec.md
```diff
@@ -0,0 +1,630 @@
...
+
+This example is non-normative, and represents just one possible way to compose [rockets](#rockets) and [astronauts](#astronauts) into CSDL.
+
+<a name=ex-rockets-and-astronauts href=#ex-rockets-and-astronauts class=listing>Example CSDL composing `rockets` and `astronauts`</a>
+```graphql
+schema @csdl(version: "^1") {
+  query: Query
+}
+
+enum csdl_Graph {
```
zionts, line 93: This is a surprising choice on first pass to use nested directives and enums to indicate the set of sub-graphs. What are the advantages of this approach vs. the existing parameterized directive `@csdl_graph` approach?

zionts: Ahh, it's because we can refer to that graph with all of its required metadata repeatably. That makes a lot of sense! I should have kept scrolling 😉 May I suggest that we use `@csdl_graph_meta` or something more generic? I suspect that we may want to have information like `revision: 1234` represented in the CSDL, and things like fall-back strategies that are at the sub-graph level down the road.

docs/source/csdl-spec.md
```diff
@@ -0,0 +1,630 @@
...
+    CSDL-->Router
+    Router-->Clients
+    Clients-->Router
+```
+
+- **Producer** the bit that generates CSDL. spec will place requirements on composers.
+- **Consumer** the bit that consumes CSDL. spec will place requirements on consumers.
+- **Composer** a kind of producer which composes subgraph schemas into CSDL. (no requirements on consumers, diagrammed for clarity.)
+- **Subgraphs** subgraphs which are composed by the composer (no requirements on subgraphs, diagrammed for clarity.)
+- **Router** a kind of consumer which serves CSDL as a graphql endpoint. no requirements on routers, these are described in the federation spec.
```
zionts, line 165: Does the federation spec call these routers nowadays?

Also, as a follow-on: Is there actually a specified way to get the CSDL _out_ of a running Router?

docs/source/csdl-spec.md
```diff
@@ -0,0 +1,630 @@
...
+
+## Structure
+
+A CSDL document must be a valid [GraphQL schema definition language](https://spec.graphql.org/draft/#sec-Type-System) document.
```
zionts, line 169: Does the GraphQL specification actually specify "schema definition language"? I was pretty sure that it was a term we made up to mean "the part of the GraphQL specification that people use in their schemas"

docs/source/csdl-spec.md
```diff
@@ -0,0 +1,630 @@
...
+CSDL removes certain SDL features, as described in the [divergence](#divergence-from-graphql-sdl) section. Nevertheless, CSDL consumers **should** be prepared to parse any valid GraphQL SDL. A CSDL document which contains disallowed SDL constructs **should** trigger validation errors after parsing.
+
+CSDL introduces a suite of types and directives. All CSDL-specific types and directives are prefixed with `csdl_` as recommended in [the GraphQL specification](https://spec.graphql.org/draft/#note-ca863).
+
+### Divergence from GraphQL SDL
+Formally, CSDL is a subset of the [GraphQL schema definition language](https://spec.graphql.org/draft/#sec-Type-System).
+
+In particular, CSDL differs from GraphQL SDL in the following ways:
+- **Extensions are forbidden.** You cannot `extend type` or `extend interface` within a CSDL document.
+- **`schema @csdl(version:)` is mandatory.** Conforming CSDL documents must have a `schema` declaration which must be annotated with a `@csdl` directive specifying, at a minimum, the semver for the CSDL spec in use.
```
zionts, line 180: YAS

docs/source/csdl-spec.md
```diff
@@ -0,0 +1,630 @@
...
+
+The `csdl_Graph` enum is used as input to csdl directives which link fields and types to subgraphs.
+
+## Scalars
+
+### `scalar csdl_SelectionSet`
+
+A GraphQL selection set represented with the same syntax as defined in [the GraphQL SDL](https://spec.graphql.org/draft/#sec-Selection-Sets).
+
+Note: Unlike the federation `_FieldSet` scalar, a `csdl_SelectionSet` must parse as a GraphQL selection set, so the selections must be surrounded by braces.
```
zionts, line 202: It might be worthwhile explaining the rationale behind such a discrepancy. Would we consider adding support for that format in federation as well & deprecating the current non-bracketed (or vice versa) for reasons of consistency?

docs/source/csdl-spec.md
```diff
@@ -0,0 +1,630 @@
...
+
+### `scalar csdl_Url`
+
+A [URL](https://www.w3.org/Addressing/URL/url-spec.html).
+
+### `scalar csdl_Version`
+
+A [Semantic versioning 2.0.0](https://semver.org/spec/v2.0.0.html) verison or range.
```
zionts, line 210: Small typo
```suggestion
A [Semantic versioning 2.0.0](https://semver.org/spec/v2.0.0.html) version or range.
```

docs/source/csdl-spec.md
```diff
@@ -0,0 +1,630 @@
...
+
+### Realized types
+
+A type is *realized* wherever a resolver from a particular *subgraph* returns that type. We use the notation `subgraph::Type` to represent `Type` realized by `subgraph` and `subgraph::Type.field` to represent the resolver for `Type.field` within `subgraph`.
+
+<a class="listing">Query with resolvers and realized types</a>
+```graphql
+query {          # resolver                    | realizes
+                 # ----------------------------|--------------
+  rockets {      # rockets::Query.rockets     -> rockets::Rocket
```
zionts, line 223: This small change would help me to understand things better 🤷 
```suggestion
  rockets {      # rockets::Query.rockets     -> rockets::[Rocket!]!
```

docs/source/csdl-spec.md
```diff
@@ -0,0 +1,630 @@
...
+## Data Model
+TK, draws heavily from federation, a little more general.
+
+### Realized types
+
+A type is *realized* wherever a resolver from a particular *subgraph* returns that type. We use the notation `subgraph::Type` to represent `Type` realized by `subgraph` and `subgraph::Type.field` to represent the resolver for `Type.field` within `subgraph`.
+
+<a class="listing">Query with resolvers and realized types</a>
+```graphql
+query {          # resolver                    | realizes
```
zionts, line 221: What does this look like for a union or interface resolver?

docs/source/csdl-spec.md
```diff
@@ -0,0 +1,630 @@
...
+                 # ----------------------------|--------------
+  rockets {      # rockets::Query.rockets     -> rockets::Rocket
+    captain {    # rockets::Rocket.astronaut  -> rockets::Astronaut
```
zionts, line 224: I wonder if it's worthwhile to distinguish the `rockets::Rocket.astronaut` resolver from the `rockets::_Entity.Rocket.astronaut`, since in practice they may have different implementations. I suspect that in theory, we would prefer that everyone have the exact same resolver for both, but I can tell you with certainty that many users don't have any authz on their `_entities` resolvers (which is a, uh, scary security concern)

docs/source/csdl-spec.md
```diff
@@ -0,0 +1,630 @@
...
+
+}
+```
+
+These distinctions are immaterial to the client. As far as clients are concerned, all selections on `Rocket` are the same.
+
+This illusion is maintained by the router, which must break incoming queries into subqueries. In the example above, the path `rockets.captain` is resolved by the `rockets::Rocket.astronaut` resolver, so its selection set is realized as a `rockets::Astronaut`. However, there is no `name` field for `rockets::Astronaut`—instead, the only resolver for `Astronaut.name` is `astronauts::Astronaut.name`. To process this query, the router must "convert" a `rockets::Astronaut` to an `astronauts::Astronaut` via an `_entities` query.
+
+### Portability
+A type is *portable to* a subgraph if it has a `@csdl_key` for that subgraph.
```
zionts, line 240: nitpick: In my head, the word I gravitate toward is "defined in" rather than "portable to" — the latter implies finality, whereas the former implies temporality.

docs/source/csdl-spec.md
```diff
@@ -0,0 +1,630 @@
...
+
+### Free / Bound fields
+
+Free fields can be resolved by any subgraph, bound fields can only be resolved by the subgraph they're bound to.
```
zionts, line 244: I like that you **bolded** things above when it was the first time you used a term, and I think it would be nicely consistent to do it here as well.

zionts: In a world where we want to move away from boundedness of fields (down the road), it may be tricky to get out of defining things like `@csdl_resolve` this way.

docs/source/csdl-spec.md
```diff
@@ -0,0 +1,630 @@
...
+  ROCKETS @csdl_endpoint(url: "https://rockets.api.com"),
+  ASTRONAUTS @csdl_endpoint(url: "https://astronauts.api.com"),
+}
+```
+
+### `@csdl_resolve`
+<code class='grammar'>
+directive @csdl_resolve(
+  graph: csdl_Graph!,
+  requires: csdl_SelectionSet,
```
zionts, line 323: Is it required that the selection set listed here is a valid selection set of the parent type? Would we say that's a convenience function, but in theory the planner could still support a root query in this place? (Somewhat prodding in a direction that might be enabling and somewhat asking about the reality of the world)

docs/source/csdl-spec.md
queerviolet: Do you mean `@csdl_resolve`?

docs/source/csdl-spec.md
queerviolet: I think if we want to attach more information to an endpoint, we can use different / more directives?

Is there anything that must be represented—i.e. is there endpoint-specific config that gateways always need which we should consider either adding to `@csdl_endpoint` or introducing new directives for?

docs/source/csdl-spec.md
queerviolet: not yet, and no.

docs/source/csdl-spec.md
queerviolet: i guess we could say "type system definition language," which is what the spec actually calls it.

queerviolet: changed it to "graphql interface description language"

docs/source/csdl-spec.md
queerviolet: So the reason the braces were introduced in the first place is that it was annoying (and introduced a copy, so some small perf impact) to add `{` and `}` to the `_FieldSet` in order to get it to parse as a [SelectionSet](https://spec.graphql.org/draft/#sec-Selection-Sets), which is specified as including the braces.

That's obviously not enough of a reason on its own, but the thing is, `_FieldSet`s don't actually correspond to any rule within the GraphQL query grammar, so (1) it's a little harder to specify, and (2) it's likely that many other implementers will also run into exactly the same issue we did in their own language's parsers.

I'd definitely consider deprecating `_FieldSet` within federation, but gosh what an annoying change for such small gains.

docs/source/csdl-spec.md
queerviolet: I need to flesh this out so it's clearer, but the point here is to conceptually break a type (`Rocket`) into different selection namespaces based on what you can actually select there.

Fields which are on a different type, or which are not known to the `rockets` subgraph, can't be queried directly.

`Query.rockets` runs in the `rockets` subgraph and returns `[Rocket!]!`. Within the selection set, the router can query any fields on `Rocket` which are known to the `rockets` subgraph. Thus, the real type is `rockets::Rocket`. In fact, all type variants—`Rocket`, `[Rocket]`, `Rocket!`, `[Rocket!]!`—have the same realized type, because it's *just* about what you can select.

docs/source/csdl-spec.md
queerviolet: good question man. good question.

docs/source/csdl-spec.md
queerviolet: They are different. `rockets::Query._entities` isn't explicitly mentioned here, it'll be implicitly used to port `rockets::Astronaut` to `astronauts::Astronaut`.

that *should* be mentioned here, but i'm really trying to not rewrite the federation spec. :/

docs/source/csdl-spec.md
queerviolet: I don't love the term portability/porting, but isn't defined in more confusing? What do you call non-federated types which are… uh… defined in a particular service, but not shared? or value types, which are *not* portable but have only free fields?

docs/source/csdl-spec.md
queerviolet: we're thinking along the same lines! I think in a future version of the spec we'll probably introduce a `csdl_FnQuery` input and more general `@csdl_resolve(query:)` form for this.

for the moment, i'd like to ship a spec version that we can support with only very minimal query planner changes.

queerviolet:
hey, after some discussion, versioning specs will be much easier if they are in [their own repo](https://github.com/apollographql/specs). moving discussion over to this pr: https://github.com/apollographql/specs/pull/1

PR #278, opened 2020-11-11
feat(gateway): Remove internal dep on `schema` from composition

trevor-scheer:
In an effort to more fully embrace CSDL within the Apollo ecosystem, this PR expands the Gateway's API to consume CSDL.

Currently, the query planner already depends on CSDL as input. This covers the majority of the Gateway's schema needs. All that remains is an internal dependency on a `GraphQLSchema` coming from composition.

It's fairly trivial to generate a `GraphQLSchema` from the CSDL itself, which unlocks the ability to provide CSDL directly to the gateway as input. This means that we can provide CSDL as configuration directly, or in the case of managed federation, host CSDL and significantly simplify the startup process of a managed gateway.

This change does introduce the ability to provide CSDL directly to the Gateway, however the remaining bits are strictly changes to internal details. With this change in place, our infra team can ship the work required to host CSDL which we can in turn depend on and simplify our waterfall of fetches and composition.

### DIFF
--- a/federation-js/src/composition/__tests__/composeAndValidate.test.ts
+++ b/federation-js/src/composition/__tests__/composeAndValidate.test.ts
@@ -12,12 +12,8 @@ import {
   typeSerializer,
   graphqlErrorSerializer,
 } from '../../snapshotSerializers';
-import {
-  compositionHasErrors,
-  CompositionResult,
-  CompositionSuccess,
-  CompositionFailure,
-} from '@apollo/federation';
+import { assertCompositionFailure, assertCompositionSuccess } from '../utils';
+import { compositionHasErrors } from '../compose';
 
 expect.addSnapshotSerializer(astSerializer);
 expect.addSnapshotSerializer(typeSerializer);
@@ -103,30 +99,6 @@ function permutateList<T>(inputArr: T[]) {
   permute(inputArr);
 
   return result;
-}
-
-// This assertion function should be used for the sake of convenient type refinement.
-// It should not be depended on for causing a test to fail. If an error is thrown
-// from here, its use should be reconsidered.
-export function assertCompositionSuccess(
-  compositionResult: CompositionResult,
-  message?: string,
-): asserts compositionResult is CompositionSuccess {
-  if (compositionHasErrors(compositionResult)) {
-    throw new Error(message || 'Unexpected test failure');
-  }
-}
-
-// This assertion function should be used for the sake of convenient type refinement.
-// It should not be depended on for causing a test to fail. If an error is thrown
-// from here, its use should be reconsidered.
-export function assertCompositionFailure(
-  compositionResult: CompositionResult,
-  message?: string,
-): asserts compositionResult is CompositionFailure {
-  if (!compositionHasErrors(compositionResult)) {
-    throw new Error(message || 'Unexpected test failure');
-  }
 }
 
 it('composes and validates all (24) permutations without error', () => {
--- a/federation-js/src/composition/__tests__/composeAndValidate.test.ts
+++ b/federation-js/src/composition/__tests__/composeAndValidate.test.ts
@@ -12,6 +12,12 @@ import {
   typeSerializer,
   graphqlErrorSerializer,
 } from '../../snapshotSerializers';
+import {
+  compositionHasErrors,
+  CompositionResult,
+  CompositionSuccess,
+  CompositionFailure,
+} from '@apollo/federation';
 
 expect.addSnapshotSerializer(astSerializer);
 expect.addSnapshotSerializer(typeSerializer);
@@ -99,6 +105,30 @@ function permutateList<T>(inputArr: T[]) {
   return result;
 }
 
+// This assertion function should be used for the sake of convenient type refinement.
+// It should not be depended on for causing a test to fail. If an error is thrown
+// from here, its use should be reconsidered.
+export function assertCompositionSuccess(
+  compositionResult: CompositionResult,
+  message?: string,
+): asserts compositionResult is CompositionSuccess {
+  if (compositionHasErrors(compositionResult)) {
+    throw new Error(message || 'Unexpected test failure');
+  }
+}
+
+// This assertion function should be used for the sake of convenient type refinement.
+// It should not be depended on for causing a test to fail. If an error is thrown
+// from here, its use should be reconsidered.
+export function assertCompositionFailure(
+  compositionResult: CompositionResult,
+  message?: string,
+): asserts compositionResult is CompositionFailure {
+  if (!compositionHasErrors(compositionResult)) {
+    throw new Error(message || 'Unexpected test failure');
+  }
+}
+
 it('composes and validates all (24) permutations without error', () => {
   permutateList([
     inventoryService,
@@ -106,15 +136,12 @@ it('composes and validates all (24) permutations without error', () => {
     accountsService,
     productsService,
   ]).map((config) => {
-    const { errors } = composeAndValidate(config);
+    const compositionResult = composeAndValidate(config);
 
-    if (errors.length) {
-      console.error(
-        `Errors found with composition [${config.map((item) => item.name)}]`,
-      );
-    }
-
-    expect(errors).toHaveLength(0);
+    assertCompositionSuccess(
+      compositionResult,
+      `Errors found with composition [${config.map((item) => item.name)}]`,
+    );
   });
 });
 
@@ -146,9 +173,12 @@ it('errors when a type extension has no base', () => {
     name: 'serviceB',
   };
 
-  const { errors } = composeAndValidate([serviceA, serviceB]);
-  expect(errors).toHaveLength(1);
-  expect(errors).toMatchInlineSnapshot(`
+  const compositionResult = composeAndValidate([serviceA, serviceB]);
+
+  assertCompositionFailure(compositionResult);
+
+  expect(compositionResult.errors).toHaveLength(1);
+  expect(compositionResult.errors).toMatchInlineSnapshot(`
     Array [
       Object {
         "code": "EXTENSION_WITH_NO_BASE",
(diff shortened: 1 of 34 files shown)


federation-js/src/composition/__tests__/composeAndValidate.test.ts
```diff
@@ -99,22 +105,43 @@ function permutateList<T>(inputArr: T[]) {
...
+    const compositionResult = composeAndValidate(config);
 
-    if (errors.length) {
-      console.error(
-        `Errors found with composition [${config.map((item) => item.name)}]`,
-      );
-    }
-
-    expect(errors).toHaveLength(0);
+    assertCompositionSuccess(
```
trevor-scheer, line 141: This violates the intention expressed in the comments above. This can be wrapped with an `expect().not.toThrow()`.

federation-js/src/composition/compose.ts
```diff
@@ -120,9 +120,6 @@ export interface KeyDirectivesMap {
  */
 type ValueTypes = Set<string>;
 
-export type ComposedGraphQLSchema = GraphQLSchema & {
```
trevor-scheer, line 123: This type is no longer needed since the `schema` we're constructing in memory no longer needs the `extensions` field with the service info.

federation-js/src/composition/composeAndValidate.ts
```diff
@@ -7,12 +7,38 @@ import {
...
+  /** @deprecated Use composedSdl instead */
+  schema: GraphQLSchema;
+  errors: GraphQLError[];
+}
+
+export interface CompositionSuccess {
+  /** @deprecated Use composedSdl instead */
+  schema: GraphQLSchema;
+  composedSdl: string;
+}
```
trevor-scheer, line 26: Included some type refinement work in this PR. Handling of composition results is a bit more strictly typed now. You either have `errors` _or_ `composedSdl`, which is now enforced by the type system itself.

gateway-js/src/__tests__/queryPlanCucumber.test.ts
```diff
@@ -25,11 +25,8 @@ features.forEach((feature) => {
         let queryPlan: QueryPlan;
         let options: BuildQueryPlanOptions = { autoFragmentization: false };
 
-        const { schema, errors, queryPlannerPointer } = getFederatedTestingSchema();
-
-        if (errors && errors.length > 0) {
-          throw new GraphQLSchemaValidationError(errors);
-        }
+        // throws on composition errors
```
trevor-scheer, line 28: This `if` statement is redundant, as `getFederatedTestingSchema` already handled this same logic internally.

gateway-js/src/csdlToSchema.ts
```diff
@@ -0,0 +1,10 @@
+import { extendSchema, GraphQLSchema, parse } from 'graphql';
+
+export function csdlToSchema(csdl: string) {
+  let schema = new GraphQLSchema({
+    query: undefined,
+  });
+
+  const parsed = parse(csdl);
+  return extendSchema(schema, parsed, { assumeValidSDL: true });
+}
```
trevor-scheer, line 10: This is the main topic of the PR, cc @abernix

gateway-js/src/index.ts
```diff
@@ -495,19 +497,23 @@ export class ApolloGateway implements GraphQLService {
...
+      this.logger.debug('Schema loaded and ready for execution');
+
+      // This is a workaround for automatic wrapping of all fields, which Apollo
+      // Server does in the case of implementing resolver wrapping for plugins.
+      // Here we wrap all fields with support for resolving aliases as part of the
+      // root value which happens because aliases are resolved by sub services and
+      // the shape of the root value already contains the aliased fields as
+      // responseNames
+      return {
+        schema: wrapSchemaWithAliasResolver(csdlToSchema(composedSdl)),
```
trevor-scheer, line 513: Here is the new usage of `csdlToSchema`

federation-js/src/composition/utils.ts
```diff
@@ -59,7 +60,12 @@ export function mapFieldNamesToServiceName<Node extends { name: NameNode }>(
 }
 
 export function findDirectivesOnTypeOrField(
```
trevor-scheer, line 62: Rename

package-lock.json
```diff
@@ -28,6 +28,107 @@
...
         "make-fetch-happen": "^8.0.0",
         "pretty-format": "^26.0.0"
+      },
+      "dependencies": {
+        "apollo-cache-control": {
+          "version": "0.11.4",
+          "resolved": "https://registry.npmjs.org/apollo-cache-control/-/apollo-cache-control-0.11.4.tgz",
+          "integrity": "sha512-FUKE8ASr8GxVq5rmky/tY8bsf++cleGT591lfLiqnPsP1fo3kAfgRfWA2QRHTCKFNlQxzUhVOEDv+PaysqiOjw==",
+          "requires": {
+            "apollo-server-env": "^2.4.5",
```
abernix, line 38: The entirety of this `package-lock.json` change seems unexpected.

package-lock.json
```diff
@@ -3923,6 +4024,7 @@
       "version": "0.11.2",
       "resolved": "https://registry.npmjs.org/apollo-cache-control/-/apollo-cache-control-0.11.2.tgz",
       "integrity": "sha512-a1YdQLWzwuvZC7+xtJ8zno04pGRZG9595cX8W5YoXs2GeJqYdk+G3/tgdTgpLN827Vmgka0xkWfmxQUYm2xZUw==",
+      "dev": true,
```
abernix, line 4027: Granted I already said "entirety of this `package-lock.json`" in my above comment, this is _still_ unexpected given the changes in this change-set?

federation-js/src/composition/composeAndValidate.ts
```diff
@@ -36,10 +36,9 @@ export function composeAndValidate(
 ): CompositionResult {
   const errors = validateServicesBeforeNormalization(serviceList);
 
-  const normalizedServiceList = serviceList.map(({ name, url, typeDefs }) => ({
-    name,
-    url,
+  const normalizedServiceList = serviceList.map(({ typeDefs, ...rest }) => ({
     typeDefs: normalizeTypeDefs(typeDefs),
+    ...rest
```
abernix: This resolves one of my previous comments I already left on a previous comment!

federation-js/src/composition/compose.ts
```diff
@@ -158,10 +158,7 @@ export function buildMapsFromServiceList(serviceList: ServiceDefinition[]) {
       ) {
         const typeName = definition.name.value;
 
-        for (const keyDirective of findDirectivesOnTypeOrField(
-          definition,
-          'key',
-        )) {
+        for (const keyDirective of findDirectivesOnNode(definition, 'key')) {
```
abernix, line 161: This also resolves a previous comment on another commit.

abernix (approved):
There's a `package-lock.json` change that seems unexpected and a `CHANGELOG.md` would be good noting the breaking change on the exported type, but presuming you back that out or if I'm missing something, then LGTM.

(Squash in the end?)

federation-js/src/composition/composeAndValidate.ts
```diff
@@ -7,12 +7,38 @@ import {
...
+  return 'errors' in compositionResult;
+}
+
+export function composeAndValidate(
+  serviceList: ServiceDefinition[],
+): CompositionResult {
   const errors = validateServicesBeforeNormalization(serviceList);
 
-  const normalizedServiceList = serviceList.map(({ name, typeDefs }) => ({
+  const normalizedServiceList = serviceList.map(({ name, url, typeDefs }) => ({
```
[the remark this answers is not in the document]
abernix: Fixed in a later commit.

federation-js/src/composition/compose.ts
```diff
@@ -655,5 +649,5 @@ export function composeServices(services: ServiceDefinition[]) {
    * and every field that was extended. Fields that were _not_ extended (added on the base type by the owner),
    * there is no `serviceName`, and we should refer to the type's `serviceName`
    */
-  return { schema: schema as ComposedGraphQLSchema, errors };
+  return { schema, errors };
```
abernix, line 652: A small part of me is curious if this `composeServices` should have symmetry with `composeAndValidate`? (In that it would return `CompositionResult`, but without applying the validations.) Is that even conceivable?

federation-js/src/composition/composeAndValidate.ts
```diff
@@ -7,13 +7,38 @@ import {
...
+  /** @deprecated Use composedSdl instead */
+  schema: GraphQLSchema;
+  errors: GraphQLError[];
+}
+
+export interface CompositionSuccess {
+  /** @deprecated Use composedSdl instead */
+  schema: GraphQLSchema;
+  composedSdl: string;
+}
```
abernix, line 26: Perhaps this is no longer the way it needs to be done and maybe it's a result of the type predicates you've put in place, but I'm surprised the `CompositionSuccess` type doesn't have to explicitly have `errors?: undefined` — and conversely — the `CompositionFailure` doesn't have to have `composedSdl?: undefined`.  (I've always found this odd behavior in TypeScript to make sure these aren't commingled, but I seem to recall it in a few places in our code.)

federation-js/src/composition/compose.ts
trevor-scheer: Yeah, I like this suggestion quite a bit and nice observation!
d8aeb2a

federation-js/src/composition/composeAndValidate.ts
trevor-scheer: I was shocked to find that I could return, for example:
`{ schema, errors: [], composedSdl: ''}`
...without a TS error.

I thought the disjointedness of the two objects would be enforced by TS, but it seems that's not the case.

Implemented your suggestion in 5170542 and that seems to do the trick. Also updated the predicate to handle the new potential existence of the key with an `undefined` value.

package-lock.json
trevor-scheer: Yeah, I think there was some good ol npm fun going on here.
6937cd8

package-lock.json
trevor-scheer: 6937cd8

PR #280, opened 2020-11-15
fix(federation): differentiate field and input value names

patrick91:
This only contains the test at the moment :)

I found a bug where I had at type like this:

```graphql

type Campaign implements Node {
  id: ID!
  identifier: String
  event(identifier: String!): Event
}
```

in two services, and apollo was still complaining about the types being different.

Turns out that the diff is comparing the field `identifier` with the argument `identifier`,
and they have different types.

So, I've added a test for that. I suspect we might have a similar issue with different fields 
having the same argument but with different type. I might add a test for that later :)

EDIT: not sure why the tests didn't fail here

```
Summary of all failing tests
 FAIL  federation-js/src/composition/validate/sdl/__tests__/uniqueFieldDefinitionNames.test.ts (5.973s)
  ● UniqueFieldDefinitionNames › permits duplicate field names for › object type definitions field and arguments with different type

    expect(received).toHaveLength(expected)

    Expected length: 0
    Received length: 2
    Received array:  [[GraphQLError: Field "Campaign.identifier" can only be defined once.], [GraphQLError: Field "Campaign.event" can only be defined once.]]

      288 |         UniqueFieldDefinitionNames,
      289 |       ]);
    > 290 |       expect(errors).toHaveLength(0);
          |                      ^
      291 |     });
      292 |   });
      293 | });

      at Object.<anonymous> (src/composition/validate/sdl/__tests__/uniqueFieldDefinitionNames.test.ts:290:22)
```

Ah, the tests were not running: https://github.com/apollographql/federation/pull/281

### DIFF
--- a/federation-js/CHANGELOG.md
+++ b/federation-js/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 > The changes noted within this `vNEXT` section have not been released yet.  New PRs and commits which introduce changes should include an entry in this `vNEXT` section as part of their development.  When a release is being prepared, a new header will be (manually) created below and the appropriate changes within that release will be moved into the new section.
 
+- Fix check for value types when having fields and arguments with the same name [PR #280](https://github.com/apollographql/federation/pull/280)
+
 ## v0.20.6
 
 - No changes, but please note that `v0.20.5` was a botched release, with no update to the `@apollo/query-planner-wasm` package that was needed. If you're seeing an error similar to `This data graph is missing a valid configuration. unreachable`, please upgrade to at least this patch release.
--- a/federation-js/src/composition/__tests__/compose.test.ts
+++ b/federation-js/src/composition/__tests__/compose.test.ts
@@ -476,11 +476,11 @@ describe('composeServices', () => {
         const { schema, errors } = composeServices([serviceA, serviceB]);
         expect(schema).toBeDefined();
         expect(errors).toMatchInlineSnapshot(`
-                    Array [
-                      [GraphQLError: Field "Product.name" can only be defined once.],
-                      [GraphQLError: There can be only one type named "Product".],
-                    ]
-                `);
+          Array [
+            [GraphQLError: Field "Product.name" can only be defined once.],
+            [GraphQLError: There can be only one type named "Product".],
+          ]
+        `);
 
         const product = schema.getType('Product') as GraphQLObjectType;
 
--- a/federation-js/src/composition/utils.ts
+++ b/federation-js/src/composition/utils.ts
@@ -384,6 +384,10 @@ export function diffTypeNodes(
     [fieldName: string]: string[];
   } = Object.create(null);
 
+  const inputValuesDiff: {
+    [inputName: string]: string[];
+  } = Object.create(null);
+
   const unionTypesDiff: {
     [typeName: string]: boolean;
   } = Object.create(null);
@@ -399,7 +403,7 @@ export function diffTypeNodes(
     definitions: [firstNode, secondNode],
   };
 
-  function fieldVisitor(node: FieldDefinitionNode | InputValueDefinitionNode) {
+  function fieldVisitor(node: FieldDefinitionNode) {
     const fieldName = node.name.value;
 
     const type = print(node.type);
@@ -419,9 +423,33 @@ export function diffTypeNodes(
     }
   }
 
+  /** Similar to fieldVisitor but specific for input values, so we don't store
+   * fields and arguments in the same place.
+   */
+
+  function inputValueVisitor(node: InputValueDefinitionNode) {
+    const fieldName = node.name.value;
+
+    const type = print(node.type);
+
+    if (!inputValuesDiff[fieldName]) {
+      inputValuesDiff[fieldName] = [type];
+      return;
+    }
+
+    // If we've seen this input value twice and the types are the same,
+    // remove it from the diff result
+    const inputValueTypes = inputValuesDiff[fieldName];
+    if (inputValueTypes[0] === type) {
+      delete inputValuesDiff[fieldName];
+    } else {
+      inputValueTypes.push(type);
+    }
+  }
+
   visit(document, {
     FieldDefinition: fieldVisitor,
-    InputValueDefinition: fieldVisitor,
+    InputValueDefinition: inputValueVisitor,
     UnionTypeDefinition(node) {
       if (!node.types) return BREAK;
       for (const namedTypeNode of node.types) {
@@ -481,6 +509,7 @@ export function diffTypeNodes(
     name: typeNameDiff,
     kind: kindDiff,
     fields: fieldsDiff,
+    inputValues: inputValuesDiff,
     unionTypes: unionTypesDiff,
     locations: Array.from(locationsDiff),
     args: argumentsDiff,
@@ -497,7 +526,7 @@ export function typeNodesAreEquivalent(
   firstNode: TypeDefinitionNode | TypeExtensionNode | DirectiveDefinitionNode,
   secondNode: TypeDefinitionNode | TypeExtensionNode | DirectiveDefinitionNode,
 ) {
-  const { name, kind, fields, unionTypes, locations, args } = diffTypeNodes(
+  const { name, kind, fields, inputValues, unionTypes, locations, args } = diffTypeNodes(
     firstNode,
     secondNode,
   );
@@ -506,6 +535,7 @@ export function typeNodesAreEquivalent(
     name.length === 0 &&
     kind.length === 0 &&
     Object.keys(fields).length === 0 &&
+    Object.keys(inputValues).length === 0 &&
     Object.keys(unionTypes).length === 0 &&
     locations.length === 0 &&
     Object.keys(args).length === 0
--- a/federation-js/src/composition/validate/sdl/__tests__/uniqueFieldDefinitionNames.test.ts
+++ b/federation-js/src/composition/validate/sdl/__tests__/uniqueFieldDefinitionNames.test.ts
@@ -264,5 +264,60 @@ describe('UniqueFieldDefinitionNames', () => {
       ]);
       expect(errors).toHaveLength(0);
     });
+
+    it('object type definitions field and arguments with different type', () => {
+      const typeDefs = gql`
+        type Campaign implements Node {
+          identifier: String
+          event(identifier: String!): Event
+        }
+      `;
+
+      const [definitions] = createDocumentsForServices([
+        {
+          typeDefs,
+          name: 'serviceA',
+        },
+        {
+          typeDefs,
+          name: 'serviceB',
+        },
+      ]);
+
+      const errors = validateSDL(definitions, schema, [
+        UniqueFieldDefinitionNames,
+      ]);
+      expect(errors).toHaveLength(0);
+    });
+  });
+
+  it('object type definitions order does not impact diffing', () => {
+    const [definitions] = createDocumentsForServices([
+      {
+        typeDefs: gql`
+          type Order implements Node {
+            questions(answers: OrderAnswersInput): Questions
+            fees(answers: OrderAnswersInput): Fees
+            answers: [Answer]
+          }
+        `,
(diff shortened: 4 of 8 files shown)


trevor-scheer:
Interesting bug, and good catch! Thanks for opening this - is this something you're interested in fixing as well?

patrick91:
@trevor-scheer I can give a try on the weekend! Do you have any pointer?

trevor-scheer:
Awesome! I see the issue - let me know if you'd like me to go into any more detail. I don't want to steal the solution from you if you're interested in the solve, but happy to provide as much assistance as you're interested in.

This object seems to be doing a bit more work than it should within the validator:
https://github.com/apollographql/federation/blob/e8fae648c8fd8d523b36bbaeb480c34709ba7033/federation-js%2Fsrc%2Fcomposition%2Fvalidate%2Fsdl%2FuniqueFieldDefinitionNames.ts#L68-L70

patrick91:
@trevor-scheer I think I'm having some troubles understanding the code. I think the issue is in `diffTypeNodes` here: https://github.com/apollographql/federation/blob/main/federation-js/src/composition/utils.ts#L379-L424

I wonder why a visitor has been used for this function, aren't type definitions basically flat? If I understood correctly with the current code what happens is that both fields names and argument names are put inside `fieldsDiff` which doesn't seem to be correct. I'm not sure what the best solution would be. Do you have any suggestions?

trevor-scheer:
@patrick91 you're on the right track! `typeDefs` can be iterated over but it's an AST - it's flat in some senses but you will end up doing a fair amount of "walking" the tree yourself without the visitor.

You're right that `fieldsDiff` is doing double duty here, so we need an `inputValueDiff` as well.

As far as the function goes, I can imagine two solutions:

1) Currently this abstraction works for both visitors, but it needs to operate on 2 different diff objects. We can turn `fieldVisitor` into a function which returns a visitor function. The argument to our factory function would be the diff.

```
function fieldVisitorFactory(diff) {
  return function visitor(node) {
    //...previous fieldVisitor logic captured here
  }
}
```

2) Split the abstraction and have two very similar visitor functions which are single purposed and operate on different diffs (one for `FieldDefinition` and one for `InputValueDefinition`).

Is that helpful? Let me know if you have any more questions 🙂

patrick91:
@trevor-scheer took me a while to get back to you, but I think I managed to make it work, I went for suggestion 2, let me know how it looks. The tests are passing on my machine 😊

federation-js/src/composition/utils.ts
```diff
@@ -480,7 +508,7 @@ export function diffTypeNodes(
   return {
     name: typeNameDiff,
     kind: kindDiff,
-    fields: fieldsDiff,
+    fields: { ...fieldsDiff, ...inputValuesDiff },
```
trevor-scheer, line 511: I think we want to break this out into its own key rather than combine with fields when we return.

So we might instead return:
```js
{
  // ...,
  fields: fieldsDiff,
  inputValues: inputValuesDiff,
  // ...,
}
```

When we combine the two objects like this, we still have a potential naming collision. You can imagine the case where we had a real `fieldsDiff` and no `inputValuesDiff`, but they have a name in common (similar to the case you encountered). We could have these two diffs:

```js
// fieldsDiff
{ identifier: ['String', 'String!'] }

// inputValuesDiff
{ identifier: [] } 
```

And now the resulting diff we report is:
```js
{ identifier: [] }
```
...due to the "overwriting" nature of spread and the way we've overloaded the `fields` key here.

When you make this change, please take a look at the `typeNodesAreEquivalent` function below. I'd like to see your new `inputValues` object receive a similar treatment to the `fields` object within that function.

trevor-scheer (changes requested):
Thanks @patrick91! This is looking great, I have one more comment for you to address and this should be ready to ship.

Please add a changelog entry to the `federation-js/CHANGELOG.md` file as well.

federation-js/src/composition/validate/sdl/uniqueTypeNamesWithFields.ts
```diff
@@ -65,9 +65,9 @@ export function UniqueTypeNamesWithFields(
       // By inspecting the diff, we can warn when field types mismatch.
       // A diff entry will exist when a field exists on one type and not the other, or if there is a type mismatch on the field
       // i.e. { sku: [Int, String!], color: [String] }
-      const { kind, fields } = diffTypeNodes(node, duplicateTypeNode);
+      const { kind, fields, inputValues } = diffTypeNodes(node, duplicateTypeNode);
 
-      const fieldsDiff = Object.entries(fields);
+      const fieldsDiff = Object.entries(fields).concat(Object.entries(inputValues));
```
patrick91, line 70: does this make sense @trevor-scheer ?

I think we might need to update `checkFieldUniquenessExcludingValueTypes` as well :)

trevor-scheer: I think it might _work_ but I'm not sure the error reporting will be very accurate if we group these diffs into the `fieldsDiff`. Notice down below (L97) when we iterate over `fieldsDiff`, we might push error messages that relate to a field, but the error code and message don't really hint at input values. It seems appropriate to introduce another error code, like `VALUE_TYPE_INPUT_VALUE_MISMATCH` and have some similar iteration over the `inputValuesDiff` which contributes to the `typesHaveSameShape` evaluation. Does that make sense? It's similar to what you're already doing, but it does split out the iteration over `inputValues` so we can provide a more targeted error message.

Feel free to draw inspiration from the format and wording of existing error messages, or I'd be happy to provide guidance there. There's also some documentation around our error codes which would need a matching update:
https://github.com/apollographql/federation/blob/8b5ebc6d868702dc08dc966554ae4515d2114edd/docs%2Fsource%2Ferrors.md#L89

trevor-scheer: I think you're right that an update to checkFieldUniquenessExcludingValueTypes makes sense as well.

patrick91: not really happy with the code duplication, I was thinking of moving the check to a function, but we'd need to change a bit how we report the error, so might not be worth it :)

federation-js/src/composition/__tests__/compose.test.ts
```diff
@@ -476,11 +476,10 @@ describe('composeServices', () => {
         const { schema, errors } = composeServices([serviceA, serviceB]);
         expect(schema).toBeDefined();
         expect(errors).toMatchInlineSnapshot(`
-                    Array [
-                      [GraphQLError: Field "Product.name" can only be defined once.],
```
trevor-scheer, line 480: I'm unsure why but it looks like we lost an error that should still be here

federation-js/src/composition/validate/sdl/uniqueTypeNamesWithFields.ts
```diff
@@ -121,9 +121,38 @@ export function UniqueTypeNamesWithFields(
...
 
+      const inputValuesDiff = Object.entries(inputValues);
+
+      const typesHaveSameInputValuesShape =
+        inputValuesDiff.length === 0 ||
+        inputValuesDiff.every(([name, types]) => {
+          if (types.length === 2) {
+            possibleErrors.push(
+              errorWithCode(
+                'VALUE_TYPE_INPUT_VALUE_MISMATCH',
```
trevor-scheer, line 132: Can we get a test case that exercises this code path and demonstrates what will trigger this error?

federation-js/src/composition/validate/sdl/uniqueTypeNamesWithFields.ts
```diff
@@ -121,9 +121,38 @@ export function UniqueTypeNamesWithFields(
...
+                `${logServiceAndType(
+                  duplicateTypeNode.serviceName!,
+                  typeName,
+                  name,
+                )}A field was defined differently in different services. \`${
```
trevor-scheer, line 137: I think, more specifically, "a field's input type" was defined differently

federation-js/src/composition/validate/sdl/uniqueTypeNamesWithFields.ts
patrick91: done! I wish we could access the field name to improve the error message, but alas I wasn't able to :)

federation-js/CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 > The changes noted within this `vNEXT` section have not been released yet.  New PRs and commits which introduce changes should include an entry in this `vNEXT` section as part of their development.  When a release is being prepared, a new header will be (manually) created below and the appropriate changes within that release will be moved into the new section.
 
+- Fix check identical types when having fields and arguments wit the same name [PR #280](https://github.com/apollographql/federation/pull/280)
```
trevor-scheer, line 7:
```suggestion
- Fix check for value types when having fields and arguments with the same name [PR #280](https://github.com/apollographql/federation/pull/280)
```

federation-js/src/composition/validate/sdl/__tests__/uniqueFieldDefinitionNames.test.ts
```diff
@@ -264,5 +264,60 @@ describe('UniqueFieldDefinitionNames', () => {
...
+      ]);
+
+      const errors = validateSDL(definitions, schema, [
+        UniqueFieldDefinitionNames,
+      ]);
+      expect(errors).toHaveLength(0);
+    });
+  });
+
+  it('object type definitions field different order', () => {
```
trevor-scheer, line 294: Is this test just asserting that order is irrelevant for fields when determining value types? Just want to make sure I understand this test isn't related to your changes (but let's keep it!)

federation-js/src/composition/validate/sdl/uniqueTypeNamesWithFields.ts
trevor-scheer: I see. I think we can find a way to pass that info down through the diff but I'm really happy with where we've landed here. This feels ready to land to me and I don't want to block getting this released. If you don't mind opening an issue that references this PR comment that would be wonderful!

trevor-scheer (approved):
This has come together really nicely, @patrick91. Huge thank you for your initial efforts and timely updates! 🎉 

Did you have any lingering concerns with `checkFieldUniquenessExcludingValueTypes`? I don't think we addressed that yet. After quickly refamiliarizing myself and reading the comments, it seems right to leave it as is, but I'm open to hearing otherwise.

federation-js/src/composition/validate/sdl/__tests__/uniqueFieldDefinitionNames.test.ts
patrick91: yup! it was failing before we did the change to fieldDiff, oddly :)

I've updated the test to have a better name (and made the schema nicer too) :)

patrick91:
@trevor-scheer thank you for the patience!

Regarding `checkFieldUniquenessExcludingValueTypes` it seems fine to me now, my previous commit fixed the regression I added, so now the behaviour is the same as before, unless there are `inputValues`, see here: https://github.com/apollographql/federation/pull/280/files#diff-73bf495bb0ca51439e86b55e70ed176aa970de874bd6801cf8950fa90d22d964R161

If this is ready to merge, do you want me to squash the commits?

federation-js/src/composition/validate/sdl/uniqueTypeNamesWithFields.ts
patrick91: done :)

trevor-scheer:
@patrick91 That would be great!

trevor-scheer:
Thanks again @patrick91! I'll aim to get a patch release out for this tomorrow.

trevor-scheer:
This is released via https://github.com/apollographql/federation/commit/ab3939415d1f4cc423435e58cf2182ead3258500!

```
@apollo/federation@0.20.7
@apollo/gateway@0.21.4
```

PR #281, opened 2020-11-15
Add action for running npm tests

patrick91:
This PR adds an action to run the npm tests via GitHub actions.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,51 @@
+name: npm tests
+
+on:
+  pull_request:
+
+jobs:
+
+  test:
+    name: Tests
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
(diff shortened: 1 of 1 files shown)


trevor-scheer:
This is curious. I see why you were inclined to open this PR - normally our tests run in CircleCI and would show up on the checks list here, but I see they aren't for neither this PR nor the other one you've opened.

cc @abernix - possible that Circle is only running for members of the Apollo org? I'll look into this Monday.

abernix:
Thanks for opening this and taking the initiative! Apparently, the option to build from forks was disabled on our CircleCI configuration, which must have been overlooked when we migrated from the `apollographql/apollo-server` repository in https://github.com/apollographql/federation/pull/134. 

image

Any existing PRs probably just need an empty commit pushed to them in order to re-initiate the CircleCI run, but I'll close this as it shouldn't be necessary.  Thanks again!

PR #285, opened 2020-11-16
Apply `repeatable` to CSDL directives `@key` and `@graph`

### LINKED ISSUE #279: CSDL: repeated directives result in an invalid GraphQL document
Currently, the CSDL that we generate to pass into the Rust query planner is not always a valid GraphQL document due to multiple instances of particular directives appearing in the generate artifact (e.g., `@graph` and `@key` directives), but not having the declarations of those directives be [marked as `repeatable`](http://spec.graphql.org/draft/#sec-Type-System.Directives).

While there are cases where it doesn't result in an invalid GraphQL document (i.e., when there is only one `@graph` or one `@key` directive), this omission will otherwise result in an invalid GraphQL document.

We should mark the directive declarations as `repeatable`.

trevor-scheer:
Currently, our generated CSDL can be invalid since we have repeated usages of our
`@key` and `@graph` directives without them being marked as `repeatable`.

This small change marks them as repeatable and updates our tests and fixtures to match.

Fixes #279

### DIFF
--- a/stargate/crates/graphql-parser/src/schema/format.rs
+++ b/stargate/crates/graphql-parser/src/schema/format.rs
@@ -381,6 +381,9 @@ impl<'a> Displayable for DirectiveDefinition<'a> {
         f.write("directive @");
         f.write(self.name.as_ref());
         format_arguments(&self.arguments, f);
+        if self.is_repeatable {
+            f.write(" repeatable");
+        }
         if !self.locations.is_empty() {
             f.write(" on ");
             let mut first = true;
@@ -420,3 +423,23 @@ impl_display!(
     InputObjectTypeExtension,
     DirectiveDefinition,
 );
+
+#[cfg(test)]
+mod test {
+    use super::Document;
+    use crate::format::Style;
+    use crate::schema::grammar::parse_schema;
+
+    fn ast(s: &str) -> Document {
+        parse_schema(&s).unwrap()
+    }
+
+    #[test]
+    fn directive() {
+        assert_eq!(
+            ast("directive @key(fields: String!, graph: String!) repeatable on OBJECT")
+                .format(&Style::default()),
+            "directive @key(fields: String!, graph: String!) repeatable on OBJECT\n"
+        );
+    }
+}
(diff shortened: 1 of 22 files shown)


trevor-scheer:
During publish tarballs step:
```
#!/bin/bash -eo pipefail
DEBUG=lerna npx lerna publish \
  --no-git-tag-version \
  --no-push \
  --registry=http://localhost:4873/ \
  --yes \
  from-package

lerna notice cli v3.22.1
lerna info versioning independent
lerna info ci enabled
lerna ERR! EUNCOMMIT Working tree has uncommitted changes, please commit or remove the following changes before continuing:
lerna ERR! EUNCOMMIT  M stargate/crates/graphql-parser/tests/tests.rs
```

I have no idea why this file is being modified during CI workflows, but it's very frustrating and I'm out for the day. Ideas are welcome.

.circleci/config.yml
```diff
@@ -38,6 +38,12 @@ commands:
       - oss/install_specific_npm_version
       - checkout
       - oss/npm_clean_install_with_caching
+      # npm ci results in some unexpected FS writes, which this undoes
+      # more specifically, the monorepo-prepare / wasm-pack scripts
+      # are causing these writes in CI and can't be reproduced locally.
+      - run:
+          name: Git reset
+          command: git reset HEAD --hard
```
trevor-scheer, line 46: cc @abernix, I found a band-aid which _works_ but it certainly isn't solving the root of the issue.

stargate/crates/graphql-parser/src/schema/format.rs
```diff
@@ -420,3 +423,23 @@ impl_display!(
...
+
+    #[test]
+    fn directive() {
+        assert_eq!(
+            ast("directive @key(fields: String!, graph: String!) repeatable on OBJECT")
+                .format(&Style::default()),
+            "directive @key(fields: String!, graph: String!) repeatable on OBJECT\n"
+        );
+    }
+}
```
trevor-scheer, line 445: @queerviolet I didn't see a place where the formatter is tested. Feedback welcome on the approach here!

queerviolet: We do round trip tests on the entire test corpus (ensuring the formatted output == the input file content), so this should already be tested through that. You could try breaking the formatter and ensuring the round trip test fails.

queerviolet (approved):
looks good! you can keep or drop the formatter test as you like. i think it's sufficiently tested with the round trips, but more tests aren't a bad thing

abernix (approved):

PR #287, opened 2020-11-18
chore(infra): Add clarity to the generated tests file

trevor-scheer:
This change should make it much more difficult to naively edit this file manually. I followed the existing pattern from `tests.rs` and manually added my test case (in the wrong place), which caused some difficulties during CI. Hopefully this change brings additionally clarity to the fact that this file is autogenerated cannot be overlooked.


trevor-scheer:
FYI, `graphql-parser` tests don't run with these changes. Do not merge.

trevor-scheer:
Outdated

PR #293, opened 2020-12-01
fix(buildFederatedSchema): Export `GraphQLSchemaModule` type

jtomaszewski:
We have a big app that uses `@apollo/federation` and we have several modules defined separately, with their own tests etc. in their folders. Each module exports `const module: GraphQLSchemaModule = { typeDefs, resolvers }` which are then glued together and passed to `buildFederatedSchema`.

However currently the `GraphQLSchemaModule` must be imported from `apollo-graphql` which is a dependency of `@apollo/federation` and we see little need in adding that as a direct dependency of our app.

This however triggers such an eslint warning for us:

image

Which is annoying as we have to ignore it all over the files. ( `// eslint-disable-next-line import/no-extraneous-dependencies `)

As `GraphQLSchemaModule` is type definition of an argument of `buildFederatedSchema`, it belongs to the public API of `@apollo/federation`, and this is why I think it should be exported by the `@apollo/federation` itself.

Thus, the PR.

### DIFF
--- a/federation-js/CHANGELOG.md
+++ b/federation-js/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 > The changes noted within this `vNEXT` section have not been released yet.  New PRs and commits which introduce changes should include an entry in this `vNEXT` section as part of their development.  When a release is being prepared, a new header will be (manually) created below and the appropriate changes within that release will be moved into the new section.
 
+- Export `GraphQLSchemaModule` type.
+
 ## v0.20.6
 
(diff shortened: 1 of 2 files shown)


trevor-scheer:
Thanks for the PR @jtomaszewski - appreciate the explanation and this seems like a reasonable ask. This LGTM! Would you mind adding an entry to the `federation-js/CHANGELOG.md`?

jtomaszewski:
Done. When merging this, I suggest doing "Squash and merge", to merge the commits into one.

trevor-scheer:
Done! Sorry for the delay on this @jtomaszewski. This change will go out with the next release.

PR #297, opened 2020-12-04
feat(csdl): Add `@using` directive

trevor-scheer:
This PR introduces the `@using` directive to the CSDL. This is a preliminary step in implementing the official CSDL spec currently being written by @queerviolet.

I suspect the introduction of `@using` defeats the previous need for `@composedGraph`, so I've removed it for now - but I'm open to feedback or reasons not to.

### DIFF
--- a/stargate/crates/query-planner/tests/features/basic/csdl.graphql
+++ b/stargate/crates/query-planner/tests/features/basic/csdl.graphql
@@ -6,6 +6,7 @@ schema
 @graph(name: "product", url: "https://product.api.com")
 @graph(name: "reviews", url: "https://reviews.api.com")
 @composedGraph(version: 1)
+@using(spec: "http://specs.apollo.dev/cs/v0.1")
 {
     query: Query
     mutation: Mutation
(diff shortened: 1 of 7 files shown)


stargate/crates/query-planner/tests/features/basic/csdl.graphql
```diff
@@ -5,7 +5,7 @@ schema
 @graph(name: "inventory", url: "https://inventory.api.com")
 @graph(name: "product", url: "https://product.api.com")
 @graph(name: "reviews", url: "https://reviews.api.com")
-@composedGraph(version: 1)
+@using(spec: "http://specs.apollo.dev/cs/v0.1")
```
queerviolet, line 8: One thought is should we maybe commit to the URL before finalizing this?

abernix (commented):
To start with, should the current query planner do something to enforce (without too much of a complex/flexible/range-compatible mechanism; strictly) that it will only support a particular version today?

stargate/crates/query-planner/tests/features/basic/csdl.graphql
abernix: Worth noting that this _has_ to be `https://` because `.dev` is Strict Transport Security (HSTS) and requires `https://`.

abernix:
Superseded by #305 (which includes this)

PR #309, opened 2020-12-24
feat(federation): Include originalError in downstreamServiceError

coryvirok:
```
 new ApolloServer({
  formatError: GraphQLError => GraphQLError.originalError, // undefined
  gateway: new ApolloGateway({ ... })
  ...
})
```

The GraphQLError type expected to have originalError (if avialible)
`formatError: GraphQLError => GraphQLError.originalError` 

In the event of a downstreamServiceError, it does not include the original error.
I believe it's reasonable to expect the original error in this event.

---
Originally at: apollographql/apollo-server#3211


coryvirok:
Originally https://github.com/apollographql/apollo-server/pull/3211 but it looks like the author hasn't created the PR in this repo yet and this is a pretty important feature to help debug downstream errors. 

@bl42 I'm happy to close this if you want to open a new PR and retain authorship. Otherwise, I'd love to see this change make its way into main.

gateway-js/src/executeQueryPlan.ts
```diff
@@ -322,6 +323,7 @@ async function executeFetch<TContext>(
           variables,
           error.extensions,
           error.path,
+          error,
```
trevor-scheer, line 326: I'm aware this PR is from the previous repo and a different author, but I have just one nit. Since we're now passing down the whole error object it seems unnecessary to also pass specific properties from the same error object. We can simplify this function's signature a bit.

trevor-scheer (approved):

gateway-js/src/executeQueryPlan.ts
coryvirok: Good call and not a problem! https://github.com/apollographql/federation/pull/315

trevor-scheer: Perfect, thanks!

PR #311, opened 2020-12-30
chore(federation)!: Drop support for Node.js 8 and Node.js 10.

abernix:
BREAKING CHANGE: This PR drops support for versions of Node.js prior to v12 that were still supported in the `@apollo/federation` package, leaving only support for Node.js 12+ LTS ([Long-Term Support](https://github.com/nodejs/release)) versions.

This aligns the supported versions of `@apollo/federation` with those supported by `@apollo/gateway`, which first received this treatment in https://github.com/apollographql/apollo-server/pull/4031 for performance gains.  Conversely to the motivation for that PR, performance gains are NOT expected from this change for the `@apollo/federation` package since it's not invoked per-request and has limited runtime computational cycles.

Aside from the impending end-of-life (EOL) cycle for Node.js 10 that is coming in April 2021, and dropping support for Node.js 8 which lapsed at the end of 2019, the motivating reason for releasing this change now is to remove the need for the `@apollo/federation` package to continue to leverage `apollo-server-env`, which brings globally-loaded polyfills from `core-js` and a `fetch` polyfill in Node.js environments that causes additional struggles when trying to package the `@apollo/federation` with bundling tools like Webpack or Rollup.

If this change proves to be problematic for any users of `@apollo/federation`, I'm happy to reconsider this and possibly revert it until the end of April 2021 (again, when Node.js 10 is completely end-of-life), but I would strongly suggest that users update their Node.js runtimes to a newer version of Node.js prior to that time, as once it is end-of-life'd continuing to support old Node.js versions won't continue to be an option.

### DIFF
--- a/federation-js/jest.config.js
+++ b/federation-js/jest.config.js
@@ -1,18 +1,9 @@
 const config = require('../jest.config.base');
 
-const NODE_MAJOR_VERSION = parseInt(
-  process.versions.node.split('.', 1)[0],
-  10
-);
-
 const additionalConfig = {
-  setupFiles: [
-    'core-js/features/array/flat',
-    'core-js/features/array/flat-map',
-  ],
   testPathIgnorePatterns: [
     ...config.testPathIgnorePatterns,
-    ...NODE_MAJOR_VERSION === 6 ? ["<rootDir>"] : []
+    ...[]
   ]
 };
 
--- a/federation-js/jest.config.js
+++ b/federation-js/jest.config.js
@@ -1,10 +1,5 @@
 const config = require('../jest.config.base');
 
-const additionalConfig = {
-  testPathIgnorePatterns: [
-    ...config.testPathIgnorePatterns,
-    ...[]
-  ]
-};
+const additionalConfig = {};
 
 module.exports = Object.assign(Object.create(null), config, additionalConfig);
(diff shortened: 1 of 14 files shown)


federation-js/package.json
```diff
@@ -17,15 +17,13 @@
   "author": "Apollo <opensource@apollographql.com>",
   "license": "MIT",
   "engines": {
-    "node": ">=8"
+    "node": ">=12.13.0 <15.0"
```
trevor-scheer, line 20: Should we follow up and see if we can support / expand to 15.x? I'm not yet familiar with the set of breaking changes in v15 so this may be a naive question.

trevor-scheer (approved):

federation-js/package.json
abernix: I considered it briefly, but I'm not particularly eager to bother with Node.js 15 until it becomes an even-numbered (i.e., more stable API, destined to be LTS) release.  We'll have several months after that happens to start testing and adjusting accordingly.

federation-js/jest.config.js
```diff
@@ -1,18 +1,9 @@
...
 const additionalConfig = {
-  setupFiles: [
-    'core-js/features/array/flat',
-    'core-js/features/array/flat-map',
-  ],
   testPathIgnorePatterns: [
     ...config.testPathIgnorePatterns,
-    ...NODE_MAJOR_VERSION === 6 ? ["<rootDir>"] : []
+    ...[]
   ]
```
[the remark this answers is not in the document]
abernix: I suppose we could!

PR #315, opened 2021-01-05
Consolidate downstreamServiceError() args

coryvirok:
Now that we are passing the whole `originalError` to the `downstreamServiceError()` function, we no longer need to pass in `originalError.message`, etc. 

This change consolidates the arguments for `downstreamServiceError()` which is only used in a single place in this file and is not exported so the change shouldn't cause any compatibility issues.


trevor-scheer (approved):
Thanks!

PR #316, opened 2021-01-05
Add missing typing support to make-fetch-happen.d.ts

michael-watson:
Add `cert`, `key`, `proxy` and `noProxy` options to `make-fetch-happen` typings to support internal security requirements around custom certs

### DIFF
--- a/gateway-js/src/make-fetch-happen.d.ts
+++ b/gateway-js/src/make-fetch-happen.d.ts
@@ -33,6 +33,10 @@ declare module 'make-fetch-happen' {
           randomize?: boolean;
         };
     onRetry?(): void;
+    cert?: string;
+    key?: string;
+    proxy?: string;
+    noProxy?: string;
   }
 
   export interface CacheManager {


gateway-js/src/make-fetch-happen.d.ts
```diff
@@ -33,6 +33,10 @@ declare module 'make-fetch-happen' {
           randomize?: boolean;
         };
     onRetry?(): void;
+    cert?: string;
+    key?: string;
+    proxy?: string;
+    noProxy?: string;
```
trevor-scheer, line 39:
```suggestion
    noProxy?: string | string[];
```

gateway-js/src/make-fetch-happen.d.ts
```diff
@@ -33,6 +33,10 @@ declare module 'make-fetch-happen' {
           randomize?: boolean;
         };
     onRetry?(): void;
+    cert?: string;
+    key?: string;
+    proxy?: string;
```
trevor-scheer, line 38:
```suggestion
    proxy?: string | URL;
```

We'll also need to add the proper import at the top:
`import { URL } from 'url';`

gateway-js/src/make-fetch-happen.d.ts
```diff
@@ -33,6 +33,10 @@ declare module 'make-fetch-happen' {
           randomize?: boolean;
         };
     onRetry?(): void;
+    cert?: string;
+    key?: string;
```
trevor-scheer, line 37:
```suggestion
    cert?: SecureContextOptions['cert'];
    key?: SecureContextOptions['key'];
```
We'll also need to add the proper import at the top:
`import { SecureContextOptions } from 'tls';`

trevor-scheer: Ref: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/tls.d.ts#L563

gateway-js/src/make-fetch-happen.d.ts
trevor-scheer: Ref: https://github.com/npm/make-fetch-happen#opts-no-proxy

gateway-js/src/make-fetch-happen.d.ts
trevor-scheer: Ref: https://github.com/npm/make-fetch-happen#opts-proxy

trevor-scheer (changes requested):
Left some suggestions. Noting that these are incomplete but definitely an improvement! These will be replaced / removed entirely when @abernix is able to land his `DefinitelyTyped` PR here: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/50339.

trevor-scheer:
Please add a changelog entry, address comments and this will be good to go. Thanks @michael-watson!

abernix:
Going to close this in favor of using the DefinitelyTyped-published types [which already include these missing properties](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/d73fdfc3ff6aed8f47bd4092ada9623bcc1f4fe8/types/make-fetch-happen/index.d.ts).

Someone still needs to open the PR to change this repository to rely on that `@types/make-fetch-happen` package (and remove these baked-in types from this repo), but that's being tracked in https://github.com/apollographql/federation/issues/535.

Thanks for opening this originally! (If there are additional updates, the PRs should be opened to DefinitelyTyped.)

PR #323, opened 2021-01-12
Update entities.md

brettfishman:
There was no target defined for `#entity-migration`, but there is a `#type-migration` heading. I updated the links to target `Type Migration` instead so they work.

### DIFF
--- a/docs/source/entities.md
+++ b/docs/source/entities.md
@@ -235,11 +235,11 @@ We can perform this migration safely with the following steps:
 
 ### Field migration
 
-The steps for migrating an individual field are nearly identical in form to the steps for [migrating an entire entity](#entity-migration).
+The steps for migrating an individual field are nearly identical in form to the steps for [migrating an entire entity](#type-migration).
 
 Let's say our `products` service defines a `Product` entity, which includes the boolean field `inStock`. Then, we add an `inventory` service to our federated graph. It now makes sense for the `inStock` field to originate in the `inventory` service instead.
 
-We can perform this migration safely with the following steps (_additional commentary on each step is provided in [Entity migration](#entity-migration)_):
+We can perform this migration safely with the following steps (_additional commentary on each step is provided in [Entity migration](#type-migration)_):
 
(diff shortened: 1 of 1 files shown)


docs/source/entities.md
```diff
@@ -235,11 +235,11 @@ We can perform this migration safely with the following steps:
 
 ### Field migration
 
-The steps for migrating an individual field are nearly identical in form to the steps for [migrating an entire entity](#entity-migration).
+The steps for migrating an individual field are nearly identical in form to the steps for [migrating an entire entity](#type-migration).
```
trevor-scheer, line 238:
```suggestion
The steps for migrating an individual field are nearly identical in form to the steps for [migrating an entire type](#type-migration).
```

docs/source/entities.md
```diff
@@ -235,11 +235,11 @@ We can perform this migration safely with the following steps:
...
 
 Let's say our `products` service defines a `Product` entity, which includes the boolean field `inStock`. Then, we add an `inventory` service to our federated graph. It now makes sense for the `inStock` field to originate in the `inventory` service instead.
 
-We can perform this migration safely with the following steps (_additional commentary on each step is provided in [Entity migration](#entity-migration)_):
+We can perform this migration safely with the following steps (_additional commentary on each step is provided in [Entity migration](#type-migration)_):
```
trevor-scheer, line 242:
```suggestion
We can perform this migration safely with the following steps (_additional commentary on each step is provided in [Type migration](#type-migration)_):
```

trevor-scheer (changes requested):
Thanks @brettfishman! Let's update the verbage: entity -> type as well. I've added suggestions where appropriate.

brettfishman:
Thanks @trevor-scheer - this should be all set now. Apologies for the delay getting back to you!

brettfishman:
Hi @trevor-scheer - just following up on this. How's it looking?

StephenBarlow (approved):

StephenBarlow:
Sorry for the delay here @brettfishman! It looks like we reverted the language in this article, and the header is once again "entity migration". So I'll close this for now and then will make sure I'm more vigilant with these internal links if we switch it back! 😄

PR #325, opened 2021-01-13
chore(gateway): Update apollo-server-* deps

trevor-scheer:


package.json
```diff
@@ -38,11 +38,9 @@
     "@types/node-fetch": "2.5.4",
     "apollo-link": "1.2.14",
     "apollo-link-http": "1.5.17",
-    "apollo-server": "2.18.0",
-    "apollo-server-core": "2.18.0",
```
trevor-scheer, line 42: This just doesn't need to be specified at the root anymore. This is cruft from the migration out of `apollo-server` monorepo.

PR #305, opened 2020-12-19
Generate composite schema in spec'd form

queerviolet:
Update the composer's printer to generate a composite schema that follows the current draft spec.

This PR was based on:

- @trevor-scheer's PR that removed the internal dependency on `schema`.
- It had the `qv/using` branch (which is #326) merged into it.
- A merged in an experimental harmonizer branch that @abernix put together.

### Note

**https://github.com/apollographql/federation/pull/559 is meant to supersede this.**

PR #326, opened 2021-01-14
Rust implementation of the using spec

queerviolet:
Define a new crate, [`using`](./using), which provides Rust support for the [current draft `@using` directive](https://apollo-specs.github.io/using/draft/pre-0).

The intention here is to provide (1) a crate we can use in the query planner to process composed schema directives (2) a reference implementation of the spec which can be ported to other languages if needed.

DONE:
- [x] bootstrap `using` as per the spec
- [x] enumerate requests (_identity_, _version_, _prefix_)
  - [x] spec url parsing
- [x] version resolution
- [x] validations
  - [x] exactly one schema definition
  - [x] overlapping prefixes
  - [x] spec urls have versions
  - [x] spec urls have prefixes

TODO:
- [ ] an **implementation activator** which lets you register implementations  by (_spec identity_, _spec version_) and iterate over matches of requested implementations
- [ ] tests for url parsing / version resolution
- [ ] micro-cli devtool for extracting using information from a schema
- [ ] better docs and examples


queerviolet:
sorry about the file movement noise y'all. If you want to hide all the moved files in this PR, you can run this snippet in the dev console:

```javascript
document.querySelectorAll('a[title^="stargate"]').forEach(e => e.closest('.file').remove())
```

queerviolet:
This is ready for review

abernix:
> sorry about the file movement noise y'all. If you want to hide all the moved files in this PR, you can run this snippet in the dev console:
> 
> ```js
> document.querySelectorAll('a[title^="stargate"]').forEach(e => e.closest('.file').remove())
> ```
> 
> (there are moved files because I put `graphql-parser` and `query-planner` back in the root of the repo, since we're using them both outside of stargate)

@queerviolet Since we already agreed to hoist the `./stargate`-nested crates back to the top of the repository, I think the file move stuff should/could survive on its own and would benefit from being de-coupled from this PR and getting landed expeditiously, ensure that it gets its own separate commit for posterity, and still forgo the need for Console filter ergonomics.  (The burden in review is quite different between the two initiatives.)

Could we cherry pick [I think, just `044a6bf` (#326)](https://github.com/apollographql/federation/pull/326/commits/044a6bf3d7263df7516bade07ec862f5ef60f683) into its own branch and PR and land that separately?  (This PR could be re-based/targeted against that branch in the interim.  They'd automatically waterfall merge into `main`.)

using/src/activations.rs
```diff
@@ -0,0 +1,178 @@
...
+//! `Request`s.
+
+use crate::{Find, Found, Implementations, Request, Schema};
+
+/// An Activation references a request and contains the min and max versioned implementations
+/// available in the implementation registry which can satisfy that request, or None if no
+/// matching implementations were available.
+// type Activation<'schema, 'impls, T> = (&'schema Request, Find<'impls, T>);
+
+impl<'a> Schema<'a> {
```
lrlna, line 11: What is the reason for having this not part of `schema.rs`? It's definitely not standard, and from what I understand of this PR I don't see a reason for it to be separate, but maybe I am missing something!

using/src/activations.rs
```diff
@@ -0,0 +1,178 @@
...
+    ///   .filter_map(|(_request, mut impls)| impls.bounds())
+    ///   .collect::<Vec<_>>();
+    ///
+    /// assert_eq!(max, vec![(
+    ///     (&Version(1, 2), &"impl for A 1.2"), (&Version(1, 7), &"impl for A 1.7")
+    /// )]);
+    /// # Ok::<(), GraphQLParseError>(())
+    /// ```
+    pub fn activations<'impls, T>(
+        &'impls self,
```
lrlna, line 44: With `impl` being a reserved word, it's a bit weird to have `'impls` as a name for a lifetime. Would definitely use something else (I do understand it's trying to describe a lifetime of Implementations, which is just very close to `impl` hah)

using/src/request.rs
```diff
@@ -0,0 +1,62 @@
...
+};
+
+use crate::spec::{Spec, SpecParseError};
+
+/// Requests contain a `spec`, the `prefix` requested for that spec (which
+/// will be the spec's [default prefix](Spec.html#default_prefix) if none was
+/// explicitly specified, and the position of the directive making the request
+/// (for validation error reporting).
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Request {
```
lrlna, line 16: Wanted to double check that users are expected to use this Struct and it's values? As in: does this need to be `pub`, and do the following three lines need to be `pub`?

lrlna: This question applies to all publicly exposed structs `Spec`, `Implementations`, `Version`, `Pos`, `VersionParseError`. (Definitely see that `Schema` needs to be exposed to the users :)

using/src/request.rs
```diff
@@ -0,0 +1,62 @@
...
+//!
+//! `Request`s are derived from `Directive`s during schema bootstrapping.
+use graphql_parser::{
+    schema::{Directive, Value},
+    Pos,
+};
+
+use crate::spec::{Spec, SpecParseError};
+
+/// Requests contain a `spec`, the `prefix` requested for that spec (which
```
lrlna, line 11: nit: for doc comments, it's usually really nice to have the first line to be a bit shorter and then an added break. When the docs get built the first line is used as a description for that particular exported type. This is what it looks like now for `Request` vs `Pos`
Screen Shot 2021-01-19 at 19 12 34

using/src/schema.rs
```diff
@@ -0,0 +1,693 @@
+//! A GraphQL schema referencing one or more specifications
```
lrlna, line 1: I would either move this to right above `pub struct Schema<'a>`, so that the docs built directly for the `Schema` struct, or, add another set of comments just above there, since this page is currently empty:
Screen Shot 2021-01-19 at 19 17 30

using/src/version.rs
```diff
@@ -0,0 +1,145 @@
+use std::{fmt::Display, num::ParseIntError, str::FromStr};
+
+use lazy_static::lazy_static;
+use regex::Regex;
+use thiserror::Error;
+
+/// Versions are a (major, minor) u64 pair.
```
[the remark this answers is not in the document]
queerviolet: Yeah, we don't have patch versions. I thought about maybe hacking it (appending a patch version of `0` always), but the logic is simple enough that I'm not sure it's worth it.

using/src/spec.rs
```diff
@@ -0,0 +1,91 @@
...
+        let version = Version::parse(&version)?;
+        let default_prefix = path.last().ok_or(NoDefaultPrefix)?;
+        parsed.set_fragment(None);
+        let _ = parsed.set_username("");
+        parsed.set_query(None);
+        let _ = parsed.set_password(None);
+        parsed.set_path(&path.join("/"));
+
+        Ok(Spec {
+            identity: parsed.to_string(),
```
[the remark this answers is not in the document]
queerviolet: could you say more about this? neither side of the assignment can currently be a cow. is it just a best practice thing?

using/src/version.rs
```diff
@@ -0,0 +1,145 @@
...
+        }
+
+        if let Some(ver) = VERSION_RE.captures(input) {
+            Ok(Version(ver[1].parse()?, ver[2].parse()?))
+        } else {
+            Err(VersionParseError)
+        }
+    }
+
+    /// Return true iff this Version satisfies the `required` version
```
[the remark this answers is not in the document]
queerviolet: oh, that's intentional, "iff" = "if and only if"

but i can spell that out

using/src/request.rs
queerviolet: Users are expected to use `Request` (`activations` iterates over `Request`s). I had a general question about whether we should expose the raw fields or make getters. I went with exposing the raw fields over having getters, but could easily be talked into the converse if that's considered poor form.

using/src/spec.rs
```diff
@@ -0,0 +1,91 @@
...
+
+        let mut parsed = Url::parse(input)?;
+        let mut path: Vec<_> = parsed
+            .path_segments()
+            .ok_or(NoPath)?
+            .map(|x| x.to_owned())
+            .collect();
+        let version = path.pop().ok_or(NoVersion)?;
+        let version = Version::parse(&version)?;
+        let default_prefix = path.last().ok_or(NoDefaultPrefix)?;
```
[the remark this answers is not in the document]
queerviolet: Unfortunately, it needs to remain in the `path` (the spec's identity includes the default prefix)

using/src/implementations.rs
```diff
@@ -0,0 +1,208 @@
...
+        match versions {
+            Some(versions) => versions
+                .range(version..&Version(version.0, u64::MAX))
+                .filter(move |(impl_version, _)| impl_version.satisfies(version))
+                .into(),
+            None => Find::None,
+        }
+    }
+
+    pub fn find_req<'a>(
```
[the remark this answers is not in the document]
queerviolet: Would that return the underlying `&BTreeMap`? I sortof wanted to hide that as an implementation detail

using/src/implementations.rs
```diff
@@ -0,0 +1,208 @@
...
+
+use crate::{Request, Version};
+
+/// Implementations stores a set of implementations indexed by
+/// spec identity and version.
+pub struct Implementations<T>(HashMap<String, BTreeMap<Version, T>>);
+
+impl<T> Implementations<T> {
+    pub fn new() -> Self {
+        Implementations(HashMap::new())
```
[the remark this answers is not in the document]
queerviolet: done

using/src/constants.rs
```diff
@@ -0,0 +1,28 @@
...
+
+use crate::{request::Request, spec::Spec, version::Version};
+
+/// The well-known URL for the `using` spec
+pub const USING_SPEC_URL: &str = "https://specs.apollo.dev/using";
+
+/// Versions of the `using` spec supported by this crate
+pub const USING_VERSIONS: &[Version] = &[Version(0, 1)];
+
+lazy_static! {
```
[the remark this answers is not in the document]
queerviolet: Ah, cool! Thank you!

using/src/implementations.rs
```diff
@@ -0,0 +1,208 @@
...
+        &'a self,
+        request: &'a Request,
+    ) -> Find<'a, T, impl Iterator<Item = Found<'a, T>>> {
+        self.find(&request.spec.identity, &request.spec.version)
+    }
+}
+
+pub type Found<'a, T> = (&'a Version, &'a T);
+
+pub enum Find<'a, T: 'a, I: Iterator<Item = Found<'a, T>>> {
```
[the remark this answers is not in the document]
queerviolet: Do you mean creating an `Iterator` impl for `Implementations`?

using/src/schema.rs
```diff
@@ -0,0 +1,693 @@
...
+    OverlappingPrefix(String, Vec<Request>),
+}
+
+impl SchemaError {
+    /// Collect `Request`s referenced by this error and join their position, urls,
+    /// and versions into an indented, "\n"-separated String for purposes of
+    /// displaying the error.
+    ///
+    /// Returns the empty string if this error does not reference any `Request`s.
+    fn requests(&self) -> String {
```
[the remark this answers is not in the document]
queerviolet: well, `derive(thiserror::Error)` is taking care of implementing `fmt::Display`. We're using that for errors everywhere else, so I kindof wanted to maintain that here (it also looks nice). But specifically for the `OverlappingPrefix` case, I wanted to list all the overlapping requests as part of the error, which requires a bit more implementation than I wanted to put into the `#[error(...)]` attribute, so we call out to this fn

queerviolet:
Thanks for the review @lrlna! Fixed some of the small things, I'll integrate the `Cow` changes next

queerviolet:
@lrlna, one question before I adopt the `cow` changes: do you think it's worth the cognitive complexity of having a `Cow` rather than just a `String`?

Another approach would be to simply have a `fn default_using() -> Request` which always allocs a new `Request` with the default using spec parameters. I'd expect the performance hit to be effectively 0 (we already do a *lot* more allocations when parsing even a simple schema). Though admittedly, the cognitive cost of `Cow` vs `String` is quite low too. Thoughts?

using/src/implementations.rs
```diff
@@ -0,0 +1,208 @@
...
+            Some((&Version(0, 99), &"0.99"))
+        );
+        assert_eq!(
+            impls.find(&identity, &Version(0, 1)).next(),
+            Some((&Version(0, 1), &"0.1"))
+        );
+        assert_eq!(
+            impls.find(&identity, &Version(0, 99)).last(),
+            Some((&Version(0, 99), &"0.99"))
+        );
```
trevor-scheer, line 206: Nit: the ordering of these assertions doesn't appear to follow a pattern and seems a bit awkward to follow. Additionally, it might be more clear if we `assert_eq` against the full `.collect()` rather than just first and last which are assumed to be the only two items in the iterator.

using/src/version.rs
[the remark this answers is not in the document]
lrlna: ah yea! this approach seems good then (i _think_ semver implementation is pretty much what you have here, plus the extra five lines for `minor` anyways)

using/src/version.rs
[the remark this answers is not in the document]
lrlna: ohhhhh definitely didn't know that! ++ on spelling it out then.

using/src/activations.rs
```diff
@@ -0,0 +1,178 @@
...
+        Item = (
+            &'a Request,
+            Find<'impls, T, impl Iterator<Item = Found<'impls, T>>>,
+        ),
+    > + 'impls
+    where
+        'impls: 'a,
+        T: 'impls,
+    {
+        self.using
```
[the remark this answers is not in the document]
queerviolet: It's a field on `Schema`, which was defined elsewhere, but we've since moved everything in activations over to schema.rs, so this confusion shouldn't be an issue anymore

queerviolet:
PTAL @lrlna @trevor-scheer 

@lrlna, great cow suggestion, I've implemented your changes. As I've been working with this library in #305, I've found that it would indeed be very helpful to be able to easily define specs as top-level consts. This is really nice!

lrlna:
@queerviolet in regards to your first question:
>Another approach would be to simply have a fn default_using() -> Request which always allocs a new Request with the default using spec parameters. I'd expect the performance hit to be effectively 0 (we already do a lot more allocations when parsing even a simple schema). Though admittedly, the cognitive cost of Cow vs String is quite low too. Thoughts?

My answer here depends on how often you expect to use `Request` and `Spec` directly. It's for sure very _tidy_ consts exported for other crate users, and I will take `Cow` implementation over a `lazy_static`. However, if you need to instantiate a default `Request` and `Spec` pretty frequently, _and_ it does not need to be a const, I would consider writing a `Default` impl. That way not only do you avoid a `lazy_static`, you can also keep the `Request` implementation to still accept `String`, which is a lot more ergonomic for the users of the library. They are super common, here is, for example, [an implementation for `Vec`](https://doc.rust-lang.org/src/alloc/vec.rs.html#2602-2607). It would look something like this:

```rust
impl Default for Spec {
    fn default() -> Self {
        identity: String::From("https://lib.apollo.dev/core"),
        name: String::from("core"),
        version: Version(0, 1),
    }
}
```

TL;DR: for only internal library changes, `Cow` is brilliant, but if you're looking for ergonomics for you library users, go with `Default`.

P.S. So very sorry it took me so long to get back to you 🙈

using/src/request.rs
lrlna: I think the usage of getters is case by case specific: I usually use them for more complicated values that are likely to change drastically, or if I am modelling my API based on a `std` API that already has getters and setters. I think in this case it makes sense to just have them exposed as raw fields!

using/src/implementations.rs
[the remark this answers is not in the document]
lrlna: Hmm yea that's fair. Requiring to `iter` over a `BTreeMap` might prove problematic on the other end. You can go with `get_req` to go with a theme, but up to you.

using/src/implementations.rs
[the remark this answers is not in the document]
lrlna: Yes, I did mean `Iterator` impl for `Implementations`. Having thought about it some more, something to note is that both `HashMap` and `BTreeMap` have `iter`(and `iter_mut`) methods that might not require you to have an your own `Find` and `Found` implementations. Did using either of those not prove very fruitful?

I guess I am trying to figure out the reason for writing `Find` and `Found` 😓

queerviolet:
PTAL, I think this is ready to go?

core-schema/src/schema.rs
```diff
@@ -0,0 +1,585 @@
+//! A GraphQL schema referencing one or more specifications
```
glasser, line 1: It looks like this PR is missing all logic relating to export. Though it would be fine to merge it and add export logic as a follow-up PR.

core-schema/src/schema.rs
```diff
@@ -0,0 +1,585 @@
...
+//! via [the `@using` directive](https://apollo-specs.github.io/using/draft/pre-0).
```
glasser, line 2: Search the whole tree for `using` and update it all.

core-schema/src/schema.rs
```diff
@@ -0,0 +1,585 @@
...
+    ///    },
+    ///    Feature {
+    ///        spec: Spec::new("https://example.net/specB", "specB", (0, 1)),
+    ///        name: Cow::Borrowed("specB"),
+    ///        position: Pos { line: 5, column: 7 },
+    ///    },
+    /// ]);
+    /// # Ok::<(), ParseError>(())
+    /// ```
+    pub fn parse(source: &'a str) -> Result<Schema<'a>, ParseError> {
```
glasser, line 121: Might it be a better API to take in a parsed schema as an argument rather than only working on strings, and then offering an alternate entry point that takes care of both parsing and core-ing?

core-schema/src/schema.rs
```diff
@@ -0,0 +1,585 @@
...
+            });
+        }
+        let schema = schemas.remove(0);
+        for extraneous in schemas {
+            errors.push(SchemaError::ExtraSchema(extraneous.position));
+        }
+
+        // Collect everything which *could be* a core feature declaration.
+        // "Could be a core feature" means it's a directive on
+        // the schema which has a `spec: String` argument. We consider
```
glasser, line 152: s/spec/feature/

and that's more like "a spec argument whose value is a String" (vs wondering if the argument could be `String!`)

core-schema/src/schema.rs
```diff
@@ -0,0 +1,585 @@
...
+            .directives
+            .iter()
+            .filter_map(|dir|
+                // Parse as a using request and retain the directive.
+                // This filters out any directives which do not have a spec: argument at all.
+                Feature::from_directive(dir)
+                    .map(|res| (dir, res)))
+            .collect();
+
+        let using_features = match bootstrap(&schema) {
```
glasser, line 170: Not sure why this name is pluralized.

core-schema/src/schema.rs
```diff
@@ -0,0 +1,585 @@
...
+    ///
+    /// let impls = Implementations::new()
+    ///     .provide("https://spec.example.com/A", Version(1, 2), "impl for A 1.2")
+    ///     .provide("https://spec.example.com/A", Version(1, 3), "impl for A 1.3")
+    ///     .provide("https://spec.example.com/A", Version(1, 7), "impl for A 1.7");
+    /// let schema = Schema::parse(r#"schema @core(feature: "https://lib.apollo.dev/core/v0.1") @core(feature: "https://spec.example.com/A/v1.0") { query: Query }"#)?;
+    ///
+    /// let max = schema
+    ///   .activations(&impls)
+    ///   .filter_map(|(_request, mut impls)| impls.bounds())
```
glasser, line 245: this impls is a different type than the impls on line 237 which I find confusing

core-schema/src/schema.rs
```diff
@@ -0,0 +1,585 @@
...
+    /// ```
+    /// use core_schema::*;
+    ///
+    /// let impls = Implementations::new()
+    ///     .provide("https://spec.example.com/A", Version(1, 2), "impl for A 1.2")
+    ///     .provide("https://spec.example.com/A", Version(1, 3), "impl for A 1.3")
+    ///     .provide("https://spec.example.com/A", Version(1, 7), "impl for A 1.7");
+    /// let schema = Schema::parse(r#"schema @core(feature: "https://lib.apollo.dev/core/v0.1") @core(feature: "https://spec.example.com/A/v1.0") { query: Query }"#)?;
+    ///
+    /// let max = schema
```
glasser, line 243: why are we assigning bounds to something called "max"?

core-schema/src/schema.rs
```diff
@@ -0,0 +1,585 @@
...
+}
+/// Validation errors which may occur on a `Schema`
+#[derive(Debug, Error, PartialEq, Eq)]
+pub enum SchemaError {
+    /// Machine schemas must contain exactly one schema definition.
+    #[error("no schema definitions found in document")]
+    NoSchemas,
+
+    /// No reference to the core spec found
+    #[error("no reference to the core spec found")]
```
glasser, line 322: technically this is more like no reference to an understood version of the core spec is found. maybe the error can actually show the identity and version (or just the URL) that is understood?

core-schema/src/schema.rs
```diff
@@ -0,0 +1,585 @@
...
+    /// Multiple schema definitions were found in the document
+    /// Extra schema definitions are reported as these errors.
+    #[error("{} extra schema definition ignored", .0)]
+    ExtraSchema(Pos),
+
+    /// Multiple versions of @using were requested
+    /// by the document. One will be selected, the others
+    /// will generate these errors.
+    #[error("{} extra using spec ignored", .0)]
+    ExtraUsing(Pos, Version),
```
glasser, line 334: This is unused (since you use `find_map`)

core-schema/src/schema.rs
```diff
@@ -0,0 +1,585 @@
...
+    #[error("{} bad spec url: {}", .0, .1)]
+    BadUsingRequest(Pos, spec::SpecParseError),
+
+    /// Document is using multiple specs with the same prefix.
+    #[error("multiple requests using the {} prefix:\n{}", .0, self.requests())]
+    OverlappingPrefix(Cow<'static, str>, Vec<Feature>),
+}
+
+impl SchemaError {
+    /// Collect `Request`s referenced by this error and join their position, urls,
```
glasser, line 346: Search whole project for "request" and rename to "feature"

core-schema/src/schema.rs
```diff
@@ -0,0 +1,585 @@
...
+    /// displaying the error.
+    ///
+    /// Returns the empty string if this error does not reference any `Request`s.
+    fn requests(&self) -> String {
+        match self {
+            Self::OverlappingPrefix(_, requests) => requests
+                .iter()
+                .map(|req| {
+                    format!(
+                        "    {pos}: {url}/{ver}",
```
glasser, line 357: The `as` seems relevant to this error

core-schema/src/spec.rs
```diff
@@ -0,0 +1,100 @@
...
+use url::Url;
+
+use crate::{version, version::Version};
+
+/// Specs contain the `identity`, `name`, and `version` extracted from
+/// a spec URL.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Spec {
+    pub identity: Cow<'static, str>,
+    pub name: Cow<'static, str>,
```
glasser, line 30: I worry it might be a little too easy to get confused between `Spec.name` and `Feature.name` and accidentally use the wrong one, writing some code that works as long as you don't use `as`. How important is it that there be a stand-alone `name` field on `spec`? It looks like the only time we ever read it is when we produce a `Feature` so I think we could just validate that the name is valid at `Spec::parse` time (and save `name` or its start index into a private field?) but make actually getting the name later be a method or something?

glasser (changes requested):

core-schema/src/schema.rs
glasser: Ah I guess it also doesn't have a facility to find spec-specific elements, prefixing logic, etc. Can certainly be a follow-up PR!

PR #462, opened 2021-02-11
[RESTDataSource] refactor to allow overridding the Error class

eturino:
for example if we want to create a `NotFoundError` that will be used when receiving 404s. This way we need to override the bare minimum like

```
class MyDataSource extends RESTDataSource {
  protected errorBaseFromResponseAndMessage(response: Response, message: string): ApolloError {
    if (response.status === 401) return new NotFoundError(message);

    return super.errorBaseFromResponseAndMessage(response, message);
  }
}
```

Also adding it in the RemoteGraphQLDataSource since it has the same code as RESTDataSource, for symmetry purposes.

Signed-off-by: Eduardo Turiño <eturino@eturino.com>

---
Originally at: apollographql/apollo-server#4308


eturino:
sorry it took me so very long to reopen this PR in the new repo. I kind of lost track of this.

dariuszkuc:
Closing old PR. REST DataSource was recently moved to a separate project -> https://github.com/apollographql/datasource-rest. If issue is still valid, please open it against new repo.

PR #474, opened 2021-02-23
Introduce `harmonizer` crate and accompanying tool to compose schemas in Rust.

abernix:
> _Note_: This breaks-out part of the work that is currently included in https://github.com/apollographql/federation/pull/305 while avoiding the adoption of some newer implementation that PR relies on within https://github.com/apollographql/federation/pull/326.

> _Also Note:_ The names and final form of this is subject to change after this lands.  None of the changes introduced in this commit are meant to be acted on by users today.

> _Also, also, note_: This PR depends on https://github.com/apollographql/federation/pull/480 and https://github.com/apollographql/federation/pull/479. (both merged now)

~**Note for reviewing: I'm going to move the CI changes to another PR, which surfaced _after_ I opened this PR, but this is ready for review.  I might suggest only reviewing the first commit.**~

Shared commit authorship for this goes to @queerviolet for improving upon the initial proof-of-concept `harmonizer` in https://github.com/apollographql/federation/pull/326. (Thanks!)  This commit:

- Introduces a `harmonizer` Rust crate which composes schemas using the existing JavaScript/TypeScript composition function and returns a string representation of the composed schema -- tentatively, known as "CSDL" (composed schema definition language).  The name "CSDL" is subject to change.

  Today, CSDL is already an internal implementation detail which acts as a format that we could pass between the TypeScript runtime and the query planner, which is written in Rust.  By providing the mechanism to produce a static CSDL artifact, the gateway can be absolved from the responsibility to create this artifact in an opaque manner at runtime.  The ability to pass CSDL into a Gateway as configuration has been possible ever since the landing of https://github.com/apollographql/federation/pull/278, though until this commit, there hasn't been a way to produce it.

  As an implementation detail of the `harmonizer` library, this is accomplished by invoking a [Rollup.js]-bundled [IIFE] of the `@apollo/federation` package, in V8, via the [`deno_core`] V8 bindings.  The IIFE JavaScript bundle itself is generated by bundling a (small) stub, native ECMAScript module that depends on `@apollo/federation` and calls its `composeAndValidate` function.  The Rollup.js build is invoked via a Rust `build.rs` script ([Build Scripts]).

  While the intention exists on the team to offer composition and validation written in native Rust, this intermediary is a short-cut that allows us to use the well-tested TypeScript logic that powers Federation today to achieve an immediate "win" without blocking other work while we move toward that future.

- _(For development purposes only)_ Introduces a `harmonizer-cli` that invokes the `harmonize` lib.  Again, this will **not** be the long-term home for where the production of this artifact will take place.  Today, it acts as an easy tool within this repository for those interfacing with this project directly today in its early-phases.  It is merely for local use within this repository.

[Rollup.js]: http://rollupjs.org/
[IIFE]: https://developer.mozilla.org/en-US/docs/Glossary/IIFE
[`deno_core`]: https://crates.io/crates/deno_core
[Build Scripts]: https://doc.rust-lang.org/cargo/reference/build-scripts.html


glasser:
Is there a particular aspect of this PR you'd like me personally to focus on? I did a quick skim and nothing looked egregious but actually reviewing this in detail would require me to learn about a bunch of new things that would take a while right now. (Valuable eventually for sure.)

harmonizer/src/lib.rs
```diff
@@ -0,0 +1,207 @@
...
+  Deno.core.jsonOpSync('op_composition_result', result);
+}
+
+// Finally we register the error class used during do_compose.js.
+// so that it throws the correct class.
+Deno.core.registerErrorClass('Error', Error);
+
+// We build some of the preliminary objects that our Rollup-built package is
+// expecting to be present in the environment.
+node_fetch_1 = {}; // This is an unused external dependency we don't bundle.
```
abernix, line 134:
```suggestion
// We build some of the preliminary objects that our Rollup-built package is
// expecting to be present in the environment.
// node_fetch_1 is an unused external dependency we don't bundle.  See the
// configuration in this package's 'rollup.config.js' for where this is marked
// as an external dependency and thus not packaged into the bundle.
node_fetch_1 = {};
```

harmonizer/src/lib.rs
```diff
@@ -0,0 +1,207 @@
...
+process = { env: { "NODE_ENV": "production" }};
+global = {};
+exports = {};
```
abernix, line 137:
```suggestion
// Some JS runtime implementation specific bits that we rely on that
// need to be initialized as empty objects. 
global = {};
exports = {};
```

harmonizer/src/lib.rs
```diff
@@ -0,0 +1,207 @@
...
+}
+
+// Finally we register the error class used during do_compose.js.
+// so that it throws the correct class.
+Deno.core.registerErrorClass('Error', Error);
+
+// We build some of the preliminary objects that our Rollup-built package is
+// expecting to be present in the environment.
+node_fetch_1 = {}; // This is an unused external dependency we don't bundle.
+process = { env: { "NODE_ENV": "production" }};
```
abernix, line 135:
```suggestion
// 'process' is a Node.js ism.  We rely on process.env.NODE_ENV, in
// particular, to determine whether or not we are running in a debug
// mode.  For the purposes of harmonizer, we don't gain anything from
// running in such a mode.
process = { env: { "NODE_ENV": "production" }};
```

package.json
```diff
@@ -26,10 +26,14 @@
...
     "@apollo/gateway": "file:gateway-js",
     "@apollo/query-planner-wasm": "file:query-planner-wasm",
+    "@apollo/harmonizer": "file:harmonizer",
     "@apollographql/apollo-tools": "0.4.9",
     "apollo-federation-integration-testsuite": "file:federation-integration-testsuite-js"
   },
   "devDependencies": {
+    "@rollup/plugin-commonjs": "^17.0.0",
+    "@rollup/plugin-node-resolve": "^11.0.1",
+    "@rollup/plugin-typescript": "^8.1.0",
```
abernix, line 36: While this package was added to the `package.json` in the PR this implementation was extracted from in #305, it does not appear to be actually used in the `rollup.config.js`.

```suggestion
```

harmonizer/js/do_compose.js
```diff
@@ -0,0 +1,44 @@
...
+serviceList = serviceList.map(({ typeDefs, ...rest }) => ({
+  typeDefs: parseTypedefs(typeDefs),
+  ...rest,
+}));
+
+function parseTypedefs(source) {
+  try {
+    return composition.parseGraphqlDocument(source)
+  } catch (err) {
+    print(err.message)
```
martijnwalraven, line 32: Why are we printing these errors, but throwing other errors ourselves?

harmonizer/src/lib.rs
```diff
@@ -0,0 +1,216 @@
...
+
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
+pub struct CompositionErrorExtensions {
+    pub code: String,
+}
+
+impl CompositionError {
+    pub fn code(&self) -> &str {
+        if let Some(ref ext) = self.extensions {
+            ext.code.as_str()
```
lrlna, line 57: if you like, you can also do: 

```suggestion
        if let Some(ref ext) = self.extensions {
            &*ext.code
```

harmonizer/src/lib.rs
```diff
@@ -0,0 +1,216 @@
+use deno_core::Op;
```
lrlna, line 1: to all the crates I write, I also like adding these configs:
```rust
#![forbid(unsafe_code)]
#![deny(missing_debug_implementations, nonstandard_style)]
#![warn(missing_docs, future_incompatible, unreachable_pub, rust_2018_idioms)]
```

These will mean you will need to document some of your methods and have a top level doc comment.

harmonizer/src/lib.rs
```diff
@@ -0,0 +1,216 @@
...
+        "op_print",
+        // The op_fn callback takes a state object OpState
+        // and a vector of ZeroCopyBuf's, which are mutable references
+        // to ArrayBuffer's in JavaScript.
+        |_state, zero_copy| {
+            let mut out = std::io::stdout();
+
+            // Write the contents of every buffer to stdout
+            for buf in zero_copy {
+                out.write_all(&buf).unwrap();
```
lrlna, line 88: If `unwrap` is necessary, I'd use `expect` instead.

harmonizer/js/index.mjs
```diff
@@ -0,0 +1,2 @@
+export { composeAndValidate } from "@apollo/federation";
+export { parse as parseGraphqlDocument } from "graphql";
```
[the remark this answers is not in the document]
abernix: I'm going to stick with this for now, particularly because it's an internal API and I have difficult to reconcile reservations about snake-case with that pattern, even if `graphql` has decided it's okay themselves, particularly in a world where libraries like Rust's `serde` need to be able to reliably round-trip transformations (between camelcase and snake-case, for example) of property keys.  In my m ind, the obstacles that `GraphQL[A-Z]` produces are not nearly as egregious as the hurdle that `XMLHTTPRequest` would provide, but the results are just as disappointing.

harmonizer/build.rs
```diff
@@ -0,0 +1,13 @@
+use std::fs::metadata;
+use std::process::Command;
+
+fn main() {
+    if !metadata("dist/composition.js").is_ok() {
```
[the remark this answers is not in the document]
abernix: Ah, yes, thank you! 👈 this is not meant to negate my _thank you_, this is purely exclamatory!

Fixed in 9ba28428b43da8cd0e620062d47c751a37a0164.

harmonizer/jest.config.js
```diff
@@ -0,0 +1,10 @@
+const config = require('../jest.config.base');
```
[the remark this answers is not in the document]
abernix: Fixed with ca48ca8db3d0c39be06121156375658709478ba0, thanks!

harmonizer/src/lib.rs
abernix: Thanks! Fixed in 12102a9cb1c276ae1efdfe1fff996bf7d884fdae and 0ef75de977c67be50b11111e99c77bcd49816008 with what I hope is a ton of valuable documentation!  Though my first foray into rustdoc so probably fraught with misformatting.

harmonizer/src/lib.rs
abernix: Fixed in ea0c9a0d7deccb150fb4966a4487de1a10dd444e, which also switched this to `match`.

harmonizer-cli/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "harmonizer-cli"
```
queerviolet, line 2: This is bikeshedding at its finest, but how would you feel about calling this `harmonizer-bin` to distinguish it from a full-blown "cli tool"?

I know that I named it `harmonizer-cli` to begin with, but while working in the repo, I often found myself running `cargo run --bin harmonizer-cli` and honestly `cargo run --bin harmonizer-bin` rolls off the fingers a bit easier. Also, the aforementioned distinction between something that's a full-on CLI and something that's just a little binary.

harmonizer/src/lib.rs
abernix: Fixed in 3d9967d1ccb6323f877052ef4be433e682844220 (many fixes!)

harmonizer/jest.config.js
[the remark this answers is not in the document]
queerviolet: I structured it similarly to [the federation-js jest config](./federation-js/jest.config.js). Is that… bad?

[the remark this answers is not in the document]
queerviolet: We don't have any `testSetup` here so it seems like the `federation-js` jest is more applicable.

harmonizer-cli/Cargo.toml
abernix: Sure! Fixed in a4fc81f0

harmonizer/js/do_compose.js
abernix: Good catch! Applied 3f65f58 which will invoke `done({ Err: err })`.

cc @queerviolet who introduced this in what I have reason to believe was usefulness in debugging this during build-out.  Perhaps not necessary now, but a test could help us be more certain?

queerviolet: Yeah, I was trying to debug what turned out to be an error in parsing selection sets. Our error reporting in that case is miserable, but it appears to just be miserable throughout the typescript implementation, so we can fix it there later.

I did apply e93a308, which try/catches composition errors and reports them via the `Result` mechanism, rather than panicking.

queerviolet (approved):
This looks great! Very much digging the improved documentation, and the build steps make sense, even if I wish there were an easier way to have the tools dance together.

I fixed a few typos and added a `try/catch` to catch any errors which are *thrown* during `composeAndValidate` (rather than being returned as `errors`). This should keep `harmonizer` from panicking when e.g. there's a parse error in a subgraph schema.

harmonizer/js/do_compose.js
abernix: Nice.  Thanks for the patch / follow-up, @queerviolet.

PR #479, opened 2021-02-26
Convey missing prerequisites, but don't try to install them.

abernix:
There are various prerequisites necessary for actually installing and building this monorepo, including Node.js, a compiler toolchain with make/cmake, Python, Rust, etc.  Node.js and Rust tooling (npm and cargo, respectively) take care of installing packages that this project depends on but the burden of installing most the more core prerequisites for this repository often managed at a higher level and varies widely depending on operating system and user preferences.  For example, the Bash shell script that this commit removes doesn't work on Windows (invoking it fails, because Bash) and running the suggested install command globally installs Rust for the user.

Additionally, there are various ways that users like to manage these installations on their systems using existing package managers.  For example, Rust can be installed globally using the Rust installer, or it can be installed using Homebrew and contained to its mechanism for managing installs.  Similarly, Node.js could be installed globally, but it can also be managed by nvm.  CI environments come with either pre-built images which have these tools installed, or offer installation scripts to install them that come with caching benefits.  For example, [GitHub Actions'] or [CircleCI Rust Orb].

We should make running this repository friendly for contributors.  Suggesting that the user need to install particular prerequisites is an excellent suggestion to make, but the burden of providing installation instructions and doing installation for various tools+platforms (and maintaining those instructions!) seems best managed by the projects that offer those tools to maintain the widest compatibility to those tools, and to make sure the user knows what they're installing and how to use it.

[GitHub Actions']: https://github.com/marketplace/actions/rust-toolchain
[CircleCI Rust Orb]: https://circleci.com/developer/orbs/orb/circleci/rust
Ref: https://github.com/apollographql/federation/pull/445


trevor-scheer:
Were you able to try this out on a win machine just to confirm? The intent LGTM but I can't validate the behavior.

glasser (approved):

abernix:
@trevor-scheer Good question! I can offer a link to https://github.com/apollographql/federation/pull/474/checks?check_run_id=1990426796#step:10:21 (👈 specific line), which is a PR that is based on this PR that failed even though Rust _was_ installed.  That line proves that it now successfully asserts that the invocation of `rustup.exe -V` returns a `status` code of `0`.  While it might be returning `0` because of some other reason and I could probably demonstrate that it prints the message, I'm pretty comfortable with this, but I don't have a Windows machine that is not this CI run, unfortunately.  (Though I guess we could kick off CI on a Windows machine without Rust.)

PR #480, opened 2021-02-26
ci: Unify divergent CI configurations

abernix:
> **I believe there is benefit to reviewing commits individually on this PR; Starting [here](https://github.com/apollographql/federation/pull/480/commits/19d094d7308118612b9a00b558781954220adb80) and using the _Next_/_Previous_ buttons is a good option**. This PR was/is a prerequisite for https://github.com/apollographql/federation/pull/474

There are subtle differences between our CI workflows which largely do the same thing and could stand to benefit from being composed as a single set of instructions and invoked with parameterization to avoid getting subtly more out of sync.

Unfortunately, as you'll see in my comment on https://github.com/apollographql/federation/commit/301a3ccbaef8089a46222ed8484afc9754d76bd3 (which references the feature thread for it), this _isn't supported by GitHub Actions_, and further _it doesn't support the YAML specified anchors and aliases feature_. 😞 

Luckily, this still aligns the configurations which were slightly askew before, but I resorted to using comments to try and define intended parallels and prepare for a future where this is possible (either on another provider, or when GitHub Actions eventually supports it).

## Follow-up Todo

- [x] Convert this to run exclusively in CircleCI where the configuration can be _actually_ simplified?
      (Tracked as https://github.com/apollographql/federation/issues/540)


.github/workflows/rust-test.yml
```diff
@@ -43,47 +43,8 @@ jobs:
...
-        uses: actions-rs/toolchain@v1
-        id: toolchain
-        with:
-          toolchain: stable
-          profile: minimal
-          override: true
-
-      ### BUILD CACHE ###
-      - name: Cache Cargo registry, target, index
-        if: matrix.build != 'macos'
```
trevor-scheer, line 63: I can't say I understand the significance of the minor difference `if: matrix.build != 'macos'` that exists here vs. below.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,44 @@
...
+      - name: Install Rust
+        uses: actions-rs/toolchain@v1
+        id: toolchain
+        with:
+          toolchain: nightly
+          profile: minimal
+          components: rustfmt, clippy
+
+      ### BUILD CACHE ###
+      - name: Cache Cargo registry, target, index
```
trevor-scheer, line 55: Are the caching additions you've made elsewhere relevant here as well?
i.e.
```
id: cache-cargo
env:
  cache-name: cache-cargo
```

trevor-scheer: Answer to myself: in this commit, maybe so. But this is DRYed out in the following commit.

trevor-scheer: Follow up, post-revert: this comment may be relevant again!

.github/workflows/lint.yml
```diff
@@ -1,3 +1,11 @@
+# Unfortunately, GitHub Actions doesn't support YAML aliases OR depending
```
trevor-scheer, line 1: Huge bummer

.github/workflows/lint.yml
trevor-scheer: And hopefully full circle, it looks like they've been added. What a ride!

trevor-scheer (approved):
LGTM! Bummer about the revert. Hopefully the new comments and unification allow for a clean and DRY migration if we choose to do so?

.github/workflows/rust-test.yml
abernix: This was in the `rust-test.yml` > `Compile tests` before I started and it is technically invalid.
 https://github.com/apollographql/federation/blob/bdb117763240947b2910341240a48144e5518453/.github/workflows/rust-test.yml#L60-L72

There isn't a `matrix` in play within that step!  If GitHub Actions ran validations on these, I would have expected it to have flagged that, though since it likely always evaluated to null (or something) this step would have always run (though again, this particular job is _only_ configured to ever run on `ubuntu-latest`, which is another point of failure). 

Good eye on your part in calling this out though!

PR #564, opened 2021-03-12
run npm install before build step

EverlastingBugstopper:
can't build before you install the dependencies!


abernix (commented):
This already kicks off inside of an `npm install` though.  This would be cycles.

EverlastingBugstopper:
oh.

Issue #294, opened 2020-12-02
Interface pattern generates query plan with duplicate fetches to the same service


michael-watson:
## Accounts Schema
```
type User @key(fields: "id") {
  id: ID!
  name: String
  username: String
}
```

## Products Service
```
type Book implements Product @key(fields: "isbn") {
  isbn: String!
  title: String
  year: Int
  name: String
  price: Int
}

type TV implements Product @key(fields: "id") {
  id: String!
  title: String
  year: Int
  name: String
  price: Int
}

type Computer implements Product @key(fields: "id") {
  id: String!
  title: String
  year: Int
  name: String
  price: Int
}

type Furniture implements Product @key(fields: "sku") {
  sku: String!
  name: String
  price: Int
  weight: Int
}

interface Product {
  name: String
  price: Int
}

extend type Query {
  topProducts: [Product]
}
```

## Reviews Service
```
extend type Book implements Product @key(fields: "isbn") {
  isbn: String! @external
  reviews: [Review]
}

extend type TV implements Product @key(fields: "id") {
  id: String! @external
  reviews: [Review]
}
extend type Computer implements Product @key(fields: "id") {
  id: String! @external
  reviews: [Review]
}

extend type Furniture implements Product @key(fields: "sku") {
  sku: String! @external
  reviews: [Review]
}

extend interface Product {
  reviews: [Review]
}

type Review @key(fields: "id") {
  id: ID!
  body: String
  author: User @provides(fields: "username")
  product: Product
}

extend type User @key(fields: "id") {
  id: ID! @external
  username: String @external
  reviews: [Review]
}
```

## Given this query
```
query ios_TopProducts {
  topProducts {
    name
    price
    reviews {
      author {
        name
        username
      }
      body
      id
    }
  }
}
```

## Generates a query plan with multiple fetches to the `topProducts.@.reviews.@.author` services. It would be a duplication of **n** where **n** is the number of types that implement the given interface.
```
QueryPlan {
  Sequence {
    Fetch(service: "products") {
      {
        topProducts {
          __typename
          ... on Book {
            name
            price
            __typename
            isbn
          }
          ... on Computer {
            name
            price
            __typename
            id
          }
          ... on Furniture {
            name
            price
            __typename
            sku
          }
          ... on TV {
            name
            price
            __typename
            id
          }
        }
      }
    },
    Flatten(path: "topProducts.@") {
      Fetch(service: "reviews") {
        {
          ... on Book {
            __typename
            isbn
          }
[...]

PR #579, opened 2021-03-16
Fix duplicate fetches in query plans

### LINKED ISSUE #399: Query planner duplicate calls to services on interface types
Hello,

_Packages: apollo-gateway v0.11.6_

I'm running a microservice architecture based on the federation specs with apollo-gateway at the helm. One of the microservices defines a `BasePost` interface type, that is implemented by `RegularPost` and `FeaturedPost`, while another microservices defines a `User` type.

When I run a query to fetch blogs, their posts and authors, I get duplicated calls to the users microservices due to the `BasePost` interface. See the query planner below:

```
QueryPlan {
    Sequence {
      Fetch(service: "blogs") {
        {
          blogs {
            id
            posts {
              __typename
              ... on RegularPost {
                id
                author {
                  id
                  __typename
                }
              }
              ... on FeaturedPost {
                id
                author {
                  id
                  __typename
                }
              }
            }
          }
        }
      },
      Parallel {
        Flatten(path: "blogs.@.posts.@.author") {
          Fetch(service: "users") {
            {
              ... on User {
                __typename
                id
              }
            } =>
            {
              ... on User {
                givenName
              }
            }
          },
        },
        Flatten(path: "blogs.@.posts.@.author") {
          Fetch(service: "users") {
            {
              ... on User {
                __typename
                id
              }
            } =>
            {
              ... on User {
                givenName
              }
            }
          },
        },
      },
    },
  }
```

Given the author has the same fields between the implementing types, I would expect a single Fetch to the users microservice. Is this behaviour normal?

Thanks!

martijnwalraven:
Fixes #399 and #294.

The issue fixed by this PR requires a bit of unpacking. The visible result is that in some cases, the query plan will contain duplicate fetches to the same underlying service. For this query, for example:

```
query {
  topProducts {
    name
    price
    reviews {
      body
      author {
        name
      }
    }
  }
}
```

We end up fetching the names of authors as many times as there are different types of products. (It isn’t just that we fetch the authors of the reviews of each product type separately, but we literally execute the same query with the same representations, so all authors for all products, multiple times.)

In the case above, this is triggered as a consequence of type explosion, but it is a separate issue and also occurs in other cases where we manually include duplicate nested fields under multiple type conditions.

This query, for example, will also contain as many fetches for authors as there are type conditions (so two fetches here):

```
query {
  topProducts {
    name
    price
    ... on Book {
      reviews {
        body
        author {
          name
        }
      }
    }
    ... on Video {
      reviews {
        body
        author {
          name
        }
      }
    }
  }
}
```

The underlying cause is that we don't merge dependent fetch groups from subfields. In this case, we’ll create separate fetch groups for `{ name }` nested under the different type conditions, and even though these operate on the same path and fetch data from the same service, they are treated as separate fetches (that all have the same path however, which is why they take the same representations as input).

This PR attempts to solve this by manually merging nested fetch groups whenever they are compatible. That fixes the issue in most cases, but there are more complicated situations it doesn't address and that will require more substantial changes to the query planner and execution.

In particular, when possible types are spread over multiple services, it will not be possible to merge nested fetch groups from subfields because these should actually result in separate fetches. What you don't want however, is for those fetches to operate on the same objects just fetched by the other, but since representations taken as input are collected by path, there is no way to differentiate between these.

I believe solving this latter problem requires additional capabilities in the query plan (which is something we’ve talked about before). The most straightforward way to fix this seems to be a `TypeCondition` node in the query plan, which filters input representations based on their type.

### DIFF
--- a/gateway-js/src/__tests__/integration/abstract-types.test.ts
+++ b/gateway-js/src/__tests__/integration/abstract-types.test.ts
@@ -1278,6 +1278,252 @@ it("when including the same nested fields under different type conditions", asyn
     `);
   });
 
+it('when including multiple nested fields to the same service under different type conditions', async () => {
+  const query = `#graphql
+    query {
+      topProducts {
+        ... on Book {
+          name
+          price
+          reviews {
+            author {
+              name
+              username
+            }
+            editor {
+              name
+              username
+            }
+            body
+            id
+          }
+        }
+        ... on TV {
+          name
+          price
+          reviews {
+            author {
+              name
+              username
+            }
+            editor {
+              name
+              username
+            }
+            body
+            id
+          }
+        }
+      }
+    }
+  `;
+
+    const { queryPlan, errors } = await execute({ query }, [
+      {
+        name: 'accounts',
+        typeDefs: gql`
+          type User @key(fields: "id") {
+            id: ID!
+            name: String
+            username: String
+          }
+        `,
+      },
+      {
+        name: 'products',
+        typeDefs: gql`
+          type Book implements Product @key(fields: "isbn") {
+            isbn: String!
+            title: String
+            year: Int
+            name: String
+            price: Int
+          }
+
+          type TV implements Product @key(fields: "id") {
+            id: String!
+            title: String
+            year: Int
+            name: String
+            price: Int
+          }
+
+          type Computer implements Product @key(fields: "id") {
+            id: String!
+            title: String
+            year: Int
+            name: String
+            price: Int
+          }
+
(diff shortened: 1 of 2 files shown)


lrlna (approved):
This looks good! 

The only thing I can add is that I'd make sure y'all run `cargo clippy` and fix the warning that it gives you. There are a bunch of unused imports and a few other nits in various parts of this repo.

martijnwalraven:
Closing this in favor of #671.

PR #611, opened 2021-03-26
Investigate not re-executing for single fetch queryplans

michael-watson:
There are [commented reasons](https://github.com/apollographql/federation/blob/main/gateway-js/src/executeQueryPlan.ts#L78) about why we need to re-execute, but this isn't necessarily needed for single **Fetch** query plans, meaning:

```
QueryPlan {
  Fetch(service: "search") {
      ...
  }
}
```
For "single Fetch" queryplans, the gateway is essentially forwarding the GraphQL request to a single federated GraphQL service. This is very common when migrating from a monolith to a federated architecture and the performance bump would help ease the adoption curve.

**Note**: This is to discuss the potential performance improvement from removing this step and provide a tarball to test

### DIFF
--- a/gateway-js/src/__tests__/gateway/composedSdl.test.ts
+++ b/gateway-js/src/__tests__/gateway/composedSdl.test.ts
@@ -21,7 +21,7 @@ describe('Using csdl configuration', () => {
     const server = await getCsdlGatewayServer();
 
     fetch.mockJSONResponseOnce({
-      data: { me: { id: 1, username: '@jbaxleyiii' } },
+      data: { me: { id: "1", username: '@jbaxleyiii' } },
     });
 
     const result = await server.executeOperation({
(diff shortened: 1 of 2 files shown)


abernix:
I think this is a very worthy experiment, but I don't want to leave this open as a tempting shortcut since there could be cases in which this approach (i.e., merely _not_ running `execute` over it again) could produce invalid results because of the way we merge fields post-execution and some over-fetching we do that is coupled to that logic. (e.g., `@skip`/`@include` on inline fragments or fragment spreads are not taken into account when fetching for example, and some complex type conditions would not be preserved).  There's also some questions about implicit trust of the underlying subgraph that are worth discussing.  I'm largely summarizing what I've heard from @martijnwalraven about this and trying to clean-up outstanding PRs on this repository, but this is the short of it.

I've opened https://github.com/apollographql/federation/issues/660 to track the need to optimize single-fetch plans, as it does seem like a worthy optimization to make (single service fetches are surely somewhat common — e.g., `{ user { name } }`).  There are also circumstantial opportunities to improve it in a more general sense (e.g., for multi-fetches, too).  We'll need to dig further into that within that thread.

Thanks very much for doing this experiment!  Valuable conversation has come out of it so far.

PR #612, opened 2021-03-26
Remove undefined entities when executing `Fetch` of query plan

michael-watson:
Sometimes a given query will select concrete types that implement an interface, which in turn generates a query plan to be executed across the selected types to downstream services. In executing the query plan, it is possible to have an array of entities that are undefined. In the case of an undefined entity, we currently still send the request to the downstream service when we should return. 

This PR introduces the idea of removing undefined entities prior to [checking if the `entities` array is greater than 0](https://github.com/apollographql/federation/blob/main/gateway-js/src/executeQueryPlan.ts#L206).

### DIFF
--- a/gateway-js/src/__tests__/executeQueryPlan.test.ts
+++ b/gateway-js/src/__tests__/executeQueryPlan.test.ts
@@ -340,6 +340,127 @@ describe('executeQueryPlan', () => {
         }
       `);
     });
+
+    it(`should not send request to downstream service when entities don't match type conditions`, async () => {
+      const reviewsEntitiesResolverSpy = spyOnEntitiesResolverInService(
+        'reviews',
+      );
+
+      const operationString = `#graphql
+        query {
+          # The first 3 products are all Furniture
+          topProducts(first: 3) {
+            ... on Book {
+              reviews {
+                body
+              }
+            }
+          }
+        }
+      `;
+
+      const operationDocument = gql(operationString);
+
+      const operationContext = buildOperationContext({
+        schema,
+        operationDocument,
+        operationString,
+        queryPlannerPointer,
+      });
+
+      const queryPlan = buildQueryPlan(operationContext);
+
+      const response = await executeQueryPlan(
+        queryPlan,
+        serviceMap,
+        buildRequestContext(),
+        operationContext,
+      );
+
+      expect(reviewsEntitiesResolverSpy).not.toHaveBeenCalled();
+
+      expect(response).toMatchInlineSnapshot(`
+        Object {
+          "data": Object {
+            "topProducts": Array [
+              Object {},
+              Object {},
+              Object {},
+            ],
+          },
+        }
+      `);
+    });
+
+    it(`should send a request to downstream services for the remaining entities when some entities don't match type conditions`, async () => {
+      const reviewsEntitiesResolverSpy = spyOnEntitiesResolverInService(
+        'reviews',
+      );
+
+      const operationString = `#graphql
+        query {
+          # The first 3 products are all Furniture, but the next 2 are Books
+          topProducts(first: 5) {
+            ... on Book {
+              reviews {
+                body
+              }
+            }
+          }
+        }
+      `;
+
+      const operationDocument = gql(operationString);
+
+      const operationContext = buildOperationContext({
+        schema,
+        operationDocument,
+        operationString,
+        queryPlannerPointer,
+      });
+
+      const queryPlan = buildQueryPlan(operationContext);
+
+      const response = await executeQueryPlan(
+        queryPlan,
+        serviceMap,
+        buildRequestContext(),
+        operationContext,
+      );
+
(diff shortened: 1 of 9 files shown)


gateway-js/src/executeQueryPlan.ts
```diff
@@ -25,6 +25,7 @@ import {
   getResponseName
 } from '@apollo/query-planner';
 import { deepMerge } from './utilities/deepMerge';
+import { isNotNullOrUndefined } from '@apollographql/apollo-tools';
```
abernix, line 28: Could we use the version of this that's defined in this `@apollo/gateway`'s `./utilities/array.ts`, instead?  It's not currently `export`'d from its module, but it could be!

https://github.com/apollographql/federation/blob/3aa1b084b70a139c1ee9885e43e537be18480822/gateway-js/src/utilities/array.ts#L1-L5

cc @michael-watson or @martijnwalraven

michael-watson: Just pushed a commit with this change

gateway-js/src/__tests__/executeQueryPlan.test.ts
```diff
@@ -142,6 +142,73 @@ describe('executeQueryPlan', () => {
...
+                  "sku": "TABLE1",
+                  "upc": "1",
+                },
+              },
+            ],
+          },
+        },
+      }
+  `);
+      expect(response).toHaveProperty('data.me.reviews', expect.any(Array));
```
abernix, line 209: This `expect` could be removed, given the snapshot which has already (just) affirmed there is an `Array` on `data.me.reviews`.
```suggestion
```

gateway-js/src/__tests__/executeQueryPlan.test.ts
```diff
@@ -142,6 +142,73 @@ describe('executeQueryPlan', () => {
...
     });
 
+    it(`should not send request to downstream services when all entities are undefined`, async () => {
+      overrideResolversInService('reviews', {
+        User: {
+          reviews() {//We are using an example where a past product was removed from the system
+            return [{id:1,body: 'Love it!', product: undefined}, {id:1,body: 'Love it!', product: { __typename: 'Furniture', upc: '1'}}];
+          },
+        },
+      });
```
abernix, line 152: If you wanted to try and check the number of times this resolver is invoked (which I believe would be representative of ensuring there weren't unexpected `FetchNode` invocations), I think you could [use a `jest.fn(...)` Jest Mock function](https://jestjs.io/docs/mock-functions) with the resolver implementation passed to it so you can watch its calls.

```suggestion
	  const reviews = jest.fn(() => ([
	    { id: 1, body: 'Love it!', product: undefined },
	    { id: 1, body: 'Love it!', product: { __typename: 'Furniture', upc: '1'} }
	  ]);
	  
      overrideResolversInService('reviews', {
        User: {
          reviews,
        },
      });
```

... further down, you can:

```js
      expect(reviews.mock.calls.length).toBe(/* The expected number here */);
```

gateway-js/src/__tests__/executeQueryPlan.test.ts
```diff
@@ -142,6 +142,73 @@ describe('executeQueryPlan', () => {
       expect(response).toHaveProperty('errors.0.extensions.variables', {});
     });
 
+    it(`should not send request to downstream services when all entities are undefined`, async () => {
+      overrideResolversInService('reviews', {
+        User: {
+          reviews() {//We are using an example where a past product was removed from the system
+            return [{id:1,body: 'Love it!', product: undefined}, {id:1,body: 'Love it!', product: { __typename: 'Furniture', upc: '1'}}];
```
abernix, line 149: Not a hard ask, and not sure if this is even a reasonable suggestion, but to avoid bringing/duplciating the fixture data directly into this file (snapshots seems fine, but the data itself might be nice to define in one place), it _might be reasonable_ to `export` the `reviews` from `fixtures/reviews.ts` (perhaps renaming it to `reviewsData` in the process?) and use a transformation on it to set one of the elements to `undefined`.

e.g.,

```typescript
import { reviews } from 'apollo-federation-integration-testsuite';


/* ... down to here ... */

const reviewsData = reviews.reviewsData;

return [
  { ...reviewsData[0], product: undefined },
  { ...reviewsData[1], },
];
```

abernix (changes requested):
Looks pretty good to me!  Thanks for adding the test!  I've left a couple comments within, but also happy to make the final adjustments myself tomorrow if you don't have time.

gateway-js/src/__tests__/executeQueryPlan.test.ts
michael-watson: Good call out, the latest push has changes to expose that data set and use it in the test

abernix (approved):
It was observed that the newly introduced test didn't fail when the implementation change was made, but a couple follow-up commits have resolved that — in addition to surfacing a separate issue involving unmet type-conditions.

LGTM now.  I'll ship this.

PR #649, opened 2021-04-08
infra: alter GH Actions caching technique for Rust

abernix:
This choses a technique that is employed in the `rover` repository and doesn't cache the `target` directory in an effort to absolve ourselves from the odd caching circumstances which are causing build failures on the various branches subsequent invocations of their CI workflows.


abernix:
We're still seeing some weird caching-related failures on https://github.com/apollographql/federation/pull/648 via https://github.com/apollographql/federation/pull/648/checks?check_run_id=2305336658, which I suspect this will fix.

pcmanus (approved):
I'm far from a GH actions expert, but this look reasonable to me.

abernix:
Note I've also pulled in https://github.com/apollographql/federation/pull/649/commits/3851ce7691a3ca1f8efb5b31fb055cf57d8fcaec

PR #658, opened 2021-04-15
chore(ci): Start running Rust tests on CircleCI

abernix:
This leaves GitHub Actions for the time being, but so far this is already showing faster spin-up times and more reliable behavior (less long random waits) along with faster completion rates.  I might suggest keeping both on for a week or so for the sake of racing them / comparing their failures.

Couple that with the relative simplicity / lower cognitive overhead of only having a single provider and having all the workflows appear in a single unified format that consider all the failure cases (e.g., a Rust failure can now properly short-circuit a release if it need to).

This intentionally disables caching for the time-being, while I wait for consideration of an upstream PR that I think is useful.  Still, even with the cache disabled, it would seem to be beating our same/similar GitHub Actions setup (which has caching enabled).  While the GitHub Actions did leverage separate jobs for linting, this currently lumps them together.  Happy to consider adjusting that later, and should be a simple change.


glasser (approved):
I'm not a huge Rust or Circle expert but everything here seems plausible.

I think we should remove the GH workflow files very soon after merging this PR, or arguably just as part of the same PR.

I also think we should consider how much value we are getting out of running Rust CI on three platforms in this repo, when (at least without caching) macos and windows are notably slower than linux, which is itself notably slower (2x) than JS.  Given that there's only one bit of Rust left and most PRs won't be directly touching it, how likely is it that we will be breaking things on Mac or Windows only?  Or is there an argument to be made that since the Rust code we're testing is "packaging" rather than "algorithm", it's in fact the most likely sort of thing to be broken on just one platform? We could perhaps at least leave the "Rust-rust_macos" and "Rust-rust_windows" tasks as not necessary in order to merge PRs...

glasser:
Since your current schedule is a bit up in the air, and the current Rust CI is actively distracting and slow on PRs in this repo, I've taken it upon myself to try to finish your incomplete sentence and to delete the GH workflow files (but still leave all three CircleCI Rust jobs in); I'll get an approval from another colleague before merging but won't wait on you.

.circleci/config.yml
```diff
@@ -64,6 +79,128 @@ jobs:
...
+# The caches above on the Rust orb jobs are all disabled intentionally because
+# the `cache_version` they are offering in their keys is, I think, misplaced.
+# The misplacement is resulting in caches being restored to hosts of differing
+# architectures than the host that created the cache. While this might work for
+# some common-foundation architectures (e.g. Unix vs Linux, with the same
+# usernames), it breaks down with even the slightest variation (e.g., macOS has
+# different home directory) structures, so the cache is attempted to be
+# restored to $HOME paths with different user, which fails.
+#
+# I've opened https://github.com/CircleCI-Public/rust-orb/issues/10 to capture
```
glasser, line 200: I am now subscribed to these FYI

.circleci/config.yml
```diff
@@ -64,6 +79,128 @@ jobs:
...
+# this bug, and I've opened https://github.com/CircleCI-Public/rust-orb/pull/11
+# to hopefully fix it.  Once this lands, each `with_cache: false` line above
+# can be removed.
```
abernix, line 203:
```suggestion
# to hopefully fix it.  Once this lands, each `with_cache: false` line above
# can be changed:
#   - Only the first job in the list needs caching enabled
#     (otherwise, we'll pay the cost of restore for already-restored content)
#   - They should have 'cache_version' parameters passed with the platform:
#     e.g. 'cache_version: v1-mac'
```

abernix:
> macos and windows are notably slower than linux, which is itself notably slower (2x) than JS. [...]  We could perhaps at least leave the "Rust-rust_macos" and "Rust-rust_windows" tasks as not necessary in order to merge PRs

The jobs run in about 5 minutes.  Do we need to optimize for fully merging and landing a PR in under 5 minutes?  I'm happy to add manual caching prior to the PRs, prior to upstream cache benefits being included but what are we optimizing for here?

> how much value we are getting out of running Rust CI on three platforms in this repo [...] how likely is it that we will be breaking things on Mac or Windows only? Or is there an argument to be made that since the Rust code we're testing is "packaging"

I agree we're mainly testing packaging, but I wasn't limiting it to that in my mind.  The consumer (Rover) of this package needs to trust that it will work on the platforms they need support.  This helps avoid surprises in that regard through being complete and not possibly surprising a future developer who tries to make a change to the Rust code in a way that does change the algorithm or challenges any likelihoods we've come up with here today.

.circleci/config.yml
```diff
@@ -64,6 +79,127 @@ jobs:
...
+# The misplacement is resulting in caches being restored to hosts of differing
+# architectures than the host that created the cache. While this might work for
+# some common-foundation architectures (e.g. Unix vs Linux, with the same
+# usernames), it breaks down with even the slightest variation (e.g., macOS has
+# different home directory) structures, so the cache is attempted to be
+# restored to $HOME paths with different user, which fails.
+#
+# I've opened https://github.com/CircleCI-Public/rust-orb/issues/10 to capture
+# this bug, and I've opened https://github.com/CircleCI-Public/rust-orb/pull/11
+# to hopefully fix it.  Once this lands, each cache above
```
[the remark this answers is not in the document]
abernix: Not quite what I was going to suggest, but glad you caught it. 😄

abernix:
Ok, I cut a couple minutes off through both caching `~/.cargo` (via the Orb) and caching (directly) the `<checkout>/target` directory.  Not much win was observed by merely caching `~/.cargo`, but a substantial win was achieved through `target`.  On the other hand, caching `target` would be the first thing I'd remove if this acts up.

glasser:
I do think there's a real difference in developer experience between 90 second and 5 minute CI roundtrip, but I don't think we need to worry about that today in a PR that's already a huge improvement.

.circleci/config.yml
```diff
@@ -64,6 +79,133 @@ jobs:
...
+                command: |
+                  echo 'Node.js version:'
+                  node --version
+                  echo 'npm version:'
+                  npm --version
+            - run: npm install
+            - rust/format:
+                cache_version: v1-macos
+            - rust/clippy:
+                with_cache: false # Only the first needs to restore cache.
```
glasser, line 121: Hmm, this also means that only rust/format will *save* to the `~/.cargo` cache.  Does format actually do enough of the build for this to be worth it? (Maybe `rust/test` being first would be better, since I kinda assume test does more?) I guess those commands are designed to be run in parallel in separate jobs.

(Despite this comment, I think this PR as stands is a big improvement and have no interest in blocking merge.)

abernix: I had thought `cargo fmt` did more of the compilation but it was `cargo clippy`.  I think it's worth moving them up, though my intent was originally to fail-fast for what was probably the most common thing to be missed locally: `fmt`-ing.

abernix: There's also only one test.

glasser:
(btw we should remember when we merge to update the list of required checks at https://github.com/apollographql/federation/settings/branch_protection_rules/19580412)

abernix:
Shifting `test` up didn't help the overall runtime much, 6m4s vs 6m1s.  Run at different times of day though, so 🤷 .  Anyhow, leaving it how it is now and putting this down.

> (btw we should remember when we merge to update the list of required checks at apollographql/federation/settings/branch_protection_rules/19580412)

Yeah, we won't be able to merge this PR without overriding as an admin or doing that first.  I'd recommend the latter. 😉

glasser:
I added a commit to make "JS Package tarballs" able to run concurrently with "Rust" rather than block on it. (This doesn't seem like it should be a regression given that the old Workflow version had no dependency here.) Note that this isn't touching the dependency on Rust for the publish workflows, just for the per-PR tarballs. That reduces overall workflow duration to be equal to the slowest Rust build rather than the slowest Rust build plus "package tarballs"; the execution for this commit was total duration 4:15 rather than 6:01 for the previous commit.

.circleci/config.yml
```diff
@@ -7,6 +7,8 @@ orbs:
   # be found at http://github.com/apollographql/CircleCI-Orbs/.
   # We could use Renovate to bump this version via PR, but that's not set up now.
   oss: apollo/oss-ci-cd-tooling@0.0.16
+  rust: circleci/rust@1.5.0
+  win: circleci/windows@2.3.0
```
glasser, line 11: BTW I think this might be unused?

.circleci/config.yml
```diff
@@ -7,6 +7,8 @@ orbs:
   # be found at http://github.com/apollographql/CircleCI-Orbs/.
   # We could use Renovate to bump this version via PR, but that's not set up now.
   oss: apollo/oss-ci-cd-tooling@0.0.16
+  rust: circleci/rust@1.5.0
```
trevor-scheer, line 10:
```suggestion
  # Rust steps which are used below (like `rust/install`, `rust/test`) are
  # defined in this orb. For reference, the orb can be found here:
  # https://github.com/CircleCI-Public/rust-orb
  rust: circleci/rust@1.5.0
```

trevor-scheer (approved):

.circleci/config.yml
trevor-scheer: Poked around a bit and this seems correct, pushed 997dfee and will see how CI does.

trevor-scheer:
Just noting this likely interim state with GH actions being stuck on "expected". Going to admin merge this now.
image

glasser:
> We could perhaps at least leave the "Rust-rust_macos" and "Rust-rust_windows" tasks as not necessary in order to merge PRs...

I did do this when updating branch settings. I think that "changes which break Rust on Mac or Windows but not Linux or anything else" will be rare enough that "notice it on PRs where all the tasks have finished running" will be good enough.

glasser:
> I did do this when updating branch settings. I think that "changes which break Rust on Mac or Windows but not Linux or anything else" will be rare enough that "notice it on PRs where all the tasks have finished running" will be good enough. If this bites us we can always change it.

@trevor-scheer raised the concern that it would be bad if Renovate upgraded packages that Harmonizer depends on (Rust, Rollup, etc) without ensuring Mac and Windows builds pass. We investigated and [confirmed that Renovate doesn't care about whether status checks are required](https://github.com/renovatebot/renovate/issues/1853#issuecomment-822737152), so Renovate will still require Mac and Windows to pass.

.circleci/config.yml
abernix: Good catch!  Indeed not necessary.  I had started with this approach, but I couldn't customize one thing I needed to within that orb and didn't have time to draw up a contribution to the orb itself.

abernix:
> We investigated and confirmed that Renovate doesn't care about whether status checks are required, so Renovate will still require Mac and Windows to pass.

This was one of my concerns too so this is good to know! Thanks for checking.  As another question: How does GitHub's new Auto-merge button functionality behave in this scenario?  Will it merge as soon as all required tests pass, or will it wait until all tests pass?  If one fails, will it merge?

abernix:
@glasser:

> if we even once create a PR where all the required checks pass but Rust Mac or Windows fails
> then that would be enough for me to want to turn them back to required.

I'm curious why putting the cognitive burden of this on a future member of the team or someone who couldn't clearly glean this context (or who might miss this nuance) is desirable versus just marking the tests are required today?  I'd asked above about whether we were needing to land PRs in under 5 minutes or what we were trying to optimize for, but I don't think I've completely understood how we're not meeting the needs of developers who are contributing to this repository right now by having them all required:

> Do we need to optimize for fully merging and landing a PR in under 5 minutes? I'm happy to add manual caching prior to the PRs, prior to upstream cache benefits being included but what are we optimizing for here?

Note that caching has been added and jobs finish in about 3.5 minutes now after some follow-up work.  I'm certainly interested in understanding your perspective / dissatisfaction with having them as required better, but as observations from what I can see:

* There is still feedback in the same amount of time as before on the most important Node.js tests which are unchanged.
* I don't see this repository as a place where _minutes/seconds matter_ (e.g., different than an infrastructure repository).
* PRs should still be getting reviews from other contributors to the repository, and those will almost surely take more than 3.5 minutes to obtain.

Overall, I think it's important that our CI tests be reliable, thorough, enforcing the contracts we extend to our consumers, and that our tooling holds developers accountable to this.  Again, quoting myself above:

> This helps avoid surprises in that regard through being complete and not possibly surprising a future developer who tries to make a change to the Rust code in a way that does change the algorithm or challenges any likelihoods we've come up with here today.

I'd feel more comfortable with these jobs as all required, today, so we don't need to keep this in the back of our minds.

glasser:
I think the difference between 3.5 minutes and 1.5 minutes is a lot, especially when crunching through a bunch of PRs which may need to be slightly rebased on top of each other, or when you're thinking "oh I saw a typo in a comment/CHANGELOG, is it worth tweaking if I then have to sit there and wait", and I think that it is unlikely that blocking on Mac and Win Rust tests *in a repo that barely contains Rust* is unlikely to solve real problems. I think that any problem caused by this will very likely be caught the next time anybody looks at a PR that has been sitting for 3.5+ minutes. I think my concerns are a bit smaller in the world of auto-merge, though, and my personal focus is more on apollo-server than federation these days, so I am happy to disagree and move on.

The two checks are now required for `main`.

PR #673, opened 2021-04-20
`JSON.stringify()` should be called only once for the single HTTP call

ValeriyMaslenikov:
The current behavior of `RemoteGraphQLDataSource.ts` is to stringify the body twice. For the applications require to have low SLO for overhead created by the gateway it may be critical. 

We didn't measure the real impact on performance, but it'll take less effort to fix this small piece of code, instead of writing metrics/tests to analyze the performance.

### DIFF
--- a/gateway-js/src/datasources/RemoteGraphQLDataSource.ts
+++ b/gateway-js/src/datasources/RemoteGraphQLDataSource.ts
@@ -140,10 +140,10 @@ export class RemoteGraphQLDataSource<TContext extends Record<string, any> = Reco
     // being transmitted.  Instead, we want those to be used to indicate what
     // we're accessing (e.g. url) and what we access it with (e.g. headers).
     const { http, ...requestWithoutHttp } = request;
-    const requestBody = JSON.stringify(requestWithoutHttp);
+    const stringifiedRequestWithoutHttp = JSON.stringify(requestWithoutHttp);
     const fetchRequest = new Request(http.url, {
       ...http,
-      body: requestBody,
+      body: stringifiedRequestWithoutHttp,
     });
 
     let fetchResponse: Response | undefined;
@@ -153,7 +153,7 @@ export class RemoteGraphQLDataSource<TContext extends Record<string, any> = Reco
       // Use the fetcher's `Request` implementation for compatibility
       fetchResponse = await this.fetcher(http.url, {
         ...http,
-        body: requestBody
+        body: stringifiedRequestWithoutHttp,
       });
 
       if (!fetchResponse.ok) {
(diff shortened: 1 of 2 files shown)


gateway-js/src/datasources/RemoteGraphQLDataSource.ts
```diff
@@ -140,9 +140,10 @@ export class RemoteGraphQLDataSource<TContext extends Record<string, any> = Reco
     // being transmitted.  Instead, we want those to be used to indicate what
     // we're accessing (e.g. url) and what we access it with (e.g. headers).
     const { http, ...requestWithoutHttp } = request;
+    const requestBody = JSON.stringify(requestWithoutHttp);
```
abernix, line 143:
```suggestion
    const stringifiedRequestWithoutHttp = JSON.stringify(requestWithoutHttp);
```

gateway-js/src/datasources/RemoteGraphQLDataSource.ts
```diff
@@ -140,9 +140,10 @@ export class RemoteGraphQLDataSource<TContext extends Record<string, any> = Reco
...
     const fetchRequest = new Request(http.url, {
       ...http,
-      body: JSON.stringify(requestWithoutHttp),
+      body: requestBody,
```
abernix, line 146:
```suggestion
      body: stringifiedRequestWithoutHttp,
```

gateway-js/src/datasources/RemoteGraphQLDataSource.ts
```diff
@@ -152,7 +153,7 @@ export class RemoteGraphQLDataSource<TContext extends Record<string, any> = Reco
       // Use the fetcher's `Request` implementation for compatibility
       fetchResponse = await this.fetcher(http.url, {
         ...http,
-        body: JSON.stringify(requestWithoutHttp)
+        body: requestBody
```
abernix, line 156:
```suggestion
        body: stringifiedRequestWithoutHttp,
```

abernix, line 156: This wasn't introduced by this PR, but the adjustments here bring it to mind again since we have the opportunity / it's tempting to pass the entire object (`{ ...http, body: ... }`) to this invocation of what is meant to be a [Fetch API-standard](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) `fetcher`, rather than creating it separately.

Put another way, it wasn't immediately clear to me why this `fetcher` can no longer accept the `RequestInit` type that we have created (which is what `fetchRequest` is) above.

That recalls the note in https://github.com/apollographql/federation/pull/188 (which made the change):

> I debugged a lot and found the ultimate cause - which is we need the Request to be constructed by the fetcher's own Request class instead of node fetch's Request class as done currently)

So that's me _not asking for a change_, but just noting that I found it more peculiar that `make-fetch-happen` _couldn't_ accept `RequestInit` and that we had to dance around that in this way.  Almost seems like a bug/limitation of `make-fetch-happen`.

abernix (approved):
Thanks so much for opening this!

My only request is that we use a different name for the variable, though I did have a small realization about some related code that I've noted within.

abernix (changes requested):
Changing to "Request changes", since I did, but I'm happy to wrap this up if you aren't able to revisit this. 😉

Issue #540, opened 2021-03-05
ci: convert to using CircleCI


abernix:
This is to track an intent/suggestion on my part to change our test-suite in this repository from running within a mix of using GitHub Actions _and_ CircleCI to **just** using CircleCI.

This follows-up an outstanding TODO from https://github.com/apollographql/federation/pull/480 which revealed a number of design choices which prevent reusability of code that is necessary to repeat in a workflow/fanout design.  CircleCI has supported this (and basic YAML features, like Anchors/Aliases) for a while now plus we already run all of our other tests (not only here, but also in other repositories) in CircleCI.  GitHub Actions are certainly nifty but a single, more powerful unified tool seems like a better choice for this particular task.

benweatherman:
I think this is all done now. Please reopen or add new issues if needed!

PR #682, opened 2021-04-22
infra(ci): Adjust npm installation strategy in Rust containers

abernix:
This is a follow-up on https://github.com/apollographql/federation/issues/540
but also what I found to be a valuable precursor to an impending upgrade to
npm@7 for this monorepo.  Essentially, this brings some durability to the
npm version that we expect to be installed in the containers, rather than
leveraging the version that ships with a version of Node.js and allows those
to move forward separately.  For example, using Node.js 14 within the Rust
containers seems prudent but that still ships with npm 6.  When we upgrade
the repository to npm 7, that version will no longer be suitable for the
`package-lock.json` that will be checked into this repository, but we
probably don't want to make the version of Node.js that runs in that
container Node.js 16 for a few more months when it reaches LTS.

Generally speaking, this seems useful to have separated to spare a future
person from having to unravel any complexities that I went through in
accomplishing this.  (I could very much see this logic living in a CircleCI
Orb elsewhere and I may take on that initiative eventually but it's a
heavier lift right now.)

Thusly, I'm opening this as a pre-req for an upcoming upgrade to `npm@7` that
can live on its own and still add value.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -112,12 +112,11 @@ jobs:
                 command: |
                   sudo npm install --global npm@${NPM_VERSION}
             - run:
-                name: Debug output of Node.js and npm versions
-                command: |
-                  echo 'Node.js version:'
-                  node --version
-                  echo 'npm version:'
-                  npm --version
+                name: Assert Node.js version
+                command: test "$(node --version)" = "v${NODE_VERSION}"
+            - run:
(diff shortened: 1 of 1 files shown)


trevor-scheer (approved):
Appreciate the replacing of debug output with assertions, that's a nice touch.

PR #686, opened 2021-04-22
Add locations to composition errors

mayakoneval:
**This PR**
- adds location information to composition errors that are specific to federation. This includes pre, post and composition errors.
- There are a couple things that needed to be in place for location information to be exposed properly. The `gql` function [strips `loc` information](https://github.com/apollographql/graphql-tag/blob/main/src/index.ts#L106) from the type definitions that it returns, so all tests that I edited use `parse` instead of `gql`.
- Composition errors use a function `errorWithCode` which calls the `GraphQLError` constructor. If `node` is provided to the `GraphQLError` constructor with a `loc` field on `node` with a `source` on `loc`, the location information added to an error is the same as the location information on node. This took me a while to get :)
- This PR is organized by error. Each commit is a composition error code [as defined in the federation docs](https://www.apollographql.com/docs/federation/errors/).
- The first commit moves a couple tests from `composeAndValidate` into the `describe('error` because that seemed more fitting 🤷 I would be fine deleting this one if anyone disagrees. 
- I updated all the error code specific tests that I could find, as well as adding locations to `composeAndValidate.test.ts` and `compose.test.ts`

**I recommend reviewing by commit**

Please let me know of any concerns switching `gql` to `parse`. 

**Questions**
- Should I change all gql calls to parse calls in the test files I edited? I only changed the tests that checked the snapshot of the whole error, rather than the length of errors, since those are the ones that would check for location.


trevor-scheer:
How do you feel about using a tool like [strip-indent](https://www.npmjs.com/package/strip-indent) in order to normalize all the whitespace in these test cases before parsing? Right now the indentation is inconsistent as reflected by the `locations` columns (based on the nested-ness of the string template literal).

Maybe we can write a small test util which does everything we need it to: normalize indentation, parse to AST, and still use (a fake) `gql` tag:

```ts
import { parse } from "graphql";
import stripIndent from 'strip-indent';

export function testingGql(strings: TemplateStringsArray) {
  return parse(stripIndent(strings[0]));
};

// other file
import { testingGql as gql } from '../../testUtils';

const result = gql`
  type Query {
    abd: ID!
  }
`;
```

mayakoneval:
> How do you feel about using a tool like [strip-indent](https://www.npmjs.com/package/strip-indent) in order to normalize all the whitespace in these test cases before parsing? Right now the indentation is inconsistent as reflected by the `locations` columns (based on the nested-ness of the string template literal).

My only concern with this is that what you put in a test directly does not reflect exactly what will be in the composition error locations. If you put in an sdl with whacky spacing, the strip indent might normalize that and the columns you get back might not be what you expect.

It is definitely frustrating to count all the columns, though. No strong opinions.

federation-js/src/composition/validate/postComposition/executableDirectivesIdentical.ts
```diff
@@ -51,6 +51,7 @@ export const executableDirectivesIdentical: PostCompositionValidator = ({
                 return `\t${serviceName}: ${print(definition)}`;
               })
               .join('\n')}`,
+          directive.astNode || undefined,
```
trevor-scheer, line 54: Minor nit: `??` is a more idiomatic way to write this fallback behavior which is fairly recently supported in the version of typescript that we use.
Ref: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#nullish-coalescing
```suggestion
          directive.astNode ?? undefined,
```

federation-js/src/composition/validate/postComposition/__tests__/executableDirectivesIdentical.test.ts
```diff
@@ -81,6 +82,12 @@ describe('executableDirectivesIdentical', () => {
       Array [
         Object {
           "code": "EXECUTABLE_DIRECTIVES_IDENTICAL",
+          "locations": Array [
+            Object {
+              "column": 9,
+              "line": 2,
+            },
+          ],
```
trevor-scheer, line 90: What's the expected use case for post composition error locations? Are we going to be able to represent these in studio?

From strictly a typical gateway use case, the location is not actionable to me since the supergraph is more or less hidden away from me as an implementation detail of the gateway.

mayakoneval: I might not be understanding when post composition error rules are used. Are the composition errors we get on the frontend only the pre and during composition errors? Maybe worth nothing that the supergraph is visible to users in studio.

trevor-scheer (commented):
The structure and approach of this PR look great! One major question w.r.t. how we can use locations from post composition errors, and the lingering question about indentation in testing.

I haven't reviewed correctness of the locations yet pending this^ possible change since they might all be rewritten.

Please add a an entry to the federation `CHANGELOG.md` as well when you get the chance.

federation-js/src/composition/validate/postComposition/__tests__/executableDirectivesIdentical.test.ts
trevor-scheer: Good to know - I think we'll find these useful then (not that I would've had you remove them or anything, in any case I think it's good they're there).

Precomposition locations will apply to subgraph SDL, and postcomposition locations will apply to supergraph SDL. The biggest difference there is that the supergraph SDL is compiled, so it only sorta makes sense to point to a spot on it and say "this is wrong, fix it", since nobody _authored_ the supergraph SDL.

And on second thought, in the case of errors there won't _be_ a supergraph SDL. I'm not sure we'll be able to utilize these until we have a concept of warnings (which would give us both a supergraph SDL and a set of errors to address). That is currently being talked about though.

mayakoneval: Yes that makes sense to me. composition error hints like “hey fix this on line x” will only apply to the subgraph asks.

martijnwalraven:
> > How do you feel about using a tool like [strip-indent](https://www.npmjs.com/package/strip-indent) in order to normalize all the whitespace in these test cases before parsing? Right now the indentation is inconsistent as reflected by the `locations` columns (based on the nested-ness of the string template literal).
> 
> My only concern with this is that what you put in a test directly does not reflect exactly what will be in the composition error locations. If you put in an sdl with whacky spacing, the strip indent might normalize that and the columns you get back might not be what you expect.

I have no strong opinions either. I get that concern, but I think `strip-indent` should  be pretty safe because it doesn't do anything fancy, it just strips common leading whitespace from every line. Not sure if that makes things that much easier though. You'll still need to visually inspect whether the locations in the snapshot make sense.

(As a point of reference, it seems `graphql-js` [does rely on absolute locations in a comparable test](https://github.com/graphql/graphql-js/blob/2d48fbbeb8718e4a0152d458145a9fe2111c0f8d/src/validation/__tests__/UniqueTypeNamesRule-test.js#L118-L161).)

martijnwalraven:
Regardless of whether we remove indentation, I like the idea of sticking to `gql` and replacing the implementation in `graphql-tag` with one of our own that doesn't remove locations. That seems nicer than (selectively) replacing `gql` with `parse` in some places.

trevor-scheer:
cc @martijnwalraven
> I have no strong opinions either. I get that concern, but I think strip-indent should be pretty safe because it doesn't do anything fancy, it just strips common leading whitespace from every line. Not sure if that makes things that much easier though. You'll still need to visually inspect whether the locations in the snapshot make sense.

I'd argue this makes it a bit easier for review and less error-prone for validating.
1) Tests of the same error code will (often/sometimes?) have the same base location, so numbers will align and be more scannable.
2) Less error-prone whitespace counting from the relative normal based on current nesting than absolute (which could be upwards of 10 or so spaces that you have to count.

Can you elaborate on your suggestion?
>Regardless of whether we remove indentation, I like the idea of sticking to gql and replacing the implementation in graphql-tag with one of our own that doesn't remove locations. That seems nicer than (selectively) replacing gql with parse in some places.

Are you suggesting that we want our own `gql` in all cases which doesn't lose location data but does cache the parsed value? Or just an aliased test util which we call `gql` like I alluded to in a previous comment?

abernix:
> Are you suggesting that we want our own gql in all cases which doesn't lose location data but does cache the parsed value? Or just an aliased test util which we call gql like I alluded to in a previous comment?

I took that he wanted `gql` to be aliased to something that wasn't `graphql-tag` / as you suggested.

>  I think strip-indent should be pretty safe

I suspect that using `strip-indent` would have other benefits too, since these files are `prettier`'d every time a snapshot is updated and that comes with the very real possibility of bulk changes to this file.  Any desire to add a `describe` parent-block around any group of tests would necessitate manually updating offsets without `strip-indent`.

> this makes it a bit easier for review and less error-prone for validating.

While I think I agree, I also I think we should be careful with any tempting shortcuts here.  A more durable approach might be to further limit the possibility of these changing circumstantially and without intent (I agree that `strip-indent` helps that, but I worry about getting too comfortable).  For example, we might avoid using template literals at all (e.g., use single line strings, where possible?) to avoid the number of changes?

abernix:
@mayakoneval Certainly no hurry, but do you mind going ahead and introducing `strip-indent` and the introduction of `gql` as generic wrapper for `parse`?

Perhaps you're aware of Lerna-monorepo-isms, but in case you're not and since there aren't contribution guidelines for this repository that convey how to do that in this Lerna monorepo the installation suggestion would be to `npm install --save-dev strip-indent` **at the root**, rather than within the sub-package.  [This issue](https://github.com/apollographql/federation/issues/703) captures some nuances of that process.

Thanks so much for tackling _this whole `locations` thing.  We can get this into an upcoming release once this is all set, for sure.

mayakoneval:
> @mayakoneval Certainly no hurry, but do you mind going ahead and introducing `strip-indent` and the introduction of `gql` as generic wrapper for `parse`?

100% this is on my list for today!

federation-js/CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 -  _Nothing yet! Stay tuned!_
 
+## v0.24.0
+
+- Composition errors now include `location` corresponding to the line number & column in the subgraph sdl. [PR #686](https://github.com/apollographql/federation/pull/686/files)
```
abernix, line 11:
```suggestion
- Composition errors now include `locations` corresponding to the line number & column in the subgraph sdl. [PR #686](https://github.com/apollographql/federation/pull/686/files)
```

abernix (commented):
Drive-by comment / will defer to @trevor-scheer for final approval, I just happened to see this in the `CHANGELOG`.

federation-integration-testsuite-js/src/utils/gql.ts
```diff
@@ -0,0 +1,15 @@
+import { parse } from "graphql";
+import stripIndent from 'strip-indent';
+
+export function gql(
+  literals: string | readonly string[],
```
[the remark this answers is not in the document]
trevor-scheer: Ok, so in my confusion I ended up at the `graphql-tag` source where this is (intelligently) copied from, and I found the reason: https://github.com/apollographql/graphql-tag/commit/75d4a5554b81ed65d6a3dd64495e0d24222857c4

So the `readonly string[]` should technically be `TemplateStringsArray` but this is not a big deal either. The more interesting bit to me was why the `string` and now I understand!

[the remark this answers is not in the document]
trevor-scheer: Might just note in a comment above the function that this whole bit is copied from `graphql-tag` so any questions about its implementation can be taken up with the OG.

federation-js/src/composition/validate/postComposition/__tests__/keyFieldsSelectInvalidType.test.ts
```diff
@@ -113,12 +111,12 @@ describe('keyFieldsSelectInvalidType', () => {
     };
 
     const serviceB = {
-      typeDefs: parse(`
+      typeDefs: gql`
```
trevor-scheer: I haven't decided yet if it's unexpected that there's no change in locations for this commit. Will revisit, just scanning for now.

federation-js/src/composition/validate/sdl/__tests__/matchingEnums.test.ts
```diff
@@ -106,7 +105,11 @@ describe('matchingEnums', () => {
           "code": "ENUM_MISMATCH",
           "locations": Array [
             Object {
-              "column": 11,
+              "column": 1,
```
trevor-scheer, line 108: Noting that some of the diff here was actually expected in a previous commit when we went from passing `definitions[0]` to `definitions` the array. I almost made a comment on that commit since I was a little surprised there was no diff in tests, but figured I'd wait til later and here we are :)

trevor-scheer (commented):
Diff from previous review to now looks good! I'm going to do one final pass for correctness of locations in the AM, but otherwise this LGTM.

federation-integration-testsuite-js/src/utils/gql.ts
```diff
@@ -0,0 +1,25 @@
...
```
abernix, line 5:
```suggestion
// This function signature is mostly lifted from the upstream `graphql-tag`.
// The motivation for this `gql` implementation is the desire to preserve
// the `locations` attributes from parsing, which `graphql-tag` purges.
// We don't typically use the function-invocation (e.g., `gql("string")`),
// but the same mode is preserved from `graphql-tag`.  The similarities
// might be appreciated.  (Or not, but having symmetry in a same-named
// function is somewhat appealing, indeed.)
export function gql(
  literals: string | readonly string[],
```

abernix (dismissed):

federation-js/src/composition/validate/postComposition/__tests__/keyFieldsMissingOnBase.test.ts
```diff
@@ -76,6 +75,12 @@ describe('keyFieldsMissingOnBase', () => {
       Array [
         Object {
           "code": "KEY_FIELDS_MISSING_ON_BASE",
+          "locations": Array [
+            Object {
+              "column": 13,
```
martijnwalraven, line 80: I'm trying to validate these changes, but I'm confused what this location is supposed to refer to. I may be missing something about this PR in general. Are these locations in the composed schema SDL or one of the service SDLs? Looking at [the code that generates the error](https://github.com/apollographql/federation/blob/fbbc85e1adc974c7937929534c691f74923ea87c/federation-js/src/composition/validate/postComposition/keyFieldsMissingOnBase.ts#L41), the location seems to be relative to the document resulting from parsing `@key(fields:)`, and I'm not sure how that is mapped to either of those?

abernix: @martijnwalraven In this case, column 13, line 1 would seem to map to the left-most part of `@key` in `serviceA` above, right? e.g., pointing to the site of the directive on the base `Product` type where the omitted `uid` is?

I think the use-case here is to show users where their composition errors (as defined by the message) exist **on a subgraph** in the Studio UI?  (Note that may require #705?)

martijnwalraven: Running the tests changes the column to 6 though, so I'm not sure how that relates to this. Still wondering how the location in the parsed field set would be related to the location in the overall document. Leaving that aside, it seems to me error locations are not useful without mapping them to individual service SDLs, so shouldn't we resolve that first before landing this?

mayakoneval: > Running the tests changes the column to 6 though
Is this test out of date?

> Leaving that aside, it seems to me error locations are not useful without mapping them to individual service SDLs, so shouldn't we resolve that first before landing this?
I opened an issue to add subgraph name as structured data to composition errors on the `extensions` field of `GraphQLError`. This PR exposes locations, and 90% of the time the locations array has a single location which maps to a single service that can be parsed from the error message. For `ENUM_MISMATCH` there are 2 service names in the error message, and in that case the locations correspond to one service each. 


Is it acceptable to assume the location in locations maps to the service name in the error message for this PR?

Would the best move forward be to 
1. merge this PR and open a separate PR to add subgraph name as structured data.
2. edit this PR to include subgraph names as structured data

federation-integration-testsuite-js/src/utils/gql.ts
mayakoneval: thank you!

mayakoneval:
something is definitely off with the columns - I am looking into it 👀  .

mayakoneval:
Thanks @martijnwalraven for catching this issue. The issue was that for errors on selection sets, the locations were off. Trevor pointed out that we manually parse selections, so the location information was off when using this node as our baseline in `GraphQLError`. I just committed changes that pass the _field_'s node instead of the _selection set_'s field node, this makes the locations match the actual field, or type if there is no field.

federation-js/src/composition/validate/postComposition/__tests__/keyFieldsMissingOnBase.test.ts
martijnwalraven: > This PR exposes locations, and 90% of the time the locations array has a single location which maps to a single service that can be parsed from the error message. For ENUM_MISMATCH there are 2 service names in the error message, and in that case the locations correspond to one service each.

If we're already relying on parsing error messages, merging this and solving structured error metadata separately seems fine. I do think depending on the specific contents of error messages is dangerous, and limits our ability to evolve composition validation, so we should move away from that. But we can talk about that as part of #705.

martijnwalraven:
> Thanks @martijnwalraven for catching this issue. The issue was that for errors on selection sets, the locations were off. @trevor-scheer pointed out that we [manually parse selections](https://github.com/apollographql/federation/blob/d20d6f941e831924b3a59a0048d44e7bafac1dd6/federation-js/src/composition/utils.ts#L159-L164), so the location information was off when using this node as our baseline in `GraphQLError`. I just committed changes that pass the _field_'s node instead of the _selection set_'s field node, this makes the locations match the actual field, or type if there is no field.

That makes sense! If we do want to refer to contents of the parsed field set to give people more specific context, we should consider setting the [`locationOffset` property on `Source`](https://github.com/graphql/graphql-js/blob/917befd7a54fe99d6141fa0ed92bc2b82544625f/src/language/source.d.ts#L16) when parsing selections.

federation-js/src/composition/__tests__/compose.test.ts
```diff
@@ -489,8 +531,34 @@ describe('composeServices', () => {
...
+              "code": "MISSING_ERROR",
+              "locations": Array [
+                Object {
+                  "column": 1,
+                  "line": 2,
+                },
+                Object {
+                  "column": 6,
+                  "line": 2,
+                },
```
trevor-scheer, line 558: I would expect these both to either point at the "t" in type or the "P" in Product. It's entirely plausible this is out of our hands since this error is coming from `graphql-js`.

federation-js/src/composition/__tests__/composeAndValidate.test.ts
```diff
@@ -524,6 +559,16 @@ describe('composition of value types', () => {
...
+          "locations": Array [
+            Object {
+              "column": 1,
+              "line": 6,
+            },
+            Object {
+              "column": 1,
+              "line": 6,
+            },
+          ],
```
trevor-scheer, line 571: Should these locations point to the fields in question instead? (Or better yet the fields' return types)

federation-js/src/composition/__tests__/composeAndValidate.test.ts
```diff
@@ -614,6 +669,16 @@ describe('composition of value types', () => {
...
+          "locations": Array [
+            Object {
+              "column": 1,
+              "line": 14,
+            },
+            Object {
+              "column": 1,
+              "line": 14,
+            },
+          ],
```
trevor-scheer, line 681: Should these locations point to the specific types within the union that are in violation? It looks like we identify that in the error message so I think that info is available to us.

Something just crossed my mind. How does a person (or code) identify how to map a location in this array back to the correct service? Do we need to append some additional info to the error's `extensions` in order to provide that mapping?

federation-js/src/composition/validate/postComposition/__tests__/executableDirectivesIdentical.test.ts
```diff
@@ -81,6 +83,12 @@ describe('executableDirectivesIdentical', () => {
       Array [
         Object {
           "code": "EXECUTABLE_DIRECTIVES_IDENTICAL",
+          "locations": Array [
+            Object {
+              "column": 1,
+              "line": 2,
+            },
+          ],
```
trevor-scheer, line 91: Should there be a `locations` entry for each service which defines this directive?

federation-js/src/composition/validate/postComposition/__tests__/executableDirectivesInAllServices.test.ts
```diff
@@ -83,6 +85,12 @@ describe('executableDirectivesInAllServices', () => {
       Array [
         Object {
           "code": "EXECUTABLE_DIRECTIVES_IN_ALL_SERVICES",
+          "locations": Array [
```
trevor-scheer, line 88: Unsure if `locations` makes sense here at all. In this case, `serviceB/C` are the ones missing the implementation, but the error points back to the implementation in `serviceA`. It's probably the _most useful_ location if there is one, just not sure that it is given the nature of the error.

federation-js/src/composition/validate/postComposition/__tests__/externalTypeMismatch.test.ts
```diff
@@ -36,10 +38,22 @@ describe('validateExternalDirectivesOnSchema', () => {
       Array [
         Object {
           "code": "EXTERNAL_TYPE_MISMATCH",
+          "locations": Array [
```
trevor-scheer, line 41: Can this location point to the return type of the field instead?

federation-js/src/composition/validate/postComposition/__tests__/externalUnused.test.ts
```diff
@@ -36,6 +38,12 @@ describe('externalUnused', () => {
       Array [
         Object {
           "code": "EXTERNAL_UNUSED",
+          "locations": Array [
```
trevor-scheer, line 41: Can this location point to the `@external` directive instead?

federation-js/src/composition/validate/postComposition/__tests__/providesNotOnEntity.test.ts
```diff
@@ -127,6 +129,12 @@ describe('providesNotOnEntity', () => {
       Array [
         Object {
           "code": "PROVIDES_NOT_ON_ENTITY",
+          "locations": Array [
```
trevor-scheer, line 132: Should this location point to the return type of the field? (`LineItem`)
Alternatively, maybe on the `@provides` directive itself. Thoughts?

federation-js/src/composition/validate/postComposition/__tests__/requiresFieldsMissingExternals.test.ts
```diff
@@ -68,6 +70,12 @@ describe('requiresFieldsMissingExternal', () => {
       Array [
         Object {
           "code": "REQUIRES_FIELDS_MISSING_EXTERNAL",
+          "locations": Array [
```
trevor-scheer, line 73: This test might be a bit poorly constructed, but I see 2 possible cases here.

* In this case, `id` doesn't even exist in the extension. Not much we can do here. Maybe the location is the `fields` arg on the requires directive? I'm unsure.
* In what I would expect to be a more likely case, `id` would be defined but not marked as `@external`, in which case we could provide a location to that field instead since it's more relevant to the error.

federation-js/src/composition/validate/postComposition/__tests__/requiresFieldsMissingOnBase.test.ts
```diff
@@ -71,6 +73,12 @@ describe('requiresFieldsMissingOnBase', () => {
       Array [
         Object {
           "code": "REQUIRES_FIELDS_MISSING_ON_BASE",
+          "locations": Array [
```
trevor-scheer, line 76: Should this location point to the `Product.id` field on `serviceC` in this case?

federation-js/src/composition/validate/preComposition/__tests__/duplicateEnumOrScalar.test.ts
```diff
@@ -60,13 +62,19 @@ describe('duplicateEnumOrScalar', () => {
...
-              Object {
-                "code": "DUPLICATE_ENUM_DEFINITION",
-                "message": "[serviceA] ProductType -> The enum, \`ProductType\` was defined multiple times in this service. Remove one of the definitions for \`ProductType\`",
-              },
-            ]
-        `);
+      Array [
+        Object {
+          "code": "DUPLICATE_ENUM_DEFINITION",
+          "locations": Array [
```
trevor-scheer, line 68: Should we provide the locations of all enums that have a name collision in a single service?

This is such an odd composition error IMO, seems like this would be handled in `graphql-js`. I wonder what caused us to build this. (ignore me though, just noting / thinking out loud)

federation-js/src/composition/validate/preComposition/__tests__/duplicateEnumOrScalar.test.ts
```diff
@@ -89,6 +97,12 @@ describe('duplicateEnumOrScalar', () => {
       Array [
         Object {
           "code": "DUPLICATE_SCALAR_DEFINITION",
+          "locations": Array [
```
trevor-scheer, line 100: Repeat from above: can we capture all locations of the duped scalar?

federation-js/src/composition/validate/preComposition/__tests__/duplicateEnumValue.test.ts
```diff
@@ -66,6 +68,12 @@ describe('duplicateEnumValue', () => {
       Array [
         Object {
           "code": "DUPLICATE_ENUM_VALUE",
+          "locations": Array [
```
trevor-scheer, line 71: Repeat: can we capture all locations of the duped enum values?

federation-js/src/composition/validate/preComposition/__tests__/externalUsedOnBase.test.ts
```diff
@@ -43,6 +45,12 @@ describe('externalUsedOnBase', () => {
       Array [
         Object {
           "code": "EXTERNAL_USED_ON_BASE",
+          "locations": Array [
```
trevor-scheer, line 48: Can this location point to the `@external` directive on this field specifically?

federation-js/src/composition/validate/preComposition/__tests__/keyFieldsMissingExternal.test.ts
```diff
@@ -133,6 +141,12 @@ describe('keyFieldsMissingExternal', () => {
       Array [
         Object {
           "code": "KEY_FIELDS_MISSING_EXTERNAL",
+          "locations": Array [
+            Object {
+              "column": 35,
+              "line": 1,
+            },
```
trevor-scheer, line 148: This seems odd

federation-js/src/composition/validate/preComposition/__tests__/requiresUsedOnBase.test.ts
```diff
@@ -43,6 +45,12 @@ describe('requiresUsedOnBase', () => {
       Array [
         Object {
           "code": "REQUIRES_USED_ON_BASE",
+          "locations": Array [
```
trevor-scheer, line 48: Should this location point to the actual `@requires` directive?

federation-js/src/composition/validate/preNormalization/__tests__/rootFieldUsed.test.ts
```diff
@@ -95,6 +97,12 @@ describe('rootFieldUsed', () => {
       Array [
         Object {
           "code": "ROOT_QUERY_USED",
+          "locations": Array [
```
trevor-scheer, line 100: Should this location point to the name `Query` itself?

federation-js/src/composition/validate/sdl/__tests__/matchingUnions.test.ts
```diff
@@ -87,6 +87,16 @@ describe('MatchingUnions', () => {
     expect(errors[0]).toMatchInlineSnapshot(`
       Object {
         "code": "VALUE_TYPE_UNION_TYPES_MISMATCH",
+        "locations": Array [
```
trevor-scheer, line 90: I think this is fine as is, but tossing around the idea of having the location point to specific types on the diff of these unions.

In this case, that would mean the location points to the `Error` type, but I'm not sure we can identify whether the `Error` is extraneous in one service vs. missing in another.

I'm leaning towards keeping it as is, but flagging the thought in case anything interesting comes to mind.

federation-js/src/composition/validate/sdl/__tests__/uniqueTypeNamesWithFields.test.ts
```diff
@@ -109,10 +109,30 @@ describe('UniqueTypeNamesWithFields', () => {
         Array [
           Object {
             "code": "VALUE_TYPE_FIELD_TYPE_MISMATCH",
+            "locations": Array [
```
trevor-scheer, line 112: Should these locations point to the return types on the problematic fields?

federation-js/src/composition/validate/sdl/__tests__/uniqueTypeNamesWithFields.test.ts
```diff
@@ -147,6 +167,16 @@ describe('UniqueTypeNamesWithFields', () => {
         Array [
           Object {
             "code": "VALUE_TYPE_INPUT_VALUE_MISMATCH",
+            "locations": Array [
```
trevor-scheer, line 170: Should these locations point to the arg value's type in each service?

federation-js/src/composition/validate/sdl/__tests__/uniqueTypeNamesWithFields.test.ts
```diff
@@ -509,6 +579,16 @@ describe('UniqueTypeNamesWithFields', () => {
       expect(errors[0]).toMatchInlineSnapshot(`
         Object {
           "code": "VALUE_TYPE_NO_ENTITY",
+          "locations": Array [
```
trevor-scheer, line 582: Here's another case where it's hard to get specific with the location. Do they need a key somewhere or do they need to delete a key somewhere else? We can't know. I think these locations are fine given that.

federation-js/src/composition/__tests__/compose.test.ts
mayakoneval: line 2, column 1 is the t in type & line 2, column 6 is the P in Product, as far as I can tell.

federation-js/src/composition/__tests__/composeAndValidate.test.ts
mayakoneval: Yes definitely - I think since Martijn brought this up in an earlier comment it seems like this will be covered in the future [when we add support for adding service names to composition errors.](https://github.com/apollographql/federation/issues/705) We could have structured data with locations associated to service name.

mayakoneval: To answer the first question about the union types. Since right now we don't have a way of specifying which locations belong to which services, we won't know which of the possibly numerous locations that would re reported back here correspond to the service name. Could I add a [comment here to come back and fix this as part of this issue in the future?](https://github.com/apollographql/federation/issues/705)

federation-js/src/composition/validate/postComposition/__tests__/executableDirectivesInAllServices.test.ts
mayakoneval: that makes sense, would it be okay for this error in this case to just not have location? I will delete the node since I do think so, but making sure.

mayakoneval: got it. I do think it would be helpful for someone to see some UI on that part of the code saying "hey you need to implement this everywhere". Could this be another place that we comment in [preparation for adding service name as structured data](https://github.com/apollographql/federation/issues/705)? In the future, we could specify "hey this is pointing to this line in service A".

I don't know your teams method of tracking TODOs - I will also definitely add these to the open issue.

federation-js/src/composition/validate/postComposition/__tests__/externalTypeMismatch.test.ts
mayakoneval: Since the return type doesn't have an ast node, can we do that? externalField.type would point to the node of the definition of the type.

federation-js/src/composition/validate/postComposition/__tests__/keysMatchBaseService.test.ts
```diff
@@ -74,11 +76,63 @@ describe('keysMatchBaseService', () => {
...
+    assertCompositionSuccess(compositionResult);
+    const { schema } = compositionResult;
+
+    const validationErrors = validateKeysMatchBaseService({
+      schema,
+      serviceList,
+    });
+    expect(validationErrors).toHaveLength(1);
+    expect(validationErrors[0]).toMatchInlineSnapshot(`
+      Object {
```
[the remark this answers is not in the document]
mayakoneval: I was thinking throughout this whole process that when possible, we only return one location that corresponds to the service name in brackets in the error message. I think a lot of this shows that really we need structured data that matches locations to service name. I'll flag this as something to come back to in [the next effort](https://github.com/apollographql/federation/issues/705)

federation-js/src/composition/validate/postComposition/__tests__/providesNotOnEntity.test.ts
mayakoneval: Yeah I think the provides makes the most sense here. 👍

federation-js/src/composition/validate/postComposition/__tests__/requiresFieldsMissingExternals.test.ts
mayakoneval: Since this is a post composition validator, there is no distinction between the fields on the Product of one service and the Product of another service, the `namedType.getFields()` returns the composed `Product`'s field (all of sku, upc, id and price). B/c of this, I am confused how we could find the node for the field on serviceB in this case at all. For now I made this the argument node.

federation-js/src/composition/__tests__/compose.test.ts
trevor-scheer: Right, I would expect both locations to point to the same thing, i.e. both pointing to the "t", or both pointing to the "P"

federation-js/src/composition/__tests__/composeAndValidate.test.ts
trevor-scheer: Yes that sounds good to me! If you're collecting these in a single issue, tagging these TODOs for searchability would be nice, i.e. `TODO(<issue #>)`

federation-js/src/composition/__tests__/compose.test.ts
mayakoneval: oh duh 🤦 thank you for the clarification. That is weird, I will look into where these errors came from. Why is there two in the first place?

federation-js/src/composition/validate/postComposition/__tests__/executableDirectivesInAllServices.test.ts
trevor-scheer: Yep that sounds good to me!

federation-js/src/composition/__tests__/compose.test.ts
trevor-scheer: There are two places where type `Product` is defined, so it's pointing to both of them

federation-js/src/composition/validate/postComposition/__tests__/keysMatchBaseService.test.ts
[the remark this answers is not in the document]
trevor-scheer: I do think multiple locations is appropriate here but the current location is still helpful and it's find to punt on this one.

federation-js/src/composition/validate/postComposition/__tests__/requiresFieldsMissingOnBase.test.ts
mayakoneval: same answer as above, for now until we have structured data I'll make this the argument node

mayakoneval:
> > Thanks @martijnwalraven for catching this issue. The issue was that for errors on selection sets, the locations were off. @trevor-scheer pointed out that we [manually parse selections](https://github.com/apollographql/federation/blob/d20d6f941e831924b3a59a0048d44e7bafac1dd6/federation-js/src/composition/utils.ts#L159-L164), so the location information was off when using this node as our baseline in `GraphQLError`. I just committed changes that pass the _field_'s node instead of the _selection set_'s field node, this makes the locations match the actual field, or type if there is no field.
> 
> That makes sense! If we do want to refer to contents of the parsed field set to give people more specific context, we should consider setting the [`locationOffset` property on `Source`](https://github.com/graphql/graphql-js/blob/917befd7a54fe99d6141fa0ed92bc2b82544625f/src/language/source.d.ts#L16) when parsing selections.

@martijnwalraven it doesn't look like parse cares about location offset at all [when determining locations](https://github.com/graphql/graphql-js/blob/a4e9bc9bb8e05916ac926233394a164b5e129062/src/language/parser.js#L151): I spent a little time trying this out

federation-js/src/composition/validate/preComposition/__tests__/duplicateEnumOrScalar.test.ts
mayakoneval: I am going to punt on this as part of the effort for #705, since we don't know which service to associate those locations with right now

federation-js/src/composition/validate/preComposition/__tests__/duplicateEnumOrScalar.test.ts
mayakoneval: ^ same as above

federation-js/src/composition/validate/preComposition/__tests__/duplicateEnumValue.test.ts
mayakoneval: ^ same as above

federation-js/src/composition/validate/preComposition/__tests__/keyFieldsMissingExternal.test.ts
mayakoneval: I can't figure out why this doesn't give a reasonable location. The node seems right. I have switched it to the parent type.

federation-js/src/composition/validate/preNormalization/__tests__/rootFieldUsed.test.ts
mayakoneval: The columns for the types start at `type` for all of these I believe.

mayakoneval: (except that weird graphql error that I can't figure out)

federation-js/src/composition/validate/sdl/__tests__/uniqueTypeNamesWithFields.test.ts
mayakoneval: This one is not easy, since we don't return the input value nodes, just the names from `diffTypeNodes`

mayakoneval:
@trevor-scheer For the post-composition errors I added [this util function](https://github.com/apollographql/federation/pull/686/files#diff-3d1f1371e9896cd2764308a9ac63031b3d1222f760202d1e88c340bb9815534bR560-R568) which gets info for a given type from the serviceList (ServiceDefinition[]) so we know that is correct with respect to a single subgraph. I used this in a few places.

federation-js/src/composition/validate/postComposition/__tests__/providesNotOnEntity.test.ts
```diff
@@ -131,7 +131,7 @@ describe('providesNotOnEntity', () => {
           "code": "PROVIDES_NOT_ON_ENTITY",
           "locations": Array [
             Object {
-              "column": 3,
+              "column": 32,
```
trevor-scheer, line 134: In this case, the existence of `@provides` on this field is incorrect, so I think we should point to the directive and not its `fields` argument.

federation-js/src/composition/validate/postComposition/keysMatchBaseService.ts
```diff
@@ -50,13 +52,17 @@ export const keysMatchBaseService: PostCompositionValidator = function ({
...
               errors.push(
                 errorWithCode(
                   'MULTIPLE_KEYS_ON_EXTENSION',
                   logServiceAndType(extendingService, parentTypeName) +
                     `is extended from service ${serviceName} but specifies multiple @key directives. Extensions may only specify one @key.`,
-                  parentType.astNode ?? undefined,
+                  extendingServiceTypeNode && 'directives' in extendingServiceTypeNode ?
+                    extendingServiceTypeNode.directives?.find(directive =>
+                      directive.name.value === 'key')?.arguments?.find
+                        (argument => argument.name.value === 'fields') : undefined,
```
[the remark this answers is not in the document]
mayakoneval: For this one, since we can't select the general "directive area" I am going to select the type node, so the error can be located to the whole error where there are too many keys.

mayakoneval:
Okay I spent time today doing a full search through all the codes to make sure the locations make sense, which meant a few more changes than were requested. I'm hoping this looks pretty good 😋 . Excited about this!

federation-js/src/composition/validate/postComposition/__tests__/externalUnused.test.ts
```diff
@@ -391,4 +391,52 @@ describe('externalUnused', () => {
       ]
     `);
   });
+
+  it('point to the right location on error when multiple directives are on the field in question', () => {
```
trevor-scheer, line 395: 🎉

federation-js/src/composition/validate/postComposition/externalMissingOnBase.ts
```diff
@@ -35,7 +35,7 @@ export const externalMissingOnBase: PostCompositionValidator = ({ schema }) => {
                 'EXTERNAL_MISSING_ON_BASE',
                 logServiceAndType(serviceName, typeName, externalFieldName) +
                   `marked @external but ${externalFieldName} is not defined on the base service of ${typeName} (${typeFederationMetadata.serviceName})`,
-                externalField,
+                findDirectivesOnNode(externalField, 'external'),
```
trevor-scheer, line 38: 😎

federation-js/src/composition/utils.ts
```diff
@@ -62,20 +62,39 @@ export function mapFieldNamesToServiceName<Node extends { name: NameNode }>(
 
 export function findDirectivesOnNode(
   node: Maybe<
-    | TypeDefinitionNode
-    | TypeExtensionNode
     | FieldDefinitionNode
-    | SchemaDefinitionNode
+    | FieldNode
+    | DefinitionNode
```
trevor-scheer, line 68: I like the direction you're taking this. I will suggest, I think we should either go full bore `Maybe<ASTNode>` as the type or only permit types which at least have a `directives` property on them. I like the latter, since the typings will warn users of the fn when they're using it inappropriately. In your case, the only inappropriate usage which is now included in this union is the `DirectiveDefinitonNode`, so I'll suggest this (+ a nice little TS revamp since we're here)

```ts
export function findDirectivesOnNode(
  node: Maybe<
    Exclude<
      | FieldDefinitionNode
      | InputValueDefinitionNode
      | FieldNode
      | DefinitionNode,
      DirectiveDefinitionNode
    >
  >,
  directiveName: string,
) {
  return (
    node?.directives?.filter(
      (directive) => directive.name.value === directiveName,
    ) ?? []
  );
}
```

federation-js/src/composition/utils.ts
```diff
@@ -62,20 +62,39 @@ export function mapFieldNamesToServiceName<Node extends { name: NameNode }>(
...
     ? node.directives.filter(
         directive => directive.name.value === directiveName,
       )
     : [];
 }
 
+export function findSelectionSetOnNode(
+  node: Maybe<
+    | FieldDefinitionNode
+    | DefinitionNode
```
trevor-scheer, line 83: The same `Exclude<>` type I suggested above would give you a similar ergonomic win here.

federation-js/src/composition/utils.ts
```diff
@@ -62,20 +62,39 @@ export function mapFieldNamesToServiceName<Node extends { name: NameNode }>(
...
 ) {
-  return node && node.directives
+  return node && 'directives' in node && node.directives
     ? node.directives.filter(
         directive => directive.name.value === directiveName,
       )
     : [];
 }
 
+export function findSelectionSetOnNode(
```
trevor-scheer, line 79: This fn is a bit tough to grok. Can you please add a good description to what this fn looks for and how?

federation-js/src/composition/utils.ts
```diff
@@ -62,20 +62,39 @@ export function mapFieldNamesToServiceName<Node extends { name: NameNode }>(
...
+  // assume there is only one directive with this name that selects this element
+  return node && 'directives' in node && node.directives
+    ? node.directives.find(
+        directive =>
+          directive.name.value === directiveName && directive.arguments?.some(
+            argument => argument.value.kind === "StringValue" &&
+              argument.value.value.includes(elementInSelectionSet)
+          ))?.arguments?.find(
+            argument => argument.name.value === 'fields')?.value
+    : undefined;
```
trevor-scheer, line 97: 1) we should use `Kind.STRING` for `"StringValue"` (or use the `isStringValueNode` predicate fn!)
2) I'm not totally sure I understand this correctly but I think I do. Is the `includes` here possibly error-prone?

I'm imagining 2 keys:
`@key(fields: "{ id }")`
`@key(fields: "{ sku { id } }")`

Could this scenario break this implementation? I might be mistaken and just need to understand this a bit better.

mayakoneval: In the scenario you described:

`type SomeType @key(fields: "id") @key(fields: "sku { id }") { ` and we are looking for the field with `id` on it, yeah this would break. If this were the case, and we wanted to find a top level field, would it be sufficient to make sure that string is not wrapped in curly braces? I'll write a test for this

mayakoneval: The resolution here was to search for the entire selection set, so we know we are selecting the correct key.

federation-js/src/composition/utils.ts
```diff
@@ -64,37 +72,56 @@ export function findDirectivesOnNode(
...
-  // assume there is only one directive with this name that selects this element
-  return node && 'directives' in node && node.directives
-    ? node.directives.find(
+  return node?.directives?.find(
         directive =>
           directive.name.value === directiveName && directive.arguments?.some(
-            argument => argument.value.kind === "StringValue" &&
-              argument.value.value.includes(elementInSelectionSet)
+            argument => isStringValueNode(argument.value) &&
+              argument.value.value.includes(printedSelectionSet)
```
trevor-scheer, line 122: Can this be `===` now instead of `includes`?

trevor-scheer (approved):
Ship it!

Can we also replace the `printFieldSet` fn in `printSupergraphSdl` with the one we've moved into the `utils` file?

mayakoneval:
@trevor-scheer just pushed those last changes, I think you have to merge :) Woohoo!

mayakoneval:
`cargo test` on the rust windows test has failed twice, I force pushed again to try.

trevor-scheer:
@mayakoneval I merged main which must've introduced some good TS changes since I then had to remove a couple unused imports. Now everything is passing, so I'm going to land this now 🎉

PR #696, opened 2021-04-25
fix(federation): get rid of util pkg

### LINKED ISSUE #695: Get rid of "util" package
https://github.com/apollographql/federation/blob/main/federation-js/src/composition/validate/sdl/matchingEnums.ts#L10 can be replaced by a simple `typeof value === 'string'`. This avoids errors in bundlers.

StarpTech:
Fix #695


abernix (commented):
Thanks!

abernix (approved):

PR #698, opened 2021-04-26
[DO NOT SHIP] Naive query plan cache HashMap.

abernix:
This is merely for performance testing reasons, not because this is a
complete implementation of what a cache should look like.  It has no
eviction policy.


abernix:
I didn't intend to open this as a PR, but apparently I found the magical keyboard combination to open a PR from the compare view. 🤷 

While you're here: It's worth noting that the TypeScript query planner is already doing caching with a more complete implementation than this.  I was just curious how this would look

PR #704, opened 2021-04-27
feat(federation): Include `@join__type` directives on interface types

trevor-scheer:
Capturing additional metadata on interface types (specifically which graphs they exist within) provides the query planner with enough information to unlock an optimization.

Discussed with @martijnwalraven about leaving `key`s as empty strings for now with an issue opened to introduce this at a later time.
- [ ] open issue

By commit review will be easier on the eyes due to a snapshot prettification of the `compose.test.ts` file.


federation-js/src/composition/validate/postComposition/providesNotOnEntity.ts
```diff
@@ -58,7 +58,10 @@ export const providesNotOnEntity: PostCompositionValidator = ({ schema }) => {
         }
 
         const fieldType = types[baseType.name];
-        const selectedFieldIsEntity = getFederationMetadata(fieldType)?.keys;
+        // FIXME: validate for correctness
```
trevor-scheer, line 61: Still not sure about this one

abernix:
This was a great implementation shot at solving a problem we were having, but I'm going to close this because I don't think it worked out in the end, but we can re-open if not.  Thanks!

PR #708, opened 2021-04-29
chore(infra): Upgrade monorepo to use npm@7 and enable `engine-strict`

abernix:
This upgrades us to use `npm@7` at both the root and in `docs/` and enables `engine-strict` via a root-level `.npmrc` file in this repository.

The intent of introducing `engine-strict` is to make sure that the users of this monorepo are not led astray into using an unsupported version of Node.js or npm that might cause unexpected side-effects. (Running the wrong version of `npm` is the most motivating reason today!)

Users who wish to run this monorepo with a version of Node.js and npm that are not specified in the `engines` range may still do so by running `npm install --no-engine-strict` to acknowledge that there might be unexpected/untested results/outcomes from their actions.


abernix:
For those stuck between two lands of npm: If using `nvm`, and needing to toggle between versions of npm, it's possible to have two different versions of Node.js with different versions.  For me, I leverage this to toggle between Node.js 12 and Node.j 14, which largely behave the same for all intents and purposes, but have npm 7 installed in my 14 environment while I have 6 installed in 12.  Just a suggestion!

package.json
```diff
@@ -26,14 +26,14 @@
   },
   "engines": {
     "node": ">=6 <15",
```
martijnwalraven, line 28: I hadn't noticed this before, but >= node 6 seems like a promise we can't really keep since we only test on node 12 and 14 (and soon 16). So I wonder if node >= 12 wouldn't be a better reflection of reality.

abernix: Changing the Node.js version would indeed be a better reflection of reality, but I'll update that in the PR that updates our Node.js versions and introduces testing and support for Node.js 16.

It's also worth nothing that we're already not constrained by this promise of Node.js v6 since individual packages in this repository (which are actually installed, and are the actual manifests/contracts) state much higher minimum versions.

abernix: Done in #713

glasser (approved):
Seems reasonable enough to me, especially if we land #713 soon so the easy way of "make sure to run with npm v7 locally" is "just use Node 16".

Can you add a motivation for enabling `engines-strict` somewhere (even just a comment on the PR)? I assume something like "so we can feel the pain of users who have it enabled or use other tools with that default".

martijnwalraven (approved):
Looks good to me! (Assuming we land this together with #713, so `engines-strict` doesn't prevent running on Node >= 15.)

PR #711, opened 2021-04-29
feat(federation): update introspection endpoint

EverlastingBugstopper:
use `rover subgraph introspect` with the new rover.apollo.dev
introspection endpoint in place of the raw lambda URL

this should only be merged after [this change](https://github.com/apollographql/orbiter/pull/21) has been merged and verified

### DIFF
--- a/docs/source/quickstart.mdx
+++ b/docs/source/quickstart.mdx
@@ -137,7 +137,7 @@ We can fetch a running subgraph's schema by executing an enhanced introspection 
 From your project directory, run the following command in your terminal:
 
 ```shell
-rover subgraph introspect https://7bssbnldib.execute-api.us-east-1.amazonaws.com/Prod/graphql
+rover subgraph introspect https://rover.apollo.dev/quickstart/products/graphql
 ```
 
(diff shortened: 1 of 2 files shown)


docs/source/quickstart.mdx
```diff
@@ -283,8 +283,8 @@ type Department {
 scalar join__FieldSet
 
 enum join__Graph {
-  PRODUCTS @join__graph(name: "products" url: "https://7bssbnldib.execute-api.us-east-1.amazonaws.com/Prod/graphql")
-  REVIEWS @join__graph(name: "reviews" url: "https://w0jtezo2pa.execute-api.us-east-1.amazonaws.com/Prod/graphql")
```
EverlastingBugstopper, line 287: ooh - umm, i wonder if these URLs will actually be updated in the schema.. likely not? lemme check

EverlastingBugstopper: ohh yeah this is right! i forgot that these come from `routing_url`. we're good 😄

StephenBarlow (approved):

PR #735, opened 2021-05-05
Document didEncounterError

StephenBarlow:


docs/source/api/apollo-gateway.mdx
```diff
@@ -416,3 +416,86 @@ This enables you to modify any combination of the operation's `context` and the
 The `http` property of the `request` and `response` objects contains additional HTTP-specific properties, such as `headers`.
 
 This method must return an object that matches the structure of a [`GraphQLResponse`](https://github.com/apollographql/apollo-server/blob/2562d096/packages/apollo-server-types/src/index.ts#L43-L48).  If no modifications are necessary, return the original `response`.
+
+#### `didEncounterError`
+
+Override this method in a subclass to customize the gateway's behavior after it encounters an error while communicating with the implementing service.
```
abernix, line 422:
```suggestion
Override this method in a subclass to customize the gateway's behavior after it encounters an error while communicating with the implementing service or while parsing its response (e.g., if the response was not well-formed JSON).
```

docs/source/api/apollo-gateway.mdx
```diff
@@ -416,3 +416,86 @@ This enables you to modify any combination of the operation's `context` and the
...
+
+If you _don't_ override this method, the default implementation `throw`s the error, like so:
+
+```javascript
+class MyDataSource extends RemoteGraphQLDataSource {
+
+  didEncounterError({ error, fetchRequest, fetchResponse, context }) {
```
abernix, line 429: Unlike the other methods in this `RemoteGraphQLDataSource`, this has positional parameters, rather than receiving an object (`GraphQLRequestContext`).

It's actually particularly unique over the other signatures since this one receives `fetchRequest` and `fetchResponse`, rather than others (such as `didReceiveResponse` and `willSendRequest`) which have `request` (`GraphQLRequest`) and `response` (`GraphQLResponse`) on a `GraphQLRequestContext` object).

```suggestion
  didEncounterError(error, fetchRequest, fetchResponse, context) {
```

docs/source/api/apollo-gateway.mdx
```diff
@@ -416,3 +416,86 @@ This enables you to modify any combination of the operation's `context` and the
...
+<tr>
+<td>
+
+###### `context`
+
+`TContext`
+</td>
+<td>
+
+The context of the communication's associated GraphQL request.
```
abernix, line 497: Not suggesting that this needs to be annotated differently, but this is the "user-context" (if we have a notion of that anywhere).  Essentially, it's the thing that a user might have defined using the `context` function at the top-level of the `ApolloServer` constructor (where implementors might attach authentication details, etc., and would be available to resolvers in non-Gateway execution).

abernix (changes requested):
One necessary change, but otherwise this LGTM!

docs/source/api/apollo-gateway.mdx
StephenBarlow: ahh right big catch

PR #751, opened 2021-05-12
Fixing the documentation to remove incorrect federation features

AndrewRayCode:
Addresses https://github.com/apollographql/federation/issues/750 and https://github.com/apollographql/federation/issues/420

It looks like field migration has _never_ been supported by non-managed federation. The docs are misleading and claim that a non-federated graph can handle composition errors. This is dangerous because a composition error in a non-federated graph means the entire graph will fail to come up. If you follow the steps in the documentation, you'll take down your entire API.

### DIFF
--- a/docs/source/entities.md
+++ b/docs/source/entities.md
@@ -209,7 +209,7 @@ type Query {
 
 ## Migrating entities and fields (advanced)
 
-As your federated graph grows, you might decide that you want an entity (or a particular field of an entity) to originate in a different subgraph. Apollo Gateway helps you perform these migrations safely.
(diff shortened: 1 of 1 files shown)


AndrewRayCode:
Hi @StephenBarlow it appears that the incorrect doc is still live. I don't recall if non-federation works for paid Apollo. My concern is that you might be advertising a feature of a paid product that fundamentally can't work. Can you update the documentation to reflect that you can't migrate schemas without downtime in a non-federated graph?

abernix:
I'm going to close this as superseded by #1252.  Thank you very much for opening this originally, and I hope the (substantial) edits that were made in #1252 help to improve this.  Let us know if you don't think so!

PR #756, opened 2021-05-13
feat(federation): Initial support for `@tag` and `@inaccessible` in composition

trevor-scheer:
Introduce the existence and composition of the `@tag` and `@inaccessible` directives,
to be printed in `supergraphSdl` and used for schema filtering.

TODO:
- [x] Figure out value types / provides regression (thanks @martijnwalraven!)
- [x] Address a few TODOs in the code
- [ ] Open issues about validations and more test cases

### DIFF
--- a/federation-js/src/composition/compose.ts
+++ b/federation-js/src/composition/compose.ts
@@ -207,8 +207,6 @@ export function buildMapsFromServiceList(serviceList: ServiceDefinition[]) {
             );
             const directives = mapGetOrSet(fieldToDirectivesMap, fieldName, []);
             directives.push(inaccessibleDirectives[0]);
-          } else if (inaccessibleDirectives.length > 1) {
-            // TODO: warn about multiple applications of inaccessible?
           }
         }
       }
--- a/federation-js/src/composition/compose.ts
+++ b/federation-js/src/composition/compose.ts
@@ -17,6 +17,7 @@ import {
   TypeExtensionNode,
   ObjectTypeDefinitionNode,
   NamedTypeNode,
+  DirectiveNode,
 } from 'graphql';
 import { transformSchema } from 'apollo-graphql';
 import federationDirectives from '../directives';
@@ -45,7 +46,7 @@ import {
 import { validateSDL } from 'graphql/validation/validate';
 import { compositionRules } from './rules';
 import { printSupergraphSdl } from '../service/printSupergraphSdl';
-import { mapValues } from '../utilities';
+import { mapGetOrSet, mapValues } from '../utilities';
 
 const EmptyQueryDefinition = {
   kind: Kind.OBJECT_TYPE_DEFINITION,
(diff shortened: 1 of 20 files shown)


federation-js/src/composition/compose.ts
```diff
@@ -176,6 +183,34 @@ export function buildMapsFromServiceList(serviceList: ServiceDefinition[]) {
...
+          if (inaccessibleDirectives.length === 1) {
+            const fieldToDirectivesMap = mapGetOrSet(
+              typeNameToFieldDirectivesMap,
+              typeName,
+              new Map(),
+            );
+            const directives = mapGetOrSet(fieldToDirectivesMap, fieldName, []);
+            directives.push(inaccessibleDirectives[0]);
+          } else if (inaccessibleDirectives.length > 1) {
+            // TODO: warn about multiple applications of inaccessible?
```
martijnwalraven, line 211: I don't think we need to do that here, that should have been caught by validation of the subgraph schema.

federation-js/src/composition/compose.ts
```diff
@@ -176,6 +183,34 @@ export function buildMapsFromServiceList(serviceList: ServiceDefinition[]) {
             );
           }
         }
+
+        for (const field of definition.fields ?? []) {
+          const fieldName = field.name.value;
+          const tagDirectives = findDirectivesOnNode(field, 'tag');
+
+          // TODO: we can dry this up into one step for both cases
```
martijnwalraven, line 191: Yeah, if we drop the concern for `@inaccessible` being repeatable it seems the code would be exactly the same.

federation-js/src/service/printSupergraphSdl.ts
```diff
@@ -375,54 +384,83 @@ function printJoinFieldDirectives(
...
+function printAppliedDirectives(field: GraphQLField<any, any>) {
+  const appliedDirectives = (
+    field.extensions?.federation?.appliedDirectives ?? []
+  ) as DirectiveNode[];
+  return appliedDirectives.length > 0
+    ? ' ' + appliedDirectives.map(printDirectiveNode).join(' ')
+    : '';
+};
+
+function printDirectiveNode(directive: DirectiveNode) {
```
martijnwalraven, line 444: I think this could just be `print` from `graphql-js`? That would also make sure argument types are handled consistently.

martijnwalraven (changes requested):

federation-js/src/composition/compose.ts
martijnwalraven: And if it slips through for some reason, it would be caught by validation of the supergraph :)

federation-js/src/service/printSupergraphSdl.ts
trevor-scheer: Why didn't I think of that? 😄
86e4ed2b

federation-js/src/composition/compose.ts
trevor-scheer: Yeah, agree. Didn't think anything would be needed here but left a placeholder to come back to and be sure. Thanks for confirming.
7c32cfbe

federation-js/src/composition/compose.ts
trevor-scheer: 26c787c2

martijnwalraven (approved):
I think this is ready to be merged. Before the next release, I do think we may want to avoid outputting directive definitions for `@inaccessible` and `@tag` to avoid changes to every composed schema even if these aren't used. And we still need to define core features for `@inaccessible` and `@tag`. But I can do that as part of the core schema work that I've started on top of this. That doesn't have to block this PR I think.

PR #758, opened 2021-05-13
Apply `repeatable` to @key directive in federation-spec

sMorm:
Repeated usages of the @key directive without them being marked as repeatable will result in an invalid schema.


trevor-scheer (approved):
Thanks @sMorm! 🎉

PR #765, opened 2021-05-18
Generalised composition doc

pcmanus:
Open a PR for the WIP generalised composition doc so as to make it easier for potential comments. This is still very much a work in progress and will stay open for now.


pcmanus:
Noticed this PR hanging around and going to close it because on top of being very incomplete, it's now very very outdated. I kind of dropped the ball on this when I started focusing on the implementation, but now that the implementation is complete, released and documented, such a doc is maybe less pressing.

To be fair, I don't think we really have a replacement doc that serves the purpose of what this was meant to do, so I might try to resurrect some of this at some point, but it'll be much simpler to create a new PR when that time arrives anyway.

PR #769, opened 2021-05-20
doc: Update community links

EduardoGHdez:
Just clicked on one of the links to the community forum and noticed it's
hosted in another place now.

Actually, this is the message I saw there:

```
We ran into trouble loading this page  😣

You may be trying to view something that is deleted, or Spectrum is
just having a hiccup. If you think something has gone wrong, please
contact us.
```

This is just for updating the links to the community at federation repo, but, will submit an Issue to follow up on the update for the links in the documentation, which seems to be managed from /gatsby-theme-apollo


trevorblades (approved):
Thanks @EduardoGHdez! 🎉

PR #777, opened 2021-05-28
feat(gateway): Submit out of band error reporters when supergraph requests fail

Y-Guo:
# Motivation 
This PR adds functionality to send a post request to an Apollo out of band error reporting endpoint. OOB error reports will be sent when unexpected errors are encountered during schema requests. 
The endpoint must be configured by setting the env variable `APOLLO_OUT_OF_BAND_REPORTER_ENDPOINT`, It's recommended this be set with `APOLLO_OUT_OF_BAND_REPORTER_ENDPOINT=https://uplink.api.apollographql.com/monitoring` in most cases.


gateway-js/src/loadSupergraphSdlFromStorage.ts
```diff
@@ -67,10 +120,68 @@ export async function loadSupergraphSdlFromStorage({
         'content-type': 'application/json',
       },
     });
+
+    const oobResponse = await result.json();
+    if (!oobResponse?.data?.reportError) {
+      throw new Error("An error occured while reporting your error to Apollo: "
+      + fetchErrorMsg + result.status + ' ' + result.statusText);
+    }
```
abernix, line 128: Four things:

* `occured` missing another `r` 👉 `occurred`
* Maybe rather than "your error" we should be more precise?  In particular, it's not really _my_ error.  It's more like, "Seems like Apollo had an error and _additionally_, this program encountered an error while reporting Apollo's error to them.", right?
* This is not something that was introduced by this PR, but I find it odd that `fetchErrorMsg` is not literally _an error message that came from trying to `fetch` something_, but is instead _a generic message **prefix** that we staple on the front of other error messages_.  Further, I find it odd that we embed the trailing colon in `fetchErrorMsg` because it requires me to always go look and remember if it has the colon or not and it's used in concatenation a lot.  Like, what if I wanted to use the error without 

  i.e., it would be more intuitive to me as the developer to use this error message like:

  ```javascript
      `${errMsgPrefix}: ${result.status} ${result.statusText}`
  ```

  rather than 

  ```javascript
      `${fetchErrorMsg}${result.status} ${result.statusText}`
  ```

  Is just me? 😆 

* With the `fetchErrorMsg` included, this ends up as _really_ long error message. The existing error message will still propagate up, right? Do we want to print it again as an error of an error? e.g.,:

  ```
  An error occurred while reporting your error to Apollo: An error occurred while fetching your schema from Apollo: 418 I'm a teapot
  ```

As a straw-dog suggestion:
	  
```suggestion
    throw new Error(
      `Out-of-band error reporting failed: ${result.status} ${result.statusText}`);
  }
```

gateway-js/src/loadSupergraphSdlFromStorage.ts
```diff
@@ -36,28 +43,74 @@ const { name, version } = require('../package.json');
...
+  startedAt = '',
+  endedAt = '',
+  tags,
   fetcher,
 }: {
-  graphId: string;
-  graphVariant: string;
-  apiKey: string;
-  endpoint: string;
+  error: any;
```
abernix, line 55: Can this be:

```suggestion
  error: Error;
```

?

gateway-js/src/loadSupergraphSdlFromStorage.ts
```diff
@@ -36,28 +43,74 @@ const { name, version } = require('../package.json');
...
   try {
-    result = await fetcher(endpoint, {
+    let responseBody;
+    try{
+      responseBody = await response?.json();
+    } catch (e) {
+      responseBody = null;
+    }
+
+    const result = await fetcher(process.env.APOLLO_OUT_OF_BAND_REPORTER_ENDPOINT!!, {
```
abernix, line 90: Are we experimental right now so not documenting/enabling this by default?  Will it be set to a different URL than the regular Uplink endpoint? (Eventually?)

package-lock.json
```diff
@@ -113,7 +113,7 @@
     },
     "harmonizer": {
       "name": "@apollo/harmonizer",
-      "version": "0.2.4",
+      "version": "0.2.5",
       "license": "MIT",
       "dependencies": {
         "@apollo/federation": "file:../federation-js"
```
abernix, line 119: It would be reasonable for you to remove these changes from this PR.  We should land them, but they're not related to it and another PR may make similar conflicting changes.  We should just fix it on `main`.   (Concretely: The best thing for you to do would probably just back-out this file's changes.)

gateway-js/src/loadSupergraphSdlFromStorage.ts
```diff
@@ -89,6 +200,18 @@ export async function loadSupergraphSdlFromStorage({
...
+        error: fetchErrorMsg + result.status + ' ' + result.statusText,
+        request,
+        response: result,
+        startedAt: startTime.toISOString(),
+        endedAt: endTime.toISOString(),
+        fetcher
+      });
+    } catch(e) {
+      throw new Error(e);
+    }
```
abernix, line 214: This `try` / `catch` doesn't seem necessary as opposed to just calling it, right? (It's just re-throwing the trapped error.)

```suggestion
    await submitOutOfBandReport({
      error: fetchErrorMsg + result.status + ' ' + result.statusText,
      request,
      response: result,
      startedAt: startTime.toISOString(),
      endedAt: endTime.toISOString(),
      fetcher
    });
```

gateway-js/src/loadSupergraphSdlFromStorage.ts
```diff
@@ -36,28 +43,74 @@ const { name, version } = require('../package.json');
...
+          input: {
+            error: {
+              code: errorCode,
+              message: error.message ?? error
+            },
+            request: {
+              url: request.url,
+              headers: Object.entries(request.headers),
+              body: request.bodyUsed ? await request.json() : ''
+            },
```
abernix, line 104: Do we need to re-transmit _this_ much information from the original request? Isn't this effectively including all of the `variables` including the `apiKey`?  If we included neither the body or the headers, could we still get the signal we need?  (I didn't immediately find the _requirements_ for this, so not sure what precisely we want to capture — maybe you can help me understand?).

I can imagine sending the graphid and variant, but if we could avoid sending everything else entirely, that seems ideal; I'd prefer to avoid what might end up being sanitization methods in any number of places.

gateway-js/src/loadSupergraphSdlFromStorage.ts
abernix, line 90: Single exclamation mark enough?

```suggestion
    const result = await fetcher(process.env.APOLLO_OUT_OF_BAND_REPORTER_ENDPOINT!, {
```

gateway-js/src/loadSupergraphSdlFromStorage.ts
```diff
@@ -36,28 +43,74 @@ const { name, version } = require('../package.json');
...
+            response: response ? {
+              httpStatusCode: response.status,
+              headers: Object.entries(response.headers),
```
abernix, line 107: Similar, but lessened concern as the `request` stuff.  Perhaps just the status code could would capture what we want and limit the possibility of us re-transmitting 🏓 
 other response `headers` we might get back?

gateway-js/src/loadSupergraphSdlFromStorage.ts
```diff
@@ -36,28 +43,74 @@ const { name, version } = require('../package.json');
...
+  // some possible error situations to check against
+  if (response?.status && [400, 413, 422].includes(response?.status)) {
+    errorCode = ErrorCode.InvalidBody;
+  }
+  else if (response?.status && [408, 504].includes(response?.status)) {
+    errorCode = ErrorCode.Timeout;
+  }
+  else if (!response || [502, 503].includes(response?.status)) {
+    errorCode = ErrorCode.ConnectionFailed;
+  }
```
abernix, line 80: This various error codes and presence/absence of `response` should have tests added for them?  (There's just one test for 408 right now, yes?)

gateway-js/src/loadSupergraphSdlFromStorage.ts
```diff
@@ -36,28 +43,74 @@ const { name, version } = require('../package.json');
...
 }: {
-  graphId: string;
-  graphVariant: string;
-  apiKey: string;
-  endpoint: string;
+  error: any;
+  request: Request;
+  response?: Response;
+  startedAt: string;
+  endedAt: string;
```
abernix, line 59: I'm tempted to suggest that we perhaps keep these as native date objects until right when we ship them off (and convert them to `ISO` at that point).  Just a thought though, seems fine to keep it this way.  (String types are just common and can be more easy to mix up since TypeScript doesn't have an opaque type notion that could represent an "ISO Date string".)

Not a firm ask, just musing about myself.  This is fine.

gateway-js/src/__tests__/loadSupergraphSdlFromStorage.test.ts
```diff
@@ -356,5 +379,34 @@ describe('loadSupergraphSdlFromStorage', () => {
         `"An error occurred while fetching your schema from Apollo: 500 Internal Server Error"`,
       );
     });
+
+    it("throws on timeout status response and successfully submits an out of band error", async () => {
+      cleanUp = mockedEnv({
+        APOLLO_OUT_OF_BAND_REPORTER_ENDPOINT: mockOutOfBandReporterUrl,
+      });
```
abernix, line 386: Should we test to make sure it _doesn't_ report when it's not enabled?

gateway-js/src/loadSupergraphSdlFromStorage.ts
```diff
@@ -67,10 +120,68 @@ export async function loadSupergraphSdlFromStorage({
         'content-type': 'application/json',
       },
     });
+
+    const oobResponse = await result.json();
```
abernix, line 124: `oobResponse` vs `result` seems a bit overloaded in this context.  It seems plausible that both could even go away at this level (and maybe even the _outer_ `try` and `catch`?) by adopting an approach that `await`s the `fetch` with a `.then` which attempts to `result.json()` and throws unless `data?.reportError`?.  that could be followed by a chained `.catch(...)` and then perhaps the final `catch` + `throw` of nearly the same error (down below) could be removed entirely?

gateway-js/src/loadSupergraphSdlFromStorage.ts
```diff
@@ -36,28 +43,74 @@ const { name, version } = require('../package.json');
...
+  }
+
   try {
-    result = await fetcher(endpoint, {
+    let responseBody;
+    try{
+      responseBody = await response?.json();
+    } catch (e) {
+      responseBody = null;
+    }
```
abernix, line 88: If we can back away from re-transmitting the response we could remove this!

gateway-js/src/loadSupergraphSdlFromStorage.ts
```diff
@@ -36,28 +43,74 @@ const { name, version } = require('../package.json');
...
+    errorCode = ErrorCode.InvalidBody;
+  }
+  else if (response?.status && [408, 504].includes(response?.status)) {
+    errorCode = ErrorCode.Timeout;
+  }
+  else if (!response || [502, 503].includes(response?.status)) {
+    errorCode = ErrorCode.ConnectionFailed;
+  }
+
   try {
```
abernix, line 82: In my comment below about `oobResponse` vs `result`, I'll make a claim about removing this entire `try` / `catch`?  (It's effectively just an error message prefixer in the way it behaves, right?)

abernix (changes requested):
Thanks for putting this together!  I have a few questions within about: the data we're exchanging/capturing, some code that could maybe benefit from tests and some general error messaging, but I think this is a step in a great direction!

gateway-js/src/loadSupergraphSdlFromStorage.ts
Y-Guo: This still needs to be dogfooded so it's a bit experimental. The url will most likely be `https://uplink.api.apollographql.com/monitoring` unless the client has a custom configuration.

gateway-js/src/loadSupergraphSdlFromStorage.ts
```diff
@@ -19,6 +19,13 @@ export const SUPERGRAPH_SDL_QUERY = /* GraphQL */`#graphql
   }
 `;
 
+// Magic /* GraphQL */ comment below is for codegen, do not remove
+export const OUT_OF_BAND_REPORTER_QUERY = /* GraphQL */`#graphql
+  mutation testMutation($input: APIMonitoringReport) {
```
glasser, line 24: Use a real name instead of testMutation

gateway-js/src/loadSupergraphSdlFromStorage.ts
```diff
@@ -36,28 +43,74 @@ const { name, version } = require('../package.json');
...
+
+    const result = await fetcher(process.env.APOLLO_OUT_OF_BAND_REPORTER_ENDPOINT!!, {
       method: 'POST',
       body: JSON.stringify({
-        query: SUPERGRAPH_SDL_QUERY,
+        query: OUT_OF_BAND_REPORTER_QUERY,
         variables: {
-          ref: `${graphId}@${graphVariant}`,
-          apiKey,
+          input: {
```
glasser, line 95: You're using codegen to generate types for the variables and return type from this request, but you're never actually using those types. The existing code doesn't do it for variables either (worth fixing!) but does for return type at least.

ie, doing something like

```
const variables: TestMutationMutationVariables = {
  input: {
```
would help here. (Though as mentioned elsewhere you should have a different name than TestMutation.)

glasser (changes requested):
I just skimmed a few pieces of this and am leaving a full review to others.

As I mentioned a bit back on Slack, I am not really sure I understand how this is really "out of band" (it's just another operation on the same endpoint?) and it sure seems like issues with the main uplink API might affect the reporting API too. But I guess it will give us some more information than we have now, just not complete information.

gateway-js/src/loadSupergraphSdlFromStorage.ts
trevor-scheer: Similarly, adding / enforcing the types in the test mocks would be great to make sure the mocks themselves don't fall out of date.

gateway-js/src/__tests__/loadSupergraphSdlFromStorage.test.ts
```diff
@@ -380,20 +371,96 @@ describe('loadSupergraphSdlFromStorage', () => {
       );
     });
 
-    it("throws on timeout status response and successfully submits an out of band error", async () => {
+    it('Out of band reporting doesnt submit reports when endpoint is not configured', async () => {
+      mockSupergraphSdlRequest().reply(400);
```
trevor-scheer, line 375: We might add a comment here to help illustrate how this test actually tests the thing. i.e.
```
// if an additional request were made by the out of band reporter, nock would throw since it's unmocked
// and this test would fail
```

Without this knowledge it could be unclear as to how this test is exercising our case at all.

gateway-js/src/__tests__/loadSupergraphSdlFromStorage.test.ts
```diff
@@ -409,4 +476,92 @@ describe('loadSupergraphSdlFromStorage', () => {
...
+        apiKey,
+        endpoint: mockCloudConfigUrl,
+        fetcher,
+      }),
+    ).rejects.toThrowErrorMatchingInlineSnapshot(
+      `"An error occurred while fetching your schema from Apollo: 504 Gateway Timeout"`,
+    );
+  });
+
+  it('throws on when there is no response and successfully submits an out of band error', async () => {
```
trevor-scheer, line 502:
```suggestion
  it('throws when there is no response and successfully submits an out of band error', async () => {
```

trevor-scheer (commented):
These are strictly my own comments from reviewing the recent commits by commit. I still need to see if feedback from other reviewers has been addressed.

gateway-js/src/loadSupergraphSdlFromStorage.ts
trevor-scheer: I believe this comment remains unaddressed given that we're still transmitting the whole body. Was there a decision explicitly made to keep this @Y-Guo?

gateway-js/src/loadSupergraphSdlFromStorage.ts
trevor-scheer: This appears to me to be resolved, but I will leave it for @Y-Guo to click the button in case she feels anything remains unaccounted for.

gateway-js/src/loadSupergraphSdlFromStorage.ts
trevor-scheer: @Y-Guo it's very possible one of my suggestions (re: preference) goes exactly against a change you've made per this recommendation from @abernix. If so, please feel even more free to disregard and leave as is.

gateway-js/src/outOfBandReporter.ts
```diff
@@ -0,0 +1,108 @@
...
+      errorCode = ErrorCode.ConnectionFailed;
+    }
+
+    let oobResponse: Response;
+    let responseBody: string;
+    try{
+      responseBody = await response?.json();
+    } catch (e) {
+      responseBody = '';
+    }
```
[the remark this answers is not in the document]
trevor-scheer: I'm actually unsure if we can call `.text()` after we've called `.json()` in the failure case.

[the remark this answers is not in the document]
Y-Guo: This is a great call out, since the api is expecting the response body as string, it actually makes more since just to use body.text() instead of decoding the json!

gateway-js/src/outOfBandReporter.ts
```diff
@@ -0,0 +1,108 @@
...
+      errorCode = ErrorCode.Timeout;
+    }
+    else if (!response || [502, 503].includes(response?.status)) {
+      errorCode = ErrorCode.ConnectionFailed;
+    }
+
+    let oobResponse: Response;
+    let responseBody: string;
+    try{
+      responseBody = await response?.json();
```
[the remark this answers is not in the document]
Y-Guo: This is a great call out, since the api is expecting the response body as string, it actually makes more sense just to use body.text() instead of decoding the json!

gateway-js/src/outOfBandReporter.ts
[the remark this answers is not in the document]
Y-Guo: resolved by getting rid of of `.json()` calls

gateway-js/src/outOfBandReporter.ts
```diff
@@ -0,0 +1,108 @@
...
+  }
+`;
+
+const { name, version } = require('../package.json');
+
+export class OutOfBandReporter {
+  endpoint: string | null;
+
+  constructor() {
+    this.endpoint = process.env.APOLLO_OUT_OF_BAND_REPORTER_ENDPOINT || null;
```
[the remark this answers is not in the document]
Y-Guo: I see the point you're making here and definitely agree with making it static. I originally had `oobEndpoint` as a global variable but loading it this way causes all tests to fail. I think it's a problem with the order env variables are populated in testing so the mocked `APOLLO_OUT_OF_BAND_REPORTER_ENDPOINT` is still un-mocked when it's accessed as a global. Putting it in a class seems to delay the loading until after the mock env is applied and allows the tests to proceed.

gateway-js/src/loadSupergraphSdlFromStorage.ts
Y-Guo: We've decided to stop transmitting headers but keep the body for now. We might be able to move away from re-transmitting the response once we see what kind of error data is actually reported from clients. In the current iteration having more data is good since it gives us more visibility.

gateway-js/src/outOfBandReporter.ts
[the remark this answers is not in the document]
trevor-scheer: Works for me!

Alternatively, we could have the function optionally accept an endpoint parameter which defaults to the globally defined constant. We'd really only expect to use that argument in the testing case.

No change needed. This class may even end up being a nice way to group different API calls to the oob reporting endpoint in the future.

gateway-js/src/outOfBandReporter.ts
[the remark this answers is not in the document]
trevor-scheer: I like the outcome here!

gateway-js/src/outOfBandReporter.ts
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    const oobVariables: ApiMonitoringReport = {
+      error: {
+        code: errorCode,
+        message: error.message ?? error
+      },
+      request: {
+        url: request.url,
+        body: request.bodyUsed ? await request.json() : ''
```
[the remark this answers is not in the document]
trevor-scheer: `bodyUsed` was removed altogether since we're not expecting this request body to have been consumed anywhere.

trevor-scheer (approved):
LGTM! Please add an entry to the `gateway-js/CHANGELOG.md` file and we'll get this merged.

gateway-js/CHANGELOG.md
```diff
@@ -5,7 +5,7 @@
 > The changes noted within this `vNEXT` section have not been released yet.  New PRs and commits which introduce changes should include an entry in this `vNEXT` section as part of their development.  When a release is being prepared, a new header will be (manually) created below and the appropriate changes within that release will be moved into the new section.
 
 -  Take subtypes into account when matching type conditions to extract representations. [PR #804](https://github.com/apollographql/federation/pull/804)
-
+- Send error reports to a configurable `APOLLO_OUT_OF_BAND_REPORTER_ENDPOINT=https://uplink.api.apollographql.com/monitoring` owned by Apollo. Using the url `https://uplink.api.apollographql.com/monitoring` is recommended unless you have a custom configuration. Reports will only be sent if the endpoint value is set in `env`.  [PR #777](https://github.com/apollographql/federation/pull/777/files)
```
trevor-scheer, line 8:
```suggestion
- Send error reports to a configurable endpoint when providing the `APOLLO_OUT_OF_BAND_REPORTER_ENDPOINT` env variable. Using the Apollo URL `https://uplink.api.apollographql.com/monitoring` is recommended unless you have a custom configuration. Reports will only be sent if the env variable is set. [PR #777](https://github.com/apollographql/federation/pull/777)
```

trevor-scheer:
Merge and release set for Monday pending internal smoke test results!

PR #804, opened 2021-06-09
Take subtypes into account when matching type conditions to extract representations

martijnwalraven:
The `executeSelectionSet` function in `executeQueryPlan.ts`, which is used to extract field values from a result object to build up representations to send to a subgraph, only looked for exact matches of the `__typename` to the type condition on an inline fragment. This wasn't an issue before, because the query plans we generated always contained an inline fragment per requested object type, never an abstract type condition.

Recent changes in the way we handle scope surfaced this issue however, because it would sometimes wrap things in an additional inline fragment for a parent interface type (this has since been fixed in https://github.com/apollographql/federation/pull/805). 

Although that means the issue won't arise for now, we do expect to make other changes that would require support for abstract type conditions (avoiding exploding types when they share an interface with a `@key` directive).

### DIFF
--- a/gateway-js/CHANGELOG.md
+++ b/gateway-js/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 > The changes noted within this `vNEXT` section have not been released yet.  New PRs and commits which introduce changes should include an entry in this `vNEXT` section as part of their development.  When a release is being prepared, a new header will be (manually) created below and the appropriate changes within that release will be moved into the new section.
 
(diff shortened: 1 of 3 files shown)


pcmanus (approved):
The commit message could use some explanatory notes but lgtm otherwise :)

martijnwalraven:
> The commit message could use some explanatory notes but lgtm otherwise :)

Just added some notes to the PR description, so I'll use that as the commit message.

PR #813, opened 2021-06-14
Add clojure/lacinia

dball:
Note I am not the maintainer, but I have happily used it in the past and observe that it now supports Apollo federation.


michael-watson:
Hi @dball 👋 , thanks for this! I'm actually actively working on a revamp of this section of the docs that will give a better understanding of what exactly is supported in each of these libraries. Much more to come on this in the near future and I'm going to add this library to the list of what I'll be testing.

PR #816, opened 2021-06-15
fix(query-planner): Avoid asking for fields of types that aren't owned

barthez:
When the resolved field type is shared interface and this interface is implemented in different services by types that aren't shared query planner would create queries that cause `No such type ...` errors.

### Example

Service A:
```
type Query {
  myField: MyInterface
}

type MyInterface {
  name: String
}

type MyTypeA implements MyInterface {
  name: String
}

type MyTypeB implements MyInterface {
  name: String
}
```

Service B:
```
interface MyInterface {
 name: String
}

type MyTypeC implements MyInterface {
  name: String
}
```

The query:
```
query {
  myField: {
    ...MyIntFragment
  }
}

fragment MyIntFragment on MyInterface {
  name
}
```

result in such query to service A:
```
query {
  myField {
    ...on MyTypeA { name }
    ...on MyTypeB { name }
    ...on MyTypeC { name }
  }
}
```
which cause an error
```
Unknown type "MyTypeC".
```

### DIFF
--- a/query-planner-js/src/__tests__/features/basic/build-query-plan.feature
+++ b/query-planner-js/src/__tests__/features/basic/build-query-plan.feature
@@ -1303,7 +1303,7 @@ Scenario: supports mutations with a cross-service request
               }
             ],
             "variableUsages": [],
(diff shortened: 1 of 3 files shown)


barthez:
Apparently the suggested fix breaks something else. Closing in favor of https://github.com/apollographql/federation/issues/817

PR #823, opened 2021-06-17
Alternative GraphQL schema abstraction

pcmanus:
WIP: open to early comments, but not to be considered ready for a full review.

This adds a new abstraction that is meant to be alternative to `GrapQLSchema` for federation code with the following main differences:
- Directives (particularly thinking directive _applications_) are first class citizen.
- It is mutable, has a notion of `SchemaElement` and has links between elements that reference each others. With the goal of making transformation of schemas and schema navigation a lot easier/safer.
- It allows to work with federation specific "built-in" a bit more conveniently.

### DIFF
--- a/core-js/src/definitions.ts
+++ b/core-js/src/definitions.ts
@@ -27,10 +27,6 @@ export type NullableType = NamedType | ListType<any>;
 
 export type NamedTypeKind = NamedType['kind'];
 
-export function defaultRootTypeName(root: SchemaRoot) {
-  return root.charAt(0).toUpperCase() + root.slice(1);
-}
-
 export function isNamedType(type: Type): type is NamedType {
   return type instanceof BaseNamedType;
 }
@@ -87,7 +83,9 @@ export interface Named {
   readonly name: string;
 }
 
-// Note exposed: mostly about avoid code duplication between SchemaElement and Directive (which is not a SchemaElement as it can't
+export type ExtendableElement = SchemaDefinition | NamedType;
+
+// Not exposed: mostly about avoid code duplication between SchemaElement and Directive (which is not a SchemaElement as it can't
 // have applied directives or a description
 abstract class Element<TParent extends SchemaElement<any> | Schema> {
   protected _parent?: TParent;
@@ -112,8 +110,34 @@ abstract class Element<TParent extends SchemaElement<any> | Schema> {
   }
 
   protected setParent(parent: TParent) {
-    assert(!this._parent, "Cannot set parent of a non-detached element");
+    assert(!this._parent, "Cannot set parent of an already attached element");
     this._parent = parent;
+  }
+
+  protected checkUpdate() {
+    // Allowing to add element to a detached element would get hairy. Because that would mean that when you do attach an element,
+    // you have to recurse within that element to all children elements to check whether they are attached or not and to which
+    // schema. And if they aren't attached, attaching them as side-effect could be surprising (think that adding a single field
+    // to a schema could bring a whole hierachy of types and directives for instance). If they are attached, it only work if
+    // it's to the same schema, but you have to check.
+    // Overall, it's simpler to force attaching elements before you add other elements to them.
+    if (!this.schema()) {
+      throw buildError(`Cannot modify detached element ${this}`);
+    }
+  }
+}
+
+export class Extension<TElement extends ExtendableElement> {
+  protected _extendedElement?: TElement;
+  sourceAST?: ASTNode;
+
+  get extendedElement(): TElement | undefined {
+    return this._extendedElement;
+  }
+
+  private setExtendedElement(element: TElement) {
+    assert(!this._extendedElement, "Cannot attached already attached extension");
+    this._extendedElement = element;
   }
 }
 
@@ -169,7 +193,7 @@ export abstract class SchemaElement<TParent extends SchemaElement<any> | Schema>
     return false;
   }
 
-  protected removeTypeReferenceInternal(type: BaseNamedType<any>) {
+  protected removeTypeReferenceInternal(type: BaseNamedType<any, any>) {
     // This method is a bit of a hack: we don't want to expose it and we call it from an other class, so we call it though
     // `SchemaElement.prototype`, but we also want this to abstract as it can only be impemented by each concrete subclass.
     // As we can't have both at the same time, this method just delegate to `remoteTypeReference` which is genuinely
@@ -193,15 +217,7 @@ export abstract class SchemaElement<TParent extends SchemaElement<any> | Schema>
     if (this.isElementBuiltIn() && !Schema.prototype['canModifyBuiltIn'].call(this.schema()!)) {
       throw buildError(`Cannot modify built-in ${this}`);
     }
-    // Allowing to add element to a detached element would get hairy. Because that would mean that when you do attach an element,
-    // you have to recurse within that element to all children elements to check whether they are attached or not and to which
-    // schema. And if they aren't attached, attaching them as side-effect could be surprising (think that adding a single field
-    // to a schema could bring a whole hierachy of types and directives for instance). If they are attached, it only work if
-    // it's to the same schema, but you have to check.
-    // Overall, it's simpler to force attaching elements before you add other elements to them.
-    if (!this.schema()) {
-      throw buildError(`Cannot modify detached element ${this}`);
-    }
+    super.checkUpdate();
     if (addedElement) {
       const thatSchema = addedElement.schema();
       if (thatSchema && thatSchema != this.schema()) {
@@ -221,8 +237,9 @@ export abstract class NamedSchemaElement<TParent extends NamedSchemaElement<any,
   abstract remove(): TReferencer[];
 }
 
-abstract class BaseNamedType<TReferencer> extends NamedSchemaElement<Schema, TReferencer> {
+abstract class BaseNamedType<TReferencer, TOwnType extends NamedType> extends NamedSchemaElement<Schema, TReferencer> {
   protected readonly _referencers: Set<TReferencer> = new Set();
+  protected readonly _extensions: Set<Extension<TOwnType>> = new Set();
 
   constructor(name: string, readonly isBuiltIn: boolean = false) {
     super(name);
@@ -242,6 +259,28 @@ abstract class BaseNamedType<TReferencer> extends NamedSchemaElement<Schema, TRe
 
   *allChildElements(): Generator<NamedSchemaElement<any, any>, void, undefined> {
     // Overriden by those types that do have chidrens
+  }
+
+  extensions(): ReadonlySet<Extension<TOwnType>> {
+    return this._extensions;
+  }
+
+  newExtension(): Extension<TOwnType> {
+    return this.addExtension(new Extension<TOwnType>());
+  }
+
+  addExtension(extension: Extension<TOwnType>): Extension<TOwnType> {
+    this.checkUpdate();
+    // Let's be nice and not complaint if we add an extension already added.
+    if (this._extensions.has(extension)) {
+      return extension;
+    }
+    if (extension.extendedElement) {
+      throw buildError(`Cannot add extension to type ${this}: it is already added to another type`);
+    }
+    this._extensions.add(extension);
+    Extension.prototype['setExtendedElement'].call(extension, this);
+    return extension;
   }
 
   protected isElementBuiltIn(): boolean {
@@ -318,7 +357,32 @@ abstract class BaseNamedElementWithType<TType extends Type, P extends NamedSchem
 
 function buildError(message: string): Error {
   // Maybe not the right error for this?
-  return new GraphQLError(message);
+  return new Error(message);
+}
+
+abstract class BaseExtensionMember<TExtended extends ExtendableElement> extends Element<TExtended> {
+  private _extension?: Extension<TExtended>;
+
+  ofExtension(): Extension<TExtended> | undefined {
+    return this._extension;
+  }
+
+  setOfExtension(extension: Extension<TExtended> | undefined) {
+    this.checkUpdate();
+    // See similar comment on FieldDefinition.setOfExtension for why we have to cast.
+    if (extension && !this.parent?.extensions().has(extension as any)) {
+      throw buildError(`Cannot set object as part of the provided extension: it is not an extension of parent ${this.parent}`);
+    }
+    this._extension = extension;
+  }
+
+  remove() {
+    this.removeInner();
+    this._extension = undefined;
+    this._parent = undefined;
+  }
+
+  protected abstract removeInner(): void;
 }
 
 export class BuiltIns {
@@ -406,7 +470,7 @@ export class Schema {
     return !this.isConstructed;
   }
 
-  private removeTypeInternal(type: BaseNamedType<any>) {
+  private removeTypeInternal(type: BaseNamedType<any, any>) {
     this._types.delete(type.name);
   }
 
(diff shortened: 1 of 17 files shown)


core-js/src/definitions.ts
```diff
@@ -0,0 +1,1421 @@
...
+  get appliedDirectives(): readonly Directive[] {
+    return this._appliedDirectives;
+  }
+
+  appliedDirective(definition: DirectiveDefinition): Directive[] {
+    return this._appliedDirectives.filter(d => d.name == definition.name);
+  }
+
+  applyDirective(directive: Directive): Directive; 
+  applyDirective(definition: DirectiveDefinition, args?: Map<string, any>): Directive;
```
martijnwalraven, line 128: Although `Map` is preferable for storage nowadays, I think an object (`Record<string, any>`) makes for a more natural API for `args` because you can pass in an object literal.

martijnwalraven: I also think we'll want `DirectiveDefinition<TArgs>` so we can have (optional) strong typing for directive arguments.

core-js/src/definitions.ts
```diff
@@ -0,0 +1,1421 @@
...
+  } else {
+    throw new Error(`Type ${type} (${type.kind}) is not an input type`);
+  }
+}
+
+export interface Named {
+  readonly name: string;
+}
+
+export abstract class SchemaElement<Parent extends SchemaElement<any, any> | Schema, Referencer> {
```
martijnwalraven, line 92: Maybe `TParent` and `TReferencer`? That seems to be the convention for type arguments.

core-js/src/definitions.ts
```diff
@@ -0,0 +1,1421 @@
...
+
+  private removeReferencer(referencer: Referencer) {
+    this._referencers.delete(referencer);
+  }
+
+  get coordinate(): string {
+    return this.name;
+  }
+
+  *allChildrenElements(): Generator<SchemaElement<any, any>, void, undefined> {
```
martijnwalraven, line 230:
```suggestion
  *allChildElements(): Generator<SchemaElement<any, any>, void, undefined> {
```

core-js/src/definitions.ts
```diff
@@ -0,0 +1,1421 @@
...
+  }
+
+  protected isElementBuiltIn(): boolean {
+    return this.isBuiltIn;
+  }
+
+  /**
+   * Removes this type definition from its parent schema.
+   *
+   * After calling this method, this type will be "detached": it wil have no parent, schema, fields,
```
martijnwalraven, line 241: I understand a detached type won't have a parent, but why would fields, values, directives, etc. be removed as well?

core-js/src/definitions.ts
```diff
@@ -0,0 +1,1421 @@
...
+    return schema.addDirectiveDefinition(new DirectiveDefinition(name, true));
+  }
+}
+
+export class Schema {
+  private _schemaDefinition: SchemaDefinition;
+  private readonly _builtInTypes: Map<string, NamedType> = new Map();
+  private readonly _types: Map<string, NamedType> = new Map();
+  private readonly _builtInDirectives: Map<string, DirectiveDefinition> = new Map();
+  private readonly _directives: Map<string, DirectiveDefinition> = new Map();
```
martijnwalraven, line 354: Why the distinction between `builtInTypes`/`types` and `builtInDirectives`/`directives`? As built-in elements live in the same namespace as user-defined elements, can't they share a `Map`?

martijnwalraven (commented):
Looks great overall! I'll have to finish the review later, but wanted to submit the comments so far.

core-js/src/definitions.ts
```diff
@@ -0,0 +1,1421 @@
...
+
+export interface Named {
+  readonly name: string;
+}
+
+export abstract class SchemaElement<Parent extends SchemaElement<any, any> | Schema, Referencer> {
+  protected _parent?: Parent;
+  protected readonly _appliedDirectives: Directive[] = [];
+  description?: string;
+  source?: ASTNode;
```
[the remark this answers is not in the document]
pcmanus: Initially I was planning to generalise this somehow to a more general notion of "sources". At least, for composition, it would made sense for a given element to point back to each subgraph nodes that were merged into that particular element (so in that case, the source would have to be more like a map of subgraph name to `ASTNode`).

I also wanted to carry the notion that this `ASTNode` might not represent the state of the `Schema` anymore, it's only the state of when it was created. And to clarify, what I mean here is that if you consider an `ObjectType` for instance, the `source` will properly represent the type after parsing, but if you modify that type now (say, add a field), then the code is not clearly the `source` (because it feels useful to still being able to point to the original source for errors) but that source is not in sync with the object anymore.

Anyway, I'll probably rename to something like `sourceAST` for now. But I might generalise this soonish when I work on composition.

core-js/src/definitions.ts
```diff
@@ -0,0 +1,1421 @@
...
+
+  abstract coordinate: string;
+
+  schema(): Schema | undefined {
+    if (!this._parent) {
+      return undefined;
+    } else if ( this._parent instanceof Schema) {
+      return this._parent as any;
```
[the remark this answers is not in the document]
pcmanus: That surprised me as well, but `as Schema` does not work. And what I suspect is that this has to do with some limitation of typescript type system with generics (or I'm being stupid, which is always an option).

Namely, even after the `instanceof`, typescript says that `this._parent` is of type `Parent & Schema`, where `Parent` is the generic argument defined as `Parent extends SchemaElement | Schema`), and complains that this isn't assignable to `Schema | undefined` (and it says it's not because `Type 'Parent & Schema' is missing some properties from type 'Schema'` which does not make sense to me: either the `&` in typescript has a really weird semantic, or it sound like a bug/limitation). But due to that, it also refuse `as Schema` saying that it "may be a mistake" and we should should "convert to 'unknown" first.

Tl;dr, `as any` is the shortest way I know to make it shut-up.

I will definitively add a comment though.

core-js/src/definitions.ts
```diff
@@ -0,0 +1,1421 @@
...
+  protected setParent(parent: Parent) {
+    assert(!this._parent, "Cannot set parent of a non-detached element");
+    this._parent = parent;
+  }
+
+  get appliedDirectives(): readonly Directive[] {
+    return this._appliedDirectives;
+  }
+
+  appliedDirective(definition: DirectiveDefinition): Directive[] {
```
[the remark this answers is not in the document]
pcmanus: > Should this be `appliedDirectives` (plural)?

Yeah, I guess. I really meant something like `directiveDefinitionApplications` by this but it's better plural.

> Or perhaps there's a way to differentiate between repeatable and non-repeatable directives

Not sure there is anything that is worth the complexity. Out of curiosity, I tried:
```typescript
appliedDirective(definition: DirectiveDefinition & { repeatable: false }) {
   ...
}
```
but that doesn't seem to work, in the sense that the following still doesn't compile:
```typescript
const def = ...;
if (def.repeatable === false)
  schema.appliedDirective(def)
```

So making 2 variants of `DirectiveDefinition` just so we can make that distinction on that function don't seem worth the trouble.

core-js/src/definitions.ts
pcmanus: > I think an object (`Record<string, any>`) makes for a more natural API for args

Fair point.

> I also think we'll want `DirectiveDefinition<TArgs>` so we can have (optional) strong typing for directive arguments.

I'll check how that could work in practice (I can see how strong typing for known directives would be nice; I just wonder how carrying this between `DirectiveDefinition` and `Directive` will work in practice).

core-js/src/definitions.ts
pcmanus: There is a comment in `SchemaElement.checkUpdate()` that goes to try to explain the reasoning a bit. But long story short, having to deal with detached elements can be a bit messy.

If we only detach the parent, we may end up with a detached field whose type is still attached, and that sound messy/error prone. But more generally,  the problem is around behaviour when you "attach" (add) an detached element that itself has a deep hierarchy of elements. Think: I'm adding a field to a type, but that field link to another detached Object Type that itself may have fields linking to other detached types, implements detached interfaces and sprinkled in there there is some directe applications that indirectly link to "some" definition. First, we'd have to recursively explore that hierarchy to ensure nothing is attached somewhere else. But even then, we'd have to properly attach everything, which concretely means that the user might have call `addField` on a schema, and as a result 10 new types and whatnot are added to the schema on top of the field. Of course, maybe that's desired if you know what you're doing, but to me, this sounds like complex behaviour that is more likely to lead to surprising results.

Overall, the idea of the current patch is that while detached element are something we unfortunately have to deal, it tries to make it as innocuous/predictable as possible by ensure that a detached element is always "just an empty shell". Which is done effectively by:
1. no allowing any modification methods on a detached element (`SchemaElement.checkUpdate`).
2. cleaning up elements when they are removed.

core-js/src/definitions.ts
pcmanus: This is somewhat opinionated but I have a strong hunch that much more code that will end up using this library will primarily not want to iterate about builtins (it may care they exists, but will not iterate/manipulate them). For instance, printing doesn't print those, composition will not merge those, any inaccessible removal won't care about those, etc... I'm sure at some point some code will want to iterate over all types/directives include builtins, but those example don't come easily to mind to me.

That's why `allChildElements` don't return builtins for instance (though I suppose maybe an optional argument to include them could make sense) and why `Schema.types` and `Schema.directives` don't include them as well.

So that's the justification for the external API, and the argument can be shorten as "I think it's more useful this way and that if we only always return all types/directives, everyone will have to filter built-ins out all the time".

Of course, internally, we can do whatever we want. We could have all in the same map, yet have the same API and filter built-in out within the accessor methods. But that just feel less efficient for no good reason.

core-js/src/definitions.ts
martijnwalraven: Yeah, dealing with detached elements is messy unfortunately. I agree we should avoid bringing in a whole hierarchy when re-attaching an element. I'm not sure the 'empty shell' semantics make sense to me though. My idea was that references to other elements would always be by name, so if you remove a field for example, you can attach it to a type in another schema but it would be your responsibility to ensure referenced elements are also added.

core-js/src/definitions.ts
```diff
@@ -0,0 +1,1459 @@
...
+    return cloned;
+  }
+}
+
+export class SchemaDefinition extends SchemaElement<Schema, never>  {
+  readonly kind: 'SchemaDefinition' = 'SchemaDefinition';
+  protected readonly _roots: Map<SchemaRoot, ObjectType> = new Map();
+
+  get coordinate(): string {
+    return '';
```
martijnwalraven, line 562: I think this is another sign `SchemaDefinition` should not extend `SchemaElement`.

core-js/src/definitions.ts
```diff
@@ -0,0 +1,1459 @@
...
+
+  protected removeTypeReference(toRemove: NamedType) {
+    for (const [root, type] of this._roots) {
+      if (type == toRemove) {
+        this._roots.delete(root);
+      }
+    }
+  }
+
+  remove(): never[] {
```
martijnwalraven, line 604: I don't think there's a need to support removing the schema definition. That also avoids having to implement/invoke `resetSchemaDefinition`.

core-js/src/definitions.ts
```diff
@@ -0,0 +1,1459 @@
...
+    return [];
+  }
+
+  toString() {
+    return `schema[${[...this._roots.keys()].join(', ')}]`;
+  }
+}
+
+export class ScalarType extends BaseNamedType<OutputTypeReferencer | InputTypeReferencer> {
+  readonly kind: 'ScalarType' = 'ScalarType';
```
martijnwalraven, line 626: I think you should be able to use [`as const`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions) here to avoid specifying the string literal twice.

core-js/src/definitions.ts
```diff
@@ -0,0 +1,1459 @@
...
+    this._referencers.clear();
+    return toReturn;
+  }
+
+  toString(): string {
+    return this.name;
+  }
+}
+
+// TODO: How do we deal with default values? It feels like it would make some sense to have `argument('x')` return the default
```
martijnwalraven, line 1197: That's a good question. I would say it makes sense for `arguments` (or perhaps `argumentValues`) to return the resolved argument values that take default values into account (that's also consistent with `getArgumentValues()`/`getDirectiveValues()` in `graphql-js`. But I don't think default values should be included in the internal representation, because you want to differentiate between explicitly providing a value and relying on the default value. That also means changing the default value in the directive definition should affect `argumentValues`, making that a function that depends on the directive definition (again, like `getArgumentValues`).

core-js/src/federation.ts
```diff
@@ -0,0 +1,61 @@
+import { BuiltIns, Schema, DirectiveDefinition, NonNullType } from "./definitions";
+
+// TODO: Need a way to deal with the fact that the _Entity type is built after validation.
+export class FederationBuiltIns extends BuiltIns {
```
martijnwalraven, line 4: I don't think we should consider the Federation directives + types as built-in. I suspect this is where we'll want to introduce some notion of `Feature` to take care of the namespace mapping in core schemas. We still need to support the current composition behavior when dealing with subgraphs that do not use core schema, but perhaps we can handle that through a more general mechanism as well (optionally marking a feature as built-in?).

core-js/src/definitions.ts
martijnwalraven: Also see https://github.com/graphql/graphql-js/pull/3049 for some background on default value semantics.

core-js/src/definitions.ts
```diff
@@ -0,0 +1,1459 @@
...
+    return this._builtInDirectives.values();
+  }
+
+  directive(name: string): DirectiveDefinition | undefined {
+    const directive = this._directives.get(name);
+    return directive ? directive : this._builtInDirectives.get(name);
+  }
+
+  *allSchemaElement(): Generator<SchemaElement<any, any>, void, undefined> {
+    yield this._schemaDefinition;
```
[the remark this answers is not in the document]
pcmanus: You're right, it's not in `graphql-js` code. That said, the schema definition shares most properties with other elements, namely:
- it can have directive applied to.
- it references other stuffs (types).
- it can have a description.
- it's nice to get it out of `allSchemaElements` to avoid having to special case it (think, code that tries to act on applied directives in a generic way).

I could also argue that, from an pure english language point of view, it _is_ an "element" of a schema. 

So I felt making it a `SchemaElement` was overall a good tradeoff by improving consistency at the price of what I felt was a pretty minor divergence with `graphql-js`.

But I hesitated slightly and I'm amenable to a different opinion. But I will note that if we do change that, we'll have to either:
1. split the current `SchemaElement` into 2 abstractions: `SchemaElement` that would just provide the `coordinate` (it's all it does in `graphql-js`) and another one that has the rest of what `SchemaElement` currently has (applied directives, description, ...). And I'm not sure how to name that new thing. Which adds a new abstraction for users to care about.
2. essentially duplicate a good part of the code of `SchemaElement` into `SchemaDefinition` _and_ change signatures of most methods on `SchemaElement` to be on `SchemaElement | SchemaDefinition`. Which overall sounds ugly.

core-js/src/definitions.ts
pcmanus: See comments above. But essentially, `SchemaElement` in this code is a lot more than just the coordinate, so the tradeoff is:
1. do we prefer slightly abusing the coordinate in that one case given than `SchemaDefinition` shares all other properties with other `SchemaElement`.
2. or do we insist on sticking with `SchemaElement` for things that strictly have coordinates in the graphql sense and creates a new abstraction for the other stuffs.

I've felt slightly abusing the coordinate was worth saving an abstraction. Might be the wrong choice (and I don't deny the empty coordinate thing is a tad ugly).

core-js/src/definitions.ts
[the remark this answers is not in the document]
pcmanus: All that said, thinking about it more, I think splitting `SchemaElement` in 2 abstractions might be overall better. I do am unsure how to call it though.

core-js/src/definitions.ts
pcmanus: Thanks for the tip!

core-js/src/definitions.ts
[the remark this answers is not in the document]
martijnwalraven: Yeah, I struggled with the same question with my AST-based attempt. I considered differentiating between `SchemaElement` and a `NamedSchemaElement` subclass at one point. I don't love it, but `graphql-java` uses [a similar model](https://github.com/graphql-java/graphql-java/blob/2cea0afb4d010a92debcc70cf514b9c446cb32e5/src/main/java/graphql/schema/GraphQLNamedSchemaElement.java#L10).

core-js/src/definitions.ts
pcmanus: I definitively agree about not storing the default values internally, that wasn't my intention.

I'll note thought that you're suggestion about having optional `TArgs` for directives make this a tad less obvious. Because with that, it made sense to remove the `Directive.argument(name: string): any` method, and always rely on `get Directive.arguments(): TArgs` (I mean, what's the point of doing `directive.argument('foo')` and potentially losing type checking, when you can do `directive.arguments.foo` and not lose type checking).

But it's a tad less clear how to handle defaults in that case (API wise I mean).

I can think of a few options:
1. stop having `Directive.arguments` be a getter, and be `Directive.arguments(withDefaultValues: boolean = false): TArgs`. If you pass `withDefaultValues = true`, we'd copy the underlying args object and set undefined keys to their defaults if necessary.
2. keep `Directive.arguments` a getter, have it always include default values but have an additional `isDefaultValue(name: string)` for if you want to know if a particular value was provided.
3. keep `Directive.arguments` a getter, have it not include default values and have a separate method `argumentOrDefault(na me: string): any` that can alternative give you values with defaults. I like this less because you don't get type-checked access if you also want defaults.

Overall, leaning towards option 1.

core-js/src/federation.ts
pcmanus: I think we need both `Feature` and built-ins in some fashion (reminder that this PR doesn't pretend to be ready). And for good or bad, we do need to support existing federation directives as "built-in" for backward compatibility (I mean built-in in the very practical sense that they exists "magically" without the user having defined them in any way).

So I think we need this, at least for now. Hopefully, in some future version, we'll remove `FederationBuiltIns` and all federation types and directives will comes through `Feature`s, but we're not there yet.

I _will_ note that I include `@inacessible` in there and we may not want to expose that one as "built-in" and instead only support it as a feature. So please don't read anything into that one other than "this is work in progress".

> but perhaps we can handle that through a more general mechanism as well

Sure, maybe once we add `Feature` support in this, we've want to refactor a bit, that's ok. But I don't think the way it's currently done is too bad either, because currently, when you use federation, those types and directives _are_ essentially built-in.

core-js/src/definitions.ts
pcmanus: As I though, I can't really find a way to handle `DirectiveDefinition<TArgs>` in a safe way. I mean, the args of a _definition_ are complex objects that can have directives and whatnot, and so I don't see a way to have `DirectiveDefinition` take a generic that reflect those properly and could be transformed to type-check the applications of that directives.

What I did do is add a `DirectiveDefinition<TApplicationArgs>`, where `TApplicationArgs` is directly the type of _application_ arguments. But this does mean there is no way to ensure (statically or not really) that this generic argument is valid. However, _assuming_ it is valid, it does provide nice type-checking for the related directives and I think that's still sufficiently useful to be worth the slight risk of getting this wrong.

Please have a look at https://github.com/apollographql/federation/pull/823/commits/a969ad569478df0ec79024adc64727a312daa6b0 for what I did do, see if this look anywhere near what you had in mind.

core-js/src/definitions.ts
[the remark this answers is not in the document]
pcmanus: For posterity: I switched to `SchemaElement` and `NamedSchemaElement` naming.

core-js/src/definitions.ts
pcmanus: This was mostly done due to `remove` being abstract in `SchemaElement` and `SchemaDefinition` being a `SchemaElement`. With the change to `NamedSchemaElement`, I moved the `remove` method there and thus removed `SchemaDefinition.remove()`.

pcmanus:
Closing as this is drastically outdated (and a more up-to-date version of this is now on main).

PR #824, opened 2021-06-17
feat(federation): Deterministically sort composed schema

trevor-scheer:
This adds a final transform to the composed schema which sorts types, fields, and anything else which can be sorted lexicographically. This guarantees a deterministic output in the case of the composition algorithm changing and provides readers of the schema with a more consistent way to find what they're looking for.


martijnwalraven (approved):
I haven't reviewed the changes in tests, but the actual invocation of `lexicographicSortSchema` seems pretty straightforward :)

PR #833, opened 2021-06-24
chore: run `npm install` on `cargo build` if `node_modules` does not exist

EverlastingBugstopper:
This makes it so that anybody can run `cargo build` from the root repo from a fresh clone. Also added some nicer error messages with `expect` instead of `unwrap`.


trevor-scheer:
Thanks @EverlastingBugstopper!

PR #834, opened 2021-06-24
fix(federation): Preserve all @tags

trevor-scheer:
Different tags from the same field coming from different places should all be preserved

### DIFF
--- a/federation-js/src/composition/compose.ts
+++ b/federation-js/src/composition/compose.ts
@@ -21,7 +21,7 @@ import {
   DirectiveNode,
 } from 'graphql';
 import { transformSchema } from 'apollo-graphql';
-import federationDirectives from '../directives';
+import apolloTypeSystemDirectives from '../directives';
 import {
   findDirectivesOnNode,
   isStringValueNode,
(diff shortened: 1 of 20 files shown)


federation-js/src/composition/compose.ts
```diff
@@ -154,6 +154,21 @@ export function buildMapsFromServiceList(serviceList: ServiceDefinition[]) {
 
     externalFields.push(...strippedFields);
 
+    // We need to capture @tag directives from the external fields as well
+    for (const {parentTypeName, field} of externalFields) {
```
lennyburdette, line 158: You're appending fields to `externalFields` in each iteration of the serviceList loop, so you're redundantly checking fields you checked in the last iteration. `for (const {parentTypeName, field} of strippedFields) {` would probably work?

lennyburdette: (or you could iterate over `externalFields` outside of the serviceList loop?)

lennyburdette (approved):

lennyburdette:
Would this logic end up adding `@inaccessible` multiple times if it's defined in two subgraphs? Should you check the `isRepeatable` property of the directive definition to know if you need to dedupe?

lennyburdette:
LGTM!

PR #849, opened 2021-07-01
fix(gateway): Fix TS error in OOB variables

trevor-scheer:
This case isn't recognized by our TS as an error because it's actually a `never`. `Error.message` is always a `string` and therefore we'll never coalesce to `error`. This is causing tsc build issues for others though.

See: https://github.com/apollographql/federation/pull/830#discussion_r662613954

cc @smyrick


glasser (approved):
Sure, hopefully errors have messages

PR #859, opened 2021-07-06
fix(federation): Print @tag and @inaccessible directives conditionally

trevor-scheer:
Only print the directive definitions and `@core` declarations for the `@tag` and `@inaccessible` directives in the supergraph SDL if there are usages of them found in the subgraph schemas.


federation-js/src/composition/compose.ts
```diff
@@ -129,6 +129,11 @@ type FieldDirectivesMap = Map<string, DirectiveNode[]>;
 // TODO: rename?
 type TypeNameToFieldDirectivesMap = Map<string, FieldDirectivesMap>;
 
+/**
+ * A set of directive names that have been used at least once
+ */
+type AppliedDirectiveUsages = Set<string>;
```
martijnwalraven, line 135: This is not a blocker, but the term 'applied directive usages' is a bit weird. Both 'applied directives' or 'directive usages' are meant to differentiate between directive definitions and their use.

federation-js/src/composition/compose.ts
```diff
@@ -21,7 +21,7 @@ import {
   DirectiveNode,
 } from 'graphql';
 import { transformSchema } from 'apollo-graphql';
-import apolloTypeSystemDirectives from '../directives';
+import apolloTypeSystemDirectives, { appliedDirectives, federationDirectives } from '../directives';
```
martijnwalraven, line 24: I had missed this before, but `appliedDirectives` doesn't seem like the right name here, because these exports are directive definitions (as opposed to directive applications, which is what we find on AST nodes). Maybe something like `otherKnownDirectives` could work here?

martijnwalraven (commented):
Other than some comments about naming, this looks good to me!

trevor-scheer:
Noting the recommendations are strictly naming/internally facing related and I'm hoping to get in on a Rover release, I'm going to merge now and follow up with naming fixes today.

abernix:
@trevor-scheer agree with your decision to expedite and follow-up with the suggestions that @martijnwalraven offered.  thanks!

PR #864, opened 2021-07-08
Do not send field to the service if it cannot process it

konalegi:
### Problem descrion

When the resolved field type is shared interface and this interface is implemented in different services by types that aren't shared query planner would create queries that cause No such type ... errors.

Example

Service A:

```
type Query {
  myField: MyInterface
}

type MyInterface {
  name: String
}

type MyTypeA implements MyInterface {
  name: String
}

type MyTypeB implements MyInterface {
  name: String
}
```
Service B:

```
interface MyInterface {
 name: String
}

type MyTypeC implements MyInterface {
  name: String
}
```

The query:

```
query {
  myField {
    ...MyIntFragment
  }
}

fragment MyIntFragment on MyInterface {
  name
}
```

result in such query to service A:

```
query {
  myField {
    ...on MyTypeA { name }
    ...on MyTypeB { name }
    ...on MyTypeC { name }
  }
}
```

which cause an error

`Unknown type "MyTypeC".`

### DIFF
--- a/query-planner-js/src/buildQueryPlan.ts
+++ b/query-planner-js/src/buildQueryPlan.ts
@@ -828,7 +828,6 @@ function splitFields(
             debug.group(`For runtime parent type ${runtimeParentType}:`);
 
             const fieldDef = context.getFieldDef(
-
               runtimeParentType,
               field.fieldNode,
             );
@@ -838,16 +837,20 @@ function splitFields(
               fieldDef,
             }));
 
-            group.fields.push(
-              completeField(
-                context,
-                scope.refine(runtimeParentType),
-                group,
-                path,
-                fieldsWithRuntimeParentType,
-              ),
-            );
-            debug.groupEnd(() => `Updated fetch group: ${debugPrintGroup(group)}`);
+            if (shouldIncludeToFetchGroup(group, fieldsWithRuntimeParentType, context, runtimeParentType)) {
+              group.fields.push(
+                completeField(
+                  context,
+                  scope.refine(runtimeParentType),
+                  group,
+                  path,
+                  fieldsWithRuntimeParentType,
+                ),
+              );
+              debug.groupEnd(() => `Updated fetch group: ${debugPrintGroup(group)}`);
+            } else {
+              debug.groupEnd(() => `Group has been skipped: ${debugPrintGroup(group)}`);
+            }
           }
           debug.groupEnd();
         }
@@ -857,6 +860,28 @@ function splitFields(
       }
     }
   }
+}
+
+function shouldIncludeToFetchGroup(
+  group: FetchGroup,
+  fields: FieldSet,
+  context: QueryPlanningContext,
+  runtimeParentType: GraphQLObjectType
+  ): Boolean {
+  const { fieldDef } = fields[0];
+  const returnType = getNamedType(fieldDef.type);
+
+  if (isCompositeType(returnType)) {
+    return true;
+  }
+
+  const owningService = context.getOwningService(runtimeParentType, fieldDef);
+
+  if (owningService === undefined) {
+    return true
+  }
+
+  return owningService === group.serviceName || group.providedFields.some(f => f.fieldDef === fields[0].fieldDef)
 }
 
 function completeField(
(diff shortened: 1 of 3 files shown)


dchambers:
@konalegi thanks for providing this PR. Any chance you can sign the Apollo CLA for it?

konalegi:
@dchambers done

query-planner-js/src/buildQueryPlan.ts
```diff
@@ -688,6 +689,24 @@ function splitFields(
...
+function shouldIncludeToFetchGroup(
+  group: FetchGroup,
+  fields: FieldSet,
+  context: QueryPlanningContext,
+  runtimeParentType: GraphQLObjectType
+  ): Boolean {
+  const { fieldDef } = fields[0];
+  const returnType = getNamedType(fieldDef.type);
+
+  if (isCompositeType(returnType)) {
```
pcmanus, line 701: Curious as to why you added this exclusion? I don't mind keeping it as it limits the cases changed by this fix and is thus somewhat more conservative, but given the goal is to exclude fields from services that don't the field at all, I'm not sure to see why the field return type would influence the behaviour.

Mostly wondering if you were thinking of a case I don't have in mind.

query-planner-js/src/buildQueryPlan.ts
```diff
@@ -688,6 +689,24 @@ function splitFields(
...
+    return true;
+  }
+
+  const owningService = context.getOwningService(runtimeParentType, fieldDef);
```
pcmanus, line 705: We actually need to handle the case where `owningService === undefined` (and `return true`). Otherwise this will create an issue for value types as they don't have an `owningService` and would always be skipped, which we don't want.

An example of this would be if a subgraph has:
```graphql
type Query {
  itf: Itf
}

interface Itf {
  f: String
}

type T1 implements Itf @key(fields: "f") {
  f: String
}

type T2 implements Itf {
  f: String
}
```
With the patch, if you query:
```graphql
{
  itf {
    f
  }
}
```
then the query plan will include `T1` but not `T2` due to this problem.

pcmanus (changes requested):
Thanks for the patch.

As pointed in a comment, the patch creates a problem with value types, but this can be fixed easily enough. It's unfortunate that no test catch that, so I think we'd need to add at least 2 tests: one for the fix here, and one that show the value-type case is not broken. But with the fix and tests, I think this could be committed.

query-planner-js/src/buildQueryPlan.ts
konalegi: Yes, that was the intention, to make changes as little as possible, because I'm not very familiar with the query planner.

konalegi:
Thanks for you review @pcmanus, I've tried to address all your comments. Could you please take a look again? Thanks!

pcmanus (approved):

pcmanus:
Slowly for the slow turn-around. The last changes look good to me, thanks!

konalegi:
@pcmanus thanks for your help with the PR :) Do you know when it's going to be released?

dchambers:
Great work getting a fix for this out there!

pcmanus:
> Do you know when it's going to be released?

I believe we're planning to do the next release relatively soon (tracked at https://github.com/apollographql/federation/pull/1369). I don't have an exact date but I'd say "within a few days".

trevor-scheer:
I'll be handling the release for this likely by the end of the week. It's actually currently available via the gateway alpha (you'll also need to install the query-planner alpha to force its usage)
`npm i @apollo/gateway@alpha @apollo/query-planner@alpha`

Would someone please share a changelog entry here so I can add it before release? And is it fair to say this is just a patch (non-breaking bug fix)?

pcmanus:
> Would someone please share a changelog entry here

I already took the liberty to ninja-push one (https://github.com/apollographql/federation/commit/acf3ffe8da377e0aae5e4d44a509b2669cf7a065) on the `version-0.x` branch. But yeah, it's a non-breaking bug fix.

konalegi:
I'm not sure if the changelog is still required, but I'll provide it, "Not owning types are not added into a query on interface unpacking" not sure if it's very accurate though.

trevor-scheer:
This fix is released in the `latest` versions of federation packages. Thanks again for providing the fix!

PR #868, opened 2021-07-09
fix(federation): Skip missing types while iterating over field directive usages

### LINKED ISSUE #866: composeAndValidate() unexpectedly throws when using @tag/@inaccessible
`composeAndValidate()` in `@apollo/federation@0.26.0` currently throws when given the following subgraph schemas:
- `inventory.graphql`
    ```graphql
    extend type Product @key(fields: "id") {
      id: ID! @external @tag(name: "hi from inventory")
      dimensions: ProductDimension @external
      delivery(zip: String): DeliveryEstimates @requires(fields: "dimensions { size weight }")
    }
    type ProductDimension {
      size: String
      weight: Float @inaccessible
    }
    type DeliveryEstimates {
      estimatedDelivery: String
      fastestDelivery: String
    }
    ```
- `users.graphl`
    ```graphql
    type User @key(fields:"email") {
        email:ID!
        name: String
        totalProductsCreated: Int
    }
    ```

It specifically throws with the message `Cannot read property 'getFields' of undefined`. (Note that while we do expect composition to fail for these specific subgraph schemas, it should generally not throw and instead return `GraphQLError`s)

trevor-scheer:
With the addition of `@tag` and `@inaccessible` directives, we inadvertently began adding invalid types to our `typeNameToFieldDirectivesMap`. These "invalid" types exist in the map (but not the schema) when a type is extended but never actually defined anywhere.

We can simply check the existence of each type in the schema before performing any work there and skip any iterations where a type is not found.

Fixes #866


federation-js/src/composition/__tests__/composeAndValidate.test.ts
```diff
@@ -117,57 +117,85 @@ it('composes and validates all (24) permutations without error', () => {
...
+          ]
+      `);
+  });
+
+  it("doesn't throw errors when a type is unknown, and the type has directive usages which we've captured", () => {
+    const inventory = {
+      name: 'inventory',
+      typeDefs: gql`
+        extend type Product @key(fields: "id") {
+          id: ID! @external @tag(name: "hi from inventory")
```
martijnwalraven, line 179: Shouldn't this error out because of #869?

federation-js/src/composition/compose.ts
```diff
@@ -762,14 +762,22 @@ export function composeServices(services: ServiceDefinition[]): CompositionResul
 
   schema = lexicographicSortSchema(schema);
 
+  // Now that we have a constructed schema, we can filter this map based on types
+  // that made it into the schema successfully. Types that didn't make it into the
+  // schema will have related composition errors.
+  const filteredTypeNameToFieldDirectivesMap = filterMap(
```
martijnwalraven, line 768: Instead of filtering here, wouldn't it be easier to make `addFederationMetadataToSchemaNodes` skip types in `typeNameToFieldDirectivesMap` that don't exist in the schema? That avoids the need for `filterMap` and recreating a `Map` even if nothing gets removed. It seems to me the code [here](https://github.com/apollographql/federation/blob/000323590eb63c425d2bcac6acf9248bad3d33b2/federation-js/src/composition/compose.ts#L636) could use a check.

federation-js/src/composition/__tests__/composeAndValidate.test.ts
trevor-scheer: Well it explicitly should _not_ error out, which it was doing before. #869 updates this test with the new composition error that's returned (it's based against this branch for simplicity). So I'll leave this as is since it's properly addressed in the next PR.

trevor-scheer: https://github.com/apollographql/federation/pull/869/files#diff-9ae375eec1a337fb4d06139755b68ec0d4fe9d6168b0f207c29e6f2a31882daeR188

abernix (approved):
Assuming the feedback is addressed/responded to accordingly, this should move forward to unblock work.  Please file follow up
Issues for anything we should circle back to before preview/GA.  Thanks!

Issue #867, opened 2021-07-09
composeAndValidate() unexpectedly generates invalid GraphQL when using @tag/@inaccessible


sachindshinde:
`composeAndValidate()` in `@apollo/federation@0.26.0` currently generates a supergraph schema that is invalid GraphQL when given the following subgraph schemas:
- `inventory.graphql`
    ```graphql
    extend type Product @key(fields: "id") {
      id: ID! @external @tag(name: "hi from inventory")
      dimensions: ProductDimension @external
      delivery(zip: String): DeliveryEstimates @requires(fields: "dimensions { size weight }")
    }
    type ProductDimension {
      size: String
      weight: Float @inaccessible
    }
    type DeliveryEstimates {
      estimatedDelivery: String
      fastestDelivery: String
    }
    ```
- `users.graphl`
    ```graphql
    type User @key(fields:"email") {
        email:ID!
        name: String
        totalProductsCreated: Int
    }
    ```
- `products.graphl`
    ```graphql
    type Product @key(fields: "id") @key(fields: "sku package") @key(fields: "sku variation { id }"){
      id: ID!
      sku: String
      package: String
      variation: ProductVariation
      dimensions: ProductDimension
      createdBy: User @provides(fields: "totalProductsCreated")
    }
    type ProductVariation {
      id: ID!
    }
    enum Color {
      BLUE
      GREEN
    }
    type ProductDimension {
      size: String
      weight: Float
    }
    extend type Query {
      allProducts: [Product]
      product(id: ID!): Product
    }
    extend type User @key(fields: "email") {
      email: ID! @external
      totalProductsCreated: Int @external
    }
    ```

The supergraph schema generated is specifically:
```graphql
schema
  @core(feature: "https://specs.apollo.dev/core/v0.1"),
  @core(feature: "https://specs.apollo.dev/join/v0.1"),
  @core(feature: "https://specs.apollo.dev/inaccessible/v0.1")
{
  query: Query
}
directive @core(feature: String!) repeatable on SCHEMA
directive @join__field(graph: join__Graph, requires: join__FieldSet, provides: join__FieldSet) on FIELD_DEFINITION
directive @join__type(graph: join__Graph!, key: join__FieldSet) repeatable on OBJECT | INTERFACE
directive @join__owner(graph: join__Graph!) on OBJECT | INTERFACE
directive @join__graph(name: String!, url: String!) on ENUM_VALUE
directive @inaccessible on FIELD_DEFINITION
enum Color {
[...]

PR #869, opened 2021-07-09
fix(federation): Prevent @tag and @inaccessible usages on fields marked @external

trevor-scheer:
This PR introduces new composition errors to prevent the usage of `@tag` and `@inaccessible` on a field that is marked `@external`. Since `@tag` is part of Studio's API, we're going to block this usage for now to prevent confusion. Similarly, `@inaccessible` usage on `@external` doesn't really mean anything, so allowing it will probably lead to confusion as users may not know what to expect.

This may be relaxed in the future in order to accommodate an actual semantic usage of the combination or possibly just a warning (warnings don't exist yet, but this could be a good candidate for one if we choose to go down that path).

This PR also handles the rename of `appliedDirectives` recommended in feedback on #859.

Fixes: #867


docs/source/errors.md
```diff
@@ -53,6 +53,12 @@ If Apollo Gateway encounters an error, composition fails. This document lists co
 | `REQUIRES_FIELDS_MISSING_ON_BASE` | The `fields` argument of an entity field's `@requires` directive includes a field that is not defined in the entity's originating subgraph.`|
 | `REQUIRES_USED_ON_BASE` | An entity field is marked with `@requires` in the entity's originating subgraph, which is invalid. |
 
+## `@tag`
+
+| Code | Description |
+|---|---|
+| `TAG_USED_WITH_EXTERNAL` | Fields marked as `@external` cannot also have `@tag` usages.  |
```
martijnwalraven, line 60: This seems fine for now, but I wonder if we should generalize this to other directives.

federation-js/src/composition/types.ts
```diff
@@ -39,7 +39,7 @@ export interface FederationField {
   requires?: ReadonlyArray<SelectionNode>;
   provides?: ReadonlyArray<SelectionNode>;
   belongsToValueType?: boolean;
-  appliedDirectives?: DirectiveNode[]
+  otherKnownDirectives?: DirectiveNode[]
```
martijnwalraven, line 42: Note that these are directive usages / applications, not directive definitions. (The naming in `graphql-js` is confusing, but `GraphQLDirective` is a directive definition, while `DirectiveNode` is a usage.)

martijnwalraven (commented):
The logic looks good to me, just a small comment about naming and the difference between directive definitions and usages.

docs/source/errors.md
prasek: @martijnwalraven @trevor-scheer should we add the same handling for`@inaccessible` here as well -- even if we don't generalize just yet?

prasek: also how will we handle `@tag` and `@inaccessible` on value types referenced only from `@external` fields, like https://github.com/apollographql/supergraph-demo/pull/165/files#diff-4a1edface097b589a87522e4fc919429284c343628e3bca83170a6db5e1bc678R9?

trevor-scheer: @prasek Included inaccessible in this PR, see commit 3d51e58

Going to handle value type concerns in a separate PR for the sake of timing.

federation-js/src/composition/types.ts
trevor-scheer: Got it! Went through and clarified the naming a bit further - everything is either named usage or definition accordingly. A bit verbose but helpful I think.

martijnwalraven (approved):
Assuming we'll figure out handling of value types in a follow up PR, this looks good to me!

trevor-scheer:
Overridden by #880

PR #893, opened 2021-07-19
fix(gateway): Prevent inaccessible type names from being leaked in error messages

trevor-scheer:
For reference: https://github.com/apollographql/federation/pull/807#issuecomment-877442278

A type which is marked as `@inaccessible` and implements an interface can trigger runtime/execution errors if a resolver returns that type. That type's name will end up in the `response.errors` array, but it really shouldn't be leaked.

This PR captures all known inaccessible type names and masks them in any response errors that occur during final pass graphql execution.

Closes https://github.com/apollographql/polaris-planning/issues/94

### DIFF
--- a/gateway-js/src/__tests__/executeQueryPlan.test.ts
+++ b/gateway-js/src/__tests__/executeQueryPlan.test.ts
@@ -1357,7 +1357,7 @@ describe('executeQueryPlan', () => {
       expect(response.data?.vehicle).toEqual(null);
       expect(response.errors).toMatchInlineSnapshot(`
         Array [
-          [GraphQLError: Abstract type "Vehicle" was resolve to a type "Car" that does not exist inside schema.],
+          [GraphQLError: Abstract type "Vehicle" was resolve to a type [inaccessible type] that does not exist inside schema.],
         ]
       `);
     });
--- a/gateway-js/src/__tests__/executeQueryPlan.test.ts
+++ b/gateway-js/src/__tests__/executeQueryPlan.test.ts
@@ -1323,5 +1323,43 @@ describe('executeQueryPlan', () => {
         ]
       `);
     });
+
+    it(`doesn't leak @inaccessible typenames in error messages`, async () => {
+      const operationString = `#graphql
+        query {
+          vehicle(id: "1") {
+            id
+          }
+        }
+      `;
+
+      const operationDocument = gql(operationString);
+
+      // Vehicle ID #1 is a "Car" type.
+      // This supergraph marks the "Car" type as inaccessible.
+      schema = buildComposedSchema(superGraphWithInaccessible);
+
+      const operationContext = buildOperationContext({
+        schema,
+        operationDocument,
+      });
+
+      queryPlanner = new QueryPlanner(schema);
+      const queryPlan = queryPlanner.buildQueryPlan(operationContext);
+
+      const response = await executeQueryPlan(
+        queryPlan,
+        serviceMap,
+        buildRequestContext(),
+        operationContext,
+      );
+
+      expect(response.data?.vehicle).toEqual(null);
+      expect(response.errors).toMatchInlineSnapshot(`
+        Array [
+          [GraphQLError: Abstract type "Vehicle" was resolve to a type "Car" that does not exist inside schema.],
+        ]
+      `);
+    });
   });
 });
--- a/gateway-js/src/executeQueryPlan.ts
+++ b/gateway-js/src/executeQueryPlan.ts
@@ -31,6 +31,7 @@ import { deepMerge } from './utilities/deepMerge';
 import { isNotNullOrUndefined } from './utilities/array';
 import { SpanStatusCode } from "@opentelemetry/api";
 import { OpenTelemetrySpanNames, tracer } from "./utilities/opentelemetry";
+import { cleanErrorOfInaccessibleNames } from './utilities/cleanErrorOfInaccessibleNames';
 
 export type ServiceMap = {
   [serviceName: string]: GraphQLDataSource;
@@ -90,7 +91,7 @@ export async function executeQueryPlan<TContext>(
         // the original query.
         // It is also used to allow execution of introspection queries though.
         try {
-          const { schema, removedTypes } = toAPISchema(operationContext.schema);
+          const schema = toAPISchema(operationContext.schema);
           const executionResult = await execute({
             schema,
             document: {
@@ -107,22 +108,10 @@ export async function executeQueryPlan<TContext>(
           });
           data = executionResult.data;
           if (executionResult.errors?.length) {
-            if (removedTypes && removedTypes.size > 0) {
-              const anyInaccessibleTypename = new RegExp(
-                [...removedTypes]
-                  .map((namedType) => '("' + namedType.name + '")')
-                  .join('|'),
-                'g',
-              );
-              executionResult.errors.forEach(
-                (e) =>
-                  (e.message = e.message.replace(
-                    anyInaccessibleTypename,
-                    '[inaccessible type]',
-                  )),
-              );
-            }
-            errors.push(...executionResult.errors);
+            const cleanedErrors = executionResult.errors.map((error) =>
+              cleanErrorOfInaccessibleNames(schema, error),
+            );
+            errors.push(...cleanedErrors);
           }
         } catch (error) {
           span.setStatus({ code:SpanStatusCode.ERROR });
--- a/gateway-js/src/executeQueryPlan.ts
+++ b/gateway-js/src/executeQueryPlan.ts
@@ -90,8 +90,9 @@ export async function executeQueryPlan<TContext>(
         // the original query.
         // It is also used to allow execution of introspection queries though.
         try {
+          const { schema, removedTypes } = toAPISchema(operationContext.schema);
           const executionResult = await execute({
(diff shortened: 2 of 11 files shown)


gateway-js/src/__tests__/executeQueryPlan.test.ts
```diff
@@ -1323,5 +1323,43 @@ describe('executeQueryPlan', () => {
...
+        queryPlan,
+        serviceMap,
+        buildRequestContext(),
+        operationContext,
+      );
+
+      expect(response.data?.vehicle).toEqual(null);
+      expect(response.errors).toMatchInlineSnapshot(`
+        Array [
+          [GraphQLError: Abstract type "Vehicle" was resolve to a type [inaccessible type] that does not exist inside schema.],
```
abernix, line 1360: Can't wait for `graphql-js` v16 (It does.) to fix this grammatical error (`was resolve to`).

gateway-js/src/executeQueryPlan.ts
```diff
@@ -106,7 +107,22 @@ export async function executeQueryPlan<TContext>(
...
+            if (removedTypes && removedTypes.size > 0) {
+              const anyInaccessibleTypename = new RegExp(
+                [...removedTypes]
+                  .map((namedType) => '("' + namedType.name + '")')
+                  .join('|'),
+                'g',
+              );
+              executionResult.errors.forEach(
+                (e) =>
+                  (e.message = e.message.replace(
```
abernix, line 119: Currently, I believe that `graphql-js` can only return this type name a single time in an error message. 
 Should we be defensive for the case that might change by using `String.prototype.replaceAll` here?

query-planner-js/src/composedSchema/removeInaccessibleElements.ts
```diff
@@ -40,38 +41,41 @@ export function removeInaccessibleElements(
...
+
+        return new GraphQLUnionType({
+          ...typeConfig,
+          types: removeInaccessibleTypes(typeConfig.types),
+        });
+      } else {
+        // Keep the type as is.
+        return undefined;
+      }
+    }),
```
abernix, line 76: nit: I think an assignment to a separate `filteredSchema` with the contents of `transformSchema`, and then just `{ schema: filteredSchema, removedTypes: typesToRemove }` could be easier to read and inspect.

query-planner-js/src/composedSchema/toAPISchema.ts
```diff
@@ -14,27 +14,31 @@ declare module 'graphql' {
   }
 }
 
-export function toAPISchema(schema: GraphQLSchema): GraphQLSchema {
+export function toAPISchema(schema: GraphQLSchema): {
```
abernix, line 17: big nit (can be a follow-up as it's internal ~enough): i'm questioning the naming of this function a bit.  

reason: a `toAPISchema` function which returns an object with a `schema` property is less intuitively named to me than `toAPISchema` was.  Maybe just `toAPI`?

abernix, line 17: This wasn't introduced by this PR, but I don't understand why this function is in `@apollo/query-planner`.  Shouldn't it be in `@apollo/federation`?  Seems odd for an implementor to need the query planner for applying such a transformation.

abernix (dismissed):
LGTM.  Left some comments which are non-blocking.

query-planner-js/src/composedSchema/toAPISchema.ts
martijnwalraven: It's part of `@apollo/query-planner` because it was extracted out of `buildComposedSchema`. I agree it doesn't really belong there, but I don't think it belongs in `@apollo/federation` either. The idea was that this was temporary hacky code and we'd move it to a new `@apollo/core` library soon. (We should also move composition out of `@apollo/federation`, to cleanly separate federated subgraph authoring from supergraph composition.)

martijnwalraven (changes requested):

query-planner-js/src/composedSchema/toAPISchema.ts
abernix: Just to clarify, by _internal ~enough_ I was alluding to our cloud's usage.  We're close enough to that work to easily change it without fear.  We can't unnecessarily block this from landing right now.

abernix: Ah, `@apollo/core` does seem like a reasonable consideration.  Happy to leave it here for now.

gateway-js/src/executeQueryPlan.ts
queerviolet: it's already replacing-all, the `'g'` when we construct the regexp makes it a global pattern, calling `.replace` with a global pattern replaces all instances of it

query-planner-js/src/composedSchema/removeInaccessibleElements.ts
trevor-scheer: outdated

query-planner-js/src/composedSchema/toAPISchema.ts
trevor-scheer: outdated

trevor-scheer:
Took another path to the solution via an offline comment from @martijnwalraven which addresses all the concerns on this PR and allows me to run back the changes I made and isolate them to where the errors appear.

We now parse out "Type" or "Type.field" patterns from messages and replace them with a redacted note if they can't be found in the API schema. It's a bit heavier on the regex but does allow us to keep the `toAPISchema` API the same with no workaround and also isolates the error cleaning logic to a single unit-tested function.

gateway-js/src/utilities/cleanErrorOfInaccessibleNames.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  const typeMatches: string[] = [];
+  while ((match = typeRegex.exec(error.message)) !== null) {
+    if (match[1] !== 'resolveType' && match[1] !== 'isTypeOf') {
+      typeMatches.push(match[1]);
+    }
+  }
+  if (typeMatches && typeMatches.length > 0) {
+    const inaccessibleTypes = typeMatches.filter(
+      (match) => !schemaTypes.includes(match),
```
martijnwalraven, line 52: Since `schemaTypes.includes()` in effect performances a linear search on `Object.keys(schema.getTypeMap())`, wouldn't it be easier and more performant to use `schema.getType()` directly?

gateway-js/src/utilities/cleanErrorOfInaccessibleNames.ts
```diff
@@ -0,0 +1,69 @@
...
+  error: GraphQLError,
+): GraphQLError {
+  const typeRegex = /"([_A-Za-z][_0-9A-Za-z]*)"/g;
+  const typeDotFieldRegex =
+    /"([_A-Za-z][_0-9A-Za-z]*)\.([_A-Za-z][_0-9A-Za-z]*)"/g;
+
+  const schemaTypes = Object.keys(schema.getTypeMap());
+  let match: RegExpExecArray | null;
+  const typeDotFieldMatches: [string, string][] = [];
+  while ((match = typeDotFieldRegex.exec(error.message)) !== null) {
```
martijnwalraven, line 14: I'm no JavaScript regexp expert, but I think it should be possible to avoid this two stage approach by passing in a function to `replaceAll`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll#specifying_a_function_as_a_parameter

That way, you don't have to collect the matches first, then filter, and then replace them. Instead, I think you could pass a replacer function to `replaceAll` that would be called for every match. It would perform the same tests as the current code, but return a replacement string directly instead of performing a separate replace pass.

martijnwalraven (approved):
I suspect the implementation could be improved (see comments), but this is already much cleaner. And it also handles inaccessible fields!

gateway-js/src/utilities/cleanErrorOfInaccessibleNames.ts
trevor-scheer: I did that in the fields names bit, not sure why I didn't do it in both places! 👍

gateway-js/src/executeQueryPlan.ts
abernix: Ah, thanks/disregard my suggestion. I totally missed the `g` modifier on the `RegExp` call; I suppose I'm better at reading the `//modifier`-framed syntax.

abernix: Tangentially related nuanced note I just recalled: `replaceAll` throws if you pass it a `RegExp` without the `g` modifier, whereas `replace` accepts either with or without the global modifier.  

But the string values behave according to the function name (_all_ or _not all_).

gateway-js/src/utilities/cleanErrorOfInaccessibleNames.ts
trevor-scheer: Soooo much better!

trevor-scheer: We had to use `replace` instead of `replaceAll` due to node 15+ requirements. This is still usable with the `g` flag on the regex - repeat tests are still passing!

Good catch @martijnwalraven.

PR #900, opened 2021-07-21
fix(gateway): Remove `path`, `query` and `variables` field from subgraph error responses

trevor-scheer:
Queries and variables coming from the gateway down to subgraphs can contain sensitive information that shouldn't be passed back to the client. Currently this can happen when an error occurs during a gateway fetch to a subgraph, and those errors are subsequently returned to the client in the response's error extensions.

We should just remove these fields. There is more discussion to be had around returning error information of this sort at all, but for now this is a sensible first step and prevents leaking potentially immediately sensitive information.


gateway-js/CHANGELOG.md
```diff
@@ -6,7 +6,8 @@
...
 - Introduce support for removing @inaccessible elements from the API schema. [PR #807](https://github.com/apollographql/federation/pull/859)
-- Call `toAPISchema` within the try/catch block in `loadStatic`. [PR 894](https://github.com/apollographql/federation/pull/894)
+- Call `toAPISchema` within the try/catch block in `loadStatic`. [PR #894](https://github.com/apollographql/federation/pull/894)
+- Remove `query` and `variables` from downstream subgraph error responses. [PR #xxx](https://github.com/apollographql/federation/pull/xxx)
```
trevor-scheer, line 10:
```suggestion
- Remove `query` and `variables` from downstream subgraph error responses. [PR #900](https://github.com/apollographql/federation/pull/900)
```

gateway-js/src/executeQueryPlan.ts
```diff
@@ -551,14 +546,8 @@ function flattenResultsAtPath(value: any, path: ResponsePath): any {
...
   serviceName: string,
-  query: string,
-  variables?: Record<string, any>,
 ) {
-  let {
-    message,
-    extensions,
-    path,
-  } = originalError
+  let { message, extensions, path } = originalError;
```
abernix: I wonder if we should just remove `path` too.  In a federated request, the `path` here is still not relevant to the request that the client made.  The [specification](https://spec.graphql.org/draft/#sec-Errors.Error-result-format) states:

> If an error can be associated to a particular field in the GraphQL result, it must contain an entry with the key path that details the path of the response field which experienced the error. This allows clients to identify whether a null result is intentional or caused by a runtime error.

I'm not really sure if a client has any reasonable way to make this determination ("allows clients to identitfy ...") given the `path` is mapped to an internal request that the Gateway makes.

This needs some further thought in the federated model, and particularly if the field was inaccessible and merely used to fulfill a different client request.  This relates to https://github.com/apollographql/federation/issues/354.

Any solution here is out of scope for this issue, so I'm tempted to suggest we just remove `path` here too until we can find a principled solution.

My concrete ask is to remove `path` and offer the same suggestion for `variables` and `query` if it's necessary to be preserved.

gateway-js/CHANGELOG.md
```diff
@@ -6,7 +6,8 @@
...
 - Introduce support for removing @inaccessible elements from the API schema. [PR #807](https://github.com/apollographql/federation/pull/859)
-- Call `toAPISchema` within the try/catch block in `loadStatic`. [PR 894](https://github.com/apollographql/federation/pull/894)
+- Call `toAPISchema` within the try/catch block in `loadStatic`. [PR #894](https://github.com/apollographql/federation/pull/894)
+- Remove `query` and `variables` from downstream subgraph error responses. [PR #900](https://github.com/apollographql/federation/pull/900)
```
trevor-scheer, line 10:
```suggestion
- Remove `query` and `variables` from downstream subgraph error extensions, as well as path from the error itself in the final response. [PR #900](https://github.com/apollographql/federation/pull/900)
```

abernix (approved):

gateway-js/src/executeQueryPlan.ts
trevor-scheer: d9e2322

gateway-js/CHANGELOG.md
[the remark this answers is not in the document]
trevor-scheer: Super, lifted a bunch of this. Thanks!
acaf0b1

PR #909, opened 2021-07-26
Docs for updating `@apollo/gateway`

david-castaneda:
It would be really neat to have a direct hyper-link we can point users to for updating the version of `@apollo/gateway`.

This PR adds a new section under "Setup" for instruction on how to update the NPM package version for `@apollo/gateway`.

### DIFF
--- a/docs/source/gateway.mdx
+++ b/docs/source/gateway.mdx
@@ -63,6 +63,20 @@ As a recommended alternative, you can use the [Rover CLI](https://www.apollograp
 
 For an example of this, see the [Federation quickstart](./quickstart/).
 
+## Updating
+
+To update the `@apollo/gateway` package to the latest version:
+
+```
+npm update @apollo/gateway --latest
+```
+
+Similarly, to update to a specific version:
+
+```
+npm update @apollo/gateway@0.34.0
+```
+
 ## Customizing requests and responses
 
 The gateway can modify the details of an incoming request before executing it across your subgraphs. For example, your subgraphs might all use the same authorization token to associate an incoming request with a particular user. The gateway can add that token to each operation it sends to your subgraphs.


docs/source/gateway.mdx
```diff
@@ -63,6 +63,20 @@ As a recommended alternative, you can use the [Rover CLI](https://www.apollograp
...
+To update the `@apollo/gateway` package to the latest version:
+
+```
+npm update @apollo/gateway --latest
+```
+
+Similarly, to update to a specific version:
+
+```
+npm update @apollo/gateway@0.34.0
```
abernix, line 77: I'm also not sure how providing a `@` range to the `npm update` command behaves in practice (I usually do this with `npm install pkg@version`).

Worth noting that such an update may be constrained by the range rules (e.g., the `^`, `~`, or lacking specifier) in the `package.json`.  See the [example section](https://docs.npmjs.com/cli/v7/commands/npm-update#example) of the `npm update` command.

abernix (changes requested):
Thanks! I'm very supportive of adding instructions on how to update.  I think a very important callout in these instructions is that we point folks to the [CHANGELOG](https://github.com/apollographql/federation/blob/main/gateway-js/CHANGELOG.md) and make sure that they follow the correct _chain_ of upgrade steps.

That means, checking the difference between their current installed version (e.g., `npm view @apollo/gateway`) and the version they're updating to (which may be `latest`, or a specific version) 

I'd like to defer to @StephenBarlow as to the positioning of this instruction within the information architecture, but I'm going to _request changes_ on account of needing instructions for completing the upgrade steps with intent.

Thanks again for opening this!

StephenBarlow:
Merged #910 to resolve this 🙏

PR #910, opened 2021-07-26
Add section for updating gateway library

StephenBarlow:
Expands on #909


trevor-scheer:
Thanks @StephenBarlow, this very nicely resolves a superset of my concerns around providing this kind of guidance!

PR #943, opened 2021-08-05
Add docs for using cache hints with federation

mandiwise:
This PR adds a new Caching page to the federation docs with detailed examples showing expected `cache-control` values when applying cache hints to entities and overriding cache headers from subgraphs in the gateway.

It also creates a new "Performance" section in the federation docs and moves the "Monitoring" page into that section as well (as discussed with @StephenBarlow).


docs/source/performance/caching.mdx
```diff
@@ -0,0 +1,225 @@
+---
+title: Caching
```
StephenBarlow, line 2: Suggest something like "Server-side caching in federated graphs" for just the regular `title`

docs/source/performance/caching.mdx
```diff
@@ -0,0 +1,225 @@
...
+}
+
+directive @cacheControl(
+  maxAge: Int
+  scope: CacheControlScope
+  inheritMaxAge: Boolean
+) on FIELD_DEFINITION | OBJECT | INTERFACE | UNION
+```
+
+The subgraph will calculate and set the cache hint for the response that it sends to the gateway and the gateway will then calculate the cache hint for the overall response based on the most restrictive settings among all of the responses received from the subgraphs involved in query plan execution.
```
StephenBarlow, line 25: The docs use present tense in favor of future tense when they're equivalent in meaning:

"The subgraph will calculate" -> "The subgraph calculates and sets"

"that it sends to the gateway and the gateway will then calculate" -> "that it sends to the gateway, and the gateway then calculates"

Also recommend ending a sentence at "overall response" and starting a new sentence with "This hint is based on"

docs/source/performance/caching.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+Cache hints may also be set [dynamically in subgraph resolvers](https://www.apollographql.com/docs/apollo-server/performance/caching/#in-your-resolvers-dynamic).
```
StephenBarlow, line 27: Docs prefer "can" over "may" for reduced ambiguity

docs/source/performance/caching.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+### Setting entity cache hints
+
+Subgraph schemas contain an `_entities` root field on the `Query` type, so all query plans that require entity resolution will have a [`maxAge` of `0` set by default](https://www.apollographql.com/docs/apollo-server/performance/caching/#default-maxagee). To override this default behavior, you can add a `@cacheControl` directive to an entity's definition:
```
StephenBarlow, line 31: "will have" -> "have"

docs/source/performance/caching.mdx
```diff
@@ -0,0 +1,225 @@
...
+
+If a subgraph does not specify a `max-age`, the gateway will assume its response (and in turn, the overall response) cannot be cached. To override this behavior, you can set the `cache-control` header in the `didReceiveResponse` method of a `RemoteGraphQLDataSource`.
+
+Additionally, if the gateway should ignore `cache-control` response headers from subgraphs that will affect the operation's cache policy, then you can set the `honorSubgraphCacheControlHeader` property of a `RemoteGraphQLDataSource` to `false` (this value is `true` by default).
+
+The effect of setting `honorSubgraphCacheControlHeader` to `false` is to have no impact on the cacheability of the response in either direction. In other words, this property won’t determine whether the response can be cached, but it does exclude a subgraph's `cache-control` header from consideration in the gateway's calculation. If all subgraphs are excluded from consideration when calculating the overall `cache-control` header, the response sent to the client will not be cached.
+
+There are detailed examples below that illustrate the resulting behavior when overriding this header for a subgraph.
+
+## Example maxAge calculations
```
StephenBarlow, line 54: codefont `maxAge`

docs/source/performance/caching.mdx
```diff
@@ -0,0 +1,225 @@
...
+    }
+  }
+}
+```
+
+The response will be considered **not cacheable**.
+
+The `cache-control` header in the response from the astronauts subgraph will be `max-age=20, public` based on the cache hint applied to the `Astronaut` type, which will be taken into consideration when this subgraph responds to the `_entities` query from the gateway.
+
+The missions subgraph, however, will not include a `cache-control` header in its response to the gateway, thus indicating that it is not cacheable, and in turn, making the entire response sent to the client not cacheable. At first glance, this behavior may seem unexpected because the `@cacheControl(maxAge: 10)` directive is applied to the `mission` root query field. But upon closer inspection, we see that this field returns the `Mission` type which contains a non-scalar `Astronaut` field and this type will have a [default `maxAge` of `0` applied](https://www.apollographql.com/docs/apollo-server/performance/caching/#default-maxage).
```
StephenBarlow, line 109: "may seem" -> "might seem"

StephenBarlow: "will not include" -> "does not include"

"thus indicating that it is not cacheable, and in turn" -> "thus indicating that it is not cacheable. Consequently, the entire response sent to the client is not cacheable."

docs/source/performance/caching.mdx
```diff
@@ -0,0 +1,225 @@
...
+The second option is to set an `@cacheControl` directive for the `Astronaut` entity extension with the appropriate `maxAge` value instead:
+
+```graphql{1}:title=Missions%20Subgraph
+extend type Astronaut @key(fields: "id") @cacheControl(maxAge: 20) {
+  id: ID! @external
+  missions: [Mission]
+}
+```
+
+Setting the `maxAge` property on the extended `Astronaut` type in the missions subgraph raises an important consideration about whether subgraphs should explicitly set these values for any entity types that they extend. In most cases, it will likely be preferable to apply `@cacheControl(inheritMaxAge: true)` wherever an extended entity type is used as a field's return type to avoid ambiguity. Depending on your requirements, you may wish to disallow this entirely using custom linting for your subgraphs to ensure that entity type extensions do not set the `maxAge` or `scope` arguments when applying the `@cacheControl` directive.
```
StephenBarlow, line 134: "it will likely be preferable" -> "it's likely preferable"

"may wish" -> "might want"

StephenBarlow (approved):
Looks good with some suggestions and one thing about redirects I'll ping you about!

mandiwise:
@StephenBarlow All requested edits have now been made. Thanks for your feedback on this.

StephenBarlow (approved):

Issue #473, opened 2021-02-23
Missing `name` property in the `ApolloGateway` `serviceList` configuration resolves with "unreachable" error and wasm stacktrace.


spencerarogers:
## Summary

Opening per the request in [#439](https://github.com/apollographql/federation/issues/439#issuecomment-783367151).

When initializing `ApolloGateway` and the `serviceList` array contains an object missing a `name` property, the gateway will start up and load the implementing service's schema properly but any queries that hit this service will result in an error.

Invalid config:
```js
const gateway = new ApolloGateway({
  // NOTE: Here the we have a service with a valid `url` but a missing `name` property.
  serviceList: [{"not-name":"my-service", "url":"http://localhost:4001/graphql"}]
});
```

The stacktrace returned by the gateway server looks like this:

```json
{
  "errors": [
    {
      "message": "unreachable",
      "extensions": {
        "code": "INTERNAL_SERVER_ERROR",
        "exception": {
          "stacktrace": [
            "RuntimeError: unreachable",
            "    at <anonymous>:wasm-function[525]:0xa165d",
            "    at <anonymous>:wasm-function[583]:0xa403a",
            "    at <anonymous>:wasm-function[718]:0xa6e56",
            "    at <anonymous>:wasm-function[688]:0xa6828",
            "    at <anonymous>:wasm-function[720]:0xa6eba",
            "    at <anonymous>:wasm-function[654]:0xa5df7",
            "    at <anonymous>:wasm-function[347]:0x934a7",
            "    at <anonymous>:wasm-function[49]:0x5288f",
            "    at <anonymous>:wasm-function[31]:0x45fa7",
            "    at <anonymous>:wasm-function[561]:0xa326f"
          ]
        }
      }
    }
  ]
}
```
## Reproduction

Here is a simple working reproduction of the issue described above with instructions in the README.md:

https://github.com/spencerarogers/apollo-federation-439-repro

Then in this `multiple-implementing-services` branch there is another reproduction in which there is 1 incorrectly configured service and 1 correctly configured service. This illustrates that the gateway can continue to function for properly configured services but error whenever a query is sent to the incorrectly configured implementing service:

https://github.com/spencerarogers/apollo-federation-439-repro/tree/multiple-implementing-services

abernix:
@spencerarogers Is this still happening for you on the latest version?

spencerarogers:
[...]

PR #955, opened 2021-08-11
Throw descriptive error if service in serviceList config does not have a 'name'.

spencerarogers:
This PR adds a nicer error for the condition in which a service in the `serviceList` config is missing a 'name' property when the gateway is trying to fetch service definitions at initialization. Currently the gateway prints a one line error when this is encountered:

`This data graph is missing a valid configuration. Cannot read property 'replace' of undefined`

@abernix mentioned [here](https://github.com/apollographql/federation/issues/473#issuecomment-886564559) in #473 that it would be nice if the error handling was improved.

This test and implementation closely mirror the way it is already being handled for the service 'url' property.

I also tested this out locally against [the bug reproduction I had created](https://github.com/spencerarogers/apollo-federation-439-repro) for #473 and it worked as expected.

### DIFF
--- a/gateway-js/src/__tests__/loadServicesFromRemoteEndpoint.test.ts
+++ b/gateway-js/src/__tests__/loadServicesFromRemoteEndpoint.test.ts
@@ -17,6 +17,23 @@ describe('getServiceDefinitionsFromRemoteEndpoint', () => {
     );
   });
 
+  it("errors when no name was specified", async () => {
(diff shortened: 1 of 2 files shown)


spencerarogers:
I'm sorry for the distraction, somehow I thought my tests were passing locally.

I now don't think this PR actually even addresses the issue of the `undefined` reference error. I'll close this and open a new PR if I can.

PR #960, opened 2021-08-15
makeRequest method to modify url and method

yawningphantom:
This PR is linked to this issue #901 
This PR enables us to modify the url and the http method for every request. The url change is was easy but to enable the method change I had to change the sendRequest method as it is always sending the body to a GET request as well which throws up a type error `TypeError: Request with GET/HEAD method cannot have body` [here](https://github.com/apollographql/federation/blob/44eebf9d47db76052df4f9edeeb1e42278167615/gateway-js/src/datasources/RemoteGraphQLDataSource.ts#L150)
I would recommend that we call the makeRequest method inside the sendRequest itself so that we can create the Request Object dynamically and this can be modified

### DIFF
--- a/gateway-js/src/datasources/RemoteGraphQLDataSource.ts
+++ b/gateway-js/src/datasources/RemoteGraphQLDataSource.ts
@@ -15,6 +15,7 @@ import {
   Headers,
   Response,
 } from 'apollo-server-env';
+import url from 'url';
 import { isObject } from '../utilities/predicates';
 import { GraphQLDataSource } from './types';
 import createSHA from 'apollo-server-core/dist/utils/createSHA';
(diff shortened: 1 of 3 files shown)


gateway-js/src/datasources/RemoteGraphQLDataSource.ts
```diff
@@ -206,6 +205,23 @@ export class RemoteGraphQLDataSource<
     }
   }
 
+  public makeRequest(
```
Cretezy, line 208: `makeRequest` sounds like this function executes the request. Maybe `createRequest` or `setupRequest` might be a better name here.

Cretezy (approved):
Overall LGTM :+1:

gateway-js/src/datasources/RemoteGraphQLDataSource.ts
yawningphantom: @Cretezy setupRequest sound better. If we all agree, I can change this.

yawningphantom:
@trevor-scheer any update on this PR?

PR #962, opened 2021-08-16
Remove root redirect in docs

trevorblades:
This branch removes the root level redirects in the docs, previously used to send our root-level deploy preview links to the right place when clicked on in the Netlify GitHub check. Unfortunately, it seems like these redirects are causing issues with Netlify when we use split tests, causing users to be redirected from our home page to the docs.

This change is an attempt to fix that, at the known cost of having to manually add "/docs" or "/docs/studio" to the end of deploy previews when previewing changes.

I'm going to investigate options to conditionally add a root redirect to deploy previews only but we want to unblock split testing first, so I made this a separate PR.

In the meantime, we can use a tag like the one below to tell Netlify what path to point deploy preview links in GitHub to.

@netlify /docs/federation

### DIFF
--- a/docs/static/_redirects
+++ b/docs/static/_redirects
@@ -1,5 +1,3 @@
-/ /docs/federation/
-
 /docs/federation/core-concepts/ /docs/federation/implementing-services/
 /docs/federation/implementing-services/ /docs/federation/subgraphs/
 /docs/federation/managed-federation/advanced-topics/ /docs/federation/managed-federation/overview/


abernix:
> In the meantime, we can use a tag like the one below to tell Netlify what path to point deploy preview links in GitHub to.
> 
> @netlify /docs/federation

Whoa, what? This works?! That's cool!

abernix:
@trevorblades This gets back to one of my earliest questions about the Gatsby setup and one of the ways it departed from the patterns we used in Hexo.  I don't quite understand why we can't just have the docs rendered at `/` in the previews and mounted at a particular URL later on.  For example, we did this in the specs repositories by merely using relative links.  That seems like it could get rid of a number of interesting things (including in our build steps where we have to _create_ that structure).

Anyhow, we should take this conversation out of this issue.  I'll approve.

abernix (approved):

trevorblades:
@abernix you're totally right. I just configured our community docs to build deploy previews at `/` and it worked perfectly. I'm gonna roll out some PRs to the other repos that does the same. Can't believe I didn't think about this earlier, but thank you for the discourse :)

PR #964, opened 2021-08-18
Deploy deploy previews at /

trevorblades:
This branch tells netlify to deploy our docs deploy previews at / instead of nesting them in the location we expect them to be in prod. This was inspired by https://github.com/apollographql/federation/pull/962#issuecomment-900239383. Thanks @abernix 🙏

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -9,3 +9,5 @@
 [build.environment]
   NODE_VERSION = "14"
   NPM_VERSION  = "7"
+[context.deploy-preview]
+  command = "gatsby build"


abernix (approved):
Word!  This worked so easily, that I'm somewhat curious what breaks if we removed it from the main `build` step, but this LGTM for deploy previews for sure!

abernix:
Thanks for looking into this!

trevorblades:
> I'm somewhat curious what breaks if we removed it from the main `build` step, but this LGTM for deploy previews for sure!

I was thinking the same thing. In fact, it seems like devhub is one web property where we don't do this "copy files into the right folder structure" dance at all, and it works perfectly fine. I'm going to run a test with the community docs today trying out this method and will update you with my results!

abernix:
@trevorblades if it does break, perhaps it was a configuration which caused the  _broken link checker_ to malfunction at one point? (but perhaps it's smarter now? not sure how that's wired up.)

trevorblades:
I don't think the link checker would be affected by the change. Here's an example of the changes I just made to the community docs:

Docs: https://github.com/apollographql/community/commit/427c258cb987f07b04f4ec452a413ca83923bd78
Router: https://github.com/apollographql/website-router/commit/48009206f6f075229a1e2198804a4bb5104e06fd

And it works! https://www.apollographql.com/docs/community/

abernix:
@trevor-scheer Sweet!

abernix:
Apologies @trevor-scheer, I meant @trevorblades!

PR #753, opened 2021-05-12
Add strawberry-graphql to federation libraries

marcoacierno:
Strawberry (https://strawberry.rocks / https://github.com/strawberry-graphql/strawberry) is a Python library that has support for federation so I think it makes sense to include it in the list of frameworks that support it

Example of usage: https://github.com/strawberry-graphql/strawberry/blob/main/CHANGELOG.md#0260---2020-05-21

PR #985, opened 2021-08-26
Add absinthe_federation to the list

kdawgwilk:
Not sure if there is a way to indicate if a library is in an alpha/beta stage but I would like to get this on the list to let see some more usage so we can start getting more feedback on it


prasek:
Hi @kdawgwilk thanks for opening this! 👋

We've recently started https://github.com/apollographql/apollo-federation-subgraph-compatibility to verify the level of support in each federation subgraph framework/library
- suggest checking that out and opening a PR over there for `absinthe_federation`
- also checkout https://github.com/apollographql/federation/pull/753#issuecomment-909619448 for an example

Looking forward to getting this added and we'll figure out a way to get alpha/beta listed. 👍

PR #988, opened 2021-08-28
Fix path for errors raised for federated nodes

lndbaryshnikov:
This is a solution for the problem described in issue https://github.com/apollographql/federation/issues/354.

In our project, we needed error paths to be correct in relation to the final api schema for the reasons described in the above issue (each service in the gateway can raise an error, for example, access error based on user permissions so we have to be able to map received errors to data).

This solution seems questionable in places, and I would be glad if you could help me improve it. The solution is already used in some of our projects and has not caused any particular problems yet.

The code in this PR solves 2 problems:

1. It fixes path for errors raised for federated nodes, as described above (it composes correct error path by combining the path to the federated entity and the path of the original error);
2. It generates errors with correct path for each federated entity in case of a network error occurs (the idea is that on the client we don't care about the federation and we want to know why we did not receive each specific field)

And I don't really know how to correctly test the case when one of the services is unavailable so this has not been covered by tests yet, I hope you'll help me with this.


col:
I'd really like to see this merged as I'm currently facing the same issue. 

@lndbaryshnikov Do you have any issue signing the CLA?

lndbaryshnikov:
> I'd really like to see this merged as I'm currently facing the same issue.
> 
> @lndbaryshnikov Do you have any issue signing the CLA?

@col no problems at all, I've signed the CLA
waiting for the review only

PR #991, opened 2021-08-31
doc: remove subscription disabling

dnalborczyk:


glasser (approved):
Makes sense — we already removed the code line that this referred to. Thanks!

PR #1000, opened 2021-09-02
fix(composition): Remove composition responsibilities from supergraph printer

trevor-scheer:
Push schema construction responsibilities up and out of the supergraph printer and into composition where they belong.

* Update tests (sort related)
* Fix incorrect test for printFederatedSchema (exposed by an unexpected test update)

I encountered these issues during implementation for #996 and figured this work was a separate issue and worth breaking out. #996 will be based against this branch.

### DIFF
--- a/federation-js/src/service/printSubgraphSchema.ts
+++ b/federation-js/src/service/printSubgraphSchema.ts
@@ -37,9 +37,10 @@ import { isFederationType } from '../types';
 import { isApolloTypeSystemDirective } from '../composition/utils';
 import { federationDirectives, gatherDirectives } from '../directives';
 
-export function printSubgraphSchema(schema: GraphQLSchema): string {
-  return printSchema(schema);
-}
+/**
+ * @deprecated Use `printSubgraphSchema` instead.
+ */
+export const printSchema = printSubgraphSchema;
 
 type Options = {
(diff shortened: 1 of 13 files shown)


federation-js/src/service/printSubgraphSchema.ts
```diff
@@ -54,7 +58,7 @@ type Options = {
  *
  *    - commentDescriptions:
  *        Provide true to use preceding comments as the description.
- *
+ * @deprecated Use `printSubgraphSchema` instead.
  */
 export function printSchema(schema: GraphQLSchema, options?: Options): string {
```
martijnwalraven, line 63: Maybe turn this around: make `printSubgraphSchema` the main function, and alias it as `printSchema` for backwards compatibility?

martijnwalraven (approved):
Good catch! Moving responsibilities out of printing nicely cleans things up.

federation-js/src/service/printSubgraphSchema.ts
trevor-scheer: Ah, yes of course 😄

trevor-scheer: 5227ad5

PR #1008, opened 2021-09-07
fix(federation): Composition propagates @deprecated usages on input type object fields

trevor-scheer:
Follow-up to #996

In landing #996, @sachindshinde discovered that `@deprecated` propagation failed specifically for fields on input type objects.

We discovered this was an issue in `graphql-js` itself which has since been landed and released in `v15.5.3`
https://github.com/graphql/graphql-js/pull/3260

This PR pushes our `graphql` `peerDependencies` to `latest` and as such is a breaking change for gateway users, requiring a (relatively small) update to their project's `graphql` dependency from `15.4.0`. 

This is a breaking change for the federation, query-planner, and gateway packages and should be landed to a release branch.

TODO
- [x] Changelog entry

### DIFF
--- a/federation-integration-testsuite-js/src/fixtures/reviews.ts
+++ b/federation-integration-testsuite-js/src/fixtures/reviews.ts
@@ -74,8 +74,14 @@ export const typeDefs = gql`
     retailPrice: String @requires(fields: "price")
   }
 
+  input ReviewProduct {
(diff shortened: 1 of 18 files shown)


abernix (approved):
LGTM assuming you introduce a CHANGELOG.md!

PR #1013, opened 2021-09-10
fix(query-planner): Remove unions from API schema which have only @inaccessible types

trevor-scheer:
`removeInaccessibleElements` currently leaves unions in the schema even if all of their types are inaccessible. This change removes them from the schema.

Note that unions can consist of unions, so this action of marking unions for removal must be repeated until none are found.

### DIFF
--- a/query-planner-js/src/composedSchema/__tests__/removeInaccessibleElements.test.ts
+++ b/query-planner-js/src/composedSchema/__tests__/removeInaccessibleElements.test.ts
@@ -1,4 +1,9 @@
-import { buildSchema, assertValidSchema, GraphQLObjectType } from 'graphql';
+import {
+  buildSchema,
+  assertValidSchema,
+  GraphQLObjectType,
+  GraphQLUnionType,
+} from 'graphql';
 import { removeInaccessibleElements } from '../removeInaccessibleElements';
 
 describe('removeInaccessibleElements', () => {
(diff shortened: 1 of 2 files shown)


query-planner-js/src/composedSchema/__tests__/removeInaccessibleElements.test.ts
```diff
@@ -148,6 +153,130 @@ describe('removeInaccessibleElements', () => {
...
+      type Baz {
+        anotherField: String
+      }
+    `);
+
+    schema = removeInaccessibleElements(schema);
+
+    const fooType = schema.getType('Foo') as GraphQLUnionType;
+    expect(fooType).toBeDefined();
+    expect(fooType.getTypes()).toHaveLength(1);
```
martijnwalraven, line 193: You may want to test for the right types more explicitly, perhaps `expect(fooType.getTypes().map(t => t.name)).toEqual(['Baz'])`?

query-planner-js/src/composedSchema/removeInaccessibleElements.ts
```diff
@@ -38,6 +39,27 @@ export function removeInaccessibleElements(
     }),
   );
 
+  // remove unions which have no accessible types
+  // this is a loop since unions can contain other unions
+  let shouldRevisitSchema = true;
```
martijnwalraven, line 44: `shouldRevisitSchema` seems a bit generic as a variable name, because it's only meant to affect this particular loop. Perhaps something like `hasInaccessibleUnionMembers`, and then changing this to a `do ... while` loop?

Alternatively, perhaps you could extract all of this into a separate function, like `removeInaccessibleUnionMembers`?

martijnwalraven (approved):
The logic seems sound, so I approved the PR. But I left some comments that might help clean up the structure.

sachindshinde (changes requested):
This function changes `toAPISchema()` and tries to do some rather complex propagation logic, please don't merge these changes in until talking with Gravity. There's a Slack thread at https://apollograph.slack.com/archives/C01JTQYPRQ9/p1632857601091000?thread_ts=1632831396.082600&cid=C01JTQYPRQ9

abernix (changes requested):
I haven't reviewed anything, just stepping in to apply a "Request changes" to make sure we hold off on merging this on account of @sachindshinde's request in  https://github.com/apollographql/federation/pull/1013#pullrequestreview-766772712 😸

trevor-scheer:
10-4. This is an incomplete solution at what the conversation has more thoroughly fleshed out as well. I will keep it around for now but have no intention of merging.

trevor-scheer:
Closing based on conversations we had today, plus I'm not really convinced the partial implementation here is even helpful in the future algorithm we hope to implement.

PR #1019, opened 2021-09-16
Describe subgraph query fields

StephenBarlow:


docs/source/subgraphs.mdx
```diff
@@ -156,6 +156,85 @@ server.listen(4001).then(({ url }) => {
...
+
+When you [generate your subgraph schema](#generating-a-subgraph-schema), some federation-specific definitions are automatically added to it. In addition to directive definitions like `@key`, the most useful of these definitions for debugging are two fields of the `Query` type: `_service` and `_entities`:
+
+```graphql
+type Query {
+  # ...your field definitions...
+
+  # Added automatically
+  _service: _Service!
+  _entities(representations: [_Any!]!): [_Entity]!
```
abernix, line 179: Should we also briefly explain that the `_Entity` type is an automatically added `union` of all entity types?

```graphql
union _Entity = Review | User | Product # ... more!
```

docs/source/subgraphs.mdx
```diff
@@ -156,6 +156,85 @@ server.listen(4001).then(({ url }) => {
...
+query GetSubgraphSchema {
+  _service {
+    sdl
+  }
+}
+```
+
+The `sdl` field returns your subgraph's schema as an SDL string. This field has a couple of important differences from a standard [introspection query](https://graphql.org/learn/introspection/) that a tool like Apollo Sandbox uses:
+
+* Unlike introspection, the `sdl` field is _not_ disabled by default in production environments.
```
abernix, line 197: ... I have a temptation here to explain that this is safe on account of the subgraph being secured (ref: above).

docs/source/subgraphs.mdx
```diff
@@ -156,6 +156,85 @@ server.listen(4001).then(({ url }) => {
...
+    }
+  ]
+}
+```
+
+Whenever one subgraph references _another_ subgraph's entity, it uses an **entity representation** to do so. An entity representation is an object that includes _only_ the entity's `__typename` and the fields in the entity's `@key` (just `id` in this case).
+
+#### Using in tests and debugging
+
+If you're writing integration tests for your subgraph, you can test the return value of the `_entities` field for various entity representations that your _other_ subgraphs use.
```
abernix, line 234: Great suggestion! 🔥

abernix (approved):
LG(reat)TM! Just a couple suggestions within. Thanks, @StephenBarlow!

PR #1022, opened 2021-09-21
feat(composition): Relax @tag definition validation

### LINKED ISSUE #1021: feat(composition): Relax `@tag` directive definition validation
Users should be allowed to provide a `@tag` directive definition which is a "subset" of the spec'ed @tag directive. i.e. the validation should permit leaving out `repeatable` or any of the directive `locations`. But it should continue to enforce the non-null arg `name: String!`.

This is in response to [a lack of `repeatable` support in Kotlin](https://github.com/ExpediaGroup/graphql-kotlin/issues/590), but also seems like a reasonable pattern to follow for core directives going forward.

trevor-scheer:
Fixes #1021

### DIFF
--- a/federation-js/CHANGELOG.md
+++ b/federation-js/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 > The changes noted within this `vNEXT` section have not been released yet.  New PRs and commits which introduce changes should include an entry in this `vNEXT` section as part of their development.  When a release is being prepared, a new header will be (manually) created below and the appropriate changes within that release will be moved into the new section.
 
-- _Nothing yet! Stay tuned!_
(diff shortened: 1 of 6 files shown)


martijnwalraven (commented):
I don't know if I have all the context, but implementing an immediate workaround for the lack of `repeatable` support in Kotlin definitely makes sense to me. I also think we want to be more lenient in allowing differences between directive definitions in general. I'd prefer if we came up with a design for that first though, and think through the consequences to make sure this is something we feel comfortable supporting for all directives.

Leaving out `repeatable` has no effect on the semantics of directive applications (except that there would never be more than one tag per element for that subgraph).

Additional arguments are a somewhat different case however, because we have no way of knowing what a user's intention is. So just ignoring them might be surprising or even dangerous. We may also want to differentiate between nullable and non-nullable arguments, or take default values into account.

martijnwalraven (approved):
Looks good to me!

PR #1029, opened 2021-09-27
chore(harmonizer): Rebuild bridge.js when relevant

o0Ignition0o:
This commit allows us to run npm install and npm run compile:for-harmonizer-build-rs if and only if file or folders have in `federation-js` have been modified since the last time `dist/bridge.js` has been generated.

Bonus point we get a warning when working on the federation, that allows to know whether the bridge has been generated or not, thus making sure we're working against an up to date bridge and definitions:

When the bridge is up to date:
image

When the bridge is not up to date:
image

This warning can be displayed when working downstream using `--verbose` twice as explained [in the cargo build display options](https://doc.rust-lang.org/cargo/commands/cargo-build.html#display-options).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -199,6 +199,7 @@ dependencies = [
  "serde",
  "serde_json",
  "thiserror",
+ "which",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


EverlastingBugstopper (approved):
this is great! should take some of the guesswork out of working in here 😄

PR #1037, opened 2021-09-29
chore: Auto-update versions during publish

trevor-scheer:
Lerna fails to update the top-level `package-lock.json` during a
publish since npm@7. Running `npm i` during the `version` lifecycle hook
provided by `lerna` will trigger a natural update of the `package-lock.json` file
that we'd like to happen.

This also ensures that Harmonizer's `Cargo.toml` stays in sync with
its respective `package.json` version, as well as updates the
top-level `Cargo.lock` in the monorepo.

Related: https://github.com/lerna/lerna/issues/2832


harmonizer/Cargo.toml
```diff
@@ -1,26 +1,26 @@
...
 anyhow = "1.0.39"
 deno_core = "0.86.0"
-serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0.60"
 thiserror = "1.0.23"
 
+  [dependencies.serde]
+  version = "1.0"
+  features = [ "derive" ]
+
```
abernix, line 23: 🤔Looks like the JS TOML lib eats comments for breakfast (not surprising) and prefers the dot notation for nested hash maps.  Luckily our only comments in this file right now is totally unnecessary boilerplate￼. 

Let's definitly try to this but I feel like this could end up being a tooling battle.

package.json
```diff
@@ -12,6 +12,7 @@
     "compile": "npm run compile:stage-01 && npm run compile:stage-02 && npm run compile:stage-03",
     "compile:clean": "npm run compile:stage-01 -- --clean && npm run compile:stage-02 && npm run compile:stage-03 -- --clean",
     "watch": "npm run compile && tsc --build tsconfig.json --watch",
+    "version": "npm i",
```
abernix, line 15: I understand why the re-install on `version` (which is heavy _-ish_) bump solves the out of sync `package-lock-only.json` in a way that doesn't require manipulating it ourselves but does this still work if we merely do `npm install --package-lock-only`?  

Seems a bit lighter weight?

If not, all good.

abernix (approved):
Nice! Left a couple comments within but nothing blocking.  Let's see how this goes!

package.json
trevor-scheer: 💯

harmonizer/Cargo.toml
trevor-scheer: Yeah, I included the changes we can expect from toml parser to make sure this didn't go unnoticed. I agree it's a bit of a bummer.

On that note, the library I've selected is current-ish but hasn't seen much love in the last 6 months. Could be fun to write a TOML parser/printer 😄

trevor-scheer:
@abernix this got me thinking a little extra. What about a little CLI prompt mid-version that asks the publisher to update the changelog (and maybe points to instructions for doing so?). That would get changelog updates batched into the release commit as well, which would be nice.

Something as simple as a "press enter when done" kinda prompt.
https://www.npmjs.com/package/prompt-confirm

abernix:
@trevor-scheer Reasonable consideration, for sure!

PR #1039, opened 2021-09-29
Who is Apollo README updates

hwillson:
Small updates to align "Who is Apollo" sections across repos.


trevor-scheer:
@hwillson I appreciate the visibility of this section, just slight concern that it might fall out of sync with the other repos. Would be ideal if we could have it live in one place, though I'm not sure there's a good solution. Still LGTM though!

trevor-scheer (approved):

hwillson:
@trevor-scheer I agree; I couldn't think of a good way to share it from one common location and still have it inlined in our main repo README's. That would be a great GitHub feature!

PR #1058, opened 2021-10-06
chore: Split `@apollo/subgraph` package from `@apollo/federation`

trevor-scheer:
This PR separates the concerns of running a subgraph from composition, which the `@apollo/federation` package is currently responsible for both of. The new `@apollo/subgraph` package will be used in the same exact manner for subgraph developers.

There were some interdependent types, functions, etc. within the federation package which I had to separate to one or the other. Given that the `@apollo/federation` already had an implicit dependency on this new package (for forward-compat), I chose to migrate the few bits into the new package and deeply import them from `@apollo/federation`. I'm not particularly interested in exposing all of those common bits to the minimal public API of `@apollo/subgraph`, hence the nested import.

Other notable updates:
* Docs
* Monorepo infra (TS references, Jest, Lerna)
* Respective READMEs

TODO
- [x] New CHANGELOG entry should point back to old CHANGELOG

### DIFF
--- a/subgraph-js/CHANGELOG.md
+++ b/subgraph-js/CHANGELOG.md
@@ -4,5 +4,5 @@
 
 > The changes noted within this `vNEXT` section have not been released yet.  New PRs and commits which introduce changes should include an entry in this `vNEXT` section as part of their development.  When a release is being prepared, a new header will be (manually) created below and the appropriate changes within that release will be moved into the new section.
 
-- Initial release of new `@apollo/subgraph` package. This package is the subgraph-related slice of the `@apollo/federation` package which was previously responsible for both subgraph and composition bits. `@apollo/federation` users will experience no change in behavior for now, but our docs will suggest the usage of the `@apollo/subgraph` package going forward. [PR #1058](https://github.com/apollographql/federation/pull/1058)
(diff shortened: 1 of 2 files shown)


abernix:
Responding to @martijnwalraven's comments:

> My main comment is that it may actually be trying to do too much, because it's keeping the composition bits in `@apollo/federation` and makes sure those are still working.

I think the PR tried to actively avoid conflicts with other work in flight.

> That makes sense if we want to decouple the introduction of `@apollo/subgraph` from the Fed 2 release

We absolutely do and that was an explicit goal of handing this off.

> it seems like a lot of extra work for something that will be short lived

Perhaps, but it's done now, already and happened within a day. 😄   (And functionally speaking, we've just shared and spread out the workload rather than trying to just do everything at once.)

> I also think we should ask @pcmanus to review this, so we don't run into surprise merge conflicts later.

Definitely would support this if @pcmanus has time to afford to this, my quick smoke test says there are no conflicts that couldn't be resolved in a few seconds. (it's like, array ordering of the package members `lerna.json` and some `jest` versions in `package.json`).

martijnwalraven:
> Out of curiosity, I do see how it's keeping composition in `@apollo/federation` but I'm less what "and makes sure those are still working" refers to? I feel the main important thing here is that the new `@apollo/subgraphs` module does not depend on `@apollo/federation` but afaict that's not the case.

@pcmanus It's mostly about fixing dependencies that (after the split) go the other way, see https://github.com/apollographql/federation/pull/1058/files#r724083941 for example.

PR #1113, opened 2021-10-26
feat: Query planner skips fetches when possible (based on `@skip` and `@include` usages

trevor-scheer:
It's reasonable for the query planner to make use of top-level `@skip` and `@include` usages in order to determine whether it can omit/ignore a `FetchNode` altogether and prevent an unnecessary call to a service when possible.

This PR implements this behavior. `FetchNode`s are appended with the sum of all "inclusion conditions" of the top level fields. Each condition has a possible `skip` and `include` value, and those values can be either a `Boolean` literal (`true`/`false`) or a variable.

During execution, the executor will reduce all of the conditions based on literal and variable values into a single "skip or include" result, and will execute or ignore that `FetchNode` based on the result.

A number of edge cases were considered before and during the implementation of this improvement. The most notable:
1. Using `@skip` and `@include` on the same selection (always skip unless `skip: false` AND` include: true`). This adheres to the spec.
2. `FragmentSpread` and `InlineFragment` cases. This behavior is partially implemented. There are test cases (skipped) to demonstrate both ideal behavior and existing limitations of this improvement. Resolving these limitations requires some other, unrelated changes to the QP impl. details which I consider to be out of scope for this change. These cases are also less likely to occur and I find the current changeset to be large enough to defer solving this edge case for a later time if/when it's deemed useful/necessary.

### DIFF
--- a/gateway-js/CHANGELOG.md
+++ b/gateway-js/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 > The changes noted within this `vNEXT` section have not been released yet.  New PRs and commits which introduce changes should include an entry in this `vNEXT` section as part of their development.  When a release is being prepared, a new header will be (manually) created below and the appropriate changes within that release will be moved into the new section.
 
-- _Nothing yet. Stay tuned._
+- Based on additions to the query plan format. `FetchNode`s may now contain a property `inclusionConditions` which informs the executor whether or not a `FetchNode` can be skipped entirely at runtime. Inclusion conditions may consist of literals and variables, so must be resolved at execution time rather than during query planning. [PR #1113](https://github.com/apollographql/federation/pull/1113)
 
 ## v0.42.1
 
--- a/gateway-js/CHANGELOG.md
+++ b/gateway-js/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 > The changes noted within this `vNEXT` section have not been released yet.  New PRs and commits which introduce changes should include an entry in this `vNEXT` section as part of their development.  When a release is being prepared, a new header will be (manually) created below and the appropriate changes within that release will be moved into the new section.
 
-- Based on additions to the query plan format. `FetchNode`s may now contain a property `inclusionConditions` which informs the executor whether or not a `FetchNode` can be skipped entirely at runtime. Inclusion conditions may consist of literals and variables, so must be resolved at execution time rather than during query planning. [PR #1113](https://github.com/apollographql/federation/pull/1113)
+- Skip fetches when possible (based on @skip and @include usages). The query planner now aggregates top-level skip and include usages, which allows the gateway to skip `FetchNode`s altogether when possible during execution. [PR #1113](https://github.com/apollographql/federation/pull/1113)
 
 ## v0.42.1
 
--- a/gateway-js/src/executeQueryPlan.ts
+++ b/gateway-js/src/executeQueryPlan.ts
@@ -1,6 +1,7 @@
 import {
   GraphQLExecutionResult,
   GraphQLRequestContext,
+  VariableValues,
 } from 'apollo-server-types';
 import { Headers } from 'apollo-server-env';
 import {
@@ -232,6 +233,10 @@ async function executeNode<TContext>(
       });
     }
     case 'Fetch': {
+      if (shouldSkipFetchNode(node, context.requestContext.request.variables)) {
+        return new Trace.QueryPlanNode();
+      }
+
       const traceNode = new Trace.QueryPlanNode.FetchNode({
         serviceName: node.serviceName,
         // executeFetch will fill in the other fields if desired.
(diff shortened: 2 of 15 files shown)


gateway-js/src/executeQueryPlan.ts
```diff
@@ -252,6 +257,42 @@ async function executeNode<TContext>(
...
+
+      if (includeValue !== null && skipValue !== null) {
+        return !includeValue || skipValue;
+      } else if (includeValue !== null) {
+        return !includeValue;
+      } else if (skipValue !== null) {
+        return skipValue;
+      }
+
+      throw new Error('Programming error: unexpected value for conditional directive');
```
martijnwalraven, line 288: I suspect there might be a way to write this more concisely. If I get the semantics right (I haven't tested it), would something like this work?

```suggestion
      if (includeValue !== null && !includeValue) {
        return true;
      }
      
     if (skipValue !== null && skipValue) {
        return true;
     }
     
     return false;
```

query-planner-js/src/buildQueryPlan.ts
```diff
@@ -134,13 +139,171 @@ export function buildQueryPlan(
...
+  );
+
+  let allTopLevelsHaveConditional = true;
+  const inclusionConditions: {
+    skip: boolean | string | null;
+    include: boolean | string | null;
+  }[] = [];
+  walkConditionalTree(conditionalTree, (node) => {
+    if (node === null || (Array.isArray(node) && node[0] === null)) {
+      allTopLevelsHaveConditional = false;
```
martijnwalraven, line 174: Couldn't this short circuit the walk instead of continuing to build up `inclusionConditions`? Alternatively, perhaps you could even change `buildConditionalTree` to return `null` as soon as it encounters a `null` condition, so you could avoid the rest of the build and walk altogether.

martijnwalraven (changes requested):

gateway-js/CHANGELOG.md
```diff
@@ -4,7 +4,7 @@
 
 > The changes noted within this `vNEXT` section have not been released yet.  New PRs and commits which introduce changes should include an entry in this `vNEXT` section as part of their development.  When a release is being prepared, a new header will be (manually) created below and the appropriate changes within that release will be moved into the new section.
 
-- _Nothing yet. Stay tuned._
+- Based on additions to the query plan format. `FetchNode`s may now contain a property `inclusionConditions` which informs the executor whether or not a `FetchNode` can be skipped entirely at runtime. Inclusion conditions may consist of literals and variables, so must be resolved at execution time rather than during query planning. [PR #1113](https://github.com/apollographql/federation/pull/1113)
```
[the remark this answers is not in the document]
trevor-scheer: Agreed, I mostly swapped them and added some additional info. 4bc89acb

gateway-js/src/executeQueryPlan.ts
trevor-scheer: Ah yeah, the case where both are provided doesn't actually need to be handled specifically. This considerably simplifies the code. With strict equality checks against boolean values we don't even have to check for null, just:
`return includeValue === false || skipValue === true;`

Resolved in c8fd47df

gateway-js/src/executeQueryPlan.ts
```diff
@@ -252,6 +257,42 @@ async function executeNode<TContext>(
...
+    });
+
+    return shouldSkip;
```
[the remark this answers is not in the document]
trevor-scheer: Indirectly resolved in c8fd47df

query-planner-js/src/buildQueryPlan.ts
trevor-scheer: Agreed! A `ConditionalTree` can no longer have a `null` node, and `buildConditionalTree` will short-circuit return `null` when possible. Good call, and this simplifies the walking too!

5479a038

gateway-js/src/executeQueryPlan.ts
```diff
@@ -261,36 +261,25 @@ export function shouldSkipFetchNode(
...
-        return skipValue;
+  if (!node.inclusionConditions) return false;
+
+  return node.inclusionConditions.every((conditionals) => {
+    function resolveConditionalValue(conditional: 'skip' | 'include') {
+      const conditionalType = typeof conditionals[conditional];
+      if (conditionalType === 'boolean') {
+        return conditionals[conditional] as boolean;
+      } else if (conditionalType === 'string') {
+        return variables[conditionals[conditional] as string] as boolean;
```
trevor-scheer, line 272: The diff is hard to read so it's worth noting that I caught an error here (was casting `as string` when I should've been casting `as boolean`.

martijnwalraven (approved):
This looks good to me! The recent changes clean up the code nicely.

PR #1121, opened 2021-11-02
fix merge value receivedEntity to entities

vvakame:
fix inefficient execution.
We want merge receivedEntities into entities. and receivedEntities are sparse compared to entities.

### DIFF
--- a/gateway-js/src/executeQueryPlan.ts
+++ b/gateway-js/src/executeQueryPlan.ts
@@ -355,9 +355,9 @@ async function executeFetch<TContext>(
           );
         }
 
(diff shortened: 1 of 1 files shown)


pcmanus:
Hi @vvakame,

Sorry for not seeing this earlier, and thanks a lot for you contribution. And agreed with the fix which looks good to me. Would you mind adding a quick changelog entry in `gateway-js/CHANGELOG.md`? And maybe amending the commit message to explain the reasoning as you did in your earlier comment? But do let me know if it's not convenient for you to do so, I can deal with that if necessary.

PR #1133, opened 2021-11-03
Add eslint to project with some default settings and fixes

clenfest:
Adding lint checking to the `composition-js` `internals-js` and `query-planner-js` directories. For now use @typescript-eslint/recommended with a couple of exceptions:

Made less restrictive than default:
- no-namespace allowDeclarations:true - for Jest Matchers
- no-inferrable-types - Allow for parameters or properties
- no-unused-vars - Don't complain if var is prefixed with underscore
- no-this-alias - this is an error by default. We may want to reenable this eventually, currently 3 occurences in code

So there are 675 warnings, but only three rules are violated:
- no-explicit-any (392) - Ideally we get rid of as many of these as possible
- no-non-null-assertion (280) - These will be tricky to get rid of
- no-this-alias (3) - discussed above

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -84,7 +84,7 @@ jobs:
   GraphQL Types:
     description: "Assert generated GraphQL types are up to date"
     docker:
-      - image: cimg/node:14.16.1
+      - image: cimg/node:14.18.1
     steps:
       - oss/install_specific_npm_version:
           version: '7'
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -84,7 +84,7 @@ jobs:
   GraphQL Types:
     description: "Assert generated GraphQL types are up to date"
     docker:
-      - image: cimg/node:14.18.1
+      - image: cimg/node:14.16.1
     steps:
       - oss/install_specific_npm_version:
           version: '7'
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -71,7 +71,7 @@ jobs:
 
   NodeJS 14:
     docker:
-      - image: cimg/node:14.16.1
+      - image: cimg/node:14.18.1
     steps:
       - common_test_steps
 
@@ -84,7 +84,7 @@ jobs:
   GraphQL Types:
     description: "Assert generated GraphQL types are up to date"
     docker:
-      - image: cimg/node:14.16.1
+      - image: cimg/node:14.18.1
     steps:
       - oss/install_specific_npm_version:
           version: '7'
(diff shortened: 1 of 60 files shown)


pcmanus:
Definitively happy to add linting (and happy with PR outside of node 14 failures mentioned below). Slightly curious as to why you included some directories and not others though: was it because the others had too many error currently? If so, happy to add them later incrementally, just curious.

Also, this seem to be breaking test with Node 14. Seems eslint requires `14.7.0` but the tests are using `14.6.1`. Don't know if just bumping our minimum node version requirement to `14.7.0` is fine or not (@abernix?).

I'll also ping @martijnwalraven on this in case he has some opinions?

abernix:
I support bumping the versions of Node.js we test against to be the latest LTS versions of those major release lines when they're available!  This used to be managed by Renovate (and they were automatically bumped, but I disabled it because it needed a bit more work to accurately bump the _right_ versions (and not bump our CircleCI versions from, e.g., 12 to 14 when we were intentionally testing 12. 😄 

However, for example, both of these should be updated:

https://github.com/apollographql/federation/blob/b961a20c6eb3990e4c1693805bedb44fdd3f39d4/.circleci/config.yml#L72-L94

Note that because of the way that Windows `nvm` "works" (read: doesn't work) — in the **Rust** portion of the CircleCI test fan-out, which are defined as a `NODE_VERSION` env variable, here:

https://github.com/apollographql/federation/blob/b961a20c6eb3990e4c1693805bedb44fdd3f39d4/.circleci/config.yml#L104-L105

We use the latest version that is set in the CircleCI Windows base image — which is typically very close to the latest LTS, but not the actual latest LTS.

Happy to talk more about this, but, yes, feel comfortable to bumping to `cimg/node:14.18.1`, but strong preference to pin explicitly to that tag, rather than say just `14`, since `14` is way less of a reproducible build.

clenfest:
> Definitively happy to add linting (and happy with PR outside of node 14 failures mentioned below). Slightly curious as to why you included some directories and not others though: was it because the others had too many error currently? If so, happy to add them later incrementally, just curious.

Mainly I was not quite sure about the scope of what we owned and didn't want to step on toes outside of our team. Happy to get rid of the ignore file and make changes in the other directories as well.

@abernix - I had another branch where I made the change to the docker image, but I wasn't sure about the Rust NODE_VERSION since it didn't match. You think I should make them both 14.18.1?

pcmanus:
> Mainly I was not quite sure about the scope of what we owned and didn't want to step on toes outside of our team.

Understood. I'm myself less familiar with the code in `gateway-js` and `federation-js`/`subgraphs-js`, which is also older code in general,  but @trevor-scheer and @martijnwalraven might have opinions on whether they'd rather have those excluded. But I'd say let's at least include all the other ones.

clenfest:
Not sure why these documentation jobs are failing.

pcmanus (approved):
This look good to me. Thanks for adding it!

PR #1140, opened 2021-11-03
chore(gateway): Update `@apollo/core-schema` / fix `GraphQLError` usage

trevor-scheer:
The latest version of `graphql-js` made incompatible changes that broke the `@apollo/core-schema` errors which extended `GraphQLError`.

After updating and releasing a new version of `@apollo/core-schema`, we need to do the same here to get this branch back into a passing state.

Note this change updates the `graphql` peerDependency and thus will require a major (0.x) version bump since it's a breaking change.

### DIFF
--- a/federation-integration-testsuite-js/src/snapshotSerializers/graphqlErrorSerializer.ts
+++ b/federation-integration-testsuite-js/src/snapshotSerializers/graphqlErrorSerializer.ts
@@ -14,8 +14,8 @@ export default {
     }
     return print({
       message: value.message,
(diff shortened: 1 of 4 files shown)


glasser (approved):

trevor-scheer:
The cause of the rust test failures is pre-existing to these changes, and is well-understood and being worked on. This PR brings the state of the `version-0.x` branch to a considerably-less-failing place than before, so I'm going to land it even though CI isn't passing.

PR #1149, opened 2021-11-04
fix: PlanningErrorExtensions

o0Ignition0o:
Updates on the TypeScript query planner changed the PlanningError structure, that will now always have an `extensions` field.
No extensions are now expressed by having the extensions field value being null, or an empty Object.

This PR allows us to express these requirements by:
- Adding a new deserialize directive `none_only_if_value_is_null_or_empty_object`.
- Applying the directive to the `PlanningError::extensions` field.
- Adding comments and tests that make the new requirement more explicit.


trevor-scheer (approved):
🥳

PR #1151, opened 2021-11-04
chore: Update generated types (`main`)

trevor-scheer:


trevor-scheer:
Note: no references to the changed args yet so no changes required besides the actual types update.

PR #1152, opened 2021-11-04
Conditional schema fetch based on ifAfterId

snoopdave:
The Apollo Uplink end-point now supports conditional schema fetching via a new parameter `ifAfterId`. If this parameter is specified then the end-point will only return a schema with an ID that is newer than `ifAfterId`. This PR adds the `ifAfterId` parameter to the Gateway, makes the associated query changes and modifies the tests so that they work with this new parameter.

I tested end-to-end by downloading the Gateway tarball from the CI system and running it via the Space Camp demo. I added some console logs and verified that the Gateway initially loads the schema, sends the `ifAfterId` parameter and does not load the schema if it unchanged.


snoopdave:
@trevor-scheer I have tested this end-to-end now and I think this PR is ready for review and merge. Please take a look when you have a chance.

gateway-js/src/loadSupergraphSdlFromStorage.ts
```diff
@@ -43,12 +43,20 @@ export async function loadSupergraphSdlFromStorage({
...
   fetcher,
+  ifAfterId,
+  currentSupergraphSdl,
 }: {
   graphRef: string;
   apiKey: string;
   endpoint: string;
   fetcher: typeof fetch;
+  ifAfterId: string | null;
+  currentSupergraphSdl: string | null;
```
trevor-scheer, line 54: Just proposing an alternative - it may not be any better or worse than what you've already got!

I'm not sure this function needs to have awareness of the current sdl. We could instead do the following:
* only pass in an `inAfterId` if we have both `this.compositionId` and `this.supergraphSdl` (though I can't imagine a case we'd have one without the other [ref](https://github.com/apollographql/federation/blob/version-0.x/gateway-js/src/index.ts), so maybe just the `compositionId` is sufficient)
* in the `routerConfig.__typename === 'Unchanged'` case, return `null` instead of the previous values, and handle the `null` case at the call site

And while `ifAfterId` does make sense as a query arg, the naming is a bit more of an implementation detail of our API. I think it makes sense to name the arg coming into this function either `compositionId` or `currentCompositionId` to be more inline with the naming of things at the call site.

gateway-js/src/__tests__/integration/nockMocks.ts
```diff
@@ -70,21 +70,33 @@ export const mockCloudConfigUrl =
...
-  supergraphSdl = getTestingSupergraphSdl(),
-  id = 'originalId-1234',
+export function mockSupergraphSdlRequest(ifAfter: string | null = null) {
+  return mockSupergraphSdlRequestIfAfter(ifAfter);
+}
+
+export function mockSupergraphSdlRequestSuccessIfAfter(
+  supergraphSdl: string | null = getTestingSupergraphSdl(),
+  id: string | null = 'originalId-1234',
+  ifAfter: string | null = null,
```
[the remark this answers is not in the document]
snoopdave: nice suggestion. I committed your suggested changes.

gateway-js/src/__tests__/integration/nockMocks.ts
```diff
@@ -70,21 +70,33 @@ export const mockCloudConfigUrl =
...
+      ifAfterId: ifAfter,
     },
   });
 }
 
-export function mockSupergraphSdlRequestSuccess(
-  supergraphSdl = getTestingSupergraphSdl(),
-  id = 'originalId-1234',
+export function mockSupergraphSdlRequest(ifAfter: string | null = null) {
+  return mockSupergraphSdlRequestIfAfter(ifAfter);
```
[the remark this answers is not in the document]
snoopdave: fixed

gateway-js/src/loadSupergraphSdlFromStorage.ts
snoopdave: I don't really like my solution to this problem of not overwriting the existing schema, but I don't see how to do what you suggest without changing `SupergraphSdlUpdate` to look like this:

```
export interface SupergraphSdlUpdate {
  id: string | null;
  supergraphSdl: string | null;
}
```

Which leads to lots of other changes to make TypeScript happy with types.

I tried dealing with the null results in `isSupergraphSdlUpdate()`.

snoopdave:
@trevor-scheer thanks for spelling out the changes around parameters for `loadSupergraphSchermaFromStorage()`. It looks much better now! I downloaded the tarball and tested locally with success.

gateway-js/src/__tests__/integration/networkRequests.test.ts
```diff
@@ -130,7 +132,11 @@ it('Fetches Supergraph SDL from remote storage using a configured env variable',
 
 it('Updates Supergraph SDL from remote storage', async () => {
   mockSupergraphSdlRequestSuccess();
-  mockSupergraphSdlRequestSuccess(getTestingSupergraphSdl(fixturesWithUpdate), 'updatedId-5678');
+  mockSupergraphSdlRequestSuccessIfAfter(
+      'originalId-1234',
+    'updatedId-5678',
+      getTestingSupergraphSdl(fixturesWithUpdate),
+  );
```
trevor-scheer, line 139: Indentation consistency nit:
```suggestion
  mockSupergraphSdlRequestSuccessIfAfter(
    'originalId-1234',
    'updatedId-5678',
    getTestingSupergraphSdl(fixturesWithUpdate),
  );
```

gateway-js/src/__tests__/integration/networkRequests.test.ts
```diff
@@ -314,8 +320,12 @@ describe('Supergraph SDL update failures', () => {
 it('Rollsback to a previous schema when triggered', async () => {
   // Init
   mockSupergraphSdlRequestSuccess();
-  mockSupergraphSdlRequestSuccess(getTestingSupergraphSdl(fixturesWithUpdate), 'updatedId-5678');
-  mockSupergraphSdlRequestSuccess();
+  mockSupergraphSdlRequestSuccessIfAfter(
+      'originalId-1234',
+    'updatedId-5678',
+      getTestingSupergraphSdl(fixturesWithUpdate),
+  );
```
trevor-scheer, line 327: Indentation consistency nit:
```suggestion
  mockSupergraphSdlRequestSuccessIfAfter(
    'originalId-1234',
    'updatedId-5678',
    getTestingSupergraphSdl(fixturesWithUpdate),
  );
```

trevor-scheer (changes requested):
Can we add a test or two which exercise the new `Unchanged` branch?

https://github.com/apollographql/federation/pull/1152/files#diff-608730b0e96c4f7dea489f3ad45851513022a29a7ba2ab00e6eb23341a66be86R135-R136

gateway-js/src/loadSupergraphSdlFromStorage.ts
trevor-scheer: Resolved by c1ecd02

gateway-js/src/__tests__/integration/networkRequests.test.ts
snoopdave: yeah, IntelliJ is too smart for my own good with those little field labels it display.

gateway-js/src/__tests__/loadSupergraphSdlFromStorage.test.ts
```diff
@@ -1,13 +1,14 @@
...
-  mockOutOfBandReporterUrl,
-  mockOutOfBandReportRequestSuccess,
-  mockSupergraphSdlRequestSuccess,
+    graphRef,
+    apiKey,
+    mockCloudConfigUrl,
+    mockSupergraphSdlRequest,
+    mockOutOfBandReporterUrl,
+    mockOutOfBandReportRequestSuccess,
+    mockSupergraphSdlRequestSuccess,
```
trevor-scheer, line 10: Some more indentation noise here

trevor-scheer (approved):
This LGTM! Can you please add a CHANGELOG entry in the `gateway-js/CHANGELOG.md`?

PR #1155, opened 2021-11-04
Remove old composeServices

clenfest:
I deleted composeServices and its dependencies, but not theo other stuff that was in compose.ts since there were a lot more tests that would be affected. If you think I should just delete the whole file and all associated tests, let me know.


pcmanus:
> If you think I should just delete the whole file and all associated tests, let me know.

Yes, I think that's what we want (removing the whole file).

pcmanus (approved):
The removals look good to me. Thanks.

PR #1191, opened 2021-11-10
Bump docs theme

StephenBarlow:


docs/package.json
```diff
@@ -1,8 +1,4 @@
 {
-  "engines": {
-    "node": ">=14 <17",
-    "npm": "7.x"
-  },
```
trevor-scheer, line 5: Assuming this is kosher, nice! Autoclosed a couple renovate PRs that I just bugged @trevorblades about.

StephenBarlow: We can triple-check next week, but yeah these `engines` specifications were inconsistent across docsets, and docsets have been breaking more frequently lately it seems. I've gotten pretty much all of them consistent in their definition and they're all building fine. I'm hoping we can hold this stuff steady until we move all of the docs _generation_ code into a single place

PR #1208, opened 2021-11-15
Finish removing old composition code

pcmanus:
The core of the old (pre-federation 2.0) composition code has been removed by https://github.com/apollographql/federation/pull/1155 but this still left some tangential code that isn't used anymore and should be remove.

This commit remove the whole `federation-js/src/composition` directory, which "make sense" since the composition code has otherwise moved to a dedicated module (`composition-js` directory).

I will note that there _was_ some method within the removed files that were exported in practice by `federation-js/src/index.ts` that this remove and could theoretically have been used by federation users, but I didn't spot any obvious candidate (most of what was exported was more related to the supergraph handling). That said, could be worth a double-check during review.


pcmanus:
@trevor-scheer, @martijnwalraven: I'd appreciate if you guys could have a quick look at this to see if you something removed here that might actually be of genuine use for federation users and should not be removed. Thinking of `composition-js/src/types.ts` and `composition-js/src/utils.ts` in particular.

federation-js/src/index.ts
```diff
@@ -1,8 +1,8 @@
 import { buildSubgraphSchema } from '@apollo/subgraph';
```
trevor-scheer, line 1: If this is all that's left of the exports, should we just remove the package and stop publishing it? Its 2 purposes (subgraph and composition) are both completely owned by separate packages, seems reasonable to remove this altogether going into v2.

trevor-scheer: This comment is directed at the whole file, not just the one line.

pcmanus: > seems reasonable to remove this altogether going into v2

I would agree (and double-checking, it doesn't seem the module is used anymore within federation itself after this PR). I think the only concerns is that many people will still have imports on `@apollo/federation` for `buildSubgraphSchema` and so the question is whether we want to maintain compatibility for that. But as fed 2 is a major, I'd personally say it's a good time for that (if anything, stopping to publish the package would be a good sign that people should move to use `@apollo/subgraph`).

The type I felt was maybe imported by users (and that the PR moves directly into `federation-js/src/index.ts` for that reason) is the `ServiceDefinition` interface because it's used for local gateway configs. But it would arguably make more sense that this interface is exported by `@apollo/gateway` so we could make it so.

@abernix Opinions?

trevor-scheer: Agree with all of the above! v2 seems like exactly the time to remove deprecated bits like this, especially if it means we can stop publishing a package altogether. Also agree with the `ServiceDefinition` export finding a new home in `@apollo/gateway`, that sgtm.

trevor-scheer (approved):

trevor-scheer:
@pcmanus I finalized the cleanup mostly in the latest commit. Feel free to revert if it's decided that we won't remove it completely for now.

harmonizer/src/js_types.rs
```diff
@@ -63,10 +63,11 @@ impl Error for CompositionError {}
 /// the `extensions` property of a JavaScript [`GraphQLError`] from the
 /// [`graphql-js`] library. Such errors are created when errors have prevented
 /// successful composition, which is accomplished using [`errorWithCode`]. An
-/// [example] of this can be seen within the `federation-js` JavaScript library.
+/// [example] of this can be seen within the `composition-js` JavaScript library.
 ///
 /// [`graphql-js']: https://npm.im/graphql
 /// [`GraphQLError`]: https://github.com/graphql/graphql-js/blob/3869211/src/error/GraphQLError.js#L18-L75
+/// TODO: replace with `composition-js` equivalents?
```
trevor-scheer, line 70: Wasn't sure here

pcmanus: Pushed an additional commit that modifies those (that, for completeness sake, ensuring we ship codes will all errors and documenting them is one of the TODO for federation 2 before GA, but "some" of the errors do get code already and I've link to an example of those).

pcmanus:
Thanks @trevor-scheer, and this lgtm. I did push a final commit to address the links in `js_types.rs` you pointed, make sure `ServiceDefinition` is exported by the gateway now as mentioned above (since after all, it's part of the gateway public API) and remove a few small remaining mentions of `federation-js`/`@apollo/federation`.

trevor-scheer:
Pinged @abernix to have a look

federation-js/src/index.ts
abernix: @trevor-scheer @pcmanus I support this, but it would be worth opening an issue that ensures that we at some point run `npm deprecate <pkg> <msg>` on this package.  Doing that will start outputting the deprecation warnings during installation of the package by package managers and be a good signal to developers to make the (remarkably straightforward) change to the new package name. We might even want to do this quite soon to get the gears in motion?

It's worth calling out the impact this has on tutorials and education content that we and others have produced — and even printed books!  Those things can all be updated, but in my mind the sooner we get the deprecation warning up and the longer we keep it working (perhaps by having `@apollo/federation` depend on a caret semver range that automatically gets minor/patch updates to `@apollo/subgraph` for some time?).  We may even want to communicate this directly to those that are thinking about developer experience / docs / blogs / etc.

docs/source/api/apollo-subgraph.mdx
```diff
@@ -9,8 +9,6 @@ This API reference documents the exports from the `@apollo/subgraph` package.
 
 ## `buildSubgraphSchema`
 
-> This method previously existed in the `@apollo/federation` package and was renamed from `buildFederatedSchema` after `@apollo/federation` v0.28.0 (the previous name still works, but it might be removed in a future release).
-
```
abernix, line 13: Leaving this here seems like a good thing to maintain for the foreseeable future, particularly since it'll likely be years before everyone actually updates?   (We could, however, stop mentioning that it still works.)

harmonizer/README.md
```diff
@@ -18,7 +18,7 @@ While we intend for a future version of composition to be done natively within
 Rust, this allows us to provide a more stable transition using an already stable
 composition implementation while we work toward something else.
 
-[`@apollo/federation`]: https://npm.im/@apollo/federation
+[`@apollo/composition`]: https://npm.im/@apollo/composition
```
abernix, line 21: Good catch!

abernix (approved):
LGTM, with my most notable comment being https://github.com/apollographql/federation/pull/1208#discussion_r751136648 but one comment below about removing the callout in the docs that might be worth preserving.

federation-js/src/index.ts
trevor-scheer: I think deprecating only applies to v2 (which we maybe just shouldn't have ever published given that we're open to killing it). The v0.x gateway continues to use `@apollo/federation` for its composition capabilities, so I don't think we want to deprecate it (unless we broke out composition as well, which I'm not sure is worth the effort).

abernix: Hmm, I see what you're saying.  Perhaps what we might consider is then adding a runtime deprecation warning to the use of `@apollo/federation` for its `buildFederatedSchema` / `buildSubgraphSchema`, rather than at the npm level?

trevor-scheer: I think we want to:
* `npm deprecate` v2+ versions of `@apollo/federation`
* Runtime deprecate `buildSubgraphSchema` (currently re-exported from `@apollo/subgraph`)

Currently `buildFederatedSchema` is already runtime deprecated in `@apollo/federation` and isn't exported by `@apollo/subgraph` at all.

docs/source/api/apollo-subgraph.mdx
pcmanus: Agreed, probably worth keeping this. Added it back with a minor edit to mention that `@apollo/federation` is no more.

pcmanus:
Sorry for the slow turnaround. Rebased and committed and created https://github.com/apollographql/federation/issues/1228 to follow-up on the npm deprecation of `@apollo/federation` (at least the version we published for fed 2 alpha).

I will note that rebasing this on main, I run into a small dependency issue with `ts-jest` and I ended up fixing that in https://github.com/apollographql/federation/pull/1208/commits/3d27ce17aeb5df3f921a0a5063fd769edef7058c (essentially bumping the version slightly, which seemed to require a minor import change). Please feel free to yell at me if I did something wrong in there.

PR #1211, opened 2021-11-16
wip: router-bridge: check in .js output

### LINKED ISSUE #1210: router-bridge: generate .js files and check it in to the repo
We're currently relying on build.rs to generate .js files, which has an impact on downstream build times and dependencies.
We can do this instead:

- generate the js files by running the `npm run compile:for-router-bridge-build-rs` while developing.
- check in the output directory to the repository
- add a snapshot test that makes sure the files are always in sync
- remove the build.rs file alltogether 🎉

o0Ignition0o:
will fix #1210 eventually


o0Ignition0o:
CI behaves as expected https://app.circleci.com/pipelines/github/apollographql/federation/5935/workflows/7a71ce24-6bc4-4e81-b63e-c1f54f0fc0ad/jobs/33237 🎉

abernix:
While I think this was a good idea to consider, I'm going to close this because I think it will cause too much churn on this repository's VCS and, additionally, incur a burden on the primary contributors to this repository since it would technically require committing a generated file each time the rest of the code (or its dependencies) in the repository get updated.  😱 

We'll discuss another option in #1210.

PR #1221, opened 2021-11-18
Have Subgraphs use OrderedMap

clenfest:
Two non-functional changes:
- Rather than having Subgraphs keep track of ordering, split out logic into OrderedMap class
- latest() asserts there is a latest version rather than potentially returning undefined

### DIFF
--- a/internals-js/src/utils.ts
+++ b/internals-js/src/utils.ts
@@ -30,6 +30,88 @@ export class MultiMap<K, V> extends Map<K, V[]> {
 }
 
 /**
+ * Generic OrderedMap class that can sort keys based on an arbitrary sorting function
+ * Insert time is O(log(N))
+ * Remove is not implemented, but the trivial implementation would be O(N)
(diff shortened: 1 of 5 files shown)


internals-js/src/utils.ts
```diff
@@ -29,6 +29,88 @@ export class MultiMap<K, V> extends Map<K, V[]> {
...
+
+  get(key: K): V | undefined {
+    return this._values.get(key);
+  }
+
+  has(key: K): boolean {
+    return this._values.has(key);
+  }
+
+  size(): number {
```
pcmanus, line 72: Nitpick: we may want to make this a getter to mimick the `Map` api.

pcmanus (approved):
Lgtm (left a minor suggestion but it can be ignored).

internals-js/src/utils.ts
clenfest: Good call, I'll change it.

PR #1222, opened 2021-11-19
change Element parent type to remove undefined

clenfest:
Want to get some feedback on this before I spend too much more time on it. Basic issue is that we're doing a lot of type assertions where a client asserts that a parent (or schema) exists, but without any real type safety guarantees.

From my reading of the code, it seems that the parent is always known at creation time and we never reassign the parent. Given this, it seems like the best way to go is to just add the parent as a parameter to the constructor. This creates a bit of weirdness given that there is a lot of code that sets parents after instantiation, but I've basically just disabled that code rather than rewrite it for now. We have support for the onAttached() callback, which I don't 100% understand what it's for, so I'm not sure if that can / should be moved into the constructor or left where it is for now.

Rather than remove the parent reference, I've added a boolean _attachedToParent which tracks the state of whether the element is actually attached. 

Throwing this out there to get comments before I work on it further. I do think it's valuable to try to get rid of type assertions so that we can trust the type system, but wanted to solicit feedback on this particular approach. Thoughts?

### DIFF
--- a/composition-js/src/merging/merge.ts
+++ b/composition-js/src/merging/merge.ts
@@ -394,7 +394,7 @@ class Merger {
         }
         const previous = this.merged.type(type.name);
         if (!previous) {
-          this.merged.addType(newNamedType(type.kind, type.name));
+          this.merged.addType(newNamedType(this.merged, type.kind, type.name));
         } else if (previous.kind !== type.kind) {
           mismatchedTypes.add(type.name);
         }
@@ -414,7 +414,7 @@ class Merger {
           continue;
         }
         if (!this.merged.directive(directive.name)) {
-          this.merged.addDirectiveDefinition(new DirectiveDefinition(directive.name));
+          this.merged.addDirectiveDefinition(new DirectiveDefinition(this.merged, directive.name));
         }
       }
     }
--- a/composition-js/src/merging/merge.ts
+++ b/composition-js/src/merging/merge.ts
@@ -975,7 +975,7 @@ class Merger {
     dest: T,
     allTypesEqual: boolean
   ) {
-    if (!this.needsJoinField(sources, dest.parent!.name, allTypesEqual)) {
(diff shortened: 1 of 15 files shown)


pcmanus:
So, while I do agree that it'd be nice for `parent` and `schema()` to "externally" never return `undefined` (to remove having null-assertions on every usage), I do would like to preserve the notion that if an element has been removed, then that element does _not_ visibility point to a parent/schema it's not attached to, and from a quick read of the patch, this is not the case anymore here.

Essentially, there is some double-linking going one with elements: a schema/parent link to its elements and the elements link back to its schema/parent (whether such mirrored-linking is a good idea in the first place is imo a different discussion). And so what the code tries to do right now is ensuring that no methods leaves things in a state where one side of link exists but not the other as that's kind of a half broken state and that's imo pretty error-prone.

Concretely, say code removes an element but mistakenly continue using that element as if it hadn't been removed (maybe the element was aliased somewhere and we didn't realize it): if the removed element still "half" link to a schema it's not part of anymore, usage of the removed element has more changes tow work long enough that when it does break it's a pain to track down the root cause (and it could also make the root problem stay undetected longer).

So to sum up, the code tries to make it so that once you've removed an element, attempting to continue using that element (as if it was still part of a schema) would break immediately, and I'd like to keep that property.

And thus, clearing up the parent link on remove is imo the right thing to do. Having removed elements pretend that they are still attached to their parent/schema is just ... wrong. And that imply  we _do_ have to assert at runtime that the concrete elements the code manipulate have a parent/schema properly set.

*But*, that said, I do agree that doing the null-assertions on every use is not exactly amazing when the rule is essentially that no code should try to access a parent/schema of an element that don't have one.

So my suggestion here would simply be to add an assertion that `parent !== undefined` in the `parent` getter, making the property never returning `undefined` anymore. This would get rid of the null-assertions all over the place.

clenfest:
Yeah, that's definitely a lot cleaner. Thanks!

PR #1223, opened 2021-11-19
Make the parent getter assert that parent exists

clenfest:
Rather than forcing clients to check or assert that the parent exists, do it in the getter. The rest of the changes are either fallouts of this decision or cleaning up type assertions.

Note that in some cases, we had internal functions start accessing the _parent value directly rather than through the parent getter because they are checking to see if the parent exists while in an intermediate state

For example, code in the Element hierarchy that looked like this:
```typescript
if (this.parent?.schema()) {
}
```

Now looks like this:
```typescript
if (this._parent?.schema()) {
}
```

### DIFF
--- a/internals-js/src/definitions.ts
+++ b/internals-js/src/definitions.ts
@@ -408,7 +408,7 @@ abstract class Element<TParent extends SchemaElement<any, any> | Schema | Direct
     return this._parent;
   }
 
-  hasParent(): boolean {
+  isAttached(): boolean {
     return !!this._parent;
   }
 
@@ -430,7 +430,7 @@ abstract class Element<TParent extends SchemaElement<any, any> | Schema | Direct
     // to a schema could bring a whole hierarchy of types and directives for instance). If they are attached, it only work if
     // it's to the same schema, but you have to check.
     // Overall, it's simpler to force attaching elements before you add other elements to them.
-    if (!this.hasParent()) {
+    if (!this.isAttached()) {
       throw error(`Cannot modify detached element ${this}`);
     }
   }
@@ -546,7 +546,7 @@ export abstract class SchemaElement<TOwnType extends SchemaElement<any, TParent>
     // removals on detached internally; though of course we could refactor the code if we wanted).
   }
 
-  protected checkUpdate(addedElement?: { schema(): Schema | undefined }) {
+  protected checkUpdate(addedElement?: { schema(): Schema, isAttached(): boolean }) {
     super.checkUpdate();
     if (!Schema.prototype['canModifyBuiltIn'].call(this.schema())) {
       // Ensure this element (the modified one), is not a built-in, or part of one.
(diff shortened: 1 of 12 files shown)


internals-js/src/definitions.ts
```diff
@@ -531,7 +548,7 @@ export abstract class SchemaElement<TOwnType extends SchemaElement<any, TParent>
 
   protected checkUpdate(addedElement?: { schema(): Schema | undefined }) {
```
pcmanus, line 549: I don't think we ever pass something whose `schema()` method can be `undefined`.

But in fact, a suggestion would be to add to the parameter type that it should also have a `isAttached(): boolean` method so we avoid some `instanceof` later one. This does imply adding a `isAttached` to `BaseWrapperType` (but as its `schema` method can throw, I think it make sense to have a "can I call schema() safely" method in case it's needed) and rename `Element.hasParent` into `Element.isAttached` (that is, `hasParent` is just fine for `Element`, but it would be weird for `BaseWrapperType` because it doesn't expose a "parent", just a "schema", and so "isAttached" is a bit more general; a `hasSchema` naming would work to but I personally have a minor preference fort `isAttached`).

Pushed a commit with that suggestion at https://github.com/pcmanus/federation/commit/9aa17622194fd522782b810e84d6025487773a88 to clarify.

internals-js/src/definitions.ts
```diff
@@ -2058,7 +2072,7 @@ export class FieldDefinition<TParent extends CompositeType> extends NamedSchemaE
   }
 
   get coordinate(): string {
-    const parent = this.parent;
+    const parent = this._parent; // TODO: @pcmanus does this change look right to you?
```
pcmanus, line 2075: Yes, looks ok to me (and fwiw, the commit I link above removes the TODO, assuming you're ok with that earlier suggestion). To expand, I can't think of code calling this method on a detached element so using `this.parent` would probably not break anything, nor can I think of a definitive reason why one option would be strictly better than the other: we use this in a number of error message so there is probably cases where actually getting an actual error clearly stating the element is detached help debugging more than throwing while generating the error message, but I can imagine other cases where throwing would be better. Overall, happy either way but a slight preference for using `this._parent` as you did.

internals-js/src/definitions.ts
```diff
@@ -2516,7 +2530,7 @@ export class Directive<
   }
 
   private isAttachedToSchemaElement(): boolean {
-    return this.parent !== undefined && this.parent instanceof SchemaElement;
+    return this._parent !== undefined && this._parent instanceof SchemaElement;
```
pcmanus, line 2533: Nit: maybe replacing the first condition by `this.isAttached()` would be ever so slightly more readable.

pcmanus (approved):
Left a few minor suggestion but lgtm overall. Nice improvement.

Side-note: a related reason for a bunch of null-assertions `!` in the code is the `get type(): TType | undefined` of `NamedSchemaElementWithType`. That is, I think it's a bit less clear-cut than with the parent/schema here in the sense that "detached" elements shouldn't really be used at all (except possibly for attaching them) while there is cases where it make sense to have an element without a type set temporarily during constructions (typically when you building cyclic types). Still, there is more code "using" a fully validated schema (whose element cannot have an `undefined` type, this is checked by validation) that code modifying schema in such a way, so most code end up using `type()` with a null assertion. So maybe we can make `type()` asserts the type is set, and add a `hasTypeSet()` method for those case where you do know you might be manipulating a partially built schema. Anyway, mentioning this because it's a similar kind of change, but probably worth doing in a separate PR anyway.

internals-js/src/definitions.ts
clenfest: Is the instanceof check unnecessary here?

pcmanus: I believe so. The subtlety is that directive can actually be attached to queries 
 and in that case the parent is an `AbstractOperationElement` from `operation.ts`), but while such operation element does link back to a particular schema ultimately (the schema on which the operation applies), that link is "external" and this ensure modifying a direction on an operation isn't considered as a schema modification.

PR #1233, opened 2021-11-23
Remove undefined  return value and add assert to type getter

clenfest:
All tests pass, but I need to continue to read code to make sure there isn't an existing case that could be a problem.

One definite issue is that if someone were to add a reference to type in the set parent callback, there would definitely be an assertion failure at runtime. Not sure how we feel about that.

### DIFF
--- a/composition-js/src/__tests__/compose.test.ts
+++ b/composition-js/src/__tests__/compose.test.ts
@@ -727,13 +727,13 @@ describe('composition', () => {
         `);
 
         // Making sure we properly extract the type of `f` for both subgraphs
-        const fInA = (subgraphs.get('subgraphA')!.schema.type('T')! as ObjectType).field('f');
+        const fInA = (subgraphs.get('subgraphA')!.schema.type('T') as ObjectType).field('f');
         expect(fInA).toBeDefined();
-        expect(fInA?.type?.toString()).toBe('I');
+        expect(fInA?.type.toString()).toBe('I');
 
-        const fInB = (subgraphs.get('subgraphB')!.schema.type('T')! as ObjectType).field('f');
+        const fInB = (subgraphs.get('subgraphB')!.schema.type('T') as ObjectType).field('f');
         expect(fInB).toBeDefined();
-        expect(fInB?.type?.toString()).toBe('A');
+        expect(fInB?.type.toString()).toBe('A');
       });
 
       it('merges union subtypes', () => {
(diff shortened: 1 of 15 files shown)


pcmanus:
> One definite issue is that if someone were to add a reference to type in the set parent callback, there would definitely be an assertion failure at runtime.

I'm not sure to see which callback you are referencing to/the problem involved.

clenfest:
> > One definite issue is that if someone were to add a reference to type in the set parent callback, there would definitely be an assertion failure at runtime.
> 
> I'm not sure to see which callback you are referencing to/the problem involved.

No current problem, but I meant since we have the callback apparatus, if someone referenced type from there in the future, things could fail in ways that might be tough to diagnose. Nothing to do on this, but I thought I'd call it out.

pcmanus:
Is this ready for review? I've hold off because you mentioned still wanted to check stuffs in your first comment, but happy to have a pass if it's ready.

> No current problem, but I meant since we have the callback apparatus, if someone referenced type from there in the future, things could fail in ways that might be tough to diagnose.

I'm sure I'm just being dense, but I'm still not sure to see what you are referring to. Which "callback apparatus"? Maybe a super simple example of what would fail in hard to diagnose if done would clarify it for me.

clenfest:
Yeah, it's ready. Sorry, I was just referring to the fact that there are a lot of calls to `setParent` in each of the different subclasses, which in turn makes a call to `onAttached`. If any of the `onAttached` overriders were to make a reference to `.type()` in the future, they would get an assertion error.

pcmanus (changes requested):
There is actually a number of instance where we are checking if the type is set or not and that this patch breaks. I'll list those I noticed below and I think we should add and use a new `isTypeSet()` method:
- [this validation](https://github.com/apollographql/federation/blob/clenfest/named_schema_assert_Type/internals-js/src/validate.ts#L133) is broken by this patch and needs to be updated. Probably the biggest issue here as it's that validation that allows us to rely in most other places of the code that the type is indeed always set.
- [this line](https://github.com/apollographql/federation/blob/clenfest/named_schema_assert_Type/internals-js/src/federation.ts#L323) is also now broken.
- [this line](https://github.com/apollographql/federation/blob/clenfest/named_schema_assert_Type/internals-js/src/federation.ts#L346) was assuming the type might be `undefined`. That said, it truly cannot be `undefined` at that point (due ot the validation of the first point running prior to this), so we can just remove the first test of the `if`.
- [this check](https://github.com/apollographql/federation/blob/clenfest/named_schema_assert_Type/internals-js/src/values.ts#L181), [that one](https://github.com/apollographql/federation/blob/clenfest/named_schema_assert_Type/internals-js/src/values.ts#L209) and [that other one](https://github.com/apollographql/federation/blob/clenfest/named_schema_assert_Type/internals-js/src/values.ts#L275) cannot ever trigger anymore and can be removed.

PR #1240, opened 2021-11-24
chore: Update renovate.json

trevor-scheer:
Remove an outdated jest entry (both branches are currently on v27.x)

### DIFF
--- a/renovate.json5
+++ b/renovate.json5
@@ -107,10 +107,5 @@
       "matchPackageNames": ["gatsby"],
       "allowedVersions": "2.x",
     },
-    {
-      "matchBaseBranches": ["version-0.x"],
-      "excludePackageNames": ["deno_core"],
-      "enabled": false
-    }
   ]
 }
--- a/renovate.json5
+++ b/renovate.json5
@@ -35,18 +35,6 @@
   "enabledManagers": ["npm", "cargo"],
   postUpdateOptions: ["npmDedupe"],
   "packageRules": [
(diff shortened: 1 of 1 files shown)


renovate.json5
```diff
@@ -119,5 +107,10 @@
       "matchPackageNames": ["gatsby"],
       "allowedVersions": "2.x",
     },
+    {
+      "matchBaseBranches": ["version-0.x"],
+      "excludePackageNames": ["deno_core"],
+      "enabled": false
+    }
```
abernix, line 114: I'm confused by this versus the intention stated in the PR body: **exclude** package names `deno_core` but `enabled: false`?

Why not:
```suggestion
    {
      "matchBaseBranches": ["version-0.x"],
      "matchPackageNames": ["deno_core"],
      "allowedVersions": "<exactly the version that we want to pin to>",
    }
```

Further, we should _probably_ include a `matchManagers`  to be precise, otherwise we're matching any package called `deno_core` in any registry.  Obviously that's not likely, but better to be precise when we can be?

So, incorporating the above:

```suggestion
    {
      "matchBaseBranches": ["version-0.x"],
      "matchPackageNames": ["deno_core"],
      "allowedVersions": "<exact version we want>",
      "matchManagers": [ "cargo" ],
    }
```

And why is this only a problem on the `version-0.x` branch?

trevor-scheer:
Decided offline with @abernix to leave config as-is (re: deno_core, anyway) and get an issue open to get it updated rather than pin it.

abernix (approved):

PR #1250, opened 2021-11-29
Clenfest/airbnb-defaults

clenfest:
Adding airbnb defaults to linter rules.

A lot of these rules are somewhat pedantic, so we can absolutely debate whether we want to exclude any or all of these. There are some that I've made into warnings that look like they might be pointing out real issues, so I think it will be good to review them.

Ok, it looks like there are a million changes, but there are really only a couple that matter. We can feel free to let this PR get out of date since it's easy to rerun autofix whenever we come to consensus. Since these are all safely autofixed, the review is really the types of rules we're adding rather than reviewing the changes individually.
- Add airbnb defaults to package.json and eslintrc
- run `eslint --fix` to auto fix all auto-fixable problems
- changed all errors to either warnings or turned off entirely.

I think that the next step going forward after merging this would be to have a PR for each of the rules that we've turned into warns.


composition-js/src/__tests__/matchers/toMatchString.ts
```diff
@@ -34,25 +34,26 @@ expect.extend({
...
   },
 
   toMatchStringArray(expected: string[], received: string[]) {
     if (expected.length !== received.length) {
-      const message = () => 
-        this.utils.matcherHint('toMatchStringArray', undefined, undefined,)
-          + `\n\nExpected an array of size ${expected.length} but got one of size ${received.length}\n\n`
-          + this.utils.printDiffOrStringify(expected, received, 'Expected', 'Received', true);
-      return {received, expected, message, name: 'toMatchStringArray', pass: false};
+      const message = () => `${this.utils.matcherHint('toMatchStringArray', undefined, undefined)
```
pcmanus, line 51: Fwiw, I think the rewrites of strings in that file (this one but others as well) make things quite a bit less unreadable (of course, readability is in parts subjective).

I get it's an automatic rewrite, but I think it's dodgy to rewrite multi-components strings automatically like this.

So if it was me, I'd disable whichever rule deal with this at first, and enable it later in its own commit so it's easier to apply/review manual changes when it gets too ugly. In this file for instance, it's probably worth extracting some things into `const` really. Not strongly insisting, it's probably not the end of the world, but I do think it's a step back.

pcmanus:
I personally lack sufficient experience in both writing typescript and with the airbnb rules to have too much of an opinion on the rules. Though in general, I'm in favour of bringing more consistency, so I'm broadly good with this.

But I think it's worth bringing this to the attention of at least @trevor-scheer, @martijnwalraven  and @abernix, as people broadly involved with this codebase, to get their opinion, if they have one.

composition-js/src/__tests__/matchers/toMatchString.ts
clenfest: Np, I disabled that rule for now.

trevor-scheer:
I don't have a strong opinion on this. Historically speaking, we've tended to veer away from tools which enforce code aesthetics (think specifically `prettier`) but I understand `es/ts-lint` goes a bit further and tries to address "smells" as well.

If we _are_ going to enforce aesthetic rules (which this seems to be doing), I'd ask why not `prettier`? I'm unfamiliar with the `es/ts-lint` tooling, but I can at least speak to `prettier` in VS Code being really great to use. It's worth calling out that all the aesthetic changes are pretty "noisy" w.r.t. git, though it's more of an inconvenience than actually problematic. And, it's noisy once and then hopefully never again since it's enforced by our tooling going forward.

Aesthetics aside, I'm most interested in the rules which warn on smells - that seems like an inarguable net positive to me. Also, for the sake of consistency I'd be interested in keeping the rules and tooling in sync with the `version-0.x` branch. While I don't think we need to go well out of our way to keep `version-0.x` in perfect shape, I think there's value in having the code as similar as possible where it applies for the sake of changesets which should affect both branches.

trevor-scheer:
@clenfest no idea on the turnaround time for PRs, but: https://github.com/prettier/prettier/pull/11901

It looks like ESLint offers the same kind of fix-on-save functionality in their extension. I think if we're happy with the formatting results, then I might not favor one over the other.

clenfest:
> @clenfest no idea on the turnaround time for PRs, but: [prettier/prettier#11901](https://github.com/prettier/prettier/pull/11901)
> 
> It looks like ESLint offers the same kind of fix-on-save functionality in their extension. I think if we're happy with the formatting results, then I might not favor one over the other.

Sweet! Thanks for doing that. It occurred to me that this was an option while I was eating dinner, but you managed to get a patch up already!

I think that ESlint is more likely to yell at you rather than fix things for you in the background (at least the way I've used it in the past), so it might make sense to prefer prettier as 'option A'. I think it probably makes sense to hold off for a bit to see if that PR makes it into a release so that we don't have to do multiple major formatting changes if that makes sense to you.

trevor-scheer:
SGTM! I don't mind waiting at all 👍

abernix:
I very much appreciate the ping!

I'm not opposed to sensible lints that prevent common coding errors (_smells_, above 👃 ), however, I am strongly opposed to tooling — like prettier — that does (and forces) wholesale automatic formatting on entire repositories.  While it touches on a lot of things, my commit on `apollo-server` (which was where this `federation` repo came from) explains my perspective pretty well, I think: https://github.com/apollographql/apollo-server/commit/55312289116d4577532f77e247ed2b1bae0bddcb.

I do personally still stand by this as I'm strongly against unnecessary levels of blame on what I find to be otherwise very useful Git history; for example, with this change, most of the Git blame on this history will say whatever the commit message of this PR ends up being!

That said, if this is all what this repository needs, then you should feel free and I abstain.  This is my perspective but it is no longer the place where I contribute the most.

gateway-js/src/__tests__/integration/networkRequests.test.ts
```diff
@@ -158,10 +162,10 @@ it('Updates Supergraph SDL from remote storage', async () => {
   gateway.onSchemaLoadOrUpdate(schemaChangeCallback);
 
   await gateway.load(mockApolloConfig);
-  expect(gateway['compositionId']).toMatchInlineSnapshot(`"originalId-1234"`);
+  expect(gateway['compositionId']).toMatchInlineSnapshot('"originalId-1234"');
```
trevor-scheer, line 165: I think this is going to be a problem. `prettier` is writing these snapshots under the hood and uses backticks when it does. As snapshots update, I expect the single quotes will be auto-rewritten to backticks. Is this something we can tune, or maybe something that specifically leaves inline snapshots alone?

clenfest: I'm not 100% sure what your concern is. I think the linter expects strings to be built with backticks iff there is a variable referenced. For example `\`originalId-${id}\``, but will use single quotes if there is not a variable present. But in either case, the string that we're passing to `toMatchInlineSnapshot` would be the same, no?

gateway-js/src/__tests__/integration/networkRequests.test.ts
```diff
@@ -499,9 +507,9 @@ describe('Downstream service health checks', () => {
 
       // Update
       mockSupergraphSdlRequestSuccessIfAfter(
-          'originalId-1234',
-          'updatedId-5678',
-          getTestingSupergraphSdl(fixturesWithUpdate),
+        'originalId-1234',
+        'updatedId-5678',
+        getTestingSupergraphSdl(fixturesWithUpdate as any),
```
trevor-scheer, line 512: Seems odd this was cast to `as any`. I have a related patch on my current branch that should resolve this, though I'm not sure why there are TS errors that didn't exist before.

```ts
// federation-integration-testsuite-js/src/fixtures/index.ts
export interface Fixture {
  name: string;
  url: string;
  typeDefs: DocumentNode;
  resolvers?: GraphQLResolverMap<any>
}

const fixtures: Fixture[] = [
  accounts,
  books,
  documents,
  inventory,
  product,
  reviews,
];

const fixturesWithUpdate: Fixture[] = [
  accounts,
  books,
  documents,
  inventory,
  product,
  reviewsWithUpdate,
];

const fixturesWithoutTag: Fixture[] = [
  accountsWithoutTag,
  books,
  documents,
  inventory,
  product,
  reviewsWithoutTag,
];
```

gateway-js/src/config.ts
```diff
@@ -20,7 +20,7 @@ export type Experimental_DidResolveQueryPlanCallback = ({
   readonly serviceMap: ServiceMap;
   readonly operationContext: OperationContext;
   readonly requestContext: GraphQLRequestContextExecutionDidStart<
-    Record<string, any>
+  Record<string, any>
```
trevor-scheer, line 23: Prefer the indentation here, can we disable this rule?

gateway-js/src/config.ts
```diff
@@ -131,8 +131,8 @@ interface GatewayConfigBase {
 export interface RemoteGatewayConfig extends GatewayConfigBase {
   serviceList: ServiceEndpointDefinition[];
   introspectionHeaders?:
-    | HeadersInit
-    | ((service: ServiceEndpointDefinition) => Promise<HeadersInit> | HeadersInit);
+  | HeadersInit
```
trevor-scheer, line 134: Maybe the same rule as before, but I think the indentation here is much nicer. Also, just my preference.

gateway-js/src/datasources/LocalGraphQLDataSource.ts
```diff
@@ -11,8 +11,7 @@ import { GraphQLDataSource, GraphQLDataSourceProcessOptions } from './types';
 
 export class LocalGraphQLDataSource<
   TContext extends Record<string, any> = Record<string, any>,
```
trevor-scheer, line 13: Seems like the indentation is inconsistent with this case vs. one I pointed out previously. Maybe the one I pointed out above is a bug?

gateway-js/src/datasources/RemoteGraphQLDataSource.ts
```diff
@@ -12,22 +12,25 @@ import {
...
-> implements GraphQLDataSource<TContext>
-{
+> implements GraphQLDataSource<TContext> {
   fetcher: typeof fetch = fetch;
 
   constructor(
     config?: Partial<RemoteGraphQLDataSource<TContext>> &
       object &
-      ThisType<RemoteGraphQLDataSource<TContext>>,
+    ThisType<RemoteGraphQLDataSource<TContext>>,
```
trevor-scheer, line 33: More odd indentation behavior, why isn't the above line also de-indented?

composition-js/src/__tests__/matchers/toMatchString.ts
trevor-scheer: Looks like changes from this rule are still applied throughout unless I'm imagining the rule incorrectly.

trevor-scheer (commented):
I've expressed my on-the-fenceness to you about these changes already so hopefully this opinion isn't too surprising...

My biggest concern with strictly formatting changes is the git noise/history. There are some formatting changes and style opinions which I think add value and might be worth eating that cost, but I don't see the value add from a lot of the rules we've turned on, and in some cases the changes are significantly worse.

Would you be open to _not_ taking a wholesale airbnb approach and instead adding rules explicitly that we think inarguably add value? I'm seeing a lot of opinionated changes which are "subjectively better" that I either disagree with completely or feel like there needn't be an opinion on at all.

gateway-js/src/config.ts
clenfest: I agree that this is uglier, but if this rule is customizable for this specific case I can't figure it out and I think we do definitely want this rule in general. I think the best solution may be to just not put a line break in between <>. I'd be willing to bet without looking that this is one of the rules prettier disables so it can do its own thing

composition-js/src/__tests__/matchers/toMatchString.ts
clenfest: Sorry, I think too much time passed and I lost the commit where I disabled it.

gateway-js/src/__tests__/integration/networkRequests.test.ts
trevor-scheer: Yeah, the issue is with jest/prettier rather than the linter. It's minor, but it is an issue.

If you go to update this test with a different id: https://github.com/apollographql/federation/blob/b0c3db7451d6f868e613933f8983b7648e4bf4aa/gateway-js/src/__tests__/integration/networkRequests.test.ts#L136
And run the tests with snapshot update flag `-u` this works as expected on `main`, but `prettier` (via jest) throws an error if you try to do the same thing on this branch.

gateway-js/src/config.ts
trevor-scheer: Was*

clenfest:
Here's the list of rules that in some way touch this commit along with the number of occurrences. It'll take some time to go through them because there are also some interactions at play that might make things more difficult. For example, I think `object-curly-newline` is a very good rule that makes things more readable, but if you disable `@typescript-eslint/indent` it basically becomes impossible to use because every time `--fix` adds a newline, it won't indent properly.

```
  14 @typescript-eslint/brace-style
 395 @typescript-eslint/comma-dangle
  34 @typescript-eslint/comma-spacing
 293 @typescript-eslint/indent
   6 @typescript-eslint/keyword-spacing
  92 @typescript-eslint/lines-between-class-members
 249 @typescript-eslint/object-curly-spacing
 402 @typescript-eslint/quotes
   1 @typescript-eslint/return-await
 166 @typescript-eslint/semi
   2 @typescript-eslint/space-before-function-paren
   8 @typescript-eslint/space-infix-ops
  29 array-bracket-spacing
  31 arrow-body-style
 336 arrow-parens
   5 block-spacing
   3 curly
   1 dot-location
   5 eol-last
  45 function-paren-newline
  12 generator-star-spacing
   4 import/first
   8 import/newline-after-import
  10 import/no-duplicates
  10 import/no-useless-path-segments
  83 import/order
   9 key-spacing
   1 new-parens
  25 no-confusing-arrow
  44 no-else-return
  22 no-multi-spaces
  31 no-multiple-empty-lines
   1 no-restricted-properties
   1 no-spaced-func
   8 no-trailing-spaces
   6 no-undef-init
   7 no-unneeded-ternary
   2 no-useless-return
   3 nonblock-statement-body-position
1552 object-curly-newline
  13 object-shorthand
   3 operator-assignment
  10 padded-blocks
   8 prefer-const
   1 prefer-exponentiation-operator
   2 prefer-object-spread
 164 prefer-template
   6 quote-props
   2 semi-spacing
   5 space-before-blocks
   2 space-in-parens
   9 spaced-comment
```

clenfest:
Ok, going to quickly go through the above list with more than a handful of occurrences and either make the case or decide to pass. Spoiler alert, I keep most everything.

object-curly-newline - I think this is a great rule. But, there are a lot of occurrences of it and that's going to mess up some history. And it requires @typescript-eslint/indent which messes up formatting for other things (generics). Even with all that, I'd say that it'd be worth it, but I went through the code looking for occurrences of abuse that I think this rule would make saner, and I didn't really find a ton. So we'll pass on it.

@typescript-eslint/quotes - I like this rule for aesthetics, but I set `allowTemplateLiterals` to true which should take care of the jest snapshot thing.

@typescript-eslint/comma-dangle - I like this rule a lot, it makes moving things around within the code much easier if commas are all present. This one is a goer.

arrow-parens - If you want to add types to anonymous function parameters, it's easier if the parens are already present. 

@typescript-eslint/indent - As discussed above, there is weirdness for generics, and this rule is unmaintained because of some of the issues with it. Prettier would definitely be the preferable choice, but even with that out, we'll pass on this rule.

@typescript-eslint/object-curly-spacing - again, I just like consistency, so we'll keep this one.

@typescript-eslint/semi - I like my statements to end in semicolons

prefer-template - We discussed this above. I do think it's more readable in general, but given that you think the opposite in a couple of cases, we'll leave it out.

@typescript-eslint/lines-between-class-members - adding whitespace for readability, no issues with this one.

import/order - not having this one drives me a bit crazy. I like being able to find imports

function-paren-newline - I like this one

no-else-return - this one is a code safety thing. Exposes unnecessary else statements.

@typescript-eslint/comma-spacing - Consistency

no-multiple-empty-lines - whitespace only

arrow-body-style - this one could be considered pedantic, but I like it. (we use the `as-needed` variation). If you feel differently, let me know and I'll take it out.

array-bracket-spacing - consistency

no-confusing-arrow - no strong opinions on this one, but

no-multi-spaces - consistency

At this point we don't have anything else that has more than 14 occurrences, so I'd just be inclined to keep everything. I'm going to commit a new version that runs off of these rules. Take a look and let me know what you think.

gateway-js/src/executeQueryPlan.ts
```diff
@@ -108,10 +106,10 @@ export async function executeQueryPlan<TContext>(
             fieldResolver: defaultFieldResolverWithAliasSupport,
           }));
         } catch (error) {
-          span.setStatus({ code:SpanStatusCode.ERROR });
+          span.setStatus({ code: SpanStatusCode.ERROR });
           if (error instanceof GraphQLError) {
             return { errors: [error] };
-          } else if (error instanceof Error) {
+          } if (error instanceof Error) {
```
trevor-scheer, line 112: There are a few cases of this, can the `if`s get new lines?

gateway-js/src/executeQueryPlan.ts
```diff
@@ -121,43 +119,40 @@ export async function executeQueryPlan<TContext>(
...
             };
-          } else {
+          }
             // The above cases should cover the known cases, but if we received
             // something else in the `catch` — like an object or something, we
             // may not want to merely return this to the client.
             logger.error(
-              "Unexpected error during query plan execution: " + error);
+              'Unexpected error during query plan execution: ' + error,
+);
```
trevor-scheer, line 131: Indentation bug?

gateway-js/src/index.ts
```diff
@@ -108,10 +107,8 @@ export function getDefaultFetcher() {
   });
 }
 
-export const HEALTH_CHECK_QUERY =
-  'query __ApolloServiceHealthCheck__ { __typename }';
-export const SERVICE_DEFINITION_QUERY =
-  'query __ApolloGetServiceDefinition__ { _service { sdl } }';
+export const HEALTH_CHECK_QUERY = 'query __ApolloServiceHealthCheck__ { __typename }';
```
trevor-scheer, line 110: Confusing / inconsistent opinions: no extra whitespace between variable declarations but enforced extra whitespace between class properties?

gateway-js/src/index.ts
```diff
@@ -292,47 +304,47 @@ export class ApolloGateway implements GraphQLService {
       // only using JSON.stringify on the DocumentNode (and thus doesn't account
       // for unicode characters, etc.), but it should do a reasonable job at
       // providing a caching document store for most operations.
-      maxSize: Math.pow(2, 20) * (approximateQueryPlanStoreMiB || 30),
+      maxSize: 2 ** 20 * (approximateQueryPlanStoreMiB || 30),
```
trevor-scheer, line 307: Disagree with this rule, I'd argue it's now less readable since I have to consider order of operations

gateway-js/src/index.ts
```diff
@@ -516,8 +528,8 @@ export class ApolloGateway implements GraphQLService {
               compositionMetadata: this.compositionMetadata,
             }),
           },
-          previousServiceDefinitions &&
-            previousSchema && {
+          previousServiceDefinitions
+            && previousSchema && {
```
trevor-scheer, line 532: Interesting it didn't put the following && in front and on a new line

internals-js/src/definitions.ts
```diff
@@ -1697,7 +1711,7 @@ abstract class FieldBasedType<T extends (ObjectType | InterfaceType) & NamedSche
     return toAdd;
   }
 
-  *allChildElements(): Generator<NamedSchemaElement<any, any, any>, void, undefined> {
+  * allChildElements(): Generator<NamedSchemaElement<any, any, any>, void, undefined> {
```
trevor-scheer, line 1714: Just noisy. Can we remain unopinionated about where the generator star goes?

trevor-scheer (commented):
Stopping for now. Thank you for listing out and justifying the rules, I think that's the stance we need to take towards these.

In the spirit of that, I think it's odd for us to start with a bunch of rules on that we have to then selectively _disable_. I think the result of this is a number of arbitrary rules that land in our config which are only there because they weren't contested, _not_ because we justified their addition.

Can we instead take the approach of selectively _adding_ the rules that introduce value that's agreed upon?

clenfest:
Yeah, so I think we just have a bit of a difference of philosophy. In general I like some consistency so that when you're reading code it isn't very obvious the differences in style of the individual author and that the code basically reads the same no matter where in the codebase you are. The advantage to using the airbnb rules was supposed to be that they are fairly widely adopted and vetted and that it wouldn't be necessary to litigate each rule individually (especially the ones where an autofix is possible), but it seems like we're finding ourselves there in any case. Let's talk about it in person when we meet up tomorrow.

trevor-scheer:
Spoke offline, I'm content to disagree and commit and allow @clenfest to have the final say. I think we both agree that this is a decision that Team Atlas gets to make 👍

I'll take one final pass at important files (package.json, tsconfig, etc.) after conflicts are resolved and approve!

PR #1252, opened 2021-11-29
Edits to entity and field migration docs

StephenBarlow:


docs/source/entities.mdx
```diff
@@ -205,53 +207,352 @@ type Query {
...
+      id: ID!
+      amount: Int!
+    }
+    ```
 
-We can perform this migration safely with the following steps:
+    </CodeColumns>
 
-1. In the `billing` subgraph's schema, define the `Bill` entity just as it's defined in the `payments` subgraph (do _not_ extend it). If you're using managed federation, register this schema change with Studio.
+    * Note that if you perform composition at this point, it produces an error because the `Bill` entity can't originate in more than one subgraph. That's okay! We'll resolve this error in a couple of steps.
```
pcmanus, line 283: Maybe, for extra clarity, something like: "That's okay, we don't run composition at this point and we'll resolve this error ..."?

docs/source/entities.mdx
```diff
@@ -205,53 +207,352 @@ type Query {
...
+      id: ID!
+      amount: Int!
+    }
+    ```
+
+    </CodeColumns>
+
+2. Publish the updated Billing subgraph schema to Apollo.
+
+    * Note that you're publishing a composition error, because the `Bill` entity now originates in more than one subgraph. That's okay! Your gateway continues to use your most recent _valid_ supergraph schema.
```
pcmanus, line 357: I actually haven't tried this with managed federation, but I assume that the gateway will _try_ composing automatically through Uplink and you'll get an error showing somewhere? If so, could be nice to be precise as to where the error shows up and be extra clear that this should be ignored.

pcmanus (approved):
Left minor comments but it's already quite good as is. Great improvement I think.

pcmanus:
@StephenBarlow can you let me know when this is get published (or update #750 yourself)? Thanks.

StephenBarlow:
This is now live, and I've responded to the issue. (We have a pesky issue with the Federation docset where we need to push at least one update to `main` for it to pick up updates to `version-0.x`)

Issue #844, opened 2021-06-29
`make-fetch-happen` Doesn't Retry POST


j:
I'm not sure if this is intended, but I figured the use of `make-fetch-happen` was to take advantage of a few things, and a major one being retries.

Since GraphQL is really only POST (unless configured for GET, though not in Federation without custom data sources), the retries really never happen.

I think a custom retry implementation can actually fix a lot of startup issues people might have.

PR #1284, opened 2021-12-09
RemoteGraphQLDataSource uses `make-fetch-happen` by default

### LINKED ISSUE #192: Apollo Gateway: Add make-fetch-happen fetcher as the default fetcher for the downstream services
Currently the `RemoteGraphQlDataSources` uses node-fetch as the default fetcher, it will be good if we switch this to `make-fetch-happen` fetcher which provides us with additional features and provides support for default request options which is very much required in the real world use cases.

### LINKED ISSUE #1232: Enable connection keep-alive on `RemoteGraphQLDataSource`s by default.
In Apollo Gateway, the default fetcher that the `RemoteGraphQLDataSource` uses doesn't configure a Node.js [`http.Agent`](https://nodejs.org/api/http.html#class-httpagent) with the [`keepAlive: true` option](https://nodejs.org/api/http.html#new-agentoptions).  This means that the default behavior is to destroy the underlying TCP sockets when there are not active connections rather than maintaining a pool of established connections.

The current default fetcher in Apollo Gateway is `node-fetch`.  Some alternative fetchers support having keep-alive on by default, but it's also possible to configure `node-fetch` to have this option enabled by default, [as documented in their documentation](https://github.com/node-fetch/node-fetch#custom-agent).

We've found a number of reasons in the past to recommend [`make-fetch-happen`](https://github.com/zkat/make-fetch-happen) — notably it's out of the box support for proxies, conditional cache control headers, gzip/deflate support, and its request pooling.  We should:

- Evaluate whether it's feasible to change our default fetcher to `make-fetch-happen`, which solves this.  This is a bit of a larger change and something we've punted before.  This is actually tracked in https://github.com/apollographql/federation/issues/192, so if we take that approach (which has some comments with reading on it and its own associated references!), we may be able to just close this issue as resolved.
- If that approach is deemed too large, we may want to just keep the default fetcher as `node-fetch` but configure it to use `keepAlive: true`.

clenfest:
As the subject says. Couple of things that I'm not sure about.

1. When I make a call to `.defaults()` do I need to pass any extra parameters? Didn't look like the `node-fetch` version did, but I want to make sure we don't want to add any extra headers or anything. For reasons stated in #192, I'm being careful and turning off retries completely, but I believe this was also the case in `node-fetch`

2. Obviously this needs to get into the CHANGELOG, but I'm assuming we do that when we cut a new version?

Fixes #192
Fixes #1232

### DIFF
--- a/gateway-js/src/datasources/RemoteGraphQLDataSource.ts
+++ b/gateway-js/src/datasources/RemoteGraphQLDataSource.ts
@@ -30,6 +30,9 @@ export class RemoteGraphQLDataSource<
       ThisType<RemoteGraphQLDataSource<TContext>>,
   ) {
     this.fetcher = fetcher.defaults({
+      // although this is the default, we want to take extra care and be very
+      // explicity to ensure that mutations cannot be retried. please leave this
+      // intact.
       retry: false,
     });
     if (config) {
--- a/gateway-js/src/datasources/RemoteGraphQLDataSource.ts
+++ b/gateway-js/src/datasources/RemoteGraphQLDataSource.ts
@@ -17,18 +17,21 @@ import { isObject } from '../utilities/predicates';
 import { GraphQLDataSource, GraphQLDataSourceProcessOptions, GraphQLDataSourceRequestKind } from './types';
 import createSHA from 'apollo-server-core/dist/utils/createSHA';
 import { parseCacheControlHeader } from './parseCacheControlHeader';
-
+import fetcher from 'make-fetch-happen';
 export class RemoteGraphQLDataSource<
   TContext extends Record<string, any> = Record<string, any>,
 > implements GraphQLDataSource<TContext>
 {
-  fetcher: typeof fetch = fetch;
+  fetcher: typeof fetch;
 
   constructor(
     config?: Partial<RemoteGraphQLDataSource<TContext>> &
       object &
       ThisType<RemoteGraphQLDataSource<TContext>>,
   ) {
+    this.fetcher = fetcher.defaults({
+      retry: false,
+    });
     if (config) {
       return Object.assign(this, config);
     }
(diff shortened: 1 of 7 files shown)


gateway-js/src/datasources/RemoteGraphQLDataSource.ts
```diff
@@ -17,18 +17,21 @@ import { isObject } from '../utilities/predicates';
...
+  fetcher: typeof fetch;
 
   constructor(
     config?: Partial<RemoteGraphQLDataSource<TContext>> &
       object &
       ThisType<RemoteGraphQLDataSource<TContext>>,
   ) {
+    this.fetcher = fetcher.defaults({
+      retry: false,
+    });
```
trevor-scheer, line 34: This is already the default, but fine if you prefer to specify: https://github.com/npm/make-fetch-happen/blob/main/lib/options.js#L15
```suggestion
    this.fetcher = fetcher;
```

Edit: I just read the comment from @abernix about being very deliberate about this. If that's the case, I'd add a comment here so a person like me doesn't "clean this up" by removing it as I've suggested above.

gateway-js/src/datasources/__tests__/RemoteGraphQLDataSource.test.ts
```diff
@@ -264,7 +264,7 @@ describe('fetcher', () => {
   });
 
   it('supports a custom fetcher, like `make-fetch-happen`', async () => {
```
trevor-scheer, line 266:
```suggestion
  it('supports a custom fetcher, like `node-fetch`', async () => {
```

trevor-scheer (approved):
This LGTM. It should get a CHANGELOG entry with this PR (ping if you'd like any guidance there).

As a side note, I'd like to see us move away from mocking the fetcher objects and just using `nock` instead. Mocking the module (as is the status quo in this case) adds a layer of convention / magic which isn't well-documented and isn't really the convention for mocking network requests within this repo (typically we use `nock`). I'll open a separate issue for this.

trevor-scheer:
Also, would like to land this on `version-0.x` as well!

gateway-js/src/datasources/RemoteGraphQLDataSource.ts
abernix: I wonder if we should open an issue to force-attempt `POST` retries (typically, a no no) on `query`s.  Users have definitely asked for retries here but maybe worth creating a conversation about it.  It's possible this is covered by https://github.com/apollographql/federation/issues/844, but I'm not sure if that's referring to `RemoteGraphQLDataSource`s or some internal Apollo agent stuff.

prasek:
Agree created #1379 for the backport

PR #1292, opened 2021-12-10
Implement deno-core required snapshot support for router-bridge

garypen:
I wondered if loading snapshots to avoid repeating environment
initialisation might be a speedup. It looks like it is.

Q1:
hey -m POST -n 300000 -c 100 \
    -H 'content-type: application/json' \
    -d '{"query":"{\n  topProducts {\n    upc\n  }\n}"}' http://127.0.0.1:4000

Q2:
hey -m POST -n 300000 -c 100 \
    -H 'Content-type: application/json' \
    -d '{"query":"query  { products(limit: 2) { review { body } related { price price1 price2 price3 } } }\n","price price1 price2 price3 variables":{}}' http://127.0.0.1:4000/

Requests/Second

Query   Unmodified Result   Snapshot Result     Improvement
Q1      1378                1799                30%
Q2      2304                4214                82%


garypen:
I ran a performance check with a version of the router based on main where the only difference was that I updated the federation repo dependency to use a githash from my PR. The performance results were slightly better than on main, which is what I would expect.

baseline main : 6950 rps
router based on this PR: 7191 rps

garypen:
fixes: #1299

o0Ignition0o (approved):
LGTM.

can we replace `.args(&["install"])` with `.args(&["ci"])` in build.rs at line 21?

We can also do this as a followup

garypen:
> LGTM.
> 
> can we replace `.args(&["install"])` with `.args(&["ci"])` in build.rs at line 21?
> 
> We can also do this as a followup

Done in: https://github.com/apollographql/federation/pull/1292/commits/499c7071616cfd97fb051830fc0df07e345d9062

PR #1304, opened 2021-12-13
Removed unused `@apollographql/apollo-tools` dependency

t-botz:
I was wondering where apollo-tools was used and couldn't find any reference to it.
CI also seem to pass without it...


trevor-scheer (approved):
Thanks @t-botz !

trevor-scheer:
^ pushed to `version-0.x` branch as well.

PR #1305, opened 2021-12-14
Ensure error resolving an entity don't impact resolvable ones (for `version-0.x` branch)

pcmanus:
This is a rebased and modified version of https://github.com/apollographql/apollo-server/pull/3914. The differences with that earlier PR are that:
1. it provides a tad more test coverage, but also use tests that don't reuse existing fixtures (I happen to think that having common fixtures reused by all tests is not a good idea, and that has created issues in the past, so I'm advocating from moving away from doing so).
2. the fix is slightly different because:
  - it doesn't log an error in `executeSelectionSet` anymore: the reasoning (explained in a comment in the code) is that another error has already been logged and that error has more context, so adding a new error is redundant and I think it's confusing because that particular error lacks context and so doesn't really tell you what the problem is. So at best, you happen to know it's just a consequence of a previous error and ignore it, but better not have it then.
  - it uses a `return null` instead of a `break`: I personally think that it's unreasonable that when you require a non-nullable field then you may actually have to resolve without that required field. If you knew how to do that, then you wouldn't a `@require` for that field in the first place. So if you can't get all the requires of fetch (and thus, in particular, of a `@require`), then I think we should simply not do that fetch for the entity at all, which is what `return null` does. Using `break` instead means we'd attempt the fetch nonetheless but with only parts of the inputs present. Note that if a required field is nullable, then this code don't get trigger and we correctly do the fetch with `null` values (I added tests that show it's the case).

### DIFF
--- a/gateway-js/src/__tests__/executeQueryPlan.test.ts
+++ b/gateway-js/src/__tests__/executeQueryPlan.test.ts
@@ -1636,4 +1636,640 @@ describe('executeQueryPlan', () => {
       });
     });
   });
+
+  describe('@requires', () => {
+    test('handles null in required field correctly (with nullable fields)', async () => {
+      const s1 = {
+        name: 'S1',
+        typeDefs: gql`
+          type T1 @key(fields: "id") {
+            id: Int!
+            f1: String
+          }
+        `
+      }
+
+      const s2 = {
+        name: 'S2',
+        typeDefs: gql`
+          type Query {
+            getT1s: [T1]
+          }
+
+          extend type T1 @key(fields: "id") {
+            id: Int! @external
+            f1: String @external
+            f2: T2 @requires(fields: "f1")
+          }
+
+          type T2 {
+            a: String
+          }
+        `
+      }
+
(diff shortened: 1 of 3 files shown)


pcmanus:
@trevor-scheer wouldn't mind your opinion here since you followed earlier iterations of this.

gateway-js/src/__tests__/executeQueryPlan.test.ts
```diff
@@ -1636,4 +1636,640 @@ describe('executeQueryPlan', () => {
...
+        T1: {
+          __resolveReference(ref) {
+            // the ref has already the id and f1 is a require is triggered, and we resolve f2 below
+            return ref;
+          },
+          f2(o) {
+            return o.f1 === null ? null : { a: `t1:${o.f1}` };
+          }
+        }
+      });
```
trevor-scheer, line 1702: Nothing wrong here, but these resolvers can be added to the fixture object itself i.e.
`const s1 = { name, typeDefs, resolvers };`

trevor-scheer (approved):
LGTM!

PR #1306, opened 2021-12-14
Ensure error resolving an entity don't impact resolvable ones (for `main` branch)

pcmanus:
This PR has the same fix than in https://github.com/apollographql/federation/pull/1305, but for `main`.

_However_, testing this also uncovered another issue with `@requires` on `main` (fed 2 code), and this PR also has this additional fix. So first commit is that "other" fix (but it includes all the tests), and second commit is the fix of https://github.com/apollographql/federation/pull/1305.

### DIFF
--- a/gateway-js/src/executeQueryPlan.ts
+++ b/gateway-js/src/executeQueryPlan.ts
@@ -489,7 +489,17 @@ function executeSelectionSet(
         const selections = (selection as QueryPlanFieldNode).selections;
 
         if (typeof source[responseName] === 'undefined') {
-          throw new Error(`Field "${responseName}" was not found in response.`);
+          // This method is called to collect the inputs/requires of a fetch. So, assuming query plans are correct
+          // (and we have not reason to assume otherwise here), all inputs should be fetched beforehand and the
+          // only reason for not finding one of the inputs is that we had an error fetching it _and_ that field
+          // is non-nullable (it it was nullable, error fetching the input would have make that input `null`; not
+          // having the input means the field is non-nullable so the whole entity had to be nullified/ignored,
+          // leading use to not have the field at all).
+          // In any case, we don't have all the necessary inputs for this particular entity and should ignore it.
+          // Note that an error has already been logged for whichever issue happen while fetching the inputs we're
+          // missing here, and that error had much more context, so no reason to log a duplicate (less useful) error
+          // here.
+          return null;
         }
         if (Array.isArray(source[responseName])) {
           result[responseName] = source[responseName].map((value: any) =>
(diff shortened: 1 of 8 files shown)


pcmanus:
@clenfest: the second commit here is going to be review as part of #1305/#376 and feel free to skip it (though certainly happy with more eyes on it), but the 1st commit here is specific to the fed2, and I think you might be the most qualified to review. Do you mind?

clenfest:
Maybe it's because it's a little late in the day for me, but I think I would probably greatly benefit from a walk through of this code. Would you maybe have time tomorrow? Either way I'll dig in deeper tomorrow.

gateway-js/src/executeQueryPlan.ts
```diff
@@ -489,7 +489,17 @@ function executeSelectionSet(
         const selections = (selection as QueryPlanFieldNode).selections;
 
         if (typeof source[responseName] === 'undefined') {
```
clenfest, line 491: should we simplify to 
```typescript
if (source[responseName] === undefined)
```

query-graphs-js/src/querygraph.ts
```diff
@@ -353,6 +360,16 @@ export class QueryGraph {
     const indexes = this.typesToVertices.get(typeName);
     return indexes == undefined ? [] : indexes.map(i => this.vertices[i]);
   }
+
+  externalTester(source: string) {
```
clenfest, line 364: nit: return type

query-planner-js/src/buildPlan.ts
```diff
@@ -1637,7 +1642,9 @@ function inputsForRequire(graph: QueryGraph, entityType: ObjectType, edge: Edge,
   fullSelectionSet.add(new FieldSelection(new Field(entityType.typenameField()!)));
   fullSelectionSet.mergeIn(edge.conditions!);
   if (includeKeyInputs) {
-    fullSelectionSet.mergeIn(additionalKeyEdgeForRequireEdge(graph, edge).conditions!);
+    const keyCondition = getLocallySatisfiableKey(graph, edge.head);
+    assert(keyCondition, () => `No key found for require on ${edge}`);
```
clenfest, line 1646: Maybe a slightly better error message? Suggestion: "Use of @require mandates a @key and no key was found on ${edge}"

clenfest (commented):
Couple of super minor comments

gateway-js/src/executeQueryPlan.ts
clenfest: Oops, did you not touch this line? Never mind.

clenfest (approved):

PR #1317, opened 2021-12-16
use @apollo/client in favor of deprecated apollo-link and apollo-link-http

clenfest:
Doing this as a part of upgrading to graphql 0.16.x so that deprecated packages won't need to update peerDependencies. Given that the required functions are only used in one test, it's probably worth thinking about whether it's worth it to bring in such a big package, but it's only a dev dependency.


trevor-scheer:
Tangentially related: https://github.com/apollographql/federation/pull/1237

And adding my approval in post!

Issue #1257, opened 2021-12-01
Interface fragment expansion causes conflicting field types


MarcPorciuncula:
I have an interface with a field where there is a nullable field for which one implementer has the field marked nullable and the other field is marked non-nullable. (I've changed the names of the types from the originals, but the concept is the same).

```graphql
interface Post {
  likes: Int
}

type Tweet implements Post {
  likes: Int!
}

type Retweet implements Post {
  # A quote retweet can have likes but a normal retweet does not
  likes: Int
}

type Query {
  post(id: ID!) Post
}
```

When querying against the interface, the field is nullable regardless of the returned type.

```graphql
query GetPost($id: ID!) {
  post(id: $id) {
    likes
  }
}
```

It seems that when receiving a query like this, the gateway will use information about the implementers of the interface to expand the fields under `post` into fragments like this:

```graphql
post(id: $id) {
  __typename
  ... on Tweet {
    likes
  }
  ... on Retweet {
    likes
  }
}
```

In this query the type of `post.likes` could be `Int` or `Int!`. This results in a GraphQL validation error because there are two different types trying to be bound to the same field.

```
Fields "likes" conflict because they return conflicting types "Int!" and "Int". Use different aliases on the fields to fetch both if this was intentional.
```

I believe the way gateway expands interface fields into inline fragments is incorrect or problematic as it causes problems like these. If I ran the source query directly on the subgraph it would have no problem but the changes the gateway makes to the query result in an invalid query.

Potential solutions would be to not expand interface fields into inline fragments, or to give the common field in the fragments different aliases (as suggested by the error message), then rename them to the original name before returning the query.

---
Here are the packages I used
| Package | Version |
| -------- | -------- | 
| `@apollo/gateway` (gateway) | `0.43.1` |
| `apollo-server` (gateway) | `3.5.0` |
| `@apollo/subgraph` (subgraph) | `0.1.4` |
| `apollo-server-koa` (subgraph) | `3.4.0` |
| `graphql` (subgraph) | `15.7.2` |

pcmanus:
Thanks for the report. And you are obviously correct, this is an issue.

Unfortunately, for the current `0.x` versions of federation, I don't think there is an easy fix:
- expanding interface into their runtime types is fairly ingrained into how the query planner work (and needed in many cases), and changing it is further made harder by the fact that intermediate representation generated by composition (what we call the supergraph) lacks (in those current versions) enough information about interface to do so.
[...]

PR #1318, opened 2021-12-16
Error out if differing interface field implementations may lead to later issue

pcmanus:
As mentioned in the commit message, this not a fix for #1257 but it ensures a much cleaner "failure mode" until that issue is properly fixed.

The patch is on top of #1274 because that later PR adds a new test file that is perfect for this, and I wanted to avoid needing to rebase after it.

### DIFF
--- a/internals-js/src/federation.ts
+++ b/internals-js/src/federation.ts
@@ -26,7 +26,7 @@ import {
   InputFieldDefinition,
   isCompositeType
 } from "./definitions";
-import { assert, OrderedMap } from "./utils";
+import { assert, joinStrings, MultiMap, OrderedMap } from "./utils";
 import { SDLValidationRule } from "graphql/validation/ValidationContext";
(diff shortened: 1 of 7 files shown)


internals-js/src/federation.ts
```diff
@@ -311,6 +311,44 @@ function isFieldSatisfyingInterface(field: FieldDefinition<ObjectType | Interfac
   return field.parent.interfaces().some(itf => itf.field(field.name));
 }
 
+function validateInterfaceRuntimeImplementationFieldsTypes(
```
clenfest, line 314: Maybe just add some comments around exactly what this function does.

clenfest (approved):

clenfest (approved):

internals-js/src/federation.ts
pcmanus: Added, thanks.

PR #1319, opened 2021-12-16
Fix issue when reducing the number of plans to consider

### LINKED ISSUE #1316: Duplicate Parallel calls to federated services with Federation 2
Query planner generates - Duplicate parallel calls to the federated services
Repository to reproduce the error (https://github.com/AneethAnand/federation2)

### Graphql Query
```
query test {
      viewer {
        user {
          user1
          user2
          id
          user3
          network {
                  id
              network1
          }
          user4
          user5
          user6
          user7
          user8
        }
        viewer1
        viewer2
        viewer3
    }
}
```
#### Actual QueryPlan
```
QueryPlan {
  Parallel {
    Fetch(service: "A1") {
      {
        viewer {
          user {
            user1
            user2
            id
            user3
            network {
              id
              network1
            }
            user4
            user5
            user6
            user7
            user8
          }
          viewer1
          viewer2
          viewer3
        }
      }
    },
    Fetch(service: "V1") {
      {
        viewer {
          user {
            user2
            id
            user3
            network {
              id
              network1
            }
            user4
            user5
            user6
            user7
            user8
          }
          viewer1
          viewer2
          viewer3
        }
      }
    },
  },
}
```
#### Expectation 
```
    Fetch(service: "A1") {
      {
        viewer {
          user {
            user1
            user2
            id
            user3
            network {
              id
              network1
            }
            user4
            user5
            user6
            user7
            user8
          }
          viewer1
          viewer2
          viewer3
        }
      }
    }
```

###  Remove one of the field say "viewer3" in the query, the gateway works as expected.
```
QueryPlan {
  Fetch(service: "A1") {
    {
      viewer {
        user {
          user1
          user2
          id
          user3
          network {
            id
            network1
          }
          user4
          user5
          user6
          user7
          user8
        }
        viewer1
        viewer2
      }
    }
  },
}
```

pcmanus:
When the number of possible plans for a query is too high to generate
all plans, we use a simple strategy to discard some options, but that
strategy had a trivial bug (a simple loop that was reordering an element
in an array was mishandling the case where the element had to move to
the end) leading to invalid plans.

Fixes #1316

### DIFF
--- a/query-planner-js/src/buildPlan.ts
+++ b/query-planner-js/src/buildPlan.ts
@@ -77,7 +77,8 @@ const debug = newDebugLogger('plan');
 // at to get it below this number to void query planning running forever.
 // Note that this number is a tad arbitrary: it's a nice round number that, on my laptop, ensure query planning don't
 // take more than a handful of seconds.
-const MAX_COMPUTED_PLANS = 10000;
+// Note: exported so we can have a test that explicitly requires more than this number.
+export const MAX_COMPUTED_PLANS = 10000;
 
 function mapOptionsToSelections<RV extends Vertex>(
   selectionSet: SelectionSet,
@@ -256,6 +257,21 @@ class QueryPlanningTaversal<RV extends Vertex> {
     return FetchDependencyGraph.create(this.subgraphs);
   }
 
+  // Moves the first closed branch to after any branch having more options.
+  // This method assumes that closed branches are sorted by decreasing number of options _except_ for the first element
+  // which may be out of order, and this method restore that order.
+  private reorderFirstBranch() {
+    const firstBranch = this.closedBranches[0];
+    let i = 1;
+    while (i < this.closedBranches.length && this.closedBranches[i].length > firstBranch.length) {
+      i++;
+    }
+    // `i` is the smallest index of an element having the same number or less options than the first one,
+    // so we switch that first branch with the element "before" `i` (which has more elements). 
+    this.closedBranches[0] = this.closedBranches[i - 1];
+    this.closedBranches[i - 1] = firstBranch;
+  }
+
   private computeBestPlanFromClosedBranches() {
     if (this.closedBranches.length === 0) {
       return;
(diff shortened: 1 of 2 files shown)


pcmanus:
I'll note that the bug is actually pretty trivial and don't strictly require a full understanding of query planning to understand the issue. Essentially, the code has an array of array sorted by the inner array lengths (longest array first), and at some point we remove an element from the first inner array and try to move it to keep the outer array still sorted by inner array length. It's that very simple step that has a bug where if the array had to move to the very end of the outer array, it's not actually moved. This ends up breaking the sorting and the remainder of the code which rely on that sorting is thrown off and it results in an invalid query plan.

query-planner-js/src/buildPlan.ts
```diff
@@ -256,6 +257,21 @@ class QueryPlanningTaversal<RV extends Vertex> {
     return FetchDependencyGraph.create(this.subgraphs);
   }
 
+  // Moves the first closed branch to after any branch having more options.
+  private reorderFirstBranch() {
+    const firstBranch = this.closedBranches[0];
+    let i = 1;
+    while (i < this.closedBranches.length && this.closedBranches[i].length > firstBranch.length) {
```
clenfest, line 264: I'm not sure if this is not working or if I just don't understand what is desired.
```text
Suppose [4,2,7,9]. In that case i will stay at 1 resulting in [2,4,7,9]
Suppose [2,5,4,7]. In this case we'll increment to the end and swap, resulting in [7,5,4,2]
Suppose [2,5,1,7]. In this case we'll increment until we find `1` and swap, resulting in [1,5,2,7]
```
That's the behavior we want?

Also I think that if you have an array of two elements, they will be swapped regardless of what the values are. That seems wrong.

query-planner-js/src/buildPlan.ts
```diff
@@ -256,6 +257,21 @@ class QueryPlanningTaversal<RV extends Vertex> {
...
+      i++;
+    }
+    // `i` is the smallest index of an element having the same number or less options than the first one,
+    // so we switch that first branch with the element "before" `i` (which has more elements). 
+    if (i > 0) {
```
clenfest, line 269: This if statement doesn't do anything since i will always be greater than 1.

query-planner-js/src/buildPlan.ts
clenfest: Oh, missed your comment that said it's already sorted. I still think there is an issue where it'll always swap even if it shouldn't.

[9,8] -> [8,9]

query-planner-js/src/buildPlan.ts
pcmanus: Indeed, force of habit of starting iterations at 0. Removed it.

query-planner-js/src/buildPlan.ts
pcmanus: > missed your comment that said it's already sorted

Yeah, added a comment to the method itself as that wasn't very clear.

> I still think there is an issue where it'll always swap even if it shouldn't.

Is it? If the input is [9, 8] (which I'm interpreting as a list of length 9 followed by one of size 8), then the very first test of the while loop would fail (the 2nd test is then `8 > 9` which is false), and the post-loop code would just re-assign the first element to itself, so not swap. Am I misunderstood the case you are referring to?

query-planner-js/src/buildPlan.ts
clenfest: My bad, didn't notice you were swapping with `i-1` instead of `i`

query-planner-js/src/buildPlan.ts
clenfest: My bad, didn't notice you were swapping with `i-1` instead of `i`

clenfest (approved):

PR #1321, opened 2021-12-16
Fix: Avoid panic!() s when a deserialization error happens.

o0Ignition0o:
So far uncaught runtime errors would trigger a panic on the router-bridge.
Deserialization errors unfortunately would lead to panic's, which we want to avoid.

This commit refactors the way router-bridge errors are raised and propagated.

We now have Two layers of errors (we previously only had one):
- One representing errors raised by the script run (such as GraphQLErrors etc...)
- One representing uncaught errors while interfacing with the JS runtime.

Less panics, and more tests \o/


router-bridge/src/error.rs
```diff
@@ -0,0 +1,18 @@
...
+*/
+
+use serde::{Deserialize, Serialize};
+use thiserror::Error;
+
+#[derive(Error, Serialize, Deserialize, Debug, Clone)]
+/// An error which occurred within the bridge.
+///
+/// This does not include JS domain related errors, such as [`GraphQLError`].
+pub enum Error {
```
garypen, line 12: I feel like we should be able to put together a better name for this. Is it `DenoRuntimeError`? Also, is it really an enum or is there only ever going to be one thing? In which case:
`pub struct DenoRuntimeError(String);`

router-bridge/src/introspect.rs
```diff
@@ -123,6 +124,7 @@ mod tests {
             }",
             vec![DEFAULT_INTROSPECTION_QUERY.to_string()],
         )
+        .unwrap()
         .unwrap();
```
garypen, line 128: Just unwrap()? Or should this be expect with a message? Or even handle/propagate the error?

router-bridge/src/introspect.rs
```diff
@@ -140,6 +142,7 @@ mod tests {
...
```
garypen, line 146: Same comment as above.

router-bridge/src/js.rs
```diff
@@ -61,11 +64,17 @@ impl Js {
...
+            let message = format!(
                 "unable to invoke {} in JavaScript runtime \n error: \n {:?}",
                 source, e
-            )
+            );
+
+            tx.send(Err(Error::JSRuntimeError(message)))
+                .expect("channel must be open");
+
+            deno_core::anyhow::Error::from(e)
```
garypen, line 77: I think you can just return `e`... I'm surprised clippy didn't complain about that.

garypen (changes requested):
This is a great improvement. I've put a few cosmetic comments.

One other thing that we should probably address in here if we can. There is still a `panic!()` in the `with_parameter()` method and I feel like we could propagate an error now like we do in `execute()`.

router-bridge/src/error.rs
o0Ignition0o: Oh that's interesting, I think we might reuse it for the with_parameter, lemme check

router-bridge/src/introspect.rs
o0Ignition0o: true, it's in a test but it might help people if they fail someday 😅

router-bridge/src/js.rs
o0Ignition0o: possibly because we're not linting in this repo 🙈 lemme check real quick :D

garypen (changes requested):

harmonizer/src/lib.rs
```diff
@@ -45,7 +45,7 @@ pub fn harmonize(
     subgraph_definitions: Vec<SubgraphDefinition>,
 ) -> Result<CompositionOutput, BuildErrors> {
     // Initialize a runtime instance
-    let mut runtime = JsRuntime::new(Default::default());
+    let mut runtime = DenoRuntime::new(Default::default());
```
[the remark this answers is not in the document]
o0Ignition0o: mehhh lemme revert that :D

garypen (approved):

EverlastingBugstopper (approved):
this looks great

PR #1326, opened 2021-12-17
Convert compatible library table to HTML and add sticky

StephenBarlow:


StephenBarlow:
If we like this scrollable table treatment we can convert it into a component for future use

PR #1336, opened 2021-12-21
Add Uplink article in advance of multi-tenant

StephenBarlow:


docs/source/managed-federation/uplink.mdx
```diff
@@ -0,0 +1,90 @@
...
+  class registry secondary;
+  class uplink secondary;
+```
+
+To maximize uptime, Uplink is hosted simultaneously at _two_ endpoints, one in GCP and one in AWS:
+
+* GCP: `https://uplink.api.apollographql.com/`
+* AWS: `https://aws.uplink.api.apollographql.com/`
+
+By default in versions of `@apollo/gateway` vTODO and later, your gateway polls the GCP URL at a regular interval. Whenever a poll request fails, the gateway retries the request with the AWS URL. It continues swapping between the two URLs until the request succeeds, or until reaching the maximum number of retries.
```
zionts, line 31: I believe this will be 0.45.0, but @trevor-scheer  can correct me!

zionts, line 31: minor: perhaps we should say the "default" URL instead of the GCP URL (just whatever is the first in the list), to be more general? I'm not sure how likely users are to override the endpoints, but that would give them the understanding of how the behavior works (it just so happens that GCP is first in the list by default)

zionts: ah, nvm! You cover this very well directly below. Thank you 🙏 I should've kept reading before commenting!

docs/source/managed-federation/uplink.mdx
```diff
@@ -0,0 +1,90 @@
...
+
+## Configuring polling behavior
+
+You can configure the list of Uplink URLs your gateway uses, along with the number of retries it performs for a failed poll request.
+
+### Uplink URLs
+
+You can provide a custom list of URLs for the gateway to use when polling Uplink. For example, you can prioritize the AWS URL or omit one of the URLs entirely.
+
+You can provide this list either in the `ApolloGateway` constructor or as an environment variable.
```
zionts, line 43: What environment variable? (APOLLO_SCHEMA_CONFIG_DELIVERY_ENDPOINT)

docs/source/managed-federation/uplink.mdx
```diff
@@ -0,0 +1,90 @@
...
+const { ApolloGateway } = require('@apollo/gateway');
+
+// ...
+
+const gateway = new ApolloGateway({
+  uplinkMaxRetries: 2
+});
+```
+
+By default, the gateway retries a number of times equal to three times the number of [Uplink URLs](#uplink-urls).
```
zionts, line 90: I think we should make it clear that this is only an _internal_ retry, in that we will keep polling the endpoints in the list this number of times for _one fetch_ before emitting an error. We will always poll indefinitely when there are errors for all retries.

zionts: In other words, if you hit `maxRetries` while round-robin-ing through the endpoints, the behavior is unchanged from today. You still will continue to use the last fetched supergraph and try again at the next poll interval.

zionts (commented):
I'll leave this to @trevor-scheer to review w/ the final thumbs-up and the go-ahead after we've published a new version, but this looks great to me sans a few minor points / questions! Thanks @StephenBarlow

docs/source/managed-federation/uplink.mdx
trevor-scheer: Correct!

docs/source/api/apollo-gateway.mdx
```diff
@@ -135,6 +135,41 @@ You can specify any string value for the `name` field, which is used for identif
...
+
+###### `uplinkEndpoints`
+
+`Array<string>`
+</td>
+<td>
+
+A list of Apollo Uplink URLs the gateway uses to poll for its managed configuration. For details and defaults, see [Apollo Uplink](/managed-federation/uplink/).
+
+Provide this field **only** if you are using managed federation.
```
trevor-scheer, line 152: I think we should be extra clear here that even managed federation users probably don't need to configure this unless we explicitly say so.

docs/source/api/apollo-gateway.mdx
```diff
@@ -135,6 +135,41 @@ You can specify any string value for the `name` field, which is used for identif
...
+
+###### `uplinkMaxRetries`
+
+`number`
+</td>
+<td>
+
+The maximum number of times the gateway retries a failed poll request to Apollo Uplink, cycling through its list of [`uplinkEndpoints`](#uplinkendpoints).
+
+The default value is three times your number of `uplinkEndpoints`.
```
trevor-scheer, line 169: Might add that there are 2 endpoints by default, which should be the overwhelming majority of cases.

docs/source/managed-federation/uplink.mdx
```diff
@@ -0,0 +1,94 @@
...
+To maximize uptime, Uplink is hosted simultaneously at _two_ endpoints, one in GCP and one in AWS:
+
+* GCP: `https://uplink.api.apollographql.com/`
+* AWS: `https://aws.uplink.api.apollographql.com/`
+
+By default in versions of `@apollo/gateway` v0.45.0 and later, your gateway polls the GCP URL at a regular interval. Whenever a poll request fails, the gateway retries that request with the AWS URL. It continues swapping between the two URLs until the request succeeds, or until reaching the maximum number of retries.
+
+Even if a poll request fails all retries, the gateway continues polling at the next interval, and it continues using its most recently obtained configuration.
+
+> Verisons of `@apollo/gateway` prior to v0.45.0 don't support multiple Uplink endpoints and only use the GCP URL by default.
```
trevor-scheer, line 35:
```suggestion
> Versions of `@apollo/gateway` prior to v0.45.0 don't support multiple Uplink endpoints and only use the GCP URL by default.
```

docs/source/managed-federation/uplink.mdx
```diff
@@ -0,0 +1,94 @@
...
+```
+
+To maximize uptime, Uplink is hosted simultaneously at _two_ endpoints, one in GCP and one in AWS:
+
+* GCP: `https://uplink.api.apollographql.com/`
+* AWS: `https://aws.uplink.api.apollographql.com/`
+
+By default in versions of `@apollo/gateway` v0.45.0 and later, your gateway polls the GCP URL at a regular interval. Whenever a poll request fails, the gateway retries that request with the AWS URL. It continues swapping between the two URLs until the request succeeds, or until reaching the maximum number of retries.
+
+Even if a poll request fails all retries, the gateway continues polling at the next interval, and it continues using its most recently obtained configuration.
```
trevor-scheer, line 33: I'm wondering if there's another wording that makes this more clear (haven't thought of one). "We used up all of our retries, we'll just retry again" is a bit of a head scratcher. It's also not that bad and I _did_ understand what it meant.

docs/source/managed-federation/uplink.mdx
```diff
@@ -0,0 +1,94 @@
...
+
+> Verisons of `@apollo/gateway` prior to v0.45.0 don't support multiple Uplink endpoints and only use the GCP URL by default.
+
+## Configuring polling behavior
+
+You can configure the list of Uplink URLs your gateway uses, along with the number of retries it performs for a failed poll request.
```
trevor-scheer, line 39: `experimental_pollInterval` can be set to >=10000 (10s or more). Do we want to talk about that here?

FWIW, I plan to kill this option soon in favor of `pollIntervalInMs` (doesn't exist yet), so might be worth waiting to document it.

docs/source/managed-federation/uplink.mdx
```diff
@@ -0,0 +1,94 @@
...
+
+### Uplink URLs
+
+You can provide a custom list of URLs for the gateway to use when polling Uplink. For example, you can prioritize the AWS URL or omit one of the URLs entirely.
```
trevor-scheer, line 43: I find the use cases listed to be the only ones I can think of and also not very compelling. We might nudge users away from this option or suggest that it's unlikely they need to configure this.

docs/source/managed-federation/uplink.mdx
StephenBarlow: Great point, I'm going to make this a separate task and will leave out here for the sake of getting the PR out

docs/source/managed-federation/uplink.mdx
```diff
@@ -30,15 +30,36 @@ To maximize uptime, Uplink is hosted simultaneously at _two_ endpoints, one in G
 
 By default in versions of `@apollo/gateway` v0.45.0 and later, your gateway polls the GCP URL at a regular interval. Whenever a poll request fails, the gateway retries that request with the AWS URL. It continues swapping between the two URLs until the request succeeds, or until reaching the maximum number of retries.
 
-Even if a poll request fails all retries, the gateway continues polling at the next interval, and it continues using its most recently obtained configuration.
+Even if a particular poll request fails _all_ of its retries, the gateway continues polling as usual at the next interval (with its own set of retries if needed). In the meantime, the gateway continues using its most recently obtained configuration.
```
trevor-scheer, line 33: Love it!

trevor-scheer (approved):
LGTM! Thanks @StephenBarlow 😄

Issue #1199, opened 2021-11-12
chore(deps): Support graphql@16


trevor-scheer:
Like it says in the title, add support for `graphql@16`!  🎉  (v0.x and v2)

trevor-scheer:
Related: https://github.com/apollographql/federation/pull/1207

setchy:
yes please 🙏

m1heng:
Do we have any progress in this 🥺

m1heng:
@trevor-scheer I think your #1129 has solved this in main right?

trevor-scheer:
@m1heng yes it's finished on main. @clenfest is working on backporting the changes to the 0.x branch as well.

m1heng:
Cool, I will close my PR then. Wish to see the release from 0.x soon.

PR #1345, opened 2021-12-28
Update dev graphql to 16 and add in peerDependency

m1heng:
solve #1199

### DIFF
--- a/composition-js/package.json
+++ b/composition-js/package.json
@@ -31,6 +31,6 @@
     "@apollo/query-graphs": "file:../query-graphs-js"
   },
   "peerDependencies": {
-    "graphql": "^15.7.0"
+    "graphql": "^15.7.0 || ^16.1.0"
   }
 }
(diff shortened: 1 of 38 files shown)


m1heng:
All test reports "cannot extend type because it is not defined" like error.
Throwing from https://github.com/apollographql/federation/blob/ff079db03e504adf509e59eda3e72ea63bfb6746/subgraph-js/src/buildSubgraphSchema.ts#L65-L71

Input of this function has not changed. So weird.

m1heng:
> All test reports "cannot extend type because it is not defined" like error. Throwing from
> 
> https://github.com/apollographql/federation/blob/ff079db03e504adf509e59eda3e72ea63bfb6746/subgraph-js/src/buildSubgraphSchema.ts#L65-L71
> 
> Input of this function has not changed. So weird.

problem located https://github.com/apollographql/apollo-tooling/issues/2474

m1heng:
After manually changed `apollo-graphql` source code in `node_modules` locally,  all tests passed.
PR for the needed change is made https://github.com/apollographql/apollo-tooling/pull/2534. Once that is merged and released, let's update dependency version here, and everything should be good.

m1heng:
Job done in #1129.

PR #1351, opened 2022-01-04
Fix errors reading fed1 supergraphs

pcmanus:
A fed1 supergraphs doesn't always have enough information to know in
which subgraph a type is defined, so when extracting a subgraph from
the supergraph, we sometime "optimistically" add a type to that
subgraph, but later have to remove it if it turns out the type has
no fields in that subgraph. The issue was that this type was sometimes
referenced and removing it left the references invalid. Instead, this
patch ensures those references are also removed (we can infer those
were also added mistakenly due to a lack of information).

Additionally, the commit slightly improves the handling of errors
happing while extracting subgraphs from the supergraph, providing
clearer messages and, for debugging, allows to dump the extracted
subgraph (but only when an `APOLLO_FEDERATION_DEBUG_SUBGRAPHS` env
variable is set to avoid surprises).


internals-js/src/__tests__/extractSubgraphsFromSupergraph.test.ts
```diff
@@ -0,0 +1,110 @@
+import { buildSupergraphSchema, extractSubgraphsFromSupergraph } from "..";
+
+test('handles types having no fields in a subgraph correctly', () => {
```
queerviolet, line 3: could we add another test which exercises the path where we remove fields in a subgraph but do not remove the whole type? mainly for documentation of the behavior—i expect this is exercised in other tests.

internals-js/src/definitions.ts
```diff
@@ -1753,6 +1777,20 @@ export class ObjectType extends FieldBasedType<ObjectType, ObjectTypeReferencer>
...
   }
+
+  protected removeReferenceRecursive(ref: ObjectTypeReferencer): void {
+    // Note that the ref can also be a`SchemaDefinition`, but don't have anything to do then.
+    switch (ref.kind) {
+      case 'FieldDefinition':
+        ref.removeRecursive();
+        break;
+      case 'UnionType':
+        if (ref.membersCount() === 0) {
```
queerviolet, line 1788: do we have tests that exercise this?

queerviolet (approved):
code looks good! would like a few more tests:
  - a negative test (ensure that types with remaining fields are not removed) mainly for documentation (i suspect many things would break if that functionality were broken)
  - testing the recursive removal of empty unions, fields, and the other cases we handle

pcmanus:
Thanks for the review. Pushed a few more tests indeed.

PR #1356, opened 2022-01-05
change broken test to match error message

clenfest:
change broken test to match error message


pcmanus (approved):
Thanks for catching that.

PR #1357, opened 2022-01-06
Update code reference in logger docs

mmartinson:
Code reference was stale, moved to target:

```
export type Logger = {
  // Ordered from least-severe to most-severe.
  debug(message?: any): void;
  info(message?: any): void;
  warn(message?: any): void;
  error(message?: any): void;
};
```


pcmanus (approved):
Thank you!

PR #1366, opened 2022-01-07
Document setting gateway poll interval

StephenBarlow:
Rather than retarget the existing PR I did some cherry-picking onto a new PR to avoid any potential rebase issues 🤷


trevor-scheer (approved):
Perfect, thanks!

PR #1367, opened 2022-01-08
Document query plans

StephenBarlow:


clenfest (approved):
Great stuff Stephen! I had a couple of suggested changes, but then I went back to check and realized I was in error and you had it right. LGTM.

StephenBarlow:
Thanks so much @clenfest!

PR #1402, opened 2022-01-19
Use optional syntax to avoid array in call to buildSubgraphSchema

StephenBarlow:
Small PR for whenever you're around


trevor-scheer (approved):
LGTM! Can we land this to `version-0.x` as well?

StephenBarlow:
@trevor-scheer Sure! Does that `port-to-2.x` label do anything fancy, or is it just to track the fact that I still need to land it in `main` as well?

trevor-scheer:
@StephenBarlow just for tracking! It could potentially do a fancy thing in the future though.

PR #1403, opened 2022-01-19
Update custom directive docs for subgraphs

StephenBarlow:
@glasser We got a question in the forums asking how custom directives are added in subgraphs using the "new" `graphql-tools` method. I tested this out with an example directive and it seems to work just fine with the executable schema generated by `buildSubgraphSchema`. How do you feel about us putting this out there?


StephenBarlow:
Also should we add language around how you should make sure if multiple subgraphs can resolve the same field, that all of those fields have the exact same directives with the exact same transformation functions associated with them? Lest someone end up with inconsistent results for a field depending on resolving subgraph

PR #1410, opened 2022-01-20
Docs: bump to gatsby v3

jgarrow:
This PR tests an install of an experimental version of `gatsby-theme-apollo-docs` to see if the upgrade to Gatsby v3 is successful.

### DIFF
--- a/docs/package-lock.json
+++ b/docs/package-lock.json
@@ -9585,9 +9585,9 @@
       }
     },
     "gatsby-theme-apollo-core": {
-      "version": "3.0.38-alpha.1",
-      "resolved": "https://registry.npmjs.org/gatsby-theme-apollo-core/-/gatsby-theme-apollo-core-3.0.38-alpha.1.tgz",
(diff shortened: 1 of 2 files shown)


trevor-scheer:
Awesome! If we do land this, just note that we land the same changes over on the `version-0.x` branch. Thanks @jgarrow!

trevor-scheer:
There's some related renovate config which should be removed along with this change. Currently we block 3.x updates from renovate, but we'd (presumably) like to allow them once we've made it to v3. (We might want to unblock the `react` updates as well?)

jgarrow:
> There's some related renovate config which should be removed along with this change. Currently we block 3.x updates from renovate, but we'd (presumably) like to allow them once we've made it to v3. (We might want to unblock the `react` updates as well?)

@trevor-scheer ooh yes, we'll want to allow 3.x updates once this is merged in. For `react` updates, this version will still be on v16, so perhaps leave those off? @trevorblades does this sound reasonable to you?

trevorblades:
Yup sounds good to me @jgarrow

PR #1431, opened 2022-01-25
Fix merging of fetch groups for @requires

pcmanus:
When computing entity fetches for query plans, we keep track for an
entity fetch of the path into it's "parent" group. But when doing so,
we were keeping the "cast" into the entity type if present. Not only
is that unecessary since entity fetches inputs/selections always start
with casting into the proper entity type, it was harmful as it means
that the logic for @require was sometimes not recognizing that it
could merge fetches (due to those casts), sometimes yielding (obviously)
ineffecient plans.

### DIFF
--- a/query-planner-js/src/buildPlan.ts
+++ b/query-planner-js/src/buildPlan.ts
@@ -7,7 +7,6 @@ import {
   Field,
   FieldSelection,
   FragmentElement,
-  FragmentSelection,
   isAbstractType,
   isCompositeType,
   isListType,
@@ -36,6 +35,8 @@ import {
   NamedFragments,
   operationToDocument,
   MapWithCachedArrays,
+  FragmentSelection,
+  sameType,
 } from "@apollo/federation-internals";
 import {
   advanceSimultaneousPathsWithOperation,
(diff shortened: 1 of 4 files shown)


query-planner-js/src/buildPlan.ts
```diff
@@ -1390,6 +1397,18 @@ function createNewFetchSelectionContext(type: CompositeType, selections: Selecti
...
 }
 
+function extractPathInParentForKeyFetch(type: CompositeType, path: OperationPath): OperationPath {
+  // A "key fetch" (calls to the `_entities` operation) always have to start with some type-cast into
+  // the entity fetched (`type` in this function), so we can remove a type-cast into the entity from
+  // the parent path if it is the last thing in the past. And doing that removal ensures the code
+  // later reuse fetch groups for different entities, as long as they get otherwise merged into the
+  // parent at the same place.
+  const lastElement = path[path.length - 1];
+  return lastElement && lastElement.kind === 'FragmentElement' && lastElement.typeCondition?.name === type.name
```
clenfest, line 1407: Could you put parenthesis around the ternary clause so it's a little more readable?

clenfest (approved):
Just had one minor comment

query-planner-js/src/buildPlan.ts
pcmanus: Added, thanks.

PR #1433, opened 2022-01-25
Docs: Bump Gatsby to v3, docs theme to new major version

jgarrow:
This PR bumps Gatsby to v3 and the docs theme to the latest major version. As mentioned in #1410, I'm making the upgrade both here and on `main`


jgarrow:
Resolved by #1436

Issue #1440, opened 2022-01-26
Documentation for how to run `subgraph-js` tests


dchambers:
I'd like to contribute a PR (documentation support for the `SchemaDefinition` element), but I can't find any way to run the tests. Running `npm test` at the root level does not run these tests, and I also notice that they are not run as part of the CI tasks either, so I can't copy that.

Any advice please?

dchambers:
BTW, to replicate what the CI was doing (assuming that it must work) these were the commands I ran:

```sh
docker run -it cimg/node:16.0.0 bash
sudo yarn global add npm@7
git clone https://github.com/apollographql/federation.git
cd federation
npm ci
npm run test:ci
```

dchambers:
FYI, the `subgraph-js` tests run just fine on the `version-0.x` branch, which is the version I really need a fix for anyway at this point, so I'll look to start there...

PR #1442, opened 2022-01-26
Schema definition description text support

dchambers:
Add support for description text on `SchemaDefinition` nodes, [as per](https://spec.graphql.org/October2021/#SchemaDefinition) the October 2021 Edition of the spec.


dchambers:
Note that I created this against the `0.x` branch because [I couldn't get the tests to run](https://github.com/apollographql/federation/issues/1440) on the `main` branch, but this looks like it should be a very clean forward-port to that branch when ready, so hopefully not problematic.

subgraph-js/src/printSubgraphSchema.ts
```diff
@@ -104,6 +104,10 @@ function printSchemaDefinition(schema: GraphQLSchema): string | undefined {
     operationTypes.push(`  subscription: ${subscriptionType.name}`);
   }
 
+  if (operationTypes.length === 0) {
```
trevor-scheer, line 107: Is this change necessary or related? I don't see a new or affected test to understand why it's been added. Not every subgraph must provide an entry point via a root operation (i.e. subgraphs can just extend types and nothing more) so I don't think this change is correct.

subgraph-js/src/printSubgraphSchema.ts
```diff
@@ -131,6 +135,10 @@ function isSchemaOfCommonNames(schema: GraphQLSchema): boolean {
     return false;
   }
 
+  if (schema.astNode?.description) {
+    return false
+  }
```
trevor-scheer, line 140: I don't understand why this change is needed either.

subgraph-js/src/schema-helper/buildSchemaFromSDL.ts
```diff
@@ -196,6 +196,7 @@ export function buildSchemaFromSDL(
   const schemaDefinitions: SchemaDefinitionNode[] = [];
   const schemaExtensions: SchemaExtensionNode[] = [];
   const schemaDirectives: ConstDirectiveNode[] = [];
+  const schemaDescriptions: string[] = []
```
trevor-scheer, line 199: Can't there only be one schema description? This should probably just be a `string | undefined`.

subgraph-js/src/schema-helper/buildSchemaFromSDL.ts
```diff
@@ -295,15 +299,24 @@ export function buildSchemaFromSDL(
     };
   }
 
+  const description = schemaDescriptions.length === 0 ? undefined :
```
trevor-scheer, line 302: Similar comment as above - there should only be one description so I don't think we need to join anything.

trevor-scheer (commented):
Can you help me understand the use case here? Subgraph printing is really just for the sake of composition. If we're going to print the schema description, the gateway should be doing something with it in a meaningful way as well, right? I don't think there's anything currently in place for "merging" N schema descriptions from N subgraphs into a composed schema.

dchambers:
> Can you help me understand the use case here? Subgraph printing is really just for the sake of composition. If we're going to print the schema description, the gateway should be doing something with it in a meaningful way as well, right? I don't think there's anything currently in place for "merging" N schema descriptions from N subgraphs into a composed schema.

Thanks for responding Trevor!

Our use case is that we have a poller service that periodically requests schemas from all subgraphs before linting them and (in staging & prod) confirming that they've received governance approval. Changes that pass both checks are registered with Apollo Registry, and in fact this is the only way for schemas to become registered.

Verifying that a subgraph's schema has been approved is already quite difficult because different federation libraries have their own set of schema fidelity quirks that must be accounted for. We currently have subgraphs using the following tech stacks across our company for example:

- Apollo / TypeScript
- Ariadne / Python
- DGS / Java
- `federation-jvm` / `java-graphql` / Java

and I've accounted for the various fidelity quirks across these stacks when generating subgraph schema content hashes (challenging, but do-able).

The remaining issue we have is that the approval-stamp is encoded as a `@governanceApproved` directive on the schema block (along side the `@contact` directive), except that not all stacks (i.e. Apollo & Ariadne) support custom directives, and so that the next most sensible place becomes a description on the directive until core-schemas becomes available, but Apollo is unique in not currently supporting either approach.

I could alternatively do this by adding a type that's not used anywhere, except that I don't really want to pollute the schema with types that don't need to be there, and so this seems like a nice way to avoid it the given the lack of custom directive support.

subgraph-js/src/printSubgraphSchema.ts
dchambers: This is only needed in the case where your schema-definition has no directives (e.g. `@contact`), and where your schema does not define types for `mutation` or `subscription`, or a type for `query` other than `Query`. Removing this line will cause the test I added to break for example.

subgraph-js/src/printSubgraphSchema.ts
dchambers: Similar to above in that it's not needed if you either have schema-definition directives (e.g. `@contact`) or if you define types for `mutation` or `subscription`, or a type for `query` other than `Query`. As before, removing this line will cause the test I modified to break.

subgraph-js/src/schema-helper/buildSchemaFromSDL.ts
dchambers: Yes, I think you're effectively correct in practice. The only confusion is that `buildSchemaFromSDL` actually allows you to pass `DocumentNode | DocumentNode[]`, and so in the latter case there's presumably the chance that we get more than one description?

It seems odd that it works that way though, and I notice that you've recently imported the `buildSchemaFromSDL` function from a stand-along module, and so maybe this array form is no longer needed in its new setting, in which case I'm definitely happy to get rid of as it really does seem wrong as you say?

subgraph-js/src/schema-helper/buildSchemaFromSDL.ts
dchambers: > Similar comment as above - there should only be one description so I don't think we need to join anything.

Agreed! I'll wait for a 🟢  light on the similar question above before proceeding with the suggested improvement 👍

subgraph-js/src/schema-helper/buildSchemaFromSDL.ts
trevor-scheer: In the `DocumentNode[]` case, you can have one `schema` definition as well as `extend schema` definition extensions. The `extend schema` notation is not permitted to have a description attached.

https://spec.graphql.org/draft/#sec-Schema
>When using the type system definition language, a document must include at most one schema definition.

https://spec.graphql.org/draft/#sec-Schema-Extension
Note that there's no description permitted on the grammar for extensions

subgraph-js/src/printSubgraphSchema.ts
trevor-scheer: I would say that the test case is invalid since you're referencing a type `Query` which you haven't defined. I think this approach falls short in the case that a subgraph _shouldn't_ have a schema definition at all.

trevor-scheer:
I understand what you're going for and I'm sympathetic to the need, so I'm trying to imagine a clever solution here. My biggest hesitation (as mentioned above) is requiring a schema definition to exist when it isn't needed (since the existence alone requires a root type to be defined as well). How do the other libraries you mentioned work around this issue?

For what it's worth, I see that `buildSubgraphSchema` still succeeds in this case (even though `Query` is undefined), but I don't want to exploit this just because "it works". This schema is still invalid IMO.
`buildSubgraphSchema({ typeDefs: gql`schema { query: Query }` })`

subgraph-js/src/printSubgraphSchema.ts
dchambers: Good shout, and I actually managed to get rid of one of the two pieces of code that you didn't think were necessary by adding a `Query` type 👍 .

dchambers:
> My biggest hesitation (as mentioned above) is requiring a schema definition to exist when it isn't needed

So that was just a shoddy test on my part actually, and not my intention. In our case all subgraph schema not only would define some `Query` types, but also will have to specify `@contact` directives as per our linting rules. I was just trying to make the test as minimal as possible so I could be sure that the code always worked even in the minimal case, but I went a bit too far 😁 .

subgraph-js/src/printSubgraphSchema.ts
trevor-scheer: Got it. So this entire file is as much of a fork of `graphql-js`'s `printSchema` as possible. Let's follow what they've done and handle this check outside of this function. That decision is in line with my feeling that this doesn't _quite_ belong in this function based on its name.

See here, the description is inspected before even calling `isSchemaOfCommonNames`. I'd like it if we did the same.
https://github.com/graphql/graphql-js/blob/47bd8c8897c72d3efc17ecb1599a95cee6bac5e8/src/utilities/printSchema.ts#L73

subgraph-js/src/schema-helper/buildSchemaFromSDL.ts
```diff
@@ -221,6 +222,7 @@ export function buildSchemaFromSDL(
       schemaDirectives.push(
         ...(definition.directives ? definition.directives : [])
       );
+      description = definition.description?.value
```
trevor-scheer, line 225: Can we capture the whole node here (`definition.description`) and use it below? It should even properly be a `StringValueNode | undefined`, which is what `astNode.description` is expecting.

The only uncertainty I have with this suggestion is whether or not it's important that you're setting `block: true` down below. Seems like it would be fine to allow `block` to be whatever it was originally on the node?

trevor-scheer (changes requested):

trevor-scheer (approved):
LGTM! Go ahead and add an entry to `subgraph-js/CHANGELOG.md` (in the `vNEXT` section, replacing the placeholder) and I'll get this landed. I'll also be cherry-picking this over to `main` once it's landed.

dchambers:
> LGTM! Go ahead and add an entry to `subgraph-js/CHANGELOG.md` (in the `vNEXT` section, replacing the placeholder)

Done.

> ... and I'll get this landed. I'll also be cherry-picking this over to `main` once it's landed.

Thank you kindly. 🙇

PR #1444, opened 2022-01-27
router-bridge: use npm ci in the build.rs process

### LINKED ISSUE #1313: router-bridge: use npm ci in the build.rs
We tried to use npm ci instead of npm i in the bridge build.rs, but it failed.

Given npm ci is probably the right path here, let's investigate why it failed and fix it.

> LGTM.
> 
> can we replace `.args(&["install"])` with `.args(&["ci"])` in build.rs at line 21?
> 
> We can also do this as a followup

Done in: https://github.com/apollographql/federation/pull/1292/commits/499c7071616cfd97fb051830fc0df07e345d9062

_Originally posted by @garypen in https://github.com/apollographql/federation/issues/1292#issuecomment-994755250_

o0Ignition0o:
fixes apollographql/federation-rs#29 

Use npm ci instead of npm install in the build.rs process, to make sure we're using the tested dependencies, and avoir surprises if one of them doesn't respect semver.

### DIFF
--- a/router-bridge/build.rs
+++ b/router-bridge/build.rs
@@ -18,7 +18,7 @@ fn update_bridge() {
 
     assert!(Command::new(&npm)
         .current_dir(&repo_dir)
-        .args(&["install"])
+        .args(&["ci"])
         .status()
         .unwrap()
         .success());


abernix (approved):
This _should_ definitely make things faster and is certainly a best practice.  That said, this project does have a `package-lock.json` which should still guarantee installation of the packages and versions that are specified in that file regardless of whether using `npm clean-install` or `npm install`.  While it is true that there's _added_ benefits to using `clean-install` / `ci` in that it will alert us (by failing! ❌ ) when the _top-level_ dependencies in `package.json` are not those specified in `package-lock.json`, our other tests should have already been ensuring that those match (since we do use `ci` in our other tests).

Anyhow, very supportive of this change and this is a good call, but I'm certainly kinda wondering now why this was ever `install` in the first place! All good.  :shipit:

abernix:
Now the question is, why are tests the tests all failing!?

o0Ignition0o:
> Now the question is, why are tests the tests all failing!?

Good question, it seems to happen on the harmonizer side, lemme check.

trevor-scheer:
Is this good for `version-0.x` as well?

o0Ignition0o:
It should be, modulo the failing tests ^^’

o0Ignition0o:
Closing this given is has landed on [federation-rs](https://github.com/apollographql/federation-rs/blob/14ad25067091e9114d9abae03181604b8baeea6b/router-bridge/build.rs#L21)

PR #1449, opened 2022-01-28
chore(infra): Fix and improve jest config (`main`)

### LINKED ISSUE #1440: Documentation for how to run `subgraph-js` tests

trevor-scheer:
Tests aren't currently running for `subgraph-js`. This re-enables them (and adds some coloring / display names to the test output for better vis).

Fixes #1440

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,12 +93,12 @@
   },
   "jest": {
     "projects": [
-      "<rootDir>/internals-js",
-      "<rootDir>/federation-integration-testsuite-js",
-      "<rootDir>/gateway-js",
-      "<rootDir>/query-planner-js",
       "<rootDir>/composition-js",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -93,12 +93,12 @@
   },
   "jest": {
     "projects": [
-      "<rootDir>/internals-js",
-      "<rootDir>/federation-integration-testsuite-js",
```
trevor-scheer, line 97: No actual tests in this package so I removed it

cpeacock (approved):

PR #1451, opened 2022-01-28
fix: Unbreak support for `graphql@15` (`version-0.x`)

trevor-scheer:
If we want to continue supporting `graphql@15`, we can't make references to new types that only exist in v16.


clenfest:
Do you have a merge issue? Looks like it's deleting some files.

IvanGoncharov:
@trevor-scheer @clenfest I can try to backport stuff in v15 if it helps.
I just need the steps necessary to reproduce the broken build on v15.

trevor-scheer:
@clenfest we spoke in person but for posterity, this just removes some duplication. The `federation-js` is just a backcompat wrapper of a package (which depends on the subgraph package), so no need to dupe code between those 2. The deletion is unrelated to the issue, just an opportunistic cleanup since I would've been touching the duplicated `buildSchemaFromSDL.ts` file anyway.

clenfest (approved):

PR #1463, opened 2022-02-01
fix(subgraph): Conditionally print @specifiedBy directive on scalars correctly

angelo-moreira:
When having a schema with scalars
```
scalar Date
scalar JSON
scalar HexColour
```

We end up in a situation where the scalar is `undefined` instead of `null`.

Because of that when we try to get the `sdl` we receive the `@specifiedBy` tag with an object attached to it.

```
query getIntrospection {
  _service {
    sdl
  }
}
```

image

This PR should fix that by weakly comparing `undefined == null`, I am happy to do a more strict comparison.

### DIFF
--- a/subgraph-js/CHANGELOG.md
+++ b/subgraph-js/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## vNEXT
 
+> The changes noted within this `vNEXT` section have not been released yet.  New PRs and commits which introduce changes should include an entry in this `vNEXT` section as part of their development.  When a release is being prepared, a new header will be (manually) created below and the appropriate changes within that release will be moved into the new section.
+
 - Make sure scalars don't print a `specifiedBy` directive unless specified in the subgraph. [PR #1463](https://github.com/apollographql/federation/pull/1463)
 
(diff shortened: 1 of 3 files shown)


trevor-scheer:
Thanks @angelo-moreira, this LGTM.
* I confirmed the test was previously failing before the change
* This is how the `graphql-js` printer does it (which this is a fork of) https://github.com/graphql/graphql-js/blob/47bd8c8897c72d3efc17ecb1599a95cee6bac5e8/src/utilities/printSchema.ts#L296

Would you please add a CHANGELOG entry for this in `subgraph-js/CHANGELOG.md`?

I'll get this cherry-picked / backported to the `version-0.x` branch.

angelo-moreira:
Thank you for the quick reply @trevor-scheer, I have now added a line to the changelog with hopefully a useful commit, please do let me know if you need me to do something else, it's my first time contributing to this project.

trevor-scheer:
A great first contribution @angelo-moreira! I made a few minor tweaks to save you the trouble, but otherwise this is excellent. Thanks again!

Issue #1439, opened 2022-01-26
Use a specific error class for Uplink connection errors


nogates:
Hello there!

I'd like to ask if it would be possible to have a specific error class to signal errors when pulling supergraph schema from uplink.

Right now, when this happens, a [generic error is thrown](https://github.com/apollographql/federation/blob/main/gateway-js/src/supergraphManagers/UplinkFetcher/loadSupergraphSdlFromStorage.ts#L166), and if we want to catch those errors, we need to match the error message, which seems a bit brittle (I wouldn't be surprised if that message is changed in the future).

Would it be possible to extend the Error class with a more specific name (something like `UplinkFetchError`), and export it, so we can check the name of that error instead of the message?

A bit of context, we plan on catching these errors on our Gateways. If Apollo Uplink is down, we can at least rescue from those errors and use a local schema instead, so our Gateways can still start up, even if the schema is an old one.

Thank you!

nogates:
BTW, I didn't mention it, but I'd be happy to work on this and submit a PR if that's ok with you folks :)

cpeacock:
Absolutely, we'd love to have you take a crack at it if you'd like!  We did place this on a team's backlog but would love your contribution if you're able to give the time.

PR #1473, opened 2022-02-03
Create specific error class to signal uplink fetcher errors

nogates:
Hello there,

As described in #1439, we are suggesting here to use a specific class when throwing errors while fetching the schema from Apollo Uplink.

This should allow us to capture those errors using the specific class name, instead of relying on the actual message.

I've decided to transform all the errors in that file (since they are all related to the same process of fetching the schema). Also, please let me know if you think the error class should be placed somewhere else, so we don't need to export it up to the top

Thank you!

Fixes

### DIFF
--- a/gateway-js/src/supergraphManagers/UplinkFetcher/index.ts
+++ b/gateway-js/src/supergraphManagers/UplinkFetcher/index.ts
@@ -4,6 +4,7 @@ import resolvable from '@josephg/resolvable';
 import { SupergraphManager, SupergraphSdlHookOptions } from '../../config';
 import { SubgraphHealthCheckFunction, SupergraphSdlUpdateFunction } from '../..';
 import { loadSupergraphSdlFromUplinks } from './loadSupergraphSdlFromStorage';
+export { UplinkFetcherError } from "./loadSupergraphSdlFromStorage"
 
 export interface UplinkFetcherOptions {
   pollIntervalInMs: number;
--- a/gateway-js/src/supergraphManagers/UplinkFetcher/index.ts
+++ b/gateway-js/src/supergraphManagers/UplinkFetcher/index.ts
@@ -4,7 +4,6 @@ import resolvable from '@josephg/resolvable';
 import { SupergraphManager, SupergraphSdlHookOptions } from '../../config';
 import { SubgraphHealthCheckFunction, SupergraphSdlUpdateFunction } from '../..';
 import { loadSupergraphSdlFromUplinks } from './loadSupergraphSdlFromStorage';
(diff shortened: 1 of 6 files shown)


gateway-js/src/supergraphManagers/UplinkFetcher/index.ts
```diff
@@ -4,6 +4,7 @@ import resolvable from '@josephg/resolvable';
 import { SupergraphManager, SupergraphSdlHookOptions } from '../../config';
 import { SubgraphHealthCheckFunction, SupergraphSdlUpdateFunction } from '../..';
 import { loadSupergraphSdlFromUplinks } from './loadSupergraphSdlFromStorage';
+export { UplinkFetcherError } from "./loadSupergraphSdlFromStorage"
```
trevor-scheer, line 7: It's a little funny to just export the error here for the sake of a simpler re-export from `supergraphManagers/index`.

I think _either_:
* This Error should be defined and exported from here (`UplinkFetcher` owns `UplinkFetcherError`, makes sense to me)
* Or, just skip exporting from here and re-export it from the nested path

One could probably argue that the contents of `loadSupergraphSdlFromStorage.ts` (and transitively `outOfBandReporter.ts`) simply belong in the `UplinkFetcher` class which is probably the root cause of ambiguity for "where should the Error class be defined?", though I'm not asking you to resolve that issue in this PR.

nogates: Yep. I completely agree. I think it's probably cleaner if I define the Error class in the `index`, and just get it from there in `loadSupergraphSdlFromStorage`.
Will change it. Thanks for the feedback

nogates: Actually, I was having problems defining `UplinkFetcherError` in the index file, and using it in `loadSupergraphSdlFromStorage`... so I've decided to go with the second option...

However, this makes me wonder, whether the error should belong in a different file (`errors.ts`), and then, export it from there? 

> One could probably argue that the contents of loadSupergraphSdlFromStorage.ts (and transitively outOfBandReporter.ts) simply belong in the UplinkFetcher class which is probably the root cause of ambiguity for "where should the Error class be defined?

trevor-scheer (approved):
This LGTM. Thanks for the changes! Can you please add an entry to the `gateway-js/CHANGELOG.md`? We'll port this to the `version-0.x` branch when this lands.

cc @cpeacock for awareness since you participated on the linked issue - sounds like there might be an Atlas issue we can close out when this lands?

cpeacock:
pinging @clenfest for awareness

PR #1482, opened 2022-02-04
Fix TS / runtime errors while using `graphql@15`

### LINKED ISSUE #1481: OperationTypeNode enum not compatible with graphql v15
I´ve updated to the last release  of apollo gateway and federation with graphql v15
 and I get this error:

image
Apollo source code error stack line:
image

Dependencies:
image

I found that graphql v16 package exports enum "OperationTypeNode" but graphql v15 not export it

v15 screenshot:
image
v16 screenshot:
image

clenfest:
It looks like that although OperationTypeNode existed as an exported type in graphql15, it did not exist as an enum, so the way I'm using it is incorrect in peer dependencies.

Fixes #1481

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -81,7 +81,7 @@ jobs:
     steps:
       - common_test_steps
   
-  GraphQL v15.8.0 Support:
+  GraphQL v15 Support:
     docker:
       - image: cimg/node:16.0.0
     steps:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -307,8 +307,8 @@ workflows:
       - NodeJS 16:
           name: "JS: Node 16"
           <<: *common_non_publish_filters
-      - GraphQL v15.8.0 Support:
-          name: "Backcompat support: graphql@15.8.0"
+      - GraphQL v15 Support:
+          name: "Backcompat support: graphql@15"
           <<: *common_non_publish_filters
       - GraphQL Types:
           name: "GraphQL Types (up to date)"
@@ -328,7 +328,7 @@ workflows:
             - "JS: Node 12"
             - "JS: Node 14"
             - "JS: Node 16"
-            - "Backcompat support: graphql@15.8.0"
+            - "Backcompat support: graphql@15"
             - "GraphQL Types (up to date)"
       - oss/dry_run:
           name: "JS: Dry-run"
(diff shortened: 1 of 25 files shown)


query-planner-js/src/schema-helper/transformSchema.ts
```diff
@@ -65,7 +65,7 @@ export function transformSchema(
     query: replaceMaybeType(schemaConfig.query),
     mutation: replaceMaybeType(schemaConfig.mutation),
     subscription: replaceMaybeType(schemaConfig.subscription),
-    directives: replaceDirectives(schemaConfig.directives)
+    directives: replaceDirectives(schemaConfig.directives) as GraphQLDirective[],
```
trevor-scheer, line 68: I'm surprised this works across both versions, wasn't this giving us trouble as one of the `Const...` node types?

trevor-scheer (approved):

.circleci/config.yml
```diff
@@ -80,6 +80,26 @@ jobs:
...
+      - run:
+          command: npm run clean && npm i
+      - run:
+          command: npm run test:ci
+          environment:
+            JEST_JUNIT_OUTPUT_DIR: "reports/junit/"
+      - store_test_results:
+          path: reports/junit
+      - store_artifacts:
+          path: reports/junit
```
trevor-scheer, line 102: @abernix I'd be curious to hear if you have any opinions on this. We've run into a couple v15 compat issues, this particular version straddle has a few gotchas. Seems like a reasonable way to catch this in CI.

I found the  `npm run clean && npm i` to workaround a problem where it seemed the TS build was out of date (probably was via the caching, which this breaks/works around). Maybe there's a better way that's still cacheable?

subgraph-js/package.json
```diff
@@ -23,7 +23,10 @@
   "publishConfig": {
     "access": "public"
   },
+  "dependencies": {
+    "apollo-server-types": "^0.9.0 || ^3.0.0"
```
trevor-scheer, line 27: Is this a stray addition?

query-planner-js/src/schema-helper/transformSchema.ts
clenfest: It's just casting away the readonly property. The issue we had before was that we had a function returning a `ConstValueNode` which wasn't in the same type hierarchy as `ValueNode`

subgraph-js/package.json
trevor-scheer: 32384c0

clenfest: I had to add it to get it to work, but I'm not sure why.

PR #1484, opened 2022-02-04
Creation of @override directive along with relevant tests

clenfest:
There are three new hints and two new errors to deal with some of the error conditions. Note that `override` can be defined both at the object as well as the field level.

### DIFF
--- a/composition-js/src/__tests__/hints.test.ts
+++ b/composition-js/src/__tests__/hints.test.ts
@@ -20,6 +20,9 @@ import {
   hintInconsistentExecutionDirectiveLocations,
   hintInconsistentArgumentPresence,
   hintInconsistentDescription,
+  hintFromSubgraphDoesNotExist,
+  hintOverrideDirectiveCanBeRemoved,
+  hintOverriddenFieldCanBeRemoved,
 } from '../hints';
 import { MergeResult, mergeSubgraphs } from '../merging';
 
@@ -604,4 +607,79 @@ test('hints on inconsistent description for field', () => {
     + '  I don\'t know what I\'m doing\n'
     + '  """'
   );
-})
+});
+
+describe('hint tests related to the @override directive', () => {
+  it('hint when from subgraph does not exist', () => {
+    const subgraph1 = gql`
+      type Query {
+        a: Int
+      }
+
+      type T @key(fields: "id"){
+        id: Int
+        f: Int @override(from: "Subgraph3")
+      }
+    `;
+
+    const subgraph2 = gql`
+    type T @key(fields: "id"){
+      id: Int
+    }
+    `;
+    const result = mergeDocuments(subgraph1, subgraph2);
+    expect(result).toRaiseHint(
+      hintFromSubgraphDoesNotExist,
+      `Source subgraph "Subgraph3" for field "T.f" on subgraph "Subgraph1" does not exist. Did you mean "Subgraph1" or "Subgraph2"?`,
+    );
+  });
+
+  it('hint when @override directive can be removed', () => {
+    const subgraph1 = gql`
+      type Query {
+        a: Int
+      }
+
+      type T @key(fields: "id"){
+        id: Int
+        f: Int @override(from: "Subgraph2")
+      }
+    `;
+
+    const subgraph2 = gql`
+    type T @key(fields: "id"){
+      id: Int
+    }
+    `;
+    const result = mergeDocuments(subgraph1, subgraph2);
+    expect(result).toRaiseHint(
+      hintOverrideDirectiveCanBeRemoved,
+      `Field "T.f" on subgraph "Subgraph1" no longer exists in the from subgraph. The @override directive can be removed.`,
+    );
+  });
+
+  it('hint overridden field can be removed', () => {
+    const subgraph1 = gql`
+      type Query {
+        a: Int
+      }
+
+      type T @key(fields: "id"){
+        id: Int
+        f: Int @override(from: "Subgraph2")
+      }
+    `;
+
+    const subgraph2 = gql`
+    type T @key(fields: "id"){
+      id: Int
+      f: Int
+    }
+    `;
+    const result = mergeDocuments(subgraph1, subgraph2);
+    expect(result).toRaiseHint(
+      hintOverriddenFieldCanBeRemoved,
+      `Field "T.f" on subgraph "Subgraph2" is overridden. Consider removing it.`,
+    );
+  });
+});
--- a/composition-js/src/__tests__/hints.test.ts
+++ b/composition-js/src/__tests__/hints.test.ts
@@ -682,4 +682,51 @@ describe('hint tests related to the @override directive', () => {
       `Field "T.f" on subgraph "Subgraph2" is overridden. Consider removing it.`,
     );
   });
+
+  it('hint overridden field can be made external', () => {
+    const subgraph1 = gql`
+      type Query {
+        a: Int
+      }
+
+      type T @key(fields: "id"){
+        id: Int @override(from: "Subgraph2")
+      }
+    `;
+
+    const subgraph2 = gql`
+      type T @key(fields: "id"){
+        id: Int
+      }
+    `;
+    const result = mergeDocuments(subgraph1, subgraph2);
+    expect(result).toRaiseHint(
+      hintOverriddenFieldCanBeRemoved,
+      `Field "T.id" on subgraph "Subgraph2" is overridden. It is still used in some federation directive(s) (@key, @requires, and/or @provides) and/or to satisfy interface constraint(s), but consider marking it @external explicitly or removing it along with its references.`,
+    );
+  });
+
+  it('hint when @override directive can be removed because overridden field has been marked external', () => {
+    const subgraph1 = gql`
+      type Query {
+        a: Int
+      }
+
+      type T @key(fields: "id"){
+        id: Int @override(from: "Subgraph2")
+        f: Int
+      }
+    `;
+
+    const subgraph2 = gql`
+    type T @key(fields: "id"){
+      id: Int @external
+    }
+    `;
+    const result = mergeDocuments(subgraph1, subgraph2);
+    expect(result).toRaiseHint(
+      hintOverrideDirectiveCanBeRemoved,
+      `Field "T.id" on subgraph "Subgraph1" is not resolved anymore by the from subgraph (it is marked "@external" in "Subgraph2"). The @override directive can be removed.`,
+    );
+  });
 });
--- a/composition-js/src/__tests__/override.compose.test.ts
+++ b/composition-js/src/__tests__/override.compose.test.ts
@@ -111,25 +111,6 @@ describe("composition involving @override directive", () => {
     `);
   });
 
-  // TODO: This test is very similar to the case of @key and `A.b` in Subgraph2
-  // needs to be "turned" into an @external.
-  //
-  // This does raise a related question for hints: currently, once a field is
-  // overridden, we raise an hint that says the field is safe to be removed,
-  // but it's not technically true in this example (the same applies to the @key case
-  // btw), and if the user blindly follow the hint, he will get an error due to
-  // the @provides referencing a now non-existing field. Instead, the user should
-  // either mark the field @external, or, alternative, also remove any directive
-  // referencing the now overridden field (and which one is more approapriate
-  // is likely case dependent: in the case of @key, if you've overridden a key,
-  // you're probably moving the whole entity anyway, so you'd want to remove the
-  // original key, but in this example, there is no particular reason you'd want
-  // to remove the @provides, which still make as much sense as before, if not
-  // more).
-  //
-  // Side-note: regarding hints, we probably want a test to ensure that if instead
-  // of removing the overriden field, you just mark it @external manually, then
-  // we still hint you to say that the @override can now be removed.
   it("override field in a @provides", () => {
     const subgraph1 = {
       name: "Subgraph1",
@@ -178,10 +159,32 @@ describe("composition involving @override directive", () => {
 
     const result = composeAsFed2Subgraphs([subgraph1, subgraph2]);
     assertCompositionSuccess(result);
-    // TODO: Should this
+
+    // Ensures `A.b` is marked external in Subgraph2 since it's overridden but there is still a provides mentioning it.
+    const typeA = result.schema.type("A");
+    expect(printType(typeA!)).toMatchInlineSnapshot(`
+      "type A
+        @join__type(graph: SUBGRAPH1, key: \\"id\\")
+        @join__type(graph: SUBGRAPH2, key: \\"id\\")
+      {
+        id: ID!
+        b: B @join__field(graph: SUBGRAPH1, override: \\"Subgraph2\\") @join__field(graph: SUBGRAPH2, external: true)
+      }"
+    `);
+
+    // Ensuring the provides is still here.
+    const typeT = result.schema.type("T");
+    expect(printType(typeT!)).toMatchInlineSnapshot(`
+      "type T
+        @join__type(graph: SUBGRAPH1, key: \\"k\\")
+        @join__type(graph: SUBGRAPH2, key: \\"k\\")
+      {
+        k: ID
+        a: A @join__field(graph: SUBGRAPH1) @join__field(graph: SUBGRAPH2, provides: \\"b { v }\\")
+      }"
+    `);
   });
 
-  // TODO: Similar issue that for @provides above, but for @requires.
   it("override field in a @requires", () => {
     const subgraph1 = {
       name: "Subgraph1",
@@ -227,7 +230,31 @@ describe("composition involving @override directive", () => {
 
     const result = composeAsFed2Subgraphs([subgraph1, subgraph2]);
     assertCompositionSuccess(result);
-    // TODO: more checks.
+
+    // Ensures `A.b` is marked external in Subgraph2 since it's overridden but there is still a requires mentioning it.
+    const typeA = result.schema.type("A");
+    expect(printType(typeA!)).toMatchInlineSnapshot(`
+      "type A
+        @join__type(graph: SUBGRAPH1, key: \\"id\\")
+        @join__type(graph: SUBGRAPH2, key: \\"id\\")
+      {
+        id: ID!
+        b: B @join__field(graph: SUBGRAPH1, override: \\"Subgraph2\\") @join__field(graph: SUBGRAPH2, external: true)
+      }"
+    `);
+
+    // Ensuring the requires is still here.
+    const typeT = result.schema.type("T");
+    expect(printType(typeT!)).toMatchInlineSnapshot(`
+      "type T
+        @join__type(graph: SUBGRAPH1, key: \\"k\\")
+        @join__type(graph: SUBGRAPH2, key: \\"k\\")
+      {
+        k: ID
+        a: A
+        x: Int @join__field(graph: SUBGRAPH2, requires: \\"a { b { v } }\\")
+      }"
+    `);
   });
 
   it("override from self error", () => {
--- a/composition-js/src/__tests__/override.compose.test.ts
+++ b/composition-js/src/__tests__/override.compose.test.ts
@@ -168,7 +168,7 @@ describe("composition involving @override directive", () => {
         @join__type(graph: SUBGRAPH2, key: \\"id\\")
       {
         id: ID!
-        b: B @join__field(graph: SUBGRAPH1, override: \\"Subgraph2\\") @join__field(graph: SUBGRAPH2, external: true)
+        b: B @join__field(graph: SUBGRAPH1, override: \\"Subgraph2\\") @join__field(graph: SUBGRAPH2, usedOverridden: true)
       }"
     `);
 
@@ -239,7 +239,7 @@ describe("composition involving @override directive", () => {
         @join__type(graph: SUBGRAPH2, key: \\"id\\")
       {
         id: ID!
-        b: B @join__field(graph: SUBGRAPH1, override: \\"Subgraph2\\") @join__field(graph: SUBGRAPH2, external: true)
+        b: B @join__field(graph: SUBGRAPH1, override: \\"Subgraph2\\") @join__field(graph: SUBGRAPH2, usedOverridden: true)
       }"
     `);
 
@@ -300,7 +300,7 @@ describe("composition involving @override directive", () => {
         @join__type(graph: SUBGRAPH2, key: \\"k\\")
       {
         k: ID
-        x: Int @join__field(graph: SUBGRAPH1, external: true) @join__field(graph: SUBGRAPH2, override: \\"Subgraph1\\")
+        x: Int @join__field(graph: SUBGRAPH1, usedOverridden: true) @join__field(graph: SUBGRAPH2, override: \\"Subgraph1\\")
       }"
     `);
   });
@@ -464,7 +464,7 @@ describe("composition involving @override directive", () => {
         @join__type(graph: SUBGRAPH1, key: \\"k\\")
         @join__type(graph: SUBGRAPH2, key: \\"k\\")
       {
-        k: ID @join__field(graph: SUBGRAPH1, override: \\"Subgraph2\\") @join__field(graph: SUBGRAPH2, external: true)
+        k: ID @join__field(graph: SUBGRAPH1, override: \\"Subgraph2\\") @join__field(graph: SUBGRAPH2, usedOverridden: true)
         a: Int @join__field(graph: SUBGRAPH1)
         b: Int @join__field(graph: SUBGRAPH2)
       }"
@@ -502,9 +502,6 @@ describe("composition involving @override directive", () => {
 
     const result = composeAsFed2Subgraphs([subgraph1, subgraph2]);
     expect(result.errors).toBeDefined();
-    // TODO: Those error messages might be currently a tad confusing for the user, because it essentially says that "k" is external
-    // in Subgraph2, but if a user looks at its subgraph, it's not external. Ideally, we should amend those message to recognize
-    // and indicate that the field is external _because_ it is overridden 
     expect(result.errors?.map(e => e.message)).toMatchStringArray([
       `
       The following supergraph API query:
@@ -515,8 +512,8 @@ describe("composition involving @override directive", () => {
       }
       cannot be satisfied by the subgraphs because:
       - from subgraph "Subgraph2":
-        - field "T.k" is not resolvable because marked @external.
-        - cannot move to subgraph "Subgraph1" using @key(fields: "k") of "T", the key field(s) cannot be resolved from subgraph "Subgraph2".
+        - field "T.k" is not resolvable because it is overridden by subgraph "Subgraph1".
+        - cannot move to subgraph "Subgraph1" using @key(fields: "k") of "T", the key field(s) cannot be resolved from subgraph "Subgraph2" (note that some of those key fields are overridden in "Subgraph2").
       `,
       `
       The following supergraph API query:
@@ -528,7 +525,7 @@ describe("composition involving @override directive", () => {
       cannot be satisfied by the subgraphs because:
       - from subgraph "Subgraph2":
         - cannot find field "T.a".
-        - cannot move to subgraph "Subgraph1" using @key(fields: "k") of "T", the key field(s) cannot be resolved from subgraph "Subgraph2".
+        - cannot move to subgraph "Subgraph1" using @key(fields: "k") of "T", the key field(s) cannot be resolved from subgraph "Subgraph2" (note that some of those key fields are overridden in "Subgraph2").
       `
     ]);
   });
@@ -616,7 +613,7 @@ describe("composition involving @override directive", () => {
         @join__type(graph: SUBGRAPH1)
         @join__type(graph: SUBGRAPH2)
       {
-        k: ID @join__field(graph: SUBGRAPH1, override: \\"Subgraph2\\") @join__field(graph: SUBGRAPH2, external: true)
+        k: ID @join__field(graph: SUBGRAPH1, override: \\"Subgraph2\\") @join__field(graph: SUBGRAPH2, usedOverridden: true)
         a: Int @join__field(graph: SUBGRAPH1)
         b: Int @join__field(graph: SUBGRAPH2)
       }"
--- a/composition-js/src/__tests__/override.compose.test.ts
+++ b/composition-js/src/__tests__/override.compose.test.ts
@@ -424,30 +424,14 @@ describe("composition involving @override directive", () => {
     };
 
     const result = composeAsFed2Subgraphs([subgraph1, subgraph2]);
-    assertCompositionSuccess(result);
-
-    const typeT = result.schema.type("T");
-    expect(printType(typeT!)).toMatchInlineSnapshot(`
-      "type T
-        @join__type(graph: SUBGRAPH1, key: \\"k\\")
-        @join__type(graph: SUBGRAPH2, key: \\"k\\")
-      {
-        k: ID
-        a: Int @join__field(graph: SUBGRAPH1, override: \\"Subgraph2\\")
-      }"
-    `);
-
-    const [_, api] = schemas(result);
-    expect(printSchema(api)).toMatchString(`
-      type Query {
-        t: T
-      }
-
-      type T {
-        k: ID
-        a: Int
-      }
-    `);
+    expect(result.errors?.length).toBe(1);
+    expect(result.errors).toBeDefined();
+    expect(errors(result)).toStrictEqual([
+      [
+        'FIELD_TYPE_MISMATCH',
+        'Field "T.a" has incompatible types across subgraphs: it has type "Int" in subgraph "Subgraph1" but type "String" in subgraph "Subgraph2"'
+      ]
+    ]);
   });
 
   it("override field that is a key in another type", () => {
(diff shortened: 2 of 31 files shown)


composition-js/src/__tests__/override.compose.test.ts
```diff
@@ -0,0 +1,270 @@
...
+      type Query
+        @join__type(graph: SUBGRAPH1)
+        @join__type(graph: SUBGRAPH2)
+      {
+        t: T @join__field(graph: SUBGRAPH1)
+      }
+
+      type T
+        @join__type(graph: SUBGRAPH1, key: \\"k\\")
+        @join__type(graph: SUBGRAPH2, key: \\"k\\")
```
pcmanus, line 88: I'm a little bit bothered that we have a `@join__type` for `SUBGRAPH2` here given the `@override` on the type. I do get why, and that `@override` is supposed to mostly mean "override on all of the fields", but in terms of the supergraph and join spec, this kind of mean that type `T` is in `SUBGRAPH2` but has no fields there. In fact, I was initially surprised that this didn't trigger errors, and the reason it doesn't is that [this loop](https://github.com/apollographql/federation/blob/main/internals-js/src/extractSubgraphsFromSupergraph.ts#L236) is not guarded by `isFed1`, but it could have been because before this PR, this logic should never kick in for a fed2 generated supergraph. Put another way, fed2 so far output precise information regarding where things are defined and don't need that kind of cleanup. And it bother me to steer away from this kind of property moving forward, as it makes it a bit harder to work with the supergraph.

To clarify, I agree this isn't a major concern because "it works currently" but I'd feel a tad better if we weren't generating the `join__type` in that case.

Though to be fair, we need to be careful with graphQL type extensions for this (see my other comment on the subject of type extensions).

composition-js/src/merging/merge.ts
```diff
@@ -815,6 +819,95 @@ class Merger {
     return this.metadata(sourceIdx).isFieldShareable(field);
   }
 
+  private isOverride(_sourceIdx: number, field: FieldDefinition<any>): boolean {
+    return !!this.getOverrideDirective(field);
+  }
+
+  private getOverrideDirective(field: FieldDefinition<any>): Directive<FieldDefinition<any>> | undefined {
```
pcmanus, line 826: I think there is actually subtlety here regarding `@override` on types due to the possible presence of type extensions (in the graphQL sense of the way). What I mean here is that you could have a subgraph that looks like this:
```graphql
type T {
  a: Int
  b: Int
}

extend type T @override(from: "X") {
  c: Int
  d: Int
}
```
And what we want for this is that the `@override` applies to `c` and `d` but not to `a` and `b`, but currently it would apply to all of them due to how we're dealing with extensions in `definitions.ts`. And to be clear, the handling of extensions is a tad ugly (but the rational is essentially that most code don't about them so I didn't wanted to make everything more complicated due to their presence, but the result is that when you do care about it, it's a tad weird).

Long story short, when looking at the type-level `@override` directive, the code should check if `directive.ofExtension() === field.ofExtension()` and ignore the directive (for that field) if that's not the case.

Note: I forgot to handle this properly for `@shareable`, which is why you may not have see it there. I'll create a separate PR to fix it.

composition-js/src/__tests__/override.compose.test.ts
```diff
@@ -0,0 +1,270 @@
...
+import "./matchers";
+import {
+  assertCompositionSuccess,
+  schemas,
+  errors,
+  composeAsFed2Subgraphs,
+} from "./compose.test";
+
+describe("composition involving @override directive", () => {
+  it("@override whole type", () => {
```
pcmanus, line 12: That test is the "idealised" case, but it would be nice to add tests for `@override` on type but where there is some mismatch between the fields, just to make sure this works as expected. Something like:
```graphql
# Subgraph1
type T @key(fields: "k") @override(from: "Subgraph2") {
  k: ID
  a: Int
  b: Int
}
```
```graphql
# Subgraph1
type T @key(fields: "k") {
  k: ID
  b: Int
  c: Int
}

```

composition-js/src/__tests__/override.compose.test.ts
```diff
@@ -0,0 +1,270 @@
...
+import gql from "graphql-tag";
+import "./matchers";
+import {
+  assertCompositionSuccess,
+  schemas,
+  errors,
+  composeAsFed2Subgraphs,
+} from "./compose.test";
+
+describe("composition involving @override directive", () => {
```
pcmanus, line 11: We should have a test for putting a `@override` on both a type and one of its field at the same time. Which I believe currently would only use the `@override` on the field for the field having it and would us the `@override` on the type for other fields. And I wonder if that's what we want? I think I'd rather just reject this case with an error for now.

composition-js/src/__tests__/hints.test.ts
```diff
@@ -603,4 +606,79 @@ test('hints on inconsistent description for field', () => {
     + '  I don\'t know what I\'m doing\n'
     + '  """'
   );
-})
+});
+
+describe('hint tests related to the @override directive', () => {
```
pcmanus, line 611: All the hints currently raises are field based. So in practice, if you were to move a type having 50 fields to another subgraph using a type-level `@override`, we'll generate 50 hints, which isn't an amazing user experience. Of course, generating type-level hints for this is also a bit tricky, because even if you put a `@override` on a type, the hint generated for each field of that type might actually differ in some cases (maybe the overriding type has actually more fields than the overridden one for instance). But I think overwhelming users with hints is a genuine potential problem and worth expanding some effort to improve.

That said, I'm ok punting on that problem in this PR, because we already know we need to rethink and cleanup the experience with hints and so we can defer the problem here then. But wanted to mention it nonetheless so that 1) we collectively have it in mind and 2) in case you actually prefer tackling that problem now.

composition-js/src/merging/merge.ts
```diff
@@ -815,6 +819,95 @@ class Merger {
...
+      return acc;
+    }, { subgraphsWithMoving: [], subgraphMap: {} });
+
+    subgraphsWithMoving.forEach((subgraphName) => {
+      const { overrideDirective } = subgraphMap[subgraphName];
+      const sourceSubgraphName = overrideDirective?.arguments()?.from;
+      if (!this.names.includes(sourceSubgraphName)) {
+        this.hints.push(new CompositionHint(
+          hintFromSubgraphDoesNotExist,
+          `Source subgraph '${sourceSubgraphName}' for field '${coordinate}' on subgraph '${subgraphName}' does not exist`,
```
pcmanus, line 882: On that particular hint, it could be nice to use the code from `internals-js/src/suggestions.ts` to check if the user might not have likely typoed the subgraph name and offer fix suggestions. Because while there may be rare cases where you use `@override` to sunset a full subgraph and thus ends up with some left-over `@override` point to a subgraph that don't exist anymore, I think the vast majority of case where user are going to get that message will be typos in the subgraph name.

pcmanus (changes requested):

composition-js/src/merging/merge.ts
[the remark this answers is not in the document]
clenfest: also, don't care. Double quotes are fine.

composition-js/src/__tests__/override.compose.test.ts
```diff
@@ -0,0 +1,270 @@
...
+      `,
+    };
+
+    const result = composeAsFed2Subgraphs([subgraph1, subgraph2]);
+    expect(result.errors?.length).toBe(2);
+    expect(result.errors).toBeDefined();
+    expect(errors(result)).toStrictEqual([
+      [
+        "OVERRIDE_SOURCE_HAS_OVERRIDE",
+        `Field 'T.a' on subgraph 'Subgraph1' has been previously marked with directive @override in subgraph 'Subgraph2'`,
```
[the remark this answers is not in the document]
clenfest: The main issue that I didn't do this is because the issue can be presumably happening in more than one place, and it's not really possible to say which is correct. I'll think about it a bit more.

composition-js/src/merging/merge.ts
clenfest: I really do not understand what ofExtension() is for. I'll try to dig in and see if it makes more sense to me, but I may have to ask you about this on Monday.

composition-js/src/__tests__/hints.test.ts
clenfest: Ok, I agree and think that we should punt as it seems like we might need a more general solution. Thanks for thinking about this.

composition-js/src/__tests__/override.compose.test.ts
clenfest: Created #1569 to address this

composition-js/src/merging/merge.ts
```diff
@@ -803,19 +824,123 @@ class Merger {
...
     return this.metadata(sourceIdx).isFieldShareable(field);
   }
 
-  private mergeField(sources: (FieldDefinition<any> | undefined)[], dest: FieldDefinition<any>) {
+  private getOverrideDirective(sourceIdx: number, field: FieldDefinition<any>): Directive<any> | undefined {
+    // Check the directive on the field, then on the enclosing type.
+    const overrideDirectiveName = this.metadata(sourceIdx).overrideDirective().name;
+    const fieldOverrideDirective = field.appliedDirectives.find(d => d.name === overrideDirectiveName);
+    const parent: FieldDefinition<any> | undefined = field.parent;
+    if (parent) { // should always be true
```
pcmanus, line 844: I usually prefer a `assert(parent, "should always be true")` because it avoids bumping the indentation and, probably more importantly, it does crash loudly if the assumption changes unexpectedly someday.

clenfest, line 844: Do you have any idea what type would be safe to put here?

composition-js/src/merging/merge.ts
```diff
@@ -690,8 +695,20 @@ class Merger {
   private mergeObject(sources: (ObjectType | undefined)[], dest: ObjectType) {
     const isEntity = this.hintOnInconsistentEntity(sources, dest);
     const isValueType = !isEntity && !dest.isRootType();
+    const typeOverriddenInSubgraphs = sources
+      .map((src, idx) => {
+        if (src) {
+          const overrideDirectiveName = this.metadata(idx).overrideDirective().name;
+          const overrideDirective = src.appliedDirectives.find(d => d.name === overrideDirectiveName);
```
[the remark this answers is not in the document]
clenfest: Since the type of the source is `ObjectType`, I wasn't able to figure out how to do the `ofExtension` comparison for this version. Let me know what you think.

composition-js/src/merging/merge.ts
```diff
@@ -803,19 +821,218 @@ class Merger {
...
+  }): { result: boolean, conflictingDirective?: string, subgraph?: string } {
+    // helper function to let us know if a given directive exists on a field
+    const directiveOnField = (field: FieldDefinition<any> | undefined, def: DirectiveDefinition<any> | undefined): boolean => {
+      if (!def || !field) {
+        return false;
+      }
+      const directives = field.appliedDirectivesOf(def);
+      const directiveOnField = directives.find(d => d.ofExtension() === field.ofExtension());
+      const directivesOnType: Directive<any>[] = field.parent.appliedDirectivesOf(def);
+      const directiveOnType = directivesOnType.find(d => d.ofExtension() === field.parent.ofExtension);
```
pcmanus, line 892: Same remark that in `getOverrideDirective` regarding the extension check for field directives. And more generally, this helper is only called for provides and requires and neither of them can be applied on types, so I think that helper can be removed and `directiveOnField(fromField, fromMetadata.providesDirective())` could be replaced by `fromField.hasAppliedDirective(fromMetadata.providesDirectives())`.

As an aside, I agree with not allowing overriding a field with a provides/requires (at least for now), but it would be awesome to add quick tests for those.

internals-js/src/error.ts
```diff
@@ -306,6 +306,26 @@ const SATISFIABILITY_ERROR = makeCodeDefinition(
...
+);
+
+const OVERRIDE_ON_BOTH_FIELD_AND_TYPE = makeCodeDefinition(
+  'OVERRIDE_ON_BOTH_FIELD_AND_TYPE',
+  'Field and type are both marked with @override directive. This is not currently supported.',
+);
+
+const OVERRIDE_COLLISION_WITH_ANOTHER_DIRECTIVE = makeCodeDefinition(
+  'OVERRIDE_COLLISION_WITH_ANOTHER_DIRECTIVE',
+  'Override cannot be used in conjunction with specific other directives.',
```
pcmanus, line 326: Could be nice to list the directives that have incompatibility, instead of "specific other". I'd even go as far as something like `The @override directive cannot be used on external fields, nor to override fields with either @external, @provides or @requires`. My reasoning is that it's a tad annoying when you read doc and things are vague and you need to manually experiment to figure out what is allowed/not allowed.

internals-js/src/federation.ts
```diff
@@ -486,6 +496,10 @@ export class FederationMetadata {
     return this.sharingPredicate()(field);
   }
 
+  isFieldKey(field: FieldDefinition<any>): boolean {
```
pcmanus, line 499: I'd prefer `isKeyField` for this. Because in really this method does not guarantee that the field is a key all by himself, only that it is part of some key, and I'd rather avoid any risk of confusion (I do get that this might be for consistency with `isFieldExternal` but the same problem does not apply for external).

internals-js/src/federation.ts
```diff
@@ -567,7 +585,9 @@ export class FederationMetadata {
       this.extendsDirective(),
     ];
     return this.isFed2Schema()
-      ? baseDirectives.concat(this.shareableDirective())
+      ? baseDirectives
+        .concat(this.shareableDirective())
```
pcmanus, line 589: Nit: `concat` accepts a list of arguments so it can be done with a single concat (not that it matters, and happy to keep as is if you prefer, just wondered if this was an oversight).

composition-js/src/__tests__/override.compose.test.ts
```diff
@@ -0,0 +1,678 @@
...
+      typeDefs: gql`
+        type T @key(fields: "k") {
+          k: ID
+          a: Int @override(from: "Subgraph1")
+        }
+      `,
+    };
+
+    const result = composeAsFed2Subgraphs([subgraph1, subgraph2]);
+    // TODO: This test really should not cause the shareable hint to be raised, but to fix it would be a bit of a pain, so punting
```
pcmanus, line 301: Nit: s/shareable hint/shareable error/

internals-js/src/precompute.ts
```diff
@@ -0,0 +1,78 @@
...
+} from ".";
+
+export const computeShareables = (schema: Schema) => {
+  const metadata = federationMetadata(schema);
+  return computeAllFieldsWithDirective(schema, metadata?.isFed2Schema() ? metadata.shareableDirective() : undefined, true);
+};
+
+export const computeKeys = (schema: Schema) => {
+  const metadata = federationMetadata(schema);
+  return computeAllFieldsWithDirective(schema, metadata?.isFed2Schema() ? metadata.keyDirective() : undefined, false);
```
pcmanus, line 21: I don't think that's correct for keys, even with the `includeOtherFieldsOnType == false` because the code of `computeAllFieldsWithDirective` still adds provided fields, which aren't keys. Also, I'm not sure why we pass `undefined` as `directive` for fed1 schemas? I'd assume a function that pre-compute all fields that are part of keys should work just fine for fed1 schema.

internals-js/src/precompute.ts
```diff
@@ -0,0 +1,78 @@
...
+};
+
+function computeAllFieldsWithDirective(schema: Schema, directive: DirectiveDefinition<any> | undefined, includeOtherFieldsOnType: boolean): (field: FieldDefinition<CompositeType>) => boolean {
```
pcmanus, line 24: In general, I think this method doesn't really do what the name implies and is still too tied to shareable (and there is still a bunch of mention to shareable which highlight this function is probably the wrong thing to use to compute keys).

Overall, the way I'd organise this to first have a method that precompute the list of field (coordinates) that appears in any `@key`. That code should not bother with provides or even have to care about extensions in any ways. It's probably something along the lines of just:
```typescript
metadata.keyDirective().applications().flapMap((key) => collectTargetFields(...)).map((f) => f.coordinate);
```
In fact, that method should probably be made generic to also allow pre-compute all the requires and provides fields (separately) since as mentioned in another comment, we should handle those consistently with how key fields are handled for override.

Then shareability can be defined using the precomputed list of keys since shareable fields are fields that either:
- key fields
- or provided fields
- or field with @shareable or part of a type with @shareable.

composition-js/src/__tests__/override.compose.test.ts
```diff
@@ -0,0 +1,678 @@
...
+import gql from "graphql-tag";
+import "./matchers";
+import {
+  assertCompositionSuccess,
+  schemas,
+  errors,
+  composeAsFed2Subgraphs,
+} from "./compose.test";
+
+describe("composition involving @override directive", () => {
```
pcmanus, line 11: Regarding the test of that file, I'd avoid matching the full supergraph every time. Mostly because if we do, we'll have to update those test every time we have a supergraph format change, even if the change is completely unrelated to those test, and that's annoying. I think we can just have those test check the part of the supergraph they care about, and I've push a commit to do so if you agree: https://github.com/pcmanus/federation/commit/d729eef5da6309b15f9b1c52da3d4e963c25a593.

But another point on testing is that I think we should have least a handful of `@override`-based test in the query planner, because even if the query planner happens to not need modification for this, `@override` is about making sure the query planner doesn't use the overridden fields anymore and we should have tests that test the full feature.

pcmanus (changes requested):
A few points on top of the other inline comments:
- I think `@provides` and `@requires` should essentially be handle consistently with `@key`. All those directives reference fields and so when we override fields, we might "break" of one those reference in the overridden subgraph if we remove the field (and so, like for `@key` such field should be made @external). It is true that `@provides` and `requires` usually applies to external fields in the first and hence are "less" impacted than `@key`, but due to nesting they can still reference some non-external fields, and overriding those create issues. I've push a commit with tests (and a few related notes/remarks) for both `@provides` and `@requires` if you want: https://github.com/pcmanus/federation/commit/d9cf0a4c91cd69a09f245255a9c5b40822f7b0c4.
- It's possible that overriding a key fields breaks composition. Which is ok, but I think the error message we currently issue in that case is going confusing to user and should be amended to take overrides into account. I pushed a test for that as well with some more details here: https://github.com/pcmanus/federation/commit/9dae95045819ff418809b29db8fe49a6d0fd51fb.

internals-js/src/error.ts
clenfest: good shout.

internals-js/src/precompute.ts
clenfest: I restored your original version of computeShareables. Let me know if I got the right message on computeKeys. If not, we should talk about it in person tomorrow.

federation-integration-testsuite-js/src/fixtures/accounts.ts
```diff
@@ -78,6 +78,7 @@ export const typeDefs = gql`
     id: ID!
     name: String @external
     userAccount(id: ID! = "1"): User @requires(fields: "name")
+    description: String
```
tinnou, line 81: Long story short, we implement Federation v1 internally and are planning to move to v2 and I'm facing the following issue; if an overridden field is on a type that implements an interface, composition will fail since removing `@join__field` from the supergraph sdl yields an invalid subgraph when extracting subgraphs from the  supergraph. 

To reproduce, we can add the following:

```
  type Library implements Describable @key(fields: "id") {
    ...
   }
  interface Describable {
    description: String
  }
```

and any tests in `buildQueryPlan.test.ts` will now fail with the following error:

```
Unexpected error extracting subgraph accounts from the supergraph: this is either a bug, or the supergraph has been corrupted..

    Details:
    [accounts] Interface field Describable.description expected but Library does not provide it.
    
    at extractSubgraphsFromSupergraph (../internals-js/src/extractSubgraphsFromSupergraph.ts:300:15)
      at buildFederatedQueryGraph (../query-graphs-js/src/querygraph.ts:523:51)
      at compose (../composition-js/src/compose.ts:53:55)
      at composeServices (../composition-js/src/compose.ts:85:10)
      at Object.<anonymous> (../federation-integration-testsuite-js/src/fixtures/special-cases/supergraphWithInaccessible.ts:13:42)
      at Object.<anonymous> (../federation-integration-testsuite-js/src/fixtures/index.ts:56:1)
```

I stumbled upon this PR and I'm really looking forward to the override functionality! Thank you for putting this one together 🙏

pcmanus: Thanks, and you're right, we also need to handle interfaces here.

@clenfest: that's a case we forgot. One other reason why an overridden field may need to be kept external rather than removed is if that field exists to fulfil an interface. Like for key/provides/requires, we should detect it and mark the field external in the supergraph in that case.

gateway-js/src/__tests__/buildQueryPlan.test.ts
```diff
@@ -1171,4 +1173,31 @@ describe('buildQueryPlan', () => {
...
+          }
+        }
+      `;
+
+      const queryPlan = buildPlan(operationString);
+      expect(queryPlan).toMatchInlineSnapshot(`
+        QueryPlan {
+          Fetch(service: "books") {
+            {
+              library(id: 3) {
```
clenfest, line 1193: I think this might be wrong, but I need to do some more checking today.

internals-js/src/federation.ts
```diff
@@ -486,6 +497,18 @@ export class FederationMetadata {
     return this.sharingPredicate()(field);
   }
 
+  isKeyField(field: FieldDefinition<any>): boolean {
+    return this.keysPredicate()(field);
+  }
+
+  addSyntheticExternal(coordinate: string) {
```
pcmanus, line 504: I haven't add time to check all the recent changes yet, but wanted to give early feedback on this one.

I'm worried about this, because this change what the `FederationMetadata` abstraction is, and in a way that I worry will hurt maintenance. What I mean here is that so far, `FederationMetadata` is kind of "stateless" in the sense that it's only meant to provide convenient access to federation-related information, and it does pre-compute/cache things, but it can all be recomputed from the schema and `FederationMetadata` has no information of its own (and btw, I reckon that the name, `FederationMetadata`, may not be ideal in that it does not convey that very well).

Importantly, that means that if we clone or print a schema, we don't care about `FederationMetadata`, there is no real information in it.

This change that, and that worries me.

Fwiw, I see 2 potential alternatives:
1. we modify merging to only include an overridden field as external in the subgraph if it is used by a key/provides/requires/interface. Essentially, the only place when `isSyntheticExternal` is used is to bypass the "is that external used?" check, so we can simply avoid adding the overridden field at all if it's not used. It's the solution that I initially suggested because it's imo the simplest one.
2. if we instead want to always keep overridden fields in the supergraph and have them extracted as special "synthetic external", then I'd prefer creating a new specific `@federation__internal__syntheticExternal` directive to mark the field so that all information remain contained in the schema itself. I wouldn't vote for this option as I think the previous one is simpler, at least for now, and good enough, but I wouldn't oppose it either.

clenfest: Ok, went with option 1. Let me know what you think.

gateway-js/src/__tests__/buildQueryPlan.test.ts
```diff
@@ -1171,4 +1173,31 @@ describe('buildQueryPlan', () => {
       `);
     });
   });
+
+  describe('overridden fields and type', () => {
+    it(`query plan of overridden field`, () => {
```
pcmanus, line 1178: Fwiw, I dislike the test fixtures/the tests using them. I find those tests annoying to work with because you need to navigate a bunch of files to even start to figure out what the test even does (unless you happen to know the fixtures intimately, and well, I don't). Not to mention that when tons of tests use the same fixtures, it's much too easy to modify them for some new test without realising that the changes invalidates some existing test (by which I mean, those existing test might still pass but stop actually testing what they were meant to).

Anyway, that was my rant, and it's ok to not agree (and totally fine keeping that test for this PR), but if you simply happened to have stumbled on that test file and though this was the status quo, I'll note that the query planning tests I've personally added have been added mostly to `query-planner-js/src/__tests__/buildPlan.test.ts` and are self-contained test.

But that point aside, I think this test would be more useful if the `@override` was in `accounts` rather than `books`. Because as it stand, I'd expect the query plan of this test _even_ if `@override` wasn't working properly because that's obviously more efficient that going into `accounts`. But having the `@override` in `accounts` would allow to validate that the query planner _has to_ force a jump into `accounts` and cannot just use the definition of `description` in `books`.

internals-js/src/error.ts
```diff
@@ -359,6 +359,26 @@ const SATISFIABILITY_ERROR = makeCodeDefinition(
...
+  'OVERRIDE_FROM_SELF_ERROR',
+  'Field with `@override` directive has "from" location that references its own subgraph.',
+);
+
+const OVERRIDE_SOURCE_HAS_OVERRIDE = makeCodeDefinition(
+  'OVERRIDE_SOURCE_HAS_OVERRIDE',
+  'Field which is overridden to another subgraph is also marked @override.',
+);
+
+const OVERRIDE_ON_BOTH_FIELD_AND_TYPE = makeCodeDefinition(
```
pcmanus, line 372: It's not a huge deal, but I'll note that this error code is currently unused since we don't yet support override on types, but the code still currently shows in the generated doc. I understand that we want override on types as a fast followup, but I've been burned by "guaranteed-fast-followup" a few too many times and my preference would be to remove this for now.

internals-js/src/federation.ts
```diff
@@ -487,6 +510,10 @@ export class FederationMetadata {
     return this.sharingPredicate()(field);
   }
 
+  isKeyField(field: FieldDefinition<any>): boolean {
```
pcmanus, line 513: Following my other comment on `merge.ts`, I don't think this method is currently necessary (since it's a strict subset of `isUsedField`). So maybe we can remove it and `computeKeys` for now to avoid dead code?

internals-js/src/federation.ts
```diff
@@ -463,6 +467,25 @@ export class FederationMetadata {
...
+  private keysPredicate(): (field: FieldDefinition<CompositeType>) => boolean {
+    if (!this._keysPredicate) {
+      this._keysPredicate = computeKeys(this.schema);
+    }
+    return this._keysPredicate;
+  }
+
+  private fieldUsedPredicate(): (field: FieldDefinition<CompositeType>) => boolean {
+    if (!this._fieldUsedPredicate) {
+      const usedFields = Array.from(collectUsedFields(this));
```
pcmanus, line 479: I think we should use a `Set` for this. On large schema, there could be quite a few "used" fields overall, and `isUsedField` might also be used for many fields, so the linear scan could realistically add up. It's fine (even admittedly cleaner) that `collectUsedFields` now return `FieldDefinition`s rather than just the coordinates, but I would still convert back to a `Set<string>` at that stage so that `_fieldUsedPredicate` can be a simple set lookup.

internals-js/src/federation.ts
```diff
@@ -563,7 +594,7 @@ export class FederationMetadata {
   }
 
   allFederationDirectives(): DirectiveDefinition[] {
-    const baseDirectives = [
+    const baseDirectives: DirectiveDefinition<any>[] = [
```
pcmanus, line 597: Nit: we can save the `<any>` in `DirectiveDefinition<any>[]` if we want (it avoid a `any` and would more consistent with the method signature too).

composition-js/src/merging/merge.ts
```diff
@@ -884,15 +893,176 @@ class Merger {
...
+    }
+    return { result: false };
+  }
+
+  /**
+   * Validates whether or not the use of the @override directive is correct.
+   * return a list of subgraphs to ignore for the current field
+   * return value is a list of fields that has been filtered to ignore overridden fields
+   */
+  private validateOverride(sources: FieldOrUndefinedArray, { coordinate }: FieldDefinition<any>) {
```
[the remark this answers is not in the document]
clenfest: I have my editor synthetically add them if they don't exist, so sometimes I don't notice that they aren't "really" there.

composition-js/src/merging/merge.ts
```diff
@@ -884,15 +893,176 @@ class Merger {
...
+        });
+        if (hasIncompatible) {
+          this.errors.push(ERRORS.OVERRIDE_COLLISION_WITH_ANOTHER_DIRECTIVE.err({
+            message: `@override cannot be used on field "${fromField?.coordinate}" on subgraph "${subgraphName}" since "${fromField?.coordinate}" on "${subgraph}" is marked with directive "@${conflictingDirective}"`,
+          }));
+        } else {
+          // if we get here, then the @override directive is valid
+          // if the field being overridden is used, then we need to add an @external directive
+          assert(fromField, 'fromField should not be undefined');
+          if (this.metadata(fromIdx).isFieldUsed(fromField) || this.metadata(fromIdx).isKeyField(fromField)) {
```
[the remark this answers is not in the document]
clenfest: `isKeyField` is actually necessary, but I agree that the way it's structured is confusing. I modified `isFieldUsed` to call `isKeyField` directly.

gateway-js/src/__tests__/buildQueryPlan.test.ts
clenfest: I'm not going to change out of the using the fixtures now, but I will in the future as I agree that it makes things less straightforward.

composition-js/src/merging/merge.ts
```diff
@@ -1014,10 +1004,15 @@ class Merger {
       } else if (sourceSubgraphName === subgraphName) {
         this.errors.push(ERRORS.OVERRIDE_FROM_SELF_ERROR.err({
           message: `Source and destination subgraphs "${sourceSubgraphName}" are the same for overridden field "${coordinate}"`,
+          nodes: overrideDirective?.sourceAST,
         }));
       } else if (subgraphsWithOverride.includes(sourceSubgraphName)) {
         this.errors.push(ERRORS.OVERRIDE_SOURCE_HAS_OVERRIDE.err({
           message: `Field "${coordinate}" on subgraph "${subgraphName}" is also marked with directive @override in subgraph "${sourceSubgraphName}". Only one @override directive is allowed per field.`,
+          nodes: [
```
pcmanus, line 1012: nit: there is a `sourceASTs` function that I've been using to simply passing those. That is,
```typescript
nodes: sourceASTs(overrideDirective, subgraphMap[sourceSubgraphName].overrideDirective)
```
should do the trick.

internals-js/src/federation.ts
```diff
@@ -476,14 +476,14 @@ export class FederationMetadata {
...
-      this._fieldUsedPredicate = (field: FieldDefinition<CompositeType>) => !!usedFields.find((f) => f.coordinate === field.coordinate);
+      const usedFields = collectUsedFields(this);
+      this._fieldUsedPredicate = (field: FieldDefinition<CompositeType>) => !!usedFields.has(field);
     }
     return this._fieldUsedPredicate;
   }
 
   isFieldUsed(field: FieldDefinition<CompositeType>): boolean {
-    return this.fieldUsedPredicate()(field);
+    return this.fieldUsedPredicate()(field) || this.keysPredicate()(field);
```
pcmanus, line 486: I still don't understand why that `keysPredicate` check is needed, could you explain?

Because as far as I can tell, `collectUsedFields` collects all fields used in a `@key`, that's even the first thing it does, but that's also what `keysPredicate` does, isn't it? Is there a subtle difference here I'm not seeing?

clenfest: The existing `collectUsedFields` will only collect fields that are marked with `@external`, so we also need to make sure that we're finding anything that is referenced by a key.

pcmanus: Oh, I missed that part. Well, I think it's an issue, because for this feature we actually need to check the non-external field that are used, and this for `@provides`, `@requires` and interfaces (it's also confusing that it's called `isUsedField` and only return "external used field" versus "all used field").

What we should do is change `isUsedField` collect all used fields, and change the other consumers of `collectUsedFields` to filter non-external field out if they don't care about them. But for override, we explicitly care about only about the non-external used field.

composition-js/src/merging/merge.ts
```diff
@@ -884,15 +893,175 @@ class Merger {
...
+       }
+    }
+    if (field && this.isExternal(idx, field)) {
+      return {
+        result: true,
+        conflictingDirective: fromMetadata.externalDirective(),
+        subgraph: subgraphName,
+      };
+    }
+    if (fromField && this.isExternal(fromIdx, fromField)) {
```
pcmanus, line 944: This actually only just occurred to me, but I think we should remove that condition. The reason is that if an overridden field is still "used" in the from subgraph, then instead of removing the field, you may want to mark it `@external` instead. But the process you would then want is:
1. override some field `f` from subgraph A in subgraph B
2. mark `f` in subgraph A as `@external` explicitly.
3. now remove the `@override` in subgraph B.

Except that currently, this rule will prevent step 2.

Overall, overriding a field that is marked external is not that different from adding an overriding point to a subgraph that don't have the field at all, and we don't error out in that later case, so it make sense not too in the former either.

internals-js/src/federation.ts
clenfest: Ok, I understand now. I had stripped out the external restriction from the top level function, but not the helper function, so the caller is already doing the right thing. I'll change the naming and then get rid of the keys specific version.

composition-js/src/merging/merge.ts
clenfest: ok, that makes sense.

pcmanus (approved):
Awesome work, and thanks for the patience.

PR #1487, opened 2022-02-08
fixing misspelling issue on documentation 

ofuenmayor:


trevor-scheer:
Thanks for the PR! However this is as intended, `Product` is the type itself (in code), we don't want the 's' included in the code block.

PR #1496, opened 2022-02-09
chore: removes all rust packages

EverlastingBugstopper:
this commit removes all rust packages because they were moved to
federation-rs.

~this commit also adds integration tests that clone the federation-rs
repo and run cargo tests. these tests are only run in ci by default
and can be run locally with npm test:rs~


EverlastingBugstopper:
a follow-up PR will be incoming to `version-0.x` when this is approved by a few folks

abernix:
@EverlastingBugstopper `federation-rs` is now public. (haven't reviewed anything else here because 📲🤳)

.prettierrc.js
```diff
@@ -15,7 +15,7 @@ module.exports = {
   requirePragma: true,
   overrides: [
     {
-      files: '{docs/{,source/**},.,{gateway-js,harmonizer,federation-integration-testsuite-js,query-planner-js,router-bridge,subgraph-js}/**,test}/{*.js,*.ts}',
+      files: '{docs/{,source/**},.,{gateway-js,federation-rs-integration/federation-integration-testsuite-js,query-planner-js,subgraph-js}/**,test}/{*.js,*.ts}',
```
trevor-scheer, line 18:
```suggestion
      files: '{docs/{,source/**},.,{gateway-js,federation-rs-integration,federation-integration-testsuite-js,query-planner-js,subgraph-js}/**,test}/{*.js,*.ts}',
```
Accidental slash instead of comma?

federation-rs-integration/.gitignore
```diff
@@ -0,0 +1,2 @@
+/node_modules
```
trevor-scheer, line 1: I suspect this shouldn't be needed

federation-rs-integration/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "federation-rs-integration",
+  "version": "1.0.0",
+  "description": "",
+  "main": "shell.ts",
+  "scripts": {
+    "clean": "rm -rf ./federation-rs ./node_modules"
```
trevor-scheer, line 7: I _think_ the top level clean script should take care of this if the gitignore stuff is correct

abernix (commented):
Some comments! I would like to lean on @trevor-scheer's thoughts as much as possible though because he has a bit more first-hand experience with these things lately.

EverlastingBugstopper:
ok i've gone and just removed all of the integration tests since i can't get them to work. we'll just make sure to run tests on the rust side before releasing the crates and everything should be fine.

lerna.json
```diff
@@ -3,10 +3,9 @@
     "internals-js",
     "query-graphs-js",
     "federation-integration-testsuite-js",
+    "federation-rs-integration",
```
trevor-scheer, line 6: Is this still happening?

package.json
```diff
@@ -52,10 +45,7 @@
     "@iarna/toml": "2.2.5",
```
trevor-scheer, line 45: I think this was used by a script that I wrote for updating TOML files during release, but not sure it ever made it. I think this can go away in any case.

package.json
```diff
@@ -102,5 +90,9 @@
       "<rootDir>/query-planner-js",
       "<rootDir>/subgraph-js"
     ]
+  },
+  "volta": {
```
trevor-scheer, line 94: I'm interested to know more about this. Is it staying? What do we use it for?

tsconfig.build-stage-01.json
```diff
@@ -1,11 +0,0 @@
-{
```
trevor-scheer, line 1: I was excited for this one

trevor-scheer (approved):
Excited for this to land!

package.json
EverlastingBugstopper: so volta is what i use to manage my node runtime versions across a bunch of different projects. i'd love to keep it and it's helpful for folks who use volta and will just be ignored by folks who don't use it.

EverlastingBugstopper:
ok i've updated to address your comments @trevor-scheer. we'll need to disable the protections on `main` that require rust tests to pass since they'll never run and then i think we can merge this along with #1518

trevor-scheer:
@EverlastingBugstopper lgtm! Needs one more `npm install` to update the package-lock after removing the dev dependency. The volta stuff can stay! Was mostly just curious about it.

EverlastingBugstopper:
done! think we're good to go. i dont have admin privileges so will likely need @abernix to delete those protections from the `main` and `version-0.x` branches and we can merge it up.

PR #1499, opened 2022-02-09
fix(subgraph): Fix `GraphQLResolverMap` types - allow `args` fields to be typed by users

trevor-scheer:
The current typings are restrictive for the `TArgs` field. Users
who wish to use `graphql@^15.8.0` will encounter a default of
`{ [key: string]: any }` type for the `args` parameter, which will
effectively prevent them from typing their `args` parameter correctly.

As demonstrated in the thread:
https://community.apollographql.com/t/buildsubgraphschema-typescript-typing-conflict-with-resolver/2716/3

Changing the type of `TArgs` to `any` is effectively the same change that
v16 moves to (defaulting to `any` instead of the above mentioned object-y
type).

Reported by @Axel-p

### DIFF
--- a/federation-integration-testsuite-js/src/resolverMap.ts
+++ b/federation-integration-testsuite-js/src/resolverMap.ts
@@ -10,10 +10,10 @@ export interface GraphQLResolverMap<TContext = {}> {
   [typeName: string]:
     | {
         [fieldName: string]:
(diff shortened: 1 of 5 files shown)


clenfest (approved):

trevor-scheer:
FYI - no port needed (this only relates to v15.x users, which fed 2 doesn't support)

Issue #1500, opened 2022-02-10
Gateway 0.x/2.x Open Telemetry traces missing with otel collector 0.43.0+


prasek:
## Problem

[otel/opentelemetry-collector v0.43.0](https://github.com/open-telemetry/opentelemetry-collector/releases/tag/v0.43.0) removed deprecated legacy path ("v1/trace") support for otlp http receiver
- https://github.com/open-telemetry/opentelemetry-collector/pull/4720
- https://github.com/open-telemetry/opentelemetry-collector/issues/1968

[supergraph-demo with otel/opentelemetry-collector 0.43.0](https://github.com/apollographql/supergraph-demo/blob/main/docker-compose.otel-collector.yml#L37) shows error samples in zpages:
- http://localhost:55679/debug/tracez shows errors `http.status_code=404, http.target=/v1/trace`

<details><summary>Error Trace Spans</summary>

```
3 Requests

When                       Elapsed (sec)
----------------------------------------
2022/02/10-02:39:24.916957      .    77 trace_id: a8f6732587ff36e0e1b45c5c6e335322 span_id: 6aa3347a98d8cbbc
                                        Status{Code=Error, description=""}
                                        Attributes:{http.flavor=1.1, http.host=collector:55681, http.method=POST, http.request_content_length=5047, http.scheme=http, http.status_code=404, http.target=/v1/trace, http.wrote_bytes=19, net.host.name=collector, net.host.port=55681, net.peer.ip=172.22.0.2, net.peer.port=41562, net.transport=ip_tcp}
2022/02/10-02:39:32.619622      .    35 trace_id: 944f49a15fd55768af3df35c7f965a3f span_id: 67bb4fd243bc30ef
                                        Status{Code=Error, description=""}
                                        Attributes:{http.flavor=1.1, http.host=collector:55681, http.method=POST, http.request_content_length=30437, http.scheme=http, http.status_code=404, http.target=/v1/trace, http.wrote_bytes=19, net.host.name=collector, net.host.port=55681, net.peer.ip=172.22.0.6, net.peer.port=59220, net.transport=ip_tcp}
2022/02/10-02:39:35.647788      .    41 trace_id: 9e672ce2916c5e67929e3254dc8c3b08 span_id: c00e14f481d47c0d
                                        Status{Code=Error, description=""}
[...]

PR #1508, opened 2022-02-11
Update docs for otel 1.0

prasek:
Docs for #1500

Signed-off-by: Phil Prasek <prasek@gmail.com>

### DIFF
--- a/docs/source/opentelemetry.mdx
+++ b/docs/source/opentelemetry.mdx
@@ -1,5 +1,5 @@
 ---
-title: OpenTelemetry in Apollo Server
+title: OpenTelemetry in Apollo Federation
 sidebar_title: OpenTelemetry
 ---
 
@@ -27,11 +27,13 @@ To use OpenTelemetry in your application, you need to install a baseline set of 
 
 ```bash
 npm install \
-  @opentelemetry/api \
-  @opentelemetry/node@0.22 \
-  @opentelemetry/core@0.22 \
-  @opentelemetry/instrumentation-http@0.22 \
-  @opentelemetry/instrumentation-express@0.22
+  @opentelemetry/api@1.0 \
+  @opentelemetry/core@1.0 \
+  @opentelemetry/resources@1.0 \
+  @opentelemetry/sdk-trace-base@1.0 \
+  @opentelemetry/sdk-trace-node@1.0 \
+  @opentelemetry/instrumentation-http@0.27 \
+  @opentelemetry/instrumentation-express@0.28
 ```
 
 </ExpansionPanel>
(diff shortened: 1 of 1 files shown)


docs/source/opentelemetry.mdx
```diff
@@ -206,9 +213,39 @@ You can show the details of any operation and see a breakdown of its processing
 
 Our example telemetry configuration assumes that Zipkin is running locally, and that we want to process every span individually as it's emitted.
 
-To prepare for production, we probably at least want to conditionally set the `url` option for the `ZipkinExporter` and replace our `SimpleSpanProcessor` with a  `BatchSpanProcessor`.
+To prepare for production, we'll want to optimize performance by sending our traces to an [OpenTelemetry Collector](https://opentelemetry.io/docs/collector/) using the `OTLPTraceExporter` and replace our `SimpleSpanProcessor` with a `BatchSpanProcessor`.
+The `Collector` should be deployed as a local sidecar `Agent` to buffer traces before they're sent along to their final destination.
```
abernix, line 217: What is `Agent`?  What is `Collector`? (I mean, specifically, they are code-fenced.) They're not explained here / used in the code?

If you're talking about them in a general sense, let's choose a different visual treatment.

docs/source/opentelemetry.mdx
```diff
@@ -1,12 +1,14 @@
...
-title: OpenTelemetry in Apollo Server
+title: OpenTelemetry in Apollo Federation
 sidebar_title: OpenTelemetry
 ---
 
 import {
   ExpansionPanel,
 } from 'gatsby-theme-apollo-docs/src/components/expansion-panel';
 
+> 📣 **Open Telemetry demo:** In addition to completing this tutorial, you can [clone our Federation demo on GitHub](https://github.com/apollographql/supergraph-demo-fed2). The demo is distinct from the project you create in this tutorial, and it helps demonstrate the use of [Open Telemetry in Federation](https://github.com/apollographql/supergraph-demo-fed2#tracing-with-open-telemetry).
```
abernix, line 10: I love the fully-functional end-to-end demo, and I know it's called "Demo", butt I find it a bit incongruent to send someone off to a completely "distinct" experience.   I'm worried that it itself will become a pattern, particularly because it has other moving pieces which are not supported, including the underlying [`supergraph-demo-opentelemetry`](https://npm.im/supergraph-demo-opentelemetry) package which — while great — is still a maintenance concern and something we'll need to think about and potentially support.  We can commit to updating these docs right now and going forward, but docs were what we agreed to for this.

Thoughts?  Maybe there's another treatment we can give this? I think the happy path should ideally be outlined below, but maybe there's another way to position it?

abernix (commented):
I won't be able to actually test this or get help with testing it until next week, but just a couple thoughts right now!

docs/source/opentelemetry.mdx
prasek: fixed

prasek:
@abernix appreciate the feedback, all suggestions incorporated!

abernix (approved):

PR #1511, opened 2022-02-11
Avoid incomplete subgraph when extracting from supergraph

pcmanus:
This is a small followup to #1351: the patch removed object/interface types recursively as it should, but typo-ed union types only calling `remove` instead of `removeRecursive`.
I didn't notice it at the time because while I added a union type test, it wasn't actually exercise the part of the code I though it was.
This PR fixes that problem and adds a proper test (keeping the other union test because while it didn't test what I though it was, it still tests something useful (namely that the `removeRecursive` on object types correctly recurse to unions)).

### DIFF
--- a/internals-js/src/__tests__/extractSubgraphsFromSupergraph.test.ts
+++ b/internals-js/src/__tests__/extractSubgraphsFromSupergraph.test.ts
@@ -430,3 +430,106 @@ test('handles types having only some of their fields removed in a subgraph corre
   expect(c.type('A')).toBeDefined();
   expect(c.type('B')).toBeDefined();
 })
+
+test('handles unions types having no members in a subgraph correctly', () => {
+  /*
+   * The following supergraph has been generated on version-0.x from:
+   *  - ServiceA:
+   *      type Query {
+   *        q: A
+   *      }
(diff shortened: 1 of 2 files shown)


internals-js/src/__tests__/extractSubgraphsFromSupergraph.test.ts
```diff
@@ -430,3 +430,106 @@ test('handles types having only some of their fields removed in a subgraph corre
   expect(c.type('A')).toBeDefined();
   expect(c.type('B')).toBeDefined();
 })
+
+test('handles unions types having no members in a subgraph correctly', () => {
+  /*
+   * The following supergraph has been generated on version-0.x from:
```
abernix, line 436: Would this have a different comment if it were generated by `main`?

abernix (approved):
The justification here makes sense to me and this commit is also on https://github.com/apollographql/federation/pull/1510 which has been reviewed separately.  Landing this will unblock a use-case in the Router and we're going to try to cut a release of that soon-ish (today?).

I will follow-up this with a release myself. (And add a `CHANGELOG` there)

PR #1512, opened 2022-02-11
docs(gateway): Remove v16 caveat

trevor-scheer:
Gateway claims support for graphql v16 so this can be removed now


trevor-scheer:
FYI @StephenBarlow

PR #1528, opened 2022-02-18
Experimentation: using the gateway or router as subgraph

Geal:
This is a test to see if we can convince the query planner to act as a
subgraph. This would allow the gateway or the router to sit in front of
one or multiple subgraphs, with potential benefits:
- have an authorization layer where the main gateway and the subgraph
  gateway can coordinate to avoid exposing the subgraph directly
- make a datasource or subgraph plugin to convert the entity query to a
  normal query if we're in front of a non federated graph
- make trees of routers if we want to expose different subsets of the
  subgraphs, or have edge routers calling into the real router

The idea is to take the supergraph created from the list of
subgraphs, manually create the _Entity type that unifies all the
federated types, and create the _entities query that will use it.

When receiving an _entities query, the query plan somehow understands
that it has must forward it to the subgraph. It can even dispatch
subqueries to multiple subgraphs depending on the requested fields.
I have no ideas why :D but it suggests a clean implementation would not
be too far.

### DIFF
--- a/composition-js/src/compose.ts
+++ b/composition-js/src/compose.ts
@@ -7,9 +7,14 @@ import {
   ServiceDefinition,
   subgraphsFromServiceList,
   ERRORS,
+  UnionType,
+  FieldDefinition,
+  ObjectType,
+  NonNullType,
+  ListType,
 } from "@apollo/federation-internals";
 import { GraphQLError } from "graphql";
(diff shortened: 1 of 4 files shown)


Geal:
of course, a lot of tests fails with the few hacks I put in the code, but the two tests I added pass! :D

PR #1529, opened 2022-02-21
Add Mercurius and Lighthouse to the compatible subgraph matrix

hwillson:
Title says it all!


StephenBarlow (approved):
These PRs both look good to me if the checks and X's are right! 😄

PR #1534, opened 2022-02-23
[docs: mistake/improvement] Concern-based separation

Namyts:
Fairly sure that the "reviews" property in the Product subgraph was supposed to be deleted in this correct example.

### DIFF
--- a/docs/source/index.mdx
+++ b/docs/source/index.mdx
@@ -312,7 +312,6 @@ type User {
 type Product {
   name: String
   price: String
-  reviews: [Review]
 }
 
 type User {


Namyts:
Additionally: would it make sense to move the:
```
# Reviews subgraph
...
author: User
product: Product
```
to the Users and Products subgraphs?    
    
Either way, I think my initial commit makes sense given the other properties are shown to follow similar logic

trevor-scheer:
The change looks correct to me but might warrant some explanation in the paragraph above. Going to tag in @StephenBarlow on this one.

StephenBarlow:
Thanks for reporting and apologies for the delay! This was resolved in another merged PR that I can't find for the life of me 🤦 , but yes this edit is absolutely correct.

PR #1539, opened 2022-02-24
fix: owned types are marked as extension when GraphQL `v16` is used

Gamote:
This PR is meant to fix the issues in which types are wrongly set as extensions when GraphQL `v16` is used.

**Issue**
[printSubgraphSchema.ts:190](https://github.com/apollographql/federation/blob/main/subgraph-js/src/printSubgraphSchema.ts#L190) (@apollo/subgraph) checks if the `extensionASTNodes` field is defined in order to determine if a type should be extended or not. GraphQL `v15` [sets it as `undefined`](https://github.com/graphql/graphql-js/blob/15.x.x/src/type/definition.js#L753) while GraphQL `v16` [always includes](https://github.com/graphql/graphql-js/blob/main/src/type/definition.ts#L770) `extensionASTNodes` in all `GraphQLObjectType` as `[]`.

**Fix**
I order to fix this we had to also check the length of the array, not only just if it's defined.

This will also fix: [#2029](https://github.com/nestjs/graphql/issues/2029).

_Let me know if you need more details. We will provide any information in order to speed up the process of releasing the fix as this is a **major issue** for consumers that are using **Federation**._

### DIFF
--- a/subgraph-js/src/printSubgraphSchema.ts
+++ b/subgraph-js/src/printSubgraphSchema.ts
@@ -178,22 +178,25 @@ function printImplementedInterfaces(
     : '';
 }
 
-function printObject(type: GraphQLObjectType): string {
-  // Apollo change: print `extend` keyword on type extensions.
-  //
-  // The implementation assumes that an owned type will have fields defined
-  // since that is required for a valid schema. Types that are *only*
-  // extensions will not have fields on the astNode since that ast doesn't
-  // exist.
-  //
-  // XXX revist extension checking
-  const isExtension =
-    type.extensionASTNodes && type.astNode && !type.astNode.fields;
+/**
+ * Method that checks if a type is an extension
+ *
+ * The implementation assumes that an owned type will have fields defined
+ * since that is required for a valid schema. Types that are *only* extensions will:
+ *  - have the `extensionASTNodes` defined and with at least 1 field. This is because prior GraphQL v16
+ *    the `extensionASTNodes` property was undefined, where now it comes as an empty {@link Array}
+ *  - not have fields on the `astNode` since that Abstract Syntax Tree doesn't exist;
+ * @param type
+ */
+function checksIfTypeIsExtension(type: GraphQLObjectType) {
+  return type.extensionASTNodes?.length && type.astNode && !type.astNode.fields;
+}
 
+function printObject(type: GraphQLObjectType): string {
   return (
     printDescription(type) +
     // Apollo addition: print `extend` keyword on type extensions
-    (isExtension ? 'extend ' : '') +
+    (checksIfTypeIsExtension(type) ? 'extend ' : '') +
     `type ${type.name}` +
     printImplementedInterfaces(type) +
     // Apollo addition: print @key usages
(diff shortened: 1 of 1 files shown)


subgraph-js/src/printSubgraphSchema.ts
```diff
@@ -178,22 +178,30 @@ function printImplementedInterfaces(
...
+ *
+ * The implementation assumes that an owned type will have fields defined
+ * since that is required for a valid schema. Types that are *only* extensions will:
+ *  - have the `extensionASTNodes` defined and with at least 1 field. This is because prior GraphQL v16
+ *    the `extensionASTNodes` property was undefined, where now it comes as an empty {@link Array}
+ *  - not have fields on the `astNode` since that Abstract Syntax Tree doesn't exist;
+ * @param type
+ * @return boolean
+ */
+function checksIfTypeIsExtension(type: GraphQLObjectType | GraphQLInterfaceType): boolean {
```
trevor-scheer, line 192: Nit: can we just name this function `isExtension`?

subgraph-js/src/printSubgraphSchema.ts
```diff
@@ -178,22 +178,30 @@ function printImplementedInterfaces(
...
+  return (
+      Array.isArray(type.extensionASTNodes) &&
+      !!type.extensionASTNodes?.length &&
```
trevor-scheer, line 195:
```suggestion
      type.extensionASTNodes.length > 0 &&
```
Is this what we actually want?
* I'm not sure the optional chain is necessary since we've asserted on the line before that it's an `Array`
* I think `!!` will yield the same result in all cases, but `length > 0` seems a bit more semantically correct and expresses exactly what we're interested in.

trevor-scheer (changes requested):
I don't doubt that this is fixing the issue but it would be super helpful to see in your commits a new test which was failing and is now passing as a consequence of your changes. At the very least, I'd like to see at least one test addition which is now passing (but that I can copy over to `main` and see fail).

Could you also please add an entry in `subgraph-js/CHANGELOG.md`?

Thanks for bringing this to our attention and tackling the issue! Once this is ready to go and landed, I'll port the fix over to the `version-0.x` branch and get it released.

trevor-scheer:
On second thought, given that this change is targeting a v15/v16 incompatibility, this PR should be opened against the `version-0.x` branch. The `0.x` versions of federation packages claim support for both v15 and v16 versions of graphql (and also run build and test against v15 as well). The `main` line (which represents federation 2) of this repo only claims support for v16.

I was pretty curious to see if there were any issues so I cherry-picked your commits over to a branch on top of `version-0.x` and it all looks good! You're welcome to use it directly or mimic the changes.
https://github.com/apollographql/federation/tree/trevor/fix-extensions-v16

Gamote:
@trevor-scheer Thank you for the insights. I will soon make the requested changes + tests.

Regarding the `version-0.x` branch, if I understand correctly this PR should target it. But what will happen with the `main` line as it has the same issue. Should we have a separated PR that mimics this one but with changes/tests that aim just for `v16` support?

trevor-scheer:
@Gamote yep - you're welcome to open both PRs if you like, or leave it to me to port the change over to `main`. Totally up to you! In any case we'll get this fixed in both places.

vinaybedre:
@Gamote can you please continue with your changes? This is blocking at least some of us, who would like to use `@apollo/subgraph` with `graphql` v16.

kdawgwilk:
@Gamote Do you think you'll have bandwidth to continue working on this?  If not I would be willing to pick this up

Gamote:
Hi @kdawgwilk, if you can please do. I really can't find the time to allocate to this issue. I appreciate 🙏

SirReiva:
Any update?

adamlarsonlee:
FYI - I was getting this issue on an Federation 1 subgraph and this issue went away completely by upgrading my packages and switching to @apollo/subgraph from @apollo/federation. Basically upgrade to Federation 2 and this shouldn't be a problem for you.

trevor-scheer:
I'm going to optimistically close this based on @adamlarsonlee 's comment, but I'm happy to reopen if this is still ongoing in latest versions of subgraph/gateway packages.

developer239:
The original issue is closed: https://github.com/nestjs/graphql/issues/2029

And this issue PR was closed.

How is this still an issue? Am I the only person in the world who is using GraphQL 16?

developer239:
This issue https://github.com/nestjs/graphql/issues/2157 also references this closed PR.

PR #1541, opened 2022-02-24
Node 17 support

pcarrier:
Node 17 declared compatible in all `package.json` and added in CI.


pcmanus:
So the changes clearly looks good. And I don't have personal opinions about this.

But I believe our current and historical policy has been to only "support" LTS versions of nodes (even numbered) and let user pass `--ignore-engines` if they wanted to use something else. Tbc, I really don't mind changing that policy, now or later, but I'll ping @abernix and @trevor-scheer just to be sure they don't have things they want to add here.

I'll simply point out that this means we'll run more CI on every PR after this. I have no clue if we care (and of course, we could also allow 17 in the `package.json` but not run it in CI as an alternative).

abernix (changes requested):
No objection, but please hook it up to the other `requires` in the CircleCI `workflows` section (can't comment on it in the PR because it's out of range).  (Search for "JS: Node 16", look for `oss/dry_run` and `oss/lerna_tarballs` sections.

abernix:
The things said above about historical policy are true.  Odd versions of Node.js used to be much less stable than they are these days.  That's changed to some degree, though.  Our general pattern has been to only be testing 3 even-numbered (LTS) versions at a time.  This puts us over that, but only for a couple months [until May 1](https://github.com/nodejs/Release) when we probably will want to: add 18, drop 17 (because it immediately goes to unsupported) and also drop v12 (which will reach EOL)

trevor-scheer (approved):

clenfest (approved):

PR #1542, opened 2022-02-25
fixing some grammar

angelo-moreira:
I was trying to find out how the Gateway calls `__resolveReference` when I've noticed a few minor grammar tweaks.


angelo-moreira:
fixed conflicts

pcmanus (approved):
Thanks!

PR #1545, opened 2022-02-26
Docs preview

StephenBarlow:


docs/source/entities-advanced.mdx
```diff
@@ -0,0 +1,535 @@
...
+      id: ID!
+      amount: Int!
+    }
+    ```
+
+    </CodeColumns>
+
+2. Publish the updated Billing subgraph schema to Apollo.
+
+    * Note that you're publishing a composition error, because multiple subgraphs define the same fields of `Bill` without marking them [`@shareable`](#using-shareable). This creates a failed [launch](https://www.apollographql.com/docs/studio/launches/) in Apollo Studio. That's okay! Apollo Uplink continues to serve the most recent _valid_ supergraph schema to your gateway.
```
pcmanus, line 195: I'm really bothered by this paragraph. We really should move away from suggesting _any_ course of action that result in a composition error and failed launches in Studio. At least for GA, we should be able to have `@override(from:)` ready so it can be used here (but, @clenfest, this does remind me that we should consider allowing that on types as a shortcut of having it on all fields, same as we allow for `@shareable`, and this exactly for that use case).

Now, we may or may not have `@override(from:)` committed for preview so there is the question of what we put in the doc in the meantime. Personally, I'd rather we recommend marking the entity `@shareable` first in the `Payments` subgraph, add it at `@shareable` in `Billing` and remove the `@shareable` only later. *And* that we explicitly document that this is a temporary solution that will be cleaned up for GA. But if we really want to keep the "go and publish a composition error" for preview, lets *please* at least remove the "That's okay!" sentence and replace it with a sentence that acknowledge that it's a ugly hack that is going to be fixed.

docs/source/entities-advanced.mdx
```diff
@@ -0,0 +1,535 @@
...
+  resolve("Can my subgraph always resolve this field,<br/>or only from certain query paths?");
+  resolve--Always-->shareable("@shareable");
+  resolve--"Only certain query paths"-->provides("@provides");
+```
+
+If you aren't sure whether your subgraph can always resolve a field, see [Using `@provides`](#using-provides) for an example of a subgraph that _can't_.
+
+> ⚠️ **Important:** If multiple subgraphs can resolve a field, **make sure each subgraph's resolver for that field behaves identically.** Otherwise, queries might return inconsistent results depending on which subgraph resolves the field.
+
+### Using `@shareable`
```
pcmanus, line 467: I'm not sure this is right place for this section (and this might need a bit of rewording/reworking) because `@shareable` is not specific to entities. In fact, most user will probably use `@shareable` first and mostly to have "shared" value types. So I don't think shareable should be in the "advanced entities" section.

docs/source/entities-advanced.mdx
```diff
@@ -0,0 +1,535 @@
...
+type Product @key(fields: "id") {
+  id: ID!
+  name: String! @shareable
+  inStock: Boolean!
+}
+```
+
+</CodeColumns>
+
+In this example, both the Products and Inventory subgraphs can resolve `Product.name`. This increases the likelihood that a client's query can be resolved by fewer total subgraphs.
```
pcmanus, line 493: I tripped a bit on this sentence. I'd probably rephrase saying something along the lines of "This is usually done as an optimisation: as multiple subgraphs are able to resolve the field, queries involving that field may be resolved by querying fewer total subgraphs". 

But I would also warn about the flip-side of the tradeoff involved. Namely, I'd first insist once against that all subgraphs resolving the field must resolve it the same way. It's a pretty important point imo and I'd suggest expanding on it; maybe with a full subsection, giving an example of what to do and what _not_ to do to clarify what we mean by "resolve the same way"). And then essentially say that not only those resolvers must work the same way initially, but they should be kept in sync if evolved, and that this require care and attention, and so that is why we advise against premature optimisation (but feel free to avoid the term "premature optimisation").

Lastly, as mentioned previously, `@shareable` is not only for entities: but all that I mention in that section does _only_ really apply to `@shareable` when applies to fields of entities. When `@shareable` is used with value types, it's not really an optimisation, it's just the only way in fed 2 to enable a shared value type and that's pretty common.

docs/source/entities-advanced.mdx
```diff
@@ -0,0 +1,535 @@
...
+  inStock: Boolean!
+}
+```
+
+Here we're using two directives in combination: `@provides` and `@external`.
+
+* The `@provides` directive tells the gateway, "This subgraph can resolve the `name` of any `Product` object returned by `InStockCount.product`."
+* The `@external` directive tells the gateway, "This subgraph _can't_ resolve the `name` of a `Product` object, _except_ wherever indicated by `@provides`."
+
+> Note that to include a field in a `@provides` directive, that field must be marked as `@shareable` or `@external` in every subgraph that defines it.
```
pcmanus, line 535: Maybe worth clarifying that if a field is in a `@provides`, then it must be `@external` in the subgraph having the `@provides`, but it must be `@shareable` in an subgraphs that resolve it?  That is, you cannot have a field that is both in a `@provides` and `@shareable` in the same subgraph and wondering if that sentence could lead one to think otherwise.

docs/source/entities.mdx
```diff
@@ -0,0 +1,321 @@
...
+}
+```
+
+</CodeColumns>
+
+Types _besides_ object types (unions, interfaces, etc.) _cannot_ be entities.
+
+## Defining an entity
+
+In a subgraph schema, you can designate any existing object type as an entity by adding the `@key` directive to its definition, like so:
```
pcmanus, line 40: I worry a bit about that kind of sentence because that notion of a type being designated as an entity does not really exists in the code, and there is use cases (see this [this discussion](https://apollographql.quip.com/bQJMAqFttef4#cETADAQUhtS) for instance), for an entity type to have a `@key` in some subgraphs but not others.

Tbc, I don't have a super simple fix to offer. Currently, "entities" is more of a usage pattern than a real internal concept, and that makes it hard to have a simple characterisation of entities that is both simple, precise and always correct. This is why I initially advocate for adding an `@entity` annotation to "force" a clear definition in the model, but we went another way (and to be fair, that proposal was not perfect either).

Anyway, I just though of something that could hopefully clarify our story here but I need to think this through a bit more.  So this comment is probably not very actionable right now, but I still wanted to flag that there is a bit of vagueness in our concepts right now, and that's certainly unfortunate.

docs/source/entities.mdx
```diff
@@ -0,0 +1,321 @@
...
+  name: String!
+  price: Int
+}
+```
+
+The `@key` directive defines the entity's **primary key**, which consists of one or more of the type's `fields`. In this example, the `Product` entity's primary key is its `id` field.
+
+**Every instance of an entity must be uniquely identifiable by its `@key` fields.** This is what enables your gateway to associate data from _different_ subgraphs with the _same_ entity instance.
+
+> An entity's `@key` cannot include fields that return a union or interface.
```
pcmanus, line 55: It also (currently) cannot include fields having arguments, which is worth specifying (this might change later, and the same is true for union and interfaces, we might allow it later, but no guarantees as of when if ever as of yet).

docs/source/entities.mdx
```diff
@@ -0,0 +1,321 @@
...
+```graphql{2}
+# Products subgraph
+type Product @key(fields: "id") {
+  id: ID!
+  name: String!
+  price: Int
+}
+```
+
+The `@key` directive defines the entity's **primary key**, which consists of one or more of the type's `fields`. In this example, the `Product` entity's primary key is its `id` field.
```
pcmanus, line 51: I'd be in favour, at this point, to mention that every time you write a `@key` directive, then you will have to write a corresponding "reference resolver" and link to the proper section. In many ways, all a `@key` is doing as far as the implementation is concerned is declaring that a reference resolver for that key will exist at runtime and so the query planner can use it, so I think creating that strong link between a key and it's resolver early is useful and a good idea.

docs/source/entities.mdx
```diff
@@ -0,0 +1,321 @@
...
+* A resolver in Reviews to generate [representations](#entity-representations) of `Product` entities
+* A [reference resolver](#reference-resolvers) in Products to return full `Product` objects from representations
+
+### Entity representations
+
+In our example, the Reviews subgraph needs to define a resolver for its stub version of the `Product` entity. The Reviews subgraph doesn't know much about `Product`s, but fortunately, it doesn't need to. All it needs to do is return data for the fields it _does_ know about, like so:
+
+```js{3-10}:title=resolvers.js
+// Reviews subgraph
+const resolvers = {
```
pcmanus, line 136: Something's wrong with that example: it's resolver for a `product` field in `Review` having one argument, but `Review.product` does not have arguments, and it references `review.productID` which isn't a field that has been defined anywhere.

What you need in the Reviews subgraph is an implementation for the `Query.latestReviews` field that has been hinted at earlier.

docs/source/entities.mdx
```diff
@@ -0,0 +1,321 @@
...
+}
+```
+
+Now we have a problem: this query starts its execution in the Reviews subgraph (where `latestReviews` is defined), but that subgraph _doesn't_ know that `Product` entities have a `price` field! Remember, the Reviews subgraph only knows about the `id` field of `Product`.
+
+Because of this, the gateway needs to fetch `price` from the Products subgraph instead. But how does the gateway know _which_ products it needs to fetch the prices for?
+
+To solve this, we add a resolver to _each_ subgraph:
+
+* A resolver in Reviews to generate [representations](#entity-representations) of `Product` entities
```
pcmanus, line 127: I'm not really sure about that one, and I'm particularly uneasy with the presentation of this alongside the "reference resolver".

I'm not sure about the concept of "a resolver to generate entity representations" in general because it's usually not 1 resolver, it's not a particularly special resolver, and I think the concept of an "entity representation" is not that clearly defined in most cases. 

What I mean here is that "entity representations" are returned by the "normal" resolvers of your subgraph. In the example above, the resolver that would return `Product` "representation" is the one of `latestReviews`, but that resolver is a normal, non-federation-specific, resolver. And What it returns is also not very federation-specific in the sense that it should return `Review` objects, and those should have a `product` field that is itself an object containing an `id`, but all that is what the graphQL schema declare. By which I mean, if `Reviews` was not a federated schema at all, that `latestReviews` would be _exactly_ the same.

Also, there is usually more than on resolver in a subgraph returning entities representations. Typically, if you had a 2nd operation, say `Query.bestReviews`, then the resolver of that would also return some `Product` "representations". But the phrasing here suggest a single resolver for all `Product` entities representations.

My personal suggestion here would be to say that there is actually nothing special to do in Reviews that is federation specific. You just need to implement resolvers as you would if Review wasn't federated in that case. And then I would say that Product does need a special resolver for its `@key`.

docs/source/entities.mdx
```diff
@@ -0,0 +1,321 @@
...
+      };
+    }
+  },
+  // ...
+}
+```
+
+This resolver's return value is a **representation** of a `Product` entity. A representation always consists of:
+
+* A `__typename` field
```
pcmanus, line 151: With the caveat that those example describe Apollo Server and other subgraph server implementation may behave differently, I'll note that the `__typename` field rarely have to be added by the user since Apollo Server usually adds it (I'm no expert with Apollo Server and I assume there is cases with Interfaces and Unions where it cannot always infer the `__typename`, but it definitively does it in most other cases).

docs/source/entities.mdx
```diff
@@ -0,0 +1,321 @@
...
+```graphql{3}
+# Reviews subgraph
+type Review {
+  product: Product!
+  score: Int!
+}
+
+# This is a required "stub" of the Product entity (see below)
+type Product {
+  id: ID!
```
pcmanus, line 83: I'll note that as of preview, _if_ this subgraph is a fed2 subgraph (and see below, but I think we should try to have all of our example be of fed2 schema), the `id` would have to be marked `@shareable` here.

Which points to another consideration: how do we make it clear in the doc whether an example if of a fed2 schema or not? Should we systematically add the proper `@link` to clarify it? I think we might want to.

docs/source/entities.mdx
```diff
@@ -0,0 +1,321 @@
...
+      return fetchProductByID(reference.id);
+    }
+  },
+  // ...
+}
+```
+
+> In the example above, `fetchProductByID` is a hypothetical function that fetches a `Product`'s full details from a data store based on its `id`.
+
+A reference resolver (always called `__resolveReference`) provides the gateway direct access to a particular entity's fields, without needing to use a custom query to reach that entity. To use a reference resolver, the gateway must provide a valid [entity representation](#entity-representations), which is why we created the resolver in the Reviews subgraph first!
```
pcmanus, line 204: "if you use typescript and @apollo/subgraphs" :). Don't know about other subgraph server implementations, but I don't think it's part of our spec that it has to be called that way in every subgraph server. So maybe worth calling this out and saying that if you use another subgraph server, you should check their document to figure out how to declare reference resolvers.

docs/source/entities.mdx
```diff
@@ -0,0 +1,321 @@
...
+
+> To learn more about `__resolveReference`, see the [API docs](/api/apollo-subgraph/#__resolvereference).
+
+After fetching the `price` field from Products via a reference resolver, the gateway can use the shared `@key` field `id` to intelligently merge the data it obtained from its two queries into a single result and return that result to the querying client.
+
+### Default reference resolver
+
+TODO
+
+## Extending entities
```
pcmanus, line 214: I'll note that the notion of "extending" an entity hints at a notion of some subgraph "defining" an entity while other extends it, but things are truly "symmetric" in fed 2. It may not be a big deal, and I don't feel this section is particularly confusing, but wondering if we could improve further.

Fwiw, the way I've been talking about this is that each subgraph "contributes" fields to a given entity. Don't know if it's a better naming, but it feels a bit more symmetric than "extends".

docs/source/entities-advanced.mdx
StephenBarlow: Totally, this will not end up the canonical location for learning about `@shareable` (it will be covered in the "value types" article), this is here to accompany `@provides`. We'll have the value type edits up soon and will let you know!

docs/source/entities.mdx
StephenBarlow: Sounds good, I incorrectly assumed `id` would be auto-`@shareable` because it was a `@key` field _somewhere_ but I see it'd need to be a key field _here_.

And yes, we definitely need to add the `@link` snippet in lots of places, I'm thinking of doing a separate pass for that

docs/source/entities.mdx
StephenBarlow: I like "contributing"! I'll make the change.

docs/source/entities.mdx
StephenBarlow: I believe this is correct but requires some extra explanation. The parameter passed to this resolver isn't a GraphQL argument, but rather the `Review` object that was resolved just prior to resolving `Review.product`. That returned `Review` object can include a field named `productID`, pulled from the same data store that populated any actual supported fields of `Review`

StephenBarlow: (Will indicate as much in the doc!)

docs/source/entities.mdx
StephenBarlow: Good point, this has been lingering here for a while, I've cleaned it up

docs/source/entities-advanced.mdx
StephenBarlow: 👍 Created a dedicated section describing these concerns and link to it from the top of each section about `@shareable` and `@provides`. Can hopefully expand it further prior to release

docs/source/entities.mdx
StephenBarlow: I've updated this section significantly to only focus on reference resolvers, and note that the already-present resolvers for the other subgraph do the trick of returning representations 👍

docs/source/entities.mdx
StephenBarlow: This section now describes the process of creating an entity as a two-step process, first adding a `@key`, then defining a reference resolver. So the connection is much stronger!

docs/source/entities.mdx
StephenBarlow: Ok, doc has been revamped pretty substantially and this example is no longer present.

StephenBarlow:
> Which is confusing, but I feel like it is an important distinction to make so people don't do this in their subgraphs and then wonder why it's working when it "shouldn't be".

Good call! Link added

docs/source/federation-2/new-in-federation-2.mdx
```diff
@@ -52,23 +53,29 @@ type Book {
 
 </CodeColumns>
 
-In Federation 2, value types can _differ_ in certain ways:
+In Federation 2, this "identical definition" constraint is removed. Types and their fields can be shared across subgraphs, even if _certain_ details differ between definitions. With this change, the term "value type" is not used in Federation 2.
```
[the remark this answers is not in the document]
StephenBarlow: Reintroduced

docs/source/entities-advanced.mdx
StephenBarlow: Updated to use `@shareable`!

docs/gatsby-config.js
```diff
@@ -16,11 +16,6 @@ module.exports = {
         subtitle: 'Federation',
         description: 'A guide to using Apollo Federation',
         githubRepo: 'apollographql/federation',
-        defaultVersion: '1',
```
StephenBarlow, line 19: This change needs to be reverted prior to release

docs/source/entities-advanced.mdx
```diff
@@ -0,0 +1,604 @@
...
+    ```graphql
+    # Payments subgraph
+    type Payment {
+      # ...
+    }
+    ```
+
+    ```graphql
+    # Billing subgraph
+    type Bill @key(fields: "id") {
```
rkoron007, line 242:
```suggestion
    type Bill @key(fields: "id") @shareable {
```

@StephenBarlow — I think this `Bill` entity in the Billing subgraph should still have the `@shareable` directive at this point? (you tell them to remove `@shareable` in step 10 below).

StephenBarlow: Great catch thank you!!

docs/source/entities.mdx
```diff
@@ -0,0 +1,316 @@
...
+# Reviews subgraph
+type Review {
+  product: Product!
+  score: Int!
+}
+```
+
+This is possible! However, this subgraph schema is currently invalid because it doesn't define the `Product` entity.
+
+To fix this, we can add a **stub** of the `Product` entity to the Reviews schema, like so:
```
pcmanus, line 164: It's a bit late notice, but we actually changed our mind on how we'd like to handle that case. For that, we added a `resolvable` argument to keys, which allow to "mark" key fields but opt-out of needing to be able to resolve references to that key. The idea is that we'll want to change the `Product` definition in the example below into:
```graphql
type Product @key(fields: "id", resolvable: false) {
  id: ID!
}
```
The pros of that is that now an entity can always have a `@key` in all subgraphs, so people's mental model can just "entity === has a @key" in all cases.

To be clear, the example here still work in practice (though it will raise a hint that `Product` is marked as an entity in some subgraphs and not others, which the `@key(..., resolvable: false)` solves as well) but I think we want to push the new pattern described above.

pcmanus: Oh, and to clarify, that `resolvable` argument *is* part of preview.

docs/source/quickstart/setup.mdx
```diff
@@ -125,7 +125,7 @@ type ReviewSummary {
   averageRating: Float
 }
 
-extend type Product @key(fields: "id") {
+type Product @key(fields: "id") {
   id: ID! @external
```
pcmanus, line 129: The `@external` on `id` should be removed.

docs/source/quickstart/local-subgraphs.mdx
```diff
@@ -47,7 +47,7 @@ And that's it! You can now perform all of the same subgraph setup on your _own_
 
 In a federated graph, each subgraph's schema can define or reference types and fields that are present in _another_ subgraph's schema.
 
-Consider the `Product` type in our composed supergraph schema from [Part 3](./quickstart/local-composition/#2-perform-composition). This type includes the following fields:
+Consider the `Product` type in our composed supergraph schema from [Part 3](./local-composition/#2-perform-composition). This type includes the following fields:
 
 ```graphql
 type Product
```
pcmanus, line 53: I believe the `extension: true` wouldn't be there on the `@join__type(graph: REVIEWS, key: "id", extension: true)` since the examples don't use `extend type` anymore.

pcmanus (commented):
I noticed that the examples of the "quickstart" section don't use fed2 subgraphs (they don't have a `@link`). Should we add it? Feels weird to have the very first fed 2 quickstart to not use fed2 schema.

docs/source/quickstart/setup.mdx
StephenBarlow: Good point on both of these, yes, the quickstart is still using v1 schemas because it's pulling from the live deployed Acephei app that is _also_ still on Fed 1. We will need to update here, I can address next week

cpeacock (approved):

StephenBarlow:
Merged in #1594 because tests didn't run on this PR, closing this one

Issue #1544, opened 2022-02-25
Provide naming of subgraph GraphQL operations instead of sending them as anonymous operations


theJC:
Federated queries when sent to the subgraphs are anonymous and not named, therefore most of the available observability tooling out of the box is not very useful in differentiating the various requests when monitoring at the subgraph level.

Please add a mechanism that names and/or provide a function to allow federation customers to provide a naming convention on the queries being sent to their subgraphs to allow observability and telemetry tooling based on operationName to work out of the box.

A few initial thoughts/ideas on ways we would like to have these operations named with the original operation name, append/suffixing it with either or both:
- the subgraph name
- a consistent/stable identifier of the step within the query plan that the call corresponds to

ie:
Query to the federated gateway:
```
query FindMyPhone($input: FindMyPhoneInput!) {
  findMyPhone(input: $input) {....
```

Currently actual behavior results in a subgraph query:  (note that it is anonymous/not named)
```
query($FindMyPhoneInput!){...
```

Wanted behavior, subgraph query is named/not anonymous:
```
query FindMyPhone_subquery_step1a($FindMyPhoneInput!){...
```

abernix:
Thanks for opening this.

This should be fixed by #1550.  This will go out in a new release of the `2.x` package next week.

Issue #195, opened 2020-09-30
Including operation name in the query execution


Glen-Moonpig:
We are using the operationName field in our client->gateway requests, [as encouraged by the GraphQL spec](https://graphql.org/learn/queries/#operation-name). When Apollo Gateway builds and executes a query plan it does not include the operation name in generated queries for federated services. This makes sense because the queries that make up the execution plan are not necessarily the operation, but a smaller part of the operation for resolving certain types/fields.
**But**, we would like the federated services to know the name of the overall operation that it is receiving requests for, to facilitate logging and monitoring purposes.

I have tried modifying my RemoteGraphQLDataSource to set the operationName field in the request, but because the query in the request does not have a matching operation name, a GraphQL error is returned:

```json
{
  "errors": [
    {
      "message": "400: Bad Request",
      "extensions": {
        "code": "INTERNAL_SERVER_ERROR",
        "response": {
          "url": "https://api.something.com/",
          "status": 400,
          "statusText": "Bad Request",
          "body": {
            "errors": [
              {
                "message": "Unknown operation named \"TestOp\".",
                "extensions": {
                  "code": "INTERNAL_SERVER_ERROR",
                  "exception": {
                    "stacktrace": [
                      "GraphQLError: Unknown operation named \"TestOp\".",
                      "    at buildExecutionContext (/var/task/bundle.js:7959:15)",
                      "    at executeImpl (/var/task/bundle.js:7875:20)",
                      "    at Module.execute (/var/task/bundle.js:7850:35)",
                      "    at /var/task/bundle.js:50331:48",
                      "    at Generator.next (<anonymous>)",
                      "    at /var/task/bundle.js:50090:71",
                      "    at new Promise (<anonymous>)",
                      "    at __awaiter (/var/task/bundle.js:50086:12)",
                      "    at execute (/var/task/bundle.js:50310:20)",
                      "    at Object.<anonymous> (/var/task/bundle.js:50258:42)"
                    ]
                  }
                }
              }
            ]
          }
        },
        "exception": {
          "stacktrace": [
            "Error: 400: Bad Request",
            "    at ...
(/var/task/node_modules/@apollo/gateway/dist/datasources/RemoteGraphQLDataSource.js:58:26)",
            "    at async sendOperation (/var/task/node_modules/@apollo/gateway/dist/executeQueryPlan.js:159:26)",
            "    at async executeFetch (/var/task/node_modules/@apollo/gateway/dist/executeQueryPlan.js:111:41)",
[...]

Issue #146, opened 2020-09-10
RFC: Should `federation-jvm` move into this repository?


abernix:
Follows-up: #134 

Currently https://github.com/apollographql/federation-jvm lives in its own repository.  It may very well make sense to leave it that way since there is currently very different ownership and contributorship to that module, but I think with the efforts to consolidate our Apollo-related federation and gateway efforts into this repository, it might be worth considering.

No action item right now, just a conversation.  (Intentionally leaving `[js-repo-migration] ` out of the subject since it's not JS related.)

abernix:
Closing as this isn't likely to happen anymore.

PR #1550, opened 2022-03-01
feat: generate operation names for subgraph fetches

lennyburdette:
Addresses #1544, #195, #146

### DIFF
--- a/gateway-js/src/__tests__/build-query-plan.feature
+++ b/gateway-js/src/__tests__/build-query-plan.feature
@@ -1009,7 +1009,8 @@ Scenario: supports basic, single-service mutation
         "password"
       ],
       "operationKind": "mutation",
-      "operation": "mutation($username:String!$password:String!){login(username:$username password:$password){id}}"
+      "operation": "mutation Login_accounts_0($username:String!$password:String!){login(username:$username password:$password){id}}",
+      "operationName": "Login_accounts_0"
     }
   }
   """
@@ -1043,7 +1044,8 @@ Scenario: supports mutations with a cross-service request
             "password"
           ],
           "operationKind": "mutation",
-          "operation": "mutation($username:String!$password:String!){login(username:$username password:$password){__typename id}}"
+          "operation": "mutation Login_accounts_0($username:String!$password:String!){login(username:$username password:$password){__typename id}}",
+          "operationName": "Login_accounts_0"
         },
         {
           "kind": "Flatten",
@@ -1071,7 +1073,8 @@ Scenario: supports mutations with a cross-service request
             ],
             "variableUsages": [],
             "operationKind": "query",
-            "operation": "query($representations:[_Any!]!){_entities(representations:$representations){...on User{reviews{product{__typename ...on Book{__typename isbn}...on Furniture{upc}}}}}}"
+            "operation": "query Login_reviews_1($representations:[_Any!]!){_entities(representations:$representations){...on User{reviews{product{__typename ...on Book{__typename isbn}...on Furniture{upc}}}}}}",
+            "operationName": "Login_reviews_1"
           }
         },
         {
@@ -1103,7 +1106,8 @@ Scenario: supports mutations with a cross-service request
             ],
             "variableUsages": [],
             "operationKind": "query",
-            "operation": "query($representations:[_Any!]!){_entities(representations:$representations){...on Book{upc}}}"
+            "operation": "query Login_product_2($representations:[_Any!]!){_entities(representations:$representations){...on Book{upc}}}",
+            "operationName": "Login_product_2"
           }
         }
       ]
@@ -1138,7 +1142,8 @@ Scenario: returning across service boundaries
             "body"
           ],
           "operationKind": "mutation",
-          "operation": "mutation($upc:String!$body:String!){reviewProduct(input:{upc:$upc body:$body}){__typename ...on Furniture{__typename upc}}}"
+          "operation": "mutation Review_reviews_0($upc:String!$body:String!){reviewProduct(input:{upc:$upc body:$body}){__typename ...on Furniture{__typename upc}}}",
+          "operationName": "Review_reviews_0"
         },
         {
           "kind": "Flatten",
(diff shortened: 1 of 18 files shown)


lennyburdette:
i didn't change any the feature tests but i'm happy to do so if my approach is acceptable

lennyburdette:
Excellent points! I talked to the customer and he thought that option 2 sounds right. I updated the code to drop operation names if the original request does not have one.

Should I start updating all the tests? Or is there anything else worth discussing/addressing first?

pcmanus:
> Should I start updating all the tests? Or is there anything else worth discussing/addressing first?

I'll do a final review pass on the final version, but I don't have anything else to bring. Seems like a great improvement with little downsides (especially if we use option 2 above which gives a de-factor option to "disable" the operation naming if for some reason someone wants that (can't really think of a reason but ...)). Re-tests, hopefully you don't have to change  that much of them if you switch to "option 2" above since I believe most tests use unnamed operations (and so won't be impacted anymore).

lennyburdette:
I can't add reviewers to the PR but I think this is good to go.

pcmanus (approved):
Lgtm, thanks!

gateway-js/src/__tests__/build-query-plan.feature
```diff
@@ -1213,7 +1219,8 @@ Scenario: supports multiple root mutations
             "password"
           ],
           "operationKind": "mutation",
-          "operation": "mutation($username:String!$password:String!){login(username:$username password:$password){__typename id}}"
+          "operation": "mutation LoginAndReview_accounts_0($username:String!$password:String!){login(username:$username password:$password){__typename id}}",
+          "operationName": "LoginAndReview_accounts_0"
```
abernix, line 1223: Very thankful you chose to add `operationName` in here!

abernix:
I closed the related issues above referenced in the original post which weren't auto-closed.  I also found https://github.com/apollographql/federation/issues/395, but I wasn't sure what the third reference was — #146  doesn't look relevant.

@pcmanus Are you keeping a running CHANGELOG for things like this?

gateway-js/src/__tests__/build-query-plan.feature
lennyburdette: 👍🏻 I had to in order to provide the operationName query param on the request, because that's probably what observability tools look at (instead of parsing the operation doc).

abernix: For context as to why i'm glad: it's only recently that we started adding `operationKind` in there, which enabled the router to make Good decisions without needing to — in the same spirit as you're describing — re-parse the `operation`.

gateway-js/src/__tests__/build-query-plan.feature
```diff
@@ -1405,7 +1418,8 @@ Scenario: multiple root mutations with correct service order
             ],
             "variableUsages": [],
             "operationKind": "query",
-            "operation": "query($representations:[_Any!]!){_entities(representations:$representations){...on User{reviews{product{__typename ...on Book{__typename isbn}...on Furniture{upc}}}}}}"
+            "operation": "query LoginAndReview_reviews_2($representations:[_Any!]!){_entities(representations:$representations){...on User{reviews{product{__typename ...on Book{__typename isbn}...on Furniture{upc}}}}}}",
+            "operationName": "LoginAndReview_reviews_2"
```
abernix, line 1422: @pcmanus @lennyburdette Picking a random place to leave this post-merge thought — would it make sense to use double-underscores as the delimiter to create a more predictable pattern for parsing and to avoid name-conflicts across operation + subgraph names?  Mostly just because it seems not completely unlikely that both an operation and a subgraph might use `_` on its own and, while not forbidden, much less likely that someone would use `__`.   We could even forbid `__` as a valid graph name (and probably should or may already have a enforced character set somewhere?).

Just trying to avoid the case of two things end up being exactly equal.

It was also very intentional that I'd suggested prefixing with `__` just to indicate the generatedness and avoid clashes, just in case people use these operations downstream in meaningful ways.

Also happy to ticket this for follow-up.   Or not, depending on thoughts.

lennyburdette: no strong feelings either way. i went with fewer characters but four extra underscores (two in the operationName query param and two in the operation document) isn't a huge deal.

smyrick: I also like the idea of the double underscore separator, and if we are not already we should enforce that subgraphs have to start and end with a number or letter

pcmanus:
Fyi, just noticed that we include the subgraph name here and that we enforce very little kind of limitations on subgraph names so this patch can currently very easily create invalid query names and break things. I'll write a followup patch for it. I might sneak in the double underscore change discussed above as well since it seems to be overall favored.

trevor-scheer:
Should this be ported to the `version-0.x` branch?

lennyburdette:
> this patch can currently very easily create invalid query names and break things

if the original operation name name is bad, wouldn't things break in the validation phase before getting to the query planner?

> Should this be ported to the version-0.x branch?

both customers i've talked to about this are fine with it being a 2.0-only improvement, but i'm sure others would be happy if you did!

pcmanus:
> if the original operation name name is bad, wouldn't things break in the validation phase

Yes, but it's not the original operation name the issue, it's the subgraph name, which is also included (as 2nd element) and is not otherwise validated. Anyway, #1568 has been pushed to fix it and has been merged (sorry for not waiting too long on feedback on that one but we want to do a new release very soon and this was a serious break so wanted to get it fixed asap).

chrskrchr:
> > Should this be ported to the version-0.x branch?
> 
> both customers i've talked to about this are fine with it being a 2.0-only improvement, but i'm sure others would be happy if you did!

Any word on if/when this change will be backported to `0.x`?  Would love to rip out our custom operation propagation code.

dakshas:
Hi folks! Can someone explain why the counter appended to the end of the operation name is necessary ? 
The counter makes tracking metrics slightly more complicated as now we need to use a wildcard to account for the changing counter values.

pcmanus:
> why the counter appended to the operation name is necessary ?

Query plans can easily have multiple different fetches to the same subgraphs. Without the counter, each of those fetches, which can very different queries (say, they may fetch completely different entity types), would have the same name. Which might be acceptable in some use cases, but isn't in many others. Typically, if you use the operation name as an identifier for metrics, not having a counter would lump together the metrics of potentially fairly unrelated subgraph queries.

dakshas:
@pcmanus thank you for the prompt reply!

PR #1554, opened 2022-03-02
add core defs to in buildSubgraphSchema

queerviolet:
always emit a core schema from `buildSubgraphSchema` for fed v2 schemas.

fed v1 schemas will not have core markings added, but will have any missing definitions of federation directives filled in

### DIFF
--- a/subgraph-js/src/__tests__/printSubgraphSchema.test.ts
+++ b/subgraph-js/src/__tests__/printSubgraphSchema.test.ts
@@ -1,0 +1,201 @@
+import { fixtures } from 'apollo-federation-integration-testsuite';
+import { buildSubgraphSchema } from '../buildSubgraphSchema';
+import { printSubgraphSchema } from '../printSubgraphSchema';
+import gql from 'graphql-tag';
+
+describe('printSubgraphSchema', () => {
+  it('prints a subgraph correctly', () => {
+    const schema = buildSubgraphSchema(fixtures[0].typeDefs);
+    expect(printSubgraphSchema(schema)).toMatchInlineSnapshot(`
+      "schema {
+        query: RootQuery
+        mutation: Mutation
+      }
+
+      extend schema @link(url: \\"https://specs.apollo.dev/link/v1.0\\") @link(url: \\"https://specs.apollo.dev/federation/v2.0\\", import: [\\"@key\\", \\"@requires\\", \\"@provides\\", \\"@external\\", \\"@tag\\", \\"@extends\\", \\"@shareable\\", \\"@inaccessible\\", \\"@override\\"])
+
+      directive @stream on FIELD
+
+      directive @transform(from: String!) on FIELD
+
+      directive @cacheControl(maxAge: Int, scope: CacheControlScope, inheritMaxAge: Boolean) on FIELD_DEFINITION | OBJECT | INTERFACE | UNION
+
+      enum CacheControlScope @tag(name: \\"from-reviews\\") {
+        PUBLIC @tag(name: \\"from-reviews\\")
+        PRIVATE
+      }
+
+      scalar JSON @specifiedBy(url: \\"https://json-spec.dev\\") @tag(name: \\"from-reviews\\")
+
+      type RootQuery {
+        _entities(representations: [_Any!]!): [_Entity]!
+        _service: _Service!
+        user(id: ID!): User
+        me: User
+      }
+
+      type PasswordAccount @key(fields: \\"email\\") {
+        email: String!
+      }
+
+      type SMSAccount @key(fields: \\"number\\") {
+        number: String
+      }
+
+      union AccountType @tag(name: \\"from-accounts\\") = PasswordAccount | SMSAccount
+
+      type UserMetadata {
+        name: String
+        address: String
+        description: String
+      }
+
+      type User @key(fields: \\"id\\") @key(fields: \\"username name { first last }\\") @tag(name: \\"from-accounts\\") {
+        id: ID! @tag(name: \\"accounts\\")
+        name: Name
+        username: String @shareable
+        birthDate(locale: String @tag(name: \\"admin\\")): String @tag(name: \\"admin\\") @tag(name: \\"dev\\")
+        account: AccountType
+        metadata: [UserMetadata]
+        ssn: String
+      }
+
+      type Name {
+        first: String
+        last: String
+      }
+
+      type Mutation {
+        login(username: String!, password: String!, userId: String @deprecated(reason: \\"Use username instead\\")): User
+      }
+
+      type Library @key(fields: \\"id\\") {
+        id: ID!
+        name: String @external
+        userAccount(id: ID! = 1): User @requires(fields: \\"name\\")
+        description: String @override(from: \\"books\\")
+      }
+
+      scalar federation__FieldSet
+      "
+    `);
+  });
+
+  it('prints a scalar without a directive correctly', () => {
+    const schema = gql`
+      scalar JSON
+    `;
+    const subgraphSchema = buildSubgraphSchema(schema);
+
+    expect(printSubgraphSchema(subgraphSchema)).toMatchInlineSnapshot(`
+      "scalar JSON
+
+      type Query {
+        _service: _Service!
+      }
+      "
+    `);
+  });
+
+  it('prints reviews subgraph correctly', () => {
+    const schema = buildSubgraphSchema(fixtures[5].typeDefs);
+    expect(printSubgraphSchema(schema)).toMatchInlineSnapshot(`
+      "extend schema @link(url: \\"https://specs.apollo.dev/link/v1.0\\") @link(url: \\"https://specs.apollo.dev/federation/v2.0\\", import: [\\"@key\\", \\"@requires\\", \\"@provides\\", \\"@external\\", \\"@tag\\", \\"@extends\\", \\"@shareable\\", \\"@inaccessible\\", \\"@override\\"])
+
+      directive @stream on FIELD
+
+      directive @transform(from: String!) on FIELD
+
+      type Query {
+        _entities(representations: [_Any!]!): [_Entity]!
+        _service: _Service!
+        topReviews(first: Int = 5): [Review]
+      }
+
+      type Review @key(fields: \\"id\\") {
+        id: ID!
+        body(format: Boolean = false): String
+        author: User @provides(fields: \\"username\\")
+        product: Product
+        metadata: [MetadataOrError]
+      }
+
+      input UpdateReviewInput @tag(name: \\"from-reviews\\") {
+        id: ID!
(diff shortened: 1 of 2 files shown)


subgraph-js/src/buildSubgraphSchema.ts
```diff
@@ -86,7 +66,6 @@ export function buildSubgraphSchema(
     modules,
     new GraphQLSchema({
       query: undefined,
-      directives: [...specifiedDirectives, ...missingFederationDirectives(modules)],
     }),
   );
```
martinbonnin, line 71: I believe the comment below this line (where I can't comment...) needs to be changed since now the directives definitions are included in `_service.sdl`?

subgraph-js/src/federation-atlas.ts
```diff
@@ -0,0 +1,62 @@
...
+  directive @external on OBJECT | FIELD_DEFINITION
+
+  scalar FieldSet
+  `),
+
+  Schema.basic(gql `${'builtin/federation/v2.0.graphql'}
+  @id(url: "https://specs.apollo.dev/federation/v2.0")
+
+  """
+  federation 2.0 key directive
```
pcmanus, line 34: It's missing:
- `@inaccessible`: which has been recently added (#1484).
- `@extend`: see comment above.
- `@tag`? Subgraphs should be able to `import` `@tag` from the fed2 feature. Does that work if `@tag` is not defined here?

subgraph-js/src/federation-atlas.ts
```diff
@@ -0,0 +1,62 @@
...
+export const SUBGRAPH_BASE = Schema.basic(
+  gql `${'builtin:subgraph-base'}
+    @link(url: "https://specs.apollo.dev/federation/v1.0", as: "",
+          import: "@key @requires @provides @external")
+    @link(url: "https://specs.apollo.dev/tag/v0.1")
+  `)
+
+export const ATLAS = Atlas.fromSchemas(
+  Schema.basic(gql `${'builtin/tag/v0.1'}
+  @id(url: "https://specs.apollo.dev/tag/v0.1")
```
pcmanus, line 12: We should also add `tag/v0.2` which landed recently (#1652) and will be supported by GA (and it is the version that `federation/v2.0` now use).

subgraph-js/src/federation-atlas.ts
```diff
@@ -0,0 +1,62 @@
...
+  @id(url: "https://specs.apollo.dev/link/v0.3")
+
+  directive @link(url: String!, as: String, import: [Import])
+    repeatable on SCHEMA
+
+  scalar Import
+  `),
+
+  Schema.basic(gql `${'builtin/id/v1.0.graphql'}
+  @id(url: "https://specs.apollo.dev/id/v1.0")
```
pcmanus, line 56: I don't know what that spec is/does. I do remember it being mentioned in a document a while ago, and I guess I can intuit the general idea, but I don't remember that document being really agreed upon in its entirely, nor further discussion/agreement on that `@id` in particular. Did I miss it?

To be clear, this seem reasonable enough at first glance, but if we ship it in GA, it becomes visible to users and we have to maintain it/have backward compatibility moving forward, and because this is new to me and we're so close to out target date for GA, that makes me quite uncomfortable (not really sure what we're committing to with this).

Obviously, if this was discussed/agreed upon before and I just missed it, that's totally fine and that's my bad.

subgraph-js/src/federation-atlas.ts
```diff
@@ -0,0 +1,62 @@
+import { Atlas, Schema, gql } from "@apollo/core-schema";
+
+export const SUBGRAPH_BASE = Schema.basic(
+  gql `${'builtin:subgraph-base'}
+    @link(url: "https://specs.apollo.dev/federation/v1.0", as: "",
+          import: "@key @requires @provides @external")
```
pcmanus, line 6: The syntax of this `import` (and the others in that file) is incompatible with the definition of `link/v0.3` made below (`import` is defined of type `[Import]` but that exclude the string syntax). I continue to think that we shouldn't have multiple different syntax for this `import` argument: it's going to confuse users, _especially_ if those different syntaxes are not universally supported.

I'll also note that if we release code as GA that support 2 different syntaxes, it's a one-way door: user my start using both and we can't remove one with breaking those users. It's not a decision we can "postpone".

subgraph-js/src/federation-atlas.ts
```diff
@@ -0,0 +1,62 @@
...
+  directive @requires(fields: FieldSet!) on FIELD_DEFINITION
+  directive @provides(fields: FieldSet!) on FIELD_DEFINITION
+  directive @external on OBJECT | FIELD_DEFINITION
+  directive @moving(to: String!) on FIELD_DEFINITION
+  directive @shareable on FIELD_DEFINITION | OBJECT
+
+  scalar FieldSet
+  `),
+
+  Schema.basic(gql `${'builtin/link/v0.3.graphql'}
```
pcmanus, line 46: This was mentioned in other places, but for completeness: we did release the preview with `link/v1.0` so changing to `v0.3` now has consequences (and as I mentioned in https://github.com/apollographql/federation/pull/1662#issuecomment-1084352362, I do think fed 2 GA is the right place for bumping the link spec to a "stable" version number).

subgraph-js/src/buildSubgraphSchema.ts
queerviolet: done

subgraph-js/src/federation-atlas.ts
```diff
@@ -0,0 +1,62 @@
...
+  Schema.basic(gql `${'builtin/tag/v0.1'}
+  @id(url: "https://specs.apollo.dev/tag/v0.1")
+  directive @tag(name: String!)
+    repeatable on FIELD_DEFINITION | INTERFACE | OBJECT | UNION
+  `),
+  Schema.basic(gql `${'builtin/federation/v1.0.graphql'}
+  @id(url: "https://specs.apollo.dev/federation/v1.0")
+
+  """
+  federation 1.0 key directive
```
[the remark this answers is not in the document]
queerviolet: done. we could import fed1's extend from fed2, but then we would end up linking both specs, which would probably confuse the composer? i just duplicated it.

subgraph-js/src/federation-atlas.ts
```diff
@@ -0,0 +1,78 @@
...
+  directive @provides(fields: FieldSet!) on FIELD_DEFINITION
+  directive @external on OBJECT | FIELD_DEFINITION
+
+  scalar FieldSet
+  `),
+
+  Schema.basic(gql `${'builtin/federation/v2.0.graphql'}
+  @id(url: "https://specs.apollo.dev/federation/v2.0")
+  @link(url: "https://specs.apollo.dev/tag/v0.1")
+  @link(url: "https://specs.apollo.dev/inaccessible/v0.1", import: "@ (as @inaccessible)")
```
pcmanus, line 36: Is that new syntax?

subgraph-js/src/federation-atlas.ts
```diff
@@ -0,0 +1,62 @@
...
+  scalar FieldSet
+  `),
+
+  Schema.basic(gql `${'builtin/federation/v2.0.graphql'}
+  @id(url: "https://specs.apollo.dev/federation/v2.0")
+
+  """
+  federation 2.0 key directive
+  """
+  directive @key(fields: FieldSet!) repeatable on OBJECT | INTERFACE
```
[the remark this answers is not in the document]
queerviolet: fixed

subgraph-js/src/federation-atlas.ts
queerviolet: fixed

subgraph-js/src/federation-atlas.ts
queerviolet: it's compiled away now, they never show up in the output.

subgraph-js/src/federation-atlas.ts
queerviolet: this directive is the dumbest thing. fixed.

subgraph-js/src/federation-atlas.ts
queerviolet: removed

subgraph-js/src/federation-atlas.ts
queerviolet: changed

subgraph-js/src/__tests__/buildSubgraphSchema.test.ts
```diff
@@ -80,7 +87,6 @@ type Money {
       }
     }`;
     const schema = buildSubgraphSchema(gql`
-      "Description text on 'SchemaDefinition' nodes supported as per the October 2021 Edition of the spec."
```
benweatherman, line 83: **question** Is supporting this no longer needed?

subgraph-js/src/index.ts
```diff
@@ -1,2 +1 @@
 export { buildSubgraphSchema } from './buildSubgraphSchema';
-export { printSubgraphSchema } from './printSubgraphSchema';
```
benweatherman, line 2: **question** Was this only used for testing? Or are we suggesting using the `graphql-js` version?

subgraph-js/src/__tests__/buildSubgraphSchema.test.ts
queerviolet: oh, i don't know why i dropped that? fixed.

subgraph-js/src/index.ts
benweatherman: I wasn't sure if this was an API we thought users would use, no need to re-add it if we don't think it was used outside of `subgraph-js`.

queerviolet: you know, i literally just spoke to a user who's using it. granted, they were using it to fix a bug that this pr fixes. but maybe let's keep it around just in case

queerviolet:
ptal, i think this is ready to go

benweatherman (approved):
🐐

queerviolet:
thanks!

ruben-nogueira:
After upgrading to `2.0.0-preview.12` I'm getting those errors.

`Error: Must provide only one schema definition.

There can be only one query type in schema.

There can be only one mutation type in schema.
`

Those types are using the `extend type Query` or `extend type Mutation`  (we have some of them). @queerviolet Can you please confirm if it's related with PR? Thanks!

cpeacock:
Hi @ruben-nogueira!  Could you open another issue and/or please send me your generated supergraph schemas?  These errors seem to be coming from graphql-js, which probably means there's a duplication of the schema definition somewhere or something similar, we're taking a look at it.

ruben-nogueira:
@cpeacock thanks for your response. I created a new issue here: https://github.com/apollographql/federation/issues/1703

PR #1556, opened 2022-03-02
Rejects composition when the options for a path have different runtim…

### LINKED ISSUE #1375: Incorrect Query Plan for Interface type queries
"@apollo/query-planner": "2.0.0-alpha.3",

### Subgraph1 Schema
```
  const subgraph1 = {
    name: 'Subgraph1',
    typeDefs: gql`
      type Query {
        node(id: ID!): Node!
      }

      interface Node {
        id: ID!
      }

      type User implements Node @key(fields: "id") {
        id: ID!
        prop1: String
      }
    `
  }
```

### Subgraph2 Schema
```
  const subgraph2 = {
    name: 'Subgraph2',
    typeDefs: gql`
      type Query {
        node(id: ID!): Node!
      }

      interface Node {
        id: ID!
      }

      type Client implements Node @key(fields: "id") {
        id: ID!
        prop2: String
      }
    `
  }
```

###GraphQL Query
```
query NodeQuery{
  node(id: "456") {
	... on User {
            prop1
        }
   }
}
```

### Actual Query plan
```
  Fetch(service: "Subgraph2") {
    {
      node(id: 456) {
        __typename
      }
    }
  },
}
```

## Expected Behavior 
1) the query plan should be formulated for "Subgraph1" where User type has been defined.

## Bug Identification
```
`${selection.selectionSet}`
'{ ... on User { prop1 } }'
`${this.stack}`
'... on User { prop1 },Query(Subgraph1) --[node]--> Node(Subgraph1) (types: [User]),Query(Subgraph2) --[node]--> Node(Subgraph2) (types: [Client])'
`${selection}`
'... on User { prop1 }'
`${options}`
'Query(Subgraph1) --[node]--> Node(Subgraph1) (types: [User]),Query(Subgraph2) --[node]--> Node(Subgraph2) (types: [Client])'

```

Per the code we consider 2 options for selection `{ ... on User { prop1 } }`
But then one of the option fails as Subgraph2 does not define the type "User"

```
`${allOptions}`
'Query(Subgraph1) --[node]--> Node(Subgraph1) --[... on User]--> User(Subgraph1) (types: [User])'
`${option}`
'Query(Subgraph1) --[node]--> Node(Subgraph1) (types: [User])'
`${option}`
'Query(Subgraph2) --[node]--> Node(Subgraph2) (types: [Client])'
`${advanceOptionsToString(options)}`
'<unsatisfiable branch>'
```

But since the "followupForOption" Subgraph2 is not possible the code returns immediately without considering the option that has been identified for Subgraph1

```
      if (followupForOption.length === 0) {

        if (operation.kind === 'FragmentElement') {
          this.closedBranches.push([option.paths.map(p => terminateWithNonRequestedTypenameField(p))]);
        }
        return;
      }
```

I think it should have been
```
        if (newOptions.length == 0 && i == options.length && operation.kind === 'FragmentElement') {
          this.closedBranches.push([option.paths.map(p => terminateWithNonRequestedTypenameField(p))]);
          return;
        }
```

If we have worked on all options and it is the last option and no other options before resulted in a solution, then terminate and return.

pcmanus:
…e types

During composition validation, a path of the supergraph can have
multiple valid options in subgraphs due to shared fields. When
that happens, this patch validates that the possible runtime types
of the various options are the same. The rational being that
shared fields should resolve the same way in all subgraphs, so
allowing 2 differen solution of shared field to have different set
of possible runtime types is a dangerous proposition: to be correct,
users would have make sure they only ever return value for that field
that are in the intersection of those possible runtime types.
Instead of asking users to reason about this, forcing the sets of
possible runtime types to always be the same (for a given shared field)
ensures any returned value is always in the intersection (since the
intersection of a set with itself is itself).

Fixes #1375


composition-js/src/validate.ts
```diff
@@ -64,25 +69,40 @@ export class ValidationError extends Error {
...
-function validationError(
+
+function satisfiabilityError(
   unsatisfiablePath: RootPath<Transition>,
   subgraphsPaths: RootPath<Transition>[],
   subgraphsPathsUnadvanceables: Unadvanceables[]
-): ValidationError {
+): GraphQLError {
   const witness = buildWitnessOperation(unsatisfiablePath);
-
```
pcmanus, line 73: Remark: removed that TODO because it is outdated, we do have more detailed error messages now. Technically unrelated to that PR though.

composition-js/src/validate.ts
```diff
@@ -257,6 +277,19 @@ function initialSubgraphPaths(kind: SchemaRootKind, subgraphs: QueryGraph): Root
   return subgraphs.outEdges(root).map(e => initialState.add(subgraphEnteringTransition, e, noConditionsResolution));
 }
 
+function possibleRuntimeTypeNamesSorted(path: RootPath<Transition>): string[] {
+  const types = path.tailPossibleRuntimeTypes().map((o) => o.name);
+  types.sort((a, b) => a.localeCompare(b));
+  return types;
+}
+
+export function extractValidationError(error: any): ValidationError | undefined {
```
pcmanus, line 286: Remark: this method is not used within federation itself, but it's useful in some tooling (planning to use it in the update to https://github.com/apollographql/composer-tool after this patch) so I'd prefer keeping it.

query-planner-js/src/buildPlan.ts
```diff
@@ -177,7 +177,7 @@ class QueryPlanningTaversal<RV extends Vertex> {
         // Do note that we'll only need that `__typename` if there is no other selections inside `foo`, and so we might include
         // it unecessarally in practice: it's a very minor inefficiency though.
         if (operation.kind === 'FragmentElement') {
-          this.closedBranches.push([option.paths.map(p => terminateWithNonRequestedTypenameField(p))]);
+          this.closedBranches.push(options.map((o) => o.paths.map(p => terminateWithNonRequestedTypenameField(p))));
```
pcmanus, line 180: Remark: this change is technically not really necessary to this patch, and in fact I'm not entirely sure if this whole branch of query planner can still be hit with the validation added (but I'm unsure and would rather keep it for now), but I noticed this while working on this patch and the new version is "more correct" (mostly, it avoids producing plans that are more inefficient than necessary if this code path is triggered).

composition-js/src/validate.ts
```diff
@@ -257,6 +277,19 @@ function initialSubgraphPaths(kind: SchemaRootKind, subgraphs: QueryGraph): Root
   return subgraphs.outEdges(root).map(e => initialState.add(subgraphEnteringTransition, e, noConditionsResolution));
 }
 
+function possibleRuntimeTypeNamesSorted(path: RootPath<Transition>): string[] {
+  const types = path.tailPossibleRuntimeTypes().map((o) => o.name);
```
clenfest, line 281: Is this already deduped?

composition-js/src/__tests__/validation_errors.test.ts
```diff
@@ -112,3 +112,57 @@ describe('@requires', () => {
     ]);
   });
 });
+
+
```
clenfest, line 116: In the comments of validation.ts you have a union as an example, did you want to have that as an explicit test as well or are the interface / union examples sufficently similar?

composition-js/src/validate.ts
```diff
@@ -306,9 +339,78 @@ export class ValidationState {
...
+    //
+    // Note that if motivated user demand for allowing the example above arises and we're not worried about users getting this wrong, then
+    // we could relax this rule. But even then, we should probably at least reject the case where the sets of runtime types for all
+    // the sources don't intersect, since there would be no good way to implement such case.
+    if (newSubgraphPaths.length > 1 && isAbstractType(newSubgraphPaths[0].tail.type)) {
+      const runtimeTypesToSubgraphs = new MultiMap<string, string>();
+      for (const path of newSubgraphPaths) {
+        // Note: we add spacing between elements because we're going use it if we display an error. This doesn't impact our set equality though
+        // since type names can have neither comma or space in graphQL.
+        runtimeTypesToSubgraphs.add(possibleRuntimeTypeNamesSorted(path).join(', '), path.tail.source);
```
clenfest, line 401: Trying to follow this, but I'm not sure if I'm getting it right. For the following example:
```
# subgraph 1
Query {
  a: A!
}

interface A {
  x: Int
}
type B implements A {
  x: Int
}

#subgraph 2
Query {
  a: A!
}

interface A {
  x: Int
}
type C implements A {
  x: Int
}
```
The things that are being compared are the internals of types B and C to make sure they match, so the string you're building will look something like "x:Int"? Is that right? If so, I think it looks good, if otherwise I may have more questions.

composition-js/src/validate.ts
pcmanus: It is. And it does take a small effort to make sure so I commented as much on the method. But essentially, `tailPossibleRuntimeTypes` in `GraphPath` is only updated through the (private) `updateRuntimeTypes` method in `graphPaths.ts`, and the only place we "push" there is guarded by I first check to ensure the element is not already there.

composition-js/src/__tests__/validation_errors.test.ts
pcmanus: You're right, it's worth having both and pulled the example from `validation.ts` into a proper test (which also ensure the comment is actually correct, which it wasn't cause I had forgotten some `@shareable` :)).

composition-js/src/validate.ts
pcmanus: So no. In that example, that code path will kick in after we've look at field `Query.a`, and we will essentially have 2 "options" (2 `newSubgraphPaths`), one per subgraph, but each one "tail" will be on interface `A`. So for the 1st path, `possibleRuntimeTypeNamesSorted` will return `B`, while  it will return `C`. And because those are different, we'll return the error. Does that help?

pcmanus:
Unfortunately, in the lead up to federation 2.0, this PR fall through the crack and never got finished/committed.

I'd like to resurrect this and get it committed ASAP however, because I think this can really help people not getting `@shareable` on entity and root fields wrong. #2228 is a recent example of this can easily trip people up at the moment, but there is some evidence that this is not at all the only case.

That said, upon further reflexion (and a bit influenced by the fact this missed 2.0), I changed the check a bit to make it a bit more permissive. That is the rebased version:
- error composition if the runtime types (for a given shareable field) have no intersection whatsoever, as this essentially is guaranteed to be incorrect for `@shareable`. This would catch example like the on in #2228, or users attempting to do the "node relay API" with `@shareable` (which cannot work; it's absolutely important we improve how easy it is to implement the "node relay API", but `@shareable` definitively isn't it and we should help user realise it quickly).
- warn (issue a hint with `WARN` level) if the runtime types intersects, but are not equal. The reasoning for a warn is that it's possible to implement such a case correctly, and one could imagine specific case where you could end up in this situation (one might be when you evolve 2 subgraphs at a different pace), but it's still a bit of a red flag so a warning seem appropriate.

clenfest (approved):

PR #1561, opened 2022-03-03
Add `resolvable` argument to `@key`

pcmanus:
This adds a new optional boolean `resolvable` argument to the `@key` directive. That argument default to `true` but if explicitly set to `false`, then the query planner does not assume that the key has an associated `__resolveReference`. In other words, this allow to mark some fields as keys of an entity but while declaring that the subgraph does not know how to resolve the entity based on those key.
This is useful for subgraph that want to "reference" an entity but don't want/can't easily resolve the entity. Before this patch, the same could be done by skipping the `@key` entirely in that case, but while doing so works in practice, it is less clear conceptually that the type is an entity type (at least looking at that subgraph).
Essentially, the goal of this patch is to make it so that we can always identify an entity type as type having a `@key` and this even when a subgraph do not need/want to have a resolver for said key.

### DIFF
--- a/composition-js/src/merging/merge.ts
+++ b/composition-js/src/merging/merge.ts
@@ -681,7 +681,8 @@ class Merger {
       } else {
         for (const key of keys) {
           const extension = key.ofExtension() || source.hasAppliedDirective(sourceMetadata.extendsDirective()) ? true : undefined;
-          dest.applyDirective(joinTypeDirective, { graph: name, key: key.arguments().fields, extension });
+          const { resolvable } = key.arguments();
+          dest.applyDirective(joinTypeDirective, { graph: name, key: key.arguments().fields, extension, resolvable });
         }
       }
     }
--- a/internals-js/src/extractSubgraphsFromSupergraph.ts
+++ b/internals-js/src/extractSubgraphsFromSupergraph.ts
@@ -101,7 +101,8 @@ export function extractSubgraphsFromSupergraph(supergraph: Schema): Subgraphs {
           subgraphType = schema.addType(newNamedType(type.kind, type.name));
         }
         if (args.key) {
-          const directive = subgraphType.applyDirective('key', {'fields': args.key});
+          const { resolvable } = args;
+          const directive = subgraphType.applyDirective('key', {'fields': args.key, resolvable});
           if (args.extension) {
             directive.setOfExtension(subgraphType.newExtension());
           }
--- a/query-graphs-js/src/graphPath.ts
+++ b/query-graphs-js/src/graphPath.ts
@@ -21,8 +21,8 @@ import {
   isObjectType,
   mapValues,
   federationMetadata,
-  isEntityType,
   isSchemaRootType,
+  Directive,
 } from "@apollo/federation-internals";
 import { OpPathTree, traversePathTree } from "./pathTree";
 import { Vertex, QueryGraph, Edge, RootVertex, isRootVertex, isFederatedGraphRootType } from "./querygraph";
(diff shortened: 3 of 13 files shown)


subgraph-js/src/directives.ts
```diff
@@ -30,6 +31,10 @@ export const KeyDirective = new GraphQLDirective({
     fields: {
       type: new GraphQLNonNull(GraphQLString),
     },
+    resolvable: {
```
clenfest, line 34: It's a little weird to me that we have to add to here as well as in the keyDirectiveSpec in `federationSpec.ts`. Is there a way to derive our GraphQLDirectives from our DirectiveSpecifications or vice versa so that we don't have to maintain both sets of types?

internals-js/src/extractSubgraphsFromSupergraph.ts
```diff
@@ -101,7 +101,8 @@ export function extractSubgraphsFromSupergraph(supergraph: Schema): Subgraphs {
           subgraphType = schema.addType(newNamedType(type.kind, type.name));
         }
         if (args.key) {
-          const directive = subgraphType.applyDirective('key', {'fields': args.key});
+          const resolvable = args.resolvable === false ? false : undefined;
```
clenfest, line 104: Maybe just this instead?
```typescript
          const { resolvable } = args;
```

composition-js/src/merging/merge.ts
```diff
@@ -681,7 +681,8 @@ class Merger {
       } else {
         for (const key of keys) {
           const extension = key.ofExtension() || source.hasAppliedDirective(sourceMetadata.extendsDirective()) ? true : undefined;
-          dest.applyDirective(joinTypeDirective, { graph: name, key: key.arguments().fields, extension });
+          const resolvable = key.arguments().resolvable === false ? false : undefined;
```
clenfest, line 684: Same here as earlier, if we just want to copy the value in `key.arguments().resolvable`, we should do so rather than use a ternary operator. `const { resolvable } = args;`

query-graphs-js/src/graphPath.ts
```diff
@@ -824,12 +824,18 @@ export function advancePathWithTransition<V extends Vertex>(
       if (type && isCompositeType(type) && type.field(fieldName)) {
         // That subgraph has the type we look for, but we have recorded no dead-ends. This means there is no edge to that type,
         // and thus that it has no keys.
-        assert(!isEntityType(type), () => `Expected type ${type} in ${subgraph} to not have keys`);
+        const metadata = federationMetadata(schema);
+        const keys: Directive<CompositeType, {fields: any, resolvable?: boolean}>[] = metadata ? type.appliedDirectivesOf(metadata.keyDirective()) : [];
```
clenfest, line 828: no need to explicitly declare the type here since we don't want to have to change in multiple places if we add a new field down the line.

clenfest (approved):
couple of minor comments, but looks good to me.

subgraph-js/src/directives.ts
pcmanus: It is a bit unfortunate and I hope we can eventually fix that, but I'd rather punt on that for now because 1) `subgraph-js` and `internals-js` don't currently depend on each other and it's worth thinking if we have to have a dependency and 2) because both use slightly different abstractions, the we'd probably have to hard-code a bunch of equivalences to derive one from the other, and that's probably a bit of effort to take in (aka, this probably deserve to be a separate dedicated task at least).

clenfest: Definitely didn't mean that we should address it now, I was just asking in general.

trevor-scheer:
Does this need an accompanying update to the join spec / version bump to a new `v0.3`?

pcmanus:
So far, my mode of operation has been to modify the `v0.2` without bumping the version since we haven't got to GA, nor published `v0.2` yet, and so in my mind we're still working on finalising that version.

I will also note that in practice this is only adding an optional argument, so this is largely backward compatible with previous alpha release (in the sense that previously generated supergraph are still going to be read just fine by versions having this patch).

PR #1567, opened 2022-03-07
Fix merging of arguments

pcmanus:
We shouldn't merge arguments unless they appear in all subgraphs, orthis lead to unclear semantic if those argument are passed by the end user.

I'll also note that the pre-existing behaviour, which was adding arguments to the supergraph if it appeared in any subgraph was potentially leading to errors when decoding the supergraph because we didn't kept information of where the argument was defined and as a result we sometimes inferred that an argument was in a subgraph that didn't have the argument type. Long story short, if we only include argument that are in all subgraphs (that have the field/directive definition), we solve that problem as well.

### DIFF
--- a/composition-js/src/merging/merge.ts
+++ b/composition-js/src/merging/merge.ts
@@ -84,7 +84,11 @@ const coreSpec = CORE_VERSIONS.latest();
 const joinSpec = JOIN_VERSIONS.latest();
 const tagSpec = TAG_VERSIONS.latest();
 
-const MERGED_TYPE_SYSTEM_DIRECTIVES = ['inaccessible', 'deprecated', 'specifiedBy', 'tag'];
+// We're currently not really "merging" any type-system directives. Though this is
+// a misleading because `tag` is included in the subgraph if and only if it is
+// used in subgraphs, so it is kind of "merged", but this is currently handled
+// specially.
+const MERGED_TYPE_SYSTEM_DIRECTIVES: string[] = [];
 
 export type MergeResult = MergeSuccess | MergeFailure;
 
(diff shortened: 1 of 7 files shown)


composition-js/src/merging/merge.ts
```diff
@@ -1102,40 +1106,55 @@ class Merger {
...
+    }
+
+    for (const argName of argNames) {
+      // We add the arguement unconditionally even if we're going to remove it in
+      // some path. Done because this helps reusing our "reportMismatchHint" method
+      // in those cases.
+      const arg = dest.addArgument(argName);
+      // If all the sources that have the field have the argument, we do merge it
+      // and we're good, but otherwise ...
+      if (!sources.every((s) => !s || s.argument(argName))) {
```
clenfest, line 1124: Is checking the name of the argument sufficent? Don't we also need to check the type or is that getting checked somewhere else?

benweatherman, line 1124: **nitpicky question** Could this be changed to

```suggestion
      if (sources.some((s) => !s || s.argument(argName))) {
```

I think that's the same logic, but it's a little easier for my brain to understand that version.

composition-js/src/merging/merge.ts
```diff
@@ -1102,40 +1106,55 @@ class Merger {
...
             (_, subgraphs) => `it is defined in ${subgraphs}`,
             (_, subgraphs) => ` but not in ${subgraphs}`,
             undefined,
             true // Do include undefined sources, that's the point
           );
-          // Note that we remove the element after the hint because we access the parent in the hint message.
-          dest.remove();
-          return;
         }
+        arg.remove();
```
clenfest, line 1152: There was a comment about why this was being removed that helped understanding of the code. Could you add it back in?

composition-js/src/merging/merge.ts
```diff
@@ -1102,40 +1106,55 @@ class Merger {
...
+      // If all the sources that have the field have the argument, we do merge it
+      // and we're good, but otherwise ...
+      if (!sources.every((s) => !s || s.argument(argName))) {
+        // ... we don't merge the argument: some subgraphs wouldn't know what
+        // to make of it and that would be dodgy at best. If the argument is
+        // optional in all sources, then we can compose properly and just issue a
+        // hint. But if it is mandatory, then we have to fail composition, otherwise
+        // the query planner would have no choice but to generate invalidate queries.
+        const nonOptionalSources = sources.map((s, i) => s && s.argument(argName)?.isRequired() ? this.names[i] : undefined).filter((s) => !!s) as string[];
+        if (nonOptionalSources.length > 0) {
```
benweatherman, line 1131: **question/nitpick** Do we normally compare against `> 0`?

```suggestion
        if (nonOptionalSources.length) {
```

internals-js/src/error.ts
```diff
@@ -301,6 +301,11 @@ const INVALID_LINK_DIRECTIVE_USAGE = makeCodeDefinition(
   'An application of the @link directive is invalid/does not respect the specification.'
 );
 
+const REQUIRED_ARGUMENT_MISSING_IN_SOME_SUBGRAPH = makeCodeDefinition(
+  'REQUIRED_ARGUMENT_MISSING_IN_SOME_SUBGRAPH',
+  'An argument (of a field or directive definition) is not defined in all the subgraphs defining the field and is mandatory in at least some of them.'
```
benweatherman, line 306: **nitpick** I think this error message could be more clear (but that might be due to my newness)

```suggestion
  'An argument of a field or directive definition is mandatory in some subgraphs, but the argument is not defined in all subgraphs that define the field or directive definition.'
```

benweatherman (approved):
Added some nitpicky comments, do with them what you will. Otherwise, things look great to me!

composition-js/src/merging/merge.ts
pcmanus: The type is checked later when we properly call `mergeArgument` (which may error out composition in its own right).

pcmanus: Right. I initially had the `if` without the negation and then a `else`, but realised I didn't need the first part. Agreed avoiding the negation is probably better now. That said, to have the same logic I believe it needs to be:
```
if (sources.some((s) => s && !s.argument(argName))) {
```
Essentially, `Not(For all x, P(x))` is `It exists x, Not(P(x))`.

composition-js/src/merging/merge.ts
pcmanus: Added back.

composition-js/src/merging/merge.ts
benweatherman: This seems fine to me, I don't feel strongly. I just wanna make sure I stay consistent when I do this sort of thing.

(Full disclosure: I'm normally in the "don't compare against 0" camp but 🤷 )

composition-js/src/merging/merge.ts
benweatherman: Right, good catch on the logic. Makes sense to me if you wanna make that tweak.

PR #1568, opened 2022-03-07
Fix operation names

pcmanus:
This is a followup to #1550: as mentioned in my last comment, that PR adds the subgraph names to the operation name, but we unfortunately allow near anything as a subgraph name, and so in particular we can have subgraph names that are invalid graphQL names. in which case the query plan we generate ends up invalid.

This PR "sanitize" the subgraph name (the other part we use for the operation name is the name of the original query, but we know _that_ is a valid graphQL name by definition). It also switch from a single `_` to `__` to separate the elements from the generate operation names so it's a bit clearer if some underscores are already used for either the original query name or the subgraph name.

### DIFF
--- a/query-planner-js/src/buildPlan.ts
+++ b/query-planner-js/src/buildPlan.ts
@@ -833,7 +833,6 @@ class FetchGroup {
           operationName,
         );
 
-    try {
     const fetchNode: FetchNode = {
       kind: 'Fetch',
       serviceName: this.subgraphName,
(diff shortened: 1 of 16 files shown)


query-planner-js/src/buildPlan.ts
```diff
@@ -834,6 +851,11 @@ class FetchGroup {
         path: this.mergeAt!,
         node: fetchNode,
       };
+    } catch (e) {
+      console.log(`Got ${e} with ${operation}`);
```
benweatherman, line 855: **question** Were these `log`s part of debugging or does it make sense to keep them in? I'm wondering if these should be changed to `.warn` or `.info` to make it more clear how bad the error is.

benweatherman (approved):
Added some unimportant comments but all looks great to me!

PR #1581, opened 2022-03-10
Aggregate errors with @inaccessible (fed1)

benweatherman:
Previously, `removeInaccessibleElements` would throw a `GraphQLError` the first time it encountered an inaccessible mismatch for types & fields.

With this change, the code will continue to parse the schema and report an `GraphQLSchemaValidationError` that contains all the mismatches.e


benweatherman:
One potential regression we could handle here: if there is only a single validation error, we could continue to throw a `GraphQLError` in case users are checking for the type of the exception being thrown. Something along the lines of:

```typescript
if (errors.length === 1) {
  throw errors[0];
}
else {
  throw new GraphQLSchemaValidationError(errors);
}
```

Does anyone have an idea of how breaking this change might be for existing callers?

query-planner-js/src/composedSchema/removeInaccessibleElements.ts
```diff
@@ -110,23 +110,28 @@ export function removeInaccessibleElements(
...
             `an @inaccessible type without being marked @inaccessible itself.`,
           fieldConfig.astNode,
-        );
+        ));
       }
 
       newFieldMapConfig[fieldName] = fieldConfig;
     }
 
+    if (errors.length) {
```
clenfest, line 131: I know this works, but I still prefer to have the `> 0` on there for type checking.

benweatherman: Oh totes! I even asked specifically about this on the fed2 PR and then totally forgot to carry the convention over here. Thanks for raising this one.

pcmanus (commented):
The one question I had on that version is whether we should try to make the newly created error be a `GraphQLError` itself (it's the case in the fed 2 version). The reason it "might" be a good idea is to make it a bit less risky on updates. Because I expect that existing code currently calling `toAPISchema` might have some existing try-catch that checks for `GraphQLError` specifically (and may treat other kind of errors are bugs or something). If that's the case, such code would break until it is updated. If our aggregate error is still a `GraphQLError`, then such code wouldn't break too badly (of course, that code would display a less useful error message, but it's better than breaking).

But honestly unsure if we should care about such thing. Just offering it because it occurred to me, and I don't mind merging as is and we feel it's not worth the trouble.

query-planner-js/src/composedSchema/removeInaccessibleElements.ts
benweatherman: Updated in https://github.com/apollographql/federation/pull/1581/commits/6b5e2feafceb4a908c2a477dfa39e1980465cbab

benweatherman:
Good idea deriving from `GraphQLError` to make it easier for clients to upgrade. That should be updated in the latest.

pcmanus (approved):
Lgtm with the last changes, good job!

Let's maybe just make sure we have commit with a changlog and give a gentle ping to our studio friends so they remember to update their code when they merge whichever version this end up in.

PR #1582, opened 2022-03-10
Fix adding descriptions to type extensions

pcmanus:
If a schema had both a type definition and a correspdonding type
extension, the printing code was mistakenly printing the description of
the type definition on the type extension as well. But since description
are not syntactically valid on type extension in graphQL, this leads to
syntactic errors.

More precisely, an input schema like:
```graphql
""" I'm a type """
type T {
  a: Int
}

extend type T {
  b: Int
}
```
was (before this patch) mistakenly printed as:
```graphql
""" I'm a type """
type T {
  a: Int
}

""" I'm a type """
extend type T {
  b: Int
}
```
which is syntactically invalid due to the description on the type
extension.

Note: the patch also fix a related additional issue where the presence
of type extension might have the type definition description ignored due
a lack of conditional when setting descriptions in `buildSchema`.

### DIFF
--- a/internals-js/src/__tests__/definitions.test.ts
+++ b/internals-js/src/__tests__/definitions.test.ts
@@ -427,6 +427,10 @@ test('handling of descriptions', () => {
       pages: Int @Important
     }
 
+    extend type Book {
+      author: String
+    }
+
     type DVD implements Product {
       id: ID!
       description: String!
--- a/internals-js/src/buildSchema.ts
+++ b/internals-js/src/buildSchema.ts
@@ -299,7 +299,9 @@ function buildNamedTypeInner(
       const enumType = type as EnumType;
       for (const enumVal of definitionNode.values ?? []) {
         const v = enumType.addValue(enumVal.name.value);
-        v.description = enumVal.description?.value;
+        if (enumVal.description) {
+          v.description = enumVal.description.value;
+        }
         v.setOfExtension(extension);
         buildAppliedDirectives(enumVal, v);
       }
@@ -315,7 +317,9 @@ function buildNamedTypeInner(
       break;
   }
   buildAppliedDirectives(definitionNode, type, extension);
-  type.description = definitionNode.description?.value;
+  if (definitionNode.description) {
+    type.description = definitionNode.description.value;
+  }
   type.sourceAST = definitionNode;
 }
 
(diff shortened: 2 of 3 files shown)


internals-js/src/buildSchema.ts
```diff
@@ -299,7 +299,9 @@ function buildNamedTypeInner(
       const enumType = type as EnumType;
       for (const enumVal of definitionNode.values ?? []) {
         const v = enumType.addValue(enumVal.name.value);
-        v.description = enumVal.description?.value;
+        if (enumVal.description) {
```
clenfest, line 302: Is this a change? Behavior looks the same to me.

internals-js/src/buildSchema.ts
```diff
@@ -315,7 +317,9 @@ function buildNamedTypeInner(
       break;
   }
   buildAppliedDirectives(definitionNode, type, extension);
-  type.description = definitionNode.description?.value;
+  if (definitionNode.description) {
```
clenfest, line 320: same as above.

internals-js/src/__tests__/definitions.test.ts
```diff
@@ -427,6 +427,10 @@ test('handling of descriptions', () => {
       pages: Int @Important
     }
 
+    extend type Book {
```
clenfest, line 430: do we need to check the result anywhere?

clenfest (approved):
I left a couple of comments, but if you're ok with them, feel free to merge.

internals-js/src/buildSchema.ts
pcmanus: That's the part I'm referring to in the "Note" of the commit message: it's not the same behaviour in that this code may run multiple times for the same `type`, so if we had a description the first time (say on the type definition), then we were erasing it if there was no description the 2nd time (say on the type extension). The change makes it so we don't erase it anymore.

internals-js/src/buildSchema.ts
pcmanus: See comment on types below. I actually don't think the change is necessary in that case of enum values, but just made it to mirror the other similar cases in the file.

internals-js/src/__tests__/definitions.test.ts
pcmanus: The test ends by comparing the result of `printSchema` and the original input. So before this patch the test would fail due the bug solved here.

Issue #1579, opened 2022-03-10
isPromise check does not work with prisma promises


hayes:
The isPromise check here: https://github.com/apollographql/federation/blob/main/subgraph-js/src/types.ts#L47-L49 does not work correctly when then is inherited. 

`'then' in value` will return false because prisma promises are wrapped with a `Proxy`.  This can be worked around by doing value['then'].  This would likely require an additional check that the value is either a promise or a function before doing that lookup.

PR #1584, opened 2022-03-10
Add missing introspection fields and fix isPromise prisma compatiblility

### LINKED ISSUE #1578: Some introspection fields are missing
Looks like there are a few things missing in https://github.com/apollographql/federation/blob/main/internals-js/src/introspection.ts#L68

Specifically:
- `includeDeprecated` on `args` (for fields, input fields, and directives)
- `isDeprecated` on `__InputValue`
- `deprecationReason` on `__InputValue`

I think thats all thats missing, but I might have missed something.

hayes:
Fixes #1578, #1579 

The isPromise fix is related to https://github.com/prisma/prisma/issues/12252.

It should probably be fixed in both places.  Since prisma promises seem to work correctly with most other libraries (graphql-js, or just native await) making it work here too is probably correct, even though not correctly handling an `in` operation is kinda weird.


CONTRIBUTING.md
```diff
@@ -2,7 +2,7 @@
 
 # Apollo Contributor Guide
 
-> If this is your first time or 100th time contributing to a project, the [Open Source Guide](https://opensource.guide/how-to-contribute/) is a fantastic place to learn about Open Source Software (OSS) and how to get involed.
+> If this is your first time or 100th time contributing to a project, the [Open Source Guide](https://opensource.guide/how-to-contribute/) is a fantastic place to learn about Open Source Software (OSS) and how to get involved.
```
benweatherman, line 5: **praise** 👏

benweatherman (changes requested):
merci

Thank you for this excellent contribution! The code is great, though it would be ideal if we could create a separate PR for the promise check vs the introspection change. Any chance you'd be willing to do that? I hate to ask you to do extra work, but I've done it anyway 😸

PR #1585, opened 2022-03-11
diffTypeNodes return type causing problems with graphql15

### LINKED ISSUE #1538: Federation library failing to compile with tsc
## Last working version
@apollo/federation@0.33.9

## Broken versions
Anything newer than @apollo/federation@0.33.9

## Expected Behaviour
Apollo federation should compile with `tsc -p tsconfig.json --noEmit`

## Actual Behaviour
Apollo federation fails to compile due to composition/utils.d.ts referencing a namespace that the compiler cannot find.
```
node_modules/@apollo/federation/dist/composition/utils.d.ts:43:371 - error TS2503: Cannot find namespace 'Kind'.

43     kind: (Kind.SCALAR_TYPE_DEFINITION | Kind.OBJECT_TYPE_DEFINITION | Kind.INTERFACE_TYPE_DEFINITION | Kind.UNION_TYPE_DEFINITION | Kind.ENUM_TYPE_DEFINITION | Kind.INPUT_OBJECT_TYPE_DEFINITION | Kind.DIRECTIVE_DEFINITION | Kind.SCALAR_TYPE_EXTENSION | Kind.OBJECT_TYPE_EXTENSION | Kind.INTERFACE_TYPE_EXTENSION | Kind.UNION_TYPE_EXTENSION | Kind.ENUM_TYPE_EXTENSION | Kind.INPUT_OBJECT_TYPE_EXTENSION)[];
```

I don't have the time right now to put together a reproduction repo, but if this is required I can look into doing it at a later point. I've seen people mentioning this same issue on stack overflow but can't see an issue anywhere in the issue tracker.

## Typescript Config
I've added the typescript config that we use as that's the main thing I can think of that might influence this to fail.
```
{
  "compilerOptions": {
    "target": "es2017",
    "module": "commonjs",
    "strict": true,
    "baseUrl": ".",
    "paths": {
      "*": ["src/*"],
    },
    "typeRoots": ["node_modules/@types"],
    "types": ["node", "jest"],
    "esModuleInterop": true,
    "inlineSourceMap": true,
    "allowJs": false,
    "resolveJsonModule": true,
    "strictFunctionTypes": false,
  }
}
```

clenfest:
Fixes #1538 

Change return type on `diffTypeNodes` so that the Kind is explicit rather than implicit. I tested this by running the following super advanced program:

```typescript
import { defaultRootOperationNameLookup, compositionHasErrors, isStringValueNode, findDirectivesOnNode } from '@apollo/federation';
console.log('hello');
```

It failed on the old version, but seems to work with the new version. I don't totally understand why the old version is failing, so another set of eyes would be helpful, especially if we think there could be other areas of the code where this might be an issue.

### DIFF
--- a/federation-js/CHANGELOG.md
+++ b/federation-js/CHANGELOG.md
@@ -6,7 +6,7 @@ This CHANGELOG pertains only to Apollo Federation packages in the `0.x` range. T
 
 > The changes noted within this `vNEXT` section have not been released yet.  New PRs and commits which introduce changes should include an entry in this `vNEXT` section as part of their development.  When a release is being prepared, a new header will be (manually) created below and the appropriate changes within that release will be moved into the new section.
 
-- Fixes typing incompatibility issue with graphql15. [PR #1585](https://github.com/apollographql/federation/pull/1585)
+- Fixes type incompatibility issue with graphql15. [PR #1585](https://github.com/apollographql/federation/pull/1585)
 
 ## v0.35.3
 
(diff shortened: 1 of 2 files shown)


benweatherman (approved):
🐐

PR #1586, opened 2022-03-11
Improve error handling on the supergraph

pcmanus:
Some errors might be thrown while extracting subgraphs from the
supergraph but without being handled gracefully, resulting in messages
to the user that were less useful than they should be.

### DIFF
--- a/internals-js/src/extractSubgraphsFromSupergraph.ts
+++ b/internals-js/src/extractSubgraphsFromSupergraph.ts
@@ -292,7 +292,7 @@ export function extractSubgraphsFromSupergraph(supergraph: Schema): Subgraphs {
       if (isFed1) {
         // Note that this could be a bug with the code handling fed1 as well, but it's more helpful to ask users to recompose their subgraphs with fed2 as either
         // it'll solve the issue and that's good, or we'll hit the other message anyway.
-        const msg = `Error extracting subgraph "${subgraph.name}" from the supergraph: this might due to errors in subgraphs that were mistakenly ignored by federation 0.x versions but are rejected by federation 2.\n`
+        const msg = `Error extracting subgraph "${subgraph.name}" from the supergraph: this might be due to errors in subgraphs that were mistakenly ignored by federation 0.x versions but are rejected by federation 2.\n`
           + 'Please try composing your subgraphs with federation 2: this should help precisely pinpoint the problems and, once fixed, generate a correct federation 2 supergraph';
         throw new Error(`${msg}.\n\nDetails:\n${errorToString(e)}`);
       } else {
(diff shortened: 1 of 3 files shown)


clenfest (approved):
LGTM

internals-js/src/extractSubgraphsFromSupergraph.ts
```diff
@@ -292,11 +292,11 @@ export function extractSubgraphsFromSupergraph(supergraph: Schema): Subgraphs {
       if (isFed1) {
         // Note that this could be a bug with the code handling fed1 as well, but it's more helpful to ask users to recompose their subgraphs with fed2 as either
         // it'll solve the issue and that's good, or we'll hit the other message anyway.
-        const msg = `Error extracting subgraph ${subgraph.name} from the supergraph: this might due to errors in subgraphs that were mistakenly ignored by federation 0.x versions but are rejected by federation 2.\n`
-          + 'Please try composing your subgraphs with federation 2: this should help precisely pinpoint the errors and generate a correct federation 2 supergraph.';
+        const msg = `Error extracting subgraph "${subgraph.name}" from the supergraph: this might due to errors in subgraphs that were mistakenly ignored by federation 0.x versions but are rejected by federation 2.\n`
```
cpeacock, line 295: change "this might due" to "this might be due"

internals-js/src/extractSubgraphsFromSupergraph.ts
```diff
@@ -292,11 +292,11 @@ export function extractSubgraphsFromSupergraph(supergraph: Schema): Subgraphs {
...
+          + 'Please try composing your subgraphs with federation 2: this should help precisely pinpoint the problems and, once fixed, generate a correct federation 2 supergraph';
         throw new Error(`${msg}.\n\nDetails:\n${errorToString(e)}`);
       } else {
-        const msg = `Unexpected error extracting subgraph ${subgraph.name} from the supergraph: this is either a bug, or the supergraph has been corrupted.`;
+        const msg = `Unexpected error extracting subgraph ${subgraph.name} from the supergraph: this is either a bug, or the supergraph has been corrupted`;
```
cpeacock, line 299: what should folks do as a next step if they hit this error message?

benweatherman (approved):
🐐

cpeacock (changes requested):

internals-js/src/extractSubgraphsFromSupergraph.ts
pcmanus: Most likely open a ticket. The only reason I didn't wrote as much is that this can be throw due to someone manually messing up with the supergraph too, so I wanted to force users to switch their [system 2](https://en.wikipedia.org/wiki/Thinking,_Fast_and_Slow) before opening a ticket. But I don't mind being more explicit.

internals-js/src/extractSubgraphsFromSupergraph.ts
pcmanus: Done. Thanks for noticing.

internals-js/src/extractSubgraphsFromSupergraph.ts
cpeacock: wfm, that's about what I thought as well.  I'm cool with leaving this as is and if we need to change it later, we can.

cpeacock (approved):

PR #1592, opened 2022-03-14
Fix propagation of @tag to supergraph and related issues

pcmanus:
The @tag directive was properly merged into supergraphs due to some
recent changes to the merging of directives and this patch fixes this.
Proper testsing for @tag is also added to avoid regressions. Those
additional tests also highligted that the code wasn't declaring the
@tag directive as repeatable, limiting its uses, and this is fixed
as well.

Lastly, this patch also raises the question of applying @tag on
an external field, and by extension of any merged directives on
external fields. Before this patch, such directives were _not_ merged
but silently so. This patch changes the behavior to reject such
directives upfront as it feels better than ignoring them silently.

### DIFF
--- a/internals-js/src/federationSpec.ts
+++ b/internals-js/src/federationSpec.ts
@@ -62,6 +62,7 @@ export const shareableDirectiveSpec = createDirectiveSpecification({
 export const tagDirectiveSpec = createDirectiveSpecification({
   name:'tag',
   locations: tagLocations,
+  repeatable: true,
   argumentFct: (schema) => {
     return [{ name: 'name', type: new NonNullType(schema.stringType()) }];
   }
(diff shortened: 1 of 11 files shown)


internals-js/src/federationSpec.ts
```diff
@@ -62,6 +62,7 @@ export const shareableDirectiveSpec = createDirectiveSpecification({
 export const tagDirectiveSpec = createDirectiveSpecification({
   name:'tag',
   locations: tagLocations,
+  repeatable: true,
```
clenfest, line 65: It's probably ok for this case, but is there a situation where we'd want a directive to be repeatable on the subgraph spec, but not the supergraph?

subgraph-js/src/__tests__/printSubgraphSchema.test.ts
```diff
@@ -130,7 +130,7 @@ describe('printSubgraphSchema', () => {
 
       type User @key(fields: \\"id\\") @tag(name: \\"from-reviews\\") {
         id: ID!
-        username: String @external @tag(name: \\"on-external\\")
+        username: String @external
```
clenfest, line 133: Do we want another test to make sure that the old case errors?  

Edit: Never mind, I see the test in `compose.test.ts`

clenfest (approved):
Overall I like the approach that @external fields are essentially incompatible with other directives. I think I agree that this should basically always be the case.

PR #1600, opened 2022-03-15
Adds `@tag` definition automatically in `buildSubgraphSchema`

pcmanus:
This patch makes `@tag` behave like any other federation directive in
`subgraph-js`/`buildSubgraphSchema`, that is, its definition is
automatically added to the schema.

But because we want to be "backward compatible", this patch also
ensure that federation directive definitions are not added by
`buildSugraphSchema` if the input document already defined a directive
with that name. Do note that while `subrgaph-js` has no validation
that such user definition of a federation directive is "correct",
this is ok because composition does do such validation and will
reject an invalid definition.

A consequence of this patch is that the `@tag` definition is not
outputed by `printSubgraphSchema` (like for other federation
directives), but it's very doubtful any external user would have
relied on this behaviour in the first place (and it's very easy
to adapt in the off chance someone does).

### DIFF
--- a/subgraph-js/src/__tests__/buildSubgraphSchema.test.ts
+++ b/subgraph-js/src/__tests__/buildSubgraphSchema.test.ts
@@ -518,30 +518,31 @@ extend type User @key(fields: "email") {
 }
 `);
     });
+  });
 
-    describe('@tag directive', () => {
-      const query = `query GetServiceDetails {
-        _service {
-          sdl
+  describe('@tag directive', () => {
+    const query = `query GetServiceDetails {
+      _service {
+        sdl
+      }
+    }`;
+
+    const validateTag = async (header: string) => {
+      const schema = buildSubgraphSchema(gql`${header}
+        type User @key(fields: "email") @tag(name: "tagOnType") {
+          email: String @tag(name: "tagOnField")
         }
-      }`;
 
-      const validateTag = async (header: string) => {
-        const schema = buildSubgraphSchema(gql`${header}
-          type User @key(fields: "email") @tag(name: "tagOnType") {
-            email: String @tag(name: "tagOnField")
-          }
+        interface Thing @tag(name: "tagOnInterface") {
+          name: String
+        }
 
-          interface Thing @tag(name: "tagOnInterface") {
-            name: String
-          }
+        union UserButAUnion @tag(name: "tagOnUnion") = User
+      `);
 
-          union UserButAUnion @tag(name: "tagOnUnion") = User
-        `);
-
-        const { data, errors } = await graphql({ schema, source: query });
-        expect(errors).toBeUndefined();
-        expect((data?._service as any).sdl).toEqual(`${header}type User @key(fields: "email") @tag(name: "tagOnType") {
+      const { data, errors } = await graphql({ schema, source: query });
+      expect(errors).toBeUndefined();
+      expect((data?._service as any).sdl).toEqual(`${header}type User @key(fields: "email") @tag(name: "tagOnType") {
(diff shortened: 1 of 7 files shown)


subgraph-js/src/directives.ts
```diff
@@ -110,23 +110,13 @@ export const federationDirectives = [
...
   ShareableDirective,
   LinkDirective,
+  TagDirective,
 ];
 
 export function isFederationDirective(directive: GraphQLDirective): boolean {
   return federationDirectives.some(({ name }) => name === directive.name);
 }
 
-export const otherKnownDirectives = [TagDirective];
```
pcmanus, line 119: Quick note for reviewers: an option could have been to keep that notion of "known but not officially federation" directives but have it empty for now. But if anything, we're trying to move away from hard-coded directives and toward `@link`-defined directives, so I'm almost certain this would never be used again, so removed it (in fact, https://github.com/apollographql/federation/pull/1554 is probably going to change all of this before GA anyway).

subgraph-js/src/__tests__/buildSubgraphSchema.test.ts
```diff
@@ -518,6 +518,50 @@ extend type User @key(fields: "email") {
...
+
+interface Thing @tag(name: "tagOnInterface") {
+  name: String
+}
+
+union UserButAUnion @tag(name: "tagOnUnion") = User
+`);
+      };
+
+      it('adds it for fed1 schema', async () => {
```
benweatherman, line 556: *nitpick/question* We're saying that the tag directive is not added for fed1, right?

benweatherman (approved):
🐐

subgraph-js/src/buildSubgraphSchema.ts
```diff
@@ -30,6 +32,23 @@ type LegacySchemaModule = {
 
 export { GraphQLSchemaModule };
 
+function missingFederationDirectives(modules: GraphQLSchemaModule[]): GraphQLDirective[] {
+  // Copying the array as we're going to modify it.
+  const missingDirectives = federationDirectives.concat();
```
clenfest, line 37: May not matter too much, but it would probably be better to use a set to avoid the O(n) lookup in a loop.

clenfest (commented):
So I understand the individual pieces, but the reasons why this PR are necessary are a little fuzzy to me. Maybe worth spending a few minutes during the meeting today to give an overview.

subgraph-js/src/buildSubgraphSchema.ts
```diff
@@ -67,7 +86,7 @@ export function buildSubgraphSchema(
     modules,
     new GraphQLSchema({
       query: undefined,
-      directives: [...specifiedDirectives, ...federationDirectives],
+      directives: [...specifiedDirectives, ...missingFederationDirectives(modules)],
```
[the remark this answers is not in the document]
pcmanus: When we visit the AST, we only look at directive _definitions_, not application. And what we're adding is the directives definition that don't already have a definition in the user provided schema.

We're not looking at applications at all. Which does mean we may be adding `@provides` definition even though it's never used by that particular schema, and we could well avoid it, but `buildSubgraphSchema` has always done that, adding all federation directive definition regardless of what is being used (which is harmless), and so keeping the code change focused.

The problem we're solving with this is that we want the `@tag` definition to be added automatically, like all other federation directive definitions, but the difference is that we historically have asked fed1 user to provide the definition themselves. So to be backward compatible, we just need to ensure we don't try to auto-add definitions that are already user defined, which wasn't something `buildSubgraphSchema` was previously bothering with.

PR #1604, opened 2022-03-15
Ensure that we aren't in the same subgraph before throwing KEY_FIELDS_MISSING_ON_BASE error.

### LINKED ISSUE #1603: Overlap of `@key(fields: a { b })` and `a: A @provides(fields: "c")` results in composition error
Reproduction: https://stackblitz.com/edit/nodemon-86wmhe?file=README.md

**Subgraph ONE containing the "parent" entity**

```graphql
type Query {
  registry: ProductRegistry
}

type ProductRegistry @key(fields: "id status") {
  id: ID!
  status: String!
  dateAdded: String
}
```

**Subgraph TWO with the "child" entity**

The child entity doesn't have its own id. Instead it uses the id of the related "parent" entity.

However, this subgraph also can `@provide` another field on the "parent" entity.

```graphql
type Query {
  product: Product
}

type Product @key(fields: "registry { id }") {
  registry: ProductRegistry! @provides(fields: "dateAdded")
  otherField: String
}

type ProductRegistry @key(fields: "id status") @extends {
  id: ID! @external
  status: String! @external
  dateAdded: String @external
}
```

**The problem**

Using both the compound key and the `@provides` causes this error:

```
[TWO] Product -> A @key selects registry, but Product.registry was either created or overwritten by TWO, not TWO
```

**Versions**

```
"@apollo/gateway": "^0.48.3",
"@apollo/subgraph": "^0.3.3",
"apollo-server": "^3.6.4",
"graphql": "^16.3.0",
```

**Expected behavior**

This should compose (it does it Fed2).

clenfest:
Fixes #1603 

Ensure that we aren't in the same subgraph before throwing KEY_FIELDS_MISSING_ON_BASE error.

So this patch will fix the issue as framed in #1603, but I would definitely like another set of eyes to make sure there's nothing I'm not considering.


federation-js/src/composition/validate/postComposition/__tests__/keyFieldsMissingOnBase.test.ts
```diff
@@ -46,6 +46,46 @@ describe('keyFieldsMissingOnBase', () => {
...
+      `,
+      name: 'serviceA',
+    };
+
+    const serviceB = {
+      typeDefs: gql`
+        type ProductCatalog @key(fields: "products { sku }"){
+          products: [Product!]! @provides(fields: "price")
+        }
+        extend type Product {
```
trevor-scheer, line 73: Kind of a nit: I know this won't cause any validation errors as-is since we're isolating the validations in these test cases. But probably best if our test cases are mostly valid. I know it's like this above as well.
```suggestion
        extend type Product @key(fields: "sku") {
```

federation-js/src/composition/validate/postComposition/__tests__/keyFieldsMissingOnBase.test.ts
```diff
@@ -46,6 +46,46 @@ describe('keyFieldsMissingOnBase', () => {
     expect(warnings).toHaveLength(0);
   });
 
+  it('returns no warnings with proper @key usage for provided field', () => {
+    const serviceA = {
+      // FIXME: add second key "upc" when duplicate directives are supported
+      // i.e. @key(fields: "sku") @key(fields: "upc")
```
trevor-scheer, line 52: This is supported so probably either remove or implement the recommendation (I know this is just copied from above)

federation-js/src/composition/validate/postComposition/__tests__/keyFieldsMissingOnBase.test.ts
```diff
@@ -46,6 +46,46 @@ describe('keyFieldsMissingOnBase', () => {
...
+    };
+
+    const serviceB = {
+      typeDefs: gql`
+        type ProductCatalog @key(fields: "products { sku }"){
+          products: [Product!]! @provides(fields: "price")
+        }
+        extend type Product {
+          sku: String! @external
+          price: Int! @requires(fields: "sku")
```
trevor-scheer, line 75: I think the `@requires` here doesn't really apply to this test case / isn't necessary. Especially since it's a key field anyway.
```suggestion
          price: Int!
```

trevor-scheer (commented):
The change seems reasonable but I can't verify that it's safe/correct e2e without at least a test case showing a query plan that exercises this path correctly.

clenfest:
> The change seems reasonable but I can't verify that it's safe/correct e2e without at least a test case showing a query plan that exercises this path correctly.

I verified the query plan manually, but given that we're in the `composition` tests, I'm not sure if verifying the plan here is possible. I'll investigate.

clenfest:
> 

@pcmanus Do you have an opinion no the query planning part of the test? Am I right in thinking there isn't a reason to explicitly check?

pcmanus:
> Do you have an opinion no the query planning part of the test?

We do allow a new case that wasn't allowed before to pass validation, so I do agree it would be valuable to ensure the query planner does not error out on that new case it never had to run on before (and it generate the expected plan). 

That said, the prior art for such a test  would be to put it in `gateway-js/src/__tests__/buildQueryPlan.test.ts` (or `executeQueryPlan.test.ts` but I don't think that's necessary here).

With all that said, I "think" the fix is overall right but I'm not entirely sure I understand the full logic of the validation in the first place so only so confident on that.

clenfest:
> 

@trevor-scheer It seems like there's not really a straightforward way of adding a test for this. Do you want me to add the screenshot for the manual test and call it good?

trevor-scheer:
I think @pcmanus nailed what I was after. This change in validation means we now have a new "allowance" as for how a schema can be composed. I think it's important that query plans built against that new path are also correct. So a minimal / contrived example in `gateway-js/src/__tests__/buildQueryPlan.test.ts` of a query which exercises the new path within a previously-invalid schema would be good to have here.

I want to say this is probably ok and just ship it, but there's just a lot of complexity in the query planner that I'd rather not gloss over and assume it's going to work as we expect it. It's arguably more important that we ensure nothing breaks this functionality it in the future.

federation-js/src/composition/validate/postComposition/__tests__/keyFieldsMissingOnBase.test.ts
```diff
@@ -46,6 +46,44 @@ describe('keyFieldsMissingOnBase', () => {
...
+          value: String!
+        }
+      `,
+      name: 'serviceA',
+    };
+
+    const serviceB = {
+      typeDefs: gql`
+        type ProductCatalog @key(fields: "products { sku }"){
+          products: [Product!]! @provides(fields: "price")
```
trevor-scheer, line 69: `price` should exist in `serviceA` for this to be valid

trevor-scheer:
Confirmed the most recent 2 commits with updates to the fixtures and `buildQueryPlan.test.ts` do fail composition (without the implemented fix). This LGTM (barring the one minor, unrelated-ish comment on the test).

PR #1609, opened 2022-03-15
feat(query-planner): Include `operationKind` and `operationName` in `FetchNode`s (backport)

trevor-scheer:
Backport of:
#1427
#1550 + #1568

This maintains parity in the query plan format with v2 by generating operation names for subgraph fetches. This also brings in the addition of `operationKind` out of convenience / maintaining parity since the changes are so closely related.

cc @chrskrchr who I know was interested in this addition for v0.x and @lennyburdette as the implementor of the name generation


trevor-scheer:
Thanks @pcmanus, good catch (I had missed those tests altogether during my cherry-pick). Needs an actual approval then I'll land!

pcmanus (approved):
Lgtm. But I'd make sure we don't forget a changelog entry that mentions that queries sent to subgraph can now be named since it's a small but visible-ish change (I could see a user have tests that needs update due to this).

clenfest (approved):

PR #1614, opened 2022-03-16
Whittle down release steps

benweatherman:
This lays out a terse set of steps for creating a release, assuming all packages get set to a specific version. It's intended to be a short-term solution to quickly make fed2.x releases.

Other detritus
- Link directly to PR with correct title & body
- Remove now unneeded script/dependencies
- Stick with the `latest-2` dist tag for now
- Add `--no-push` to versioning instructions to require a manual push to make changes


DEVELOPMENT.md
```diff
@@ -102,7 +102,7 @@ DEBUG=lerna NPM_CONFIG_USERCONFIG="$HOME/.npmrc-apollo-bot" npx lerna publish fr
 
 ### Step 4: Update GitHub Milestones
 
-The [milestones](./milestones) should be updated, as appropriatee.  If the milestone didn't didn't have a concrete version as its name, it should be renamed to the version that was finally released.
+The [milestones](./milestones) should be updated, as appropriate. If the milestone didn't didn't have a concrete version as its name, it should be renamed to the version that was finally released.
```
trevor-scheer, line 105:
```suggestion
The [milestones](./milestones) should be updated, as appropriate. If the milestone didn't have a concrete version as its name, it should be renamed to the version that was finally released.
```

RELEASING.md
```diff
@@ -0,0 +1,60 @@
...
+#### Start a release PR
+
+1. `FEDERATION_RELEASE_VERSION=X.Y.Z` (e.g. `2.0.1` or `2.0.0-beta.3` or `2.0.0-preview.99`)
+1. `git fetch origin main:main` to fetch the latest `main` branch
+1. `git checkout -b "release-$FEDERATION_RELEASE_VERSION" main`
+1. Update `CHANGELOG.md` in each repo ([more info](#changelogs))
+1. `git add ':/*CHANGELOG.md'` to add only changelogs to the commit
+1. `git commit --message "Update CHANGELOGs for v$FEDERATION_RELEASE_VERSION release"`
+1. `git push -u origin "release-$FEDERATION_RELEASE_VERSION"`
+1. `open "https://github.com/apollographql/federation/compare/main...release-$FEDERATION_RELEASE_VERSION?quick_pull=1&title=Release+$FEDERATION_RELEASE_VERSION&template=APOLLO_RELEASE_TEMPLATE.md"` to open PR in Github
```
trevor-scheer, line 14: I really like this. Can it exist as a shell script in the repo?

RELEASING.md
```diff
@@ -0,0 +1,60 @@
...
+
+#### Bump version and tag release
+
+1. `npx lerna version --no-push --force-publish=\* "$FEDERATION_RELEASE_VERSION"` ([more info](#lerna-version))
```
trevor-scheer, line 18: We've force published in the past in order to lockstep versioning of related packages but moved away from that awhile ago to reduce noise. I think there are good reasons for both but mostly just want to acknowledge this change. Is this something we'll continue to do after v2 becomes `latest`?

RELEASING.md
```diff
@@ -0,0 +1,60 @@
...
+    1. `git show --pretty="" --name-only HEAD` to see the files in the commit
+    1. `git --no-pager tag --points-at HEAD` to see the tags that were created
+1. `git push --follow-tags origin` to push the version bumps & tags created by lerna in the previous step
+1. `APOLLO_DIST_TAG=next npm run release:start-ci-publish` ([more info](#publishing))
+1. `open https://app.circleci.com/pipelines/github/apollographql/federation?filter=mine` to approve publishing to NPM
+    1. There will also be a message posted to #team-atlas in slack that has a link to the approval job
+1. `npm view @apollo/subgraph versions` to ensure the version that was just published shows up in the NPM registry
```
trevor-scheer, line 25: This command is helpful but maybe overly specific? It's at least pasteable and not hard to figure out how to validate the rest of them, so I think it's fine but I'm on the fence about it.

`npm view @apollo/subgraph@next` might also be an interesting variation with a similar result.

RELEASING.md
```diff
@@ -0,0 +1,60 @@
...
+
+![celebrate](https://media.giphy.com/media/LZElUsjl1Bu6c/giphy.gif)
+
+---
+
+## Changelogs
+
+Changelogs should be updated when creating a release. If there are no changes for a given package, add an entry that states `Released in sync with other federation packages but no changes to this package.`
+
+1. [gateway-js/CHANGELOG.md](gateway-js/CHANGELOG.md) all changes should be reflected in here, since it's essentially a rollup of all packages
```
trevor-scheer, line 39: I'm conflicted on this one but I don't disagree. This is the most interesting changelog to users. There are probably changes to internals, etc. which don't need to roll up to this changelog though, so maybe per discretion?

Also `subgraph-js` is an exception to this.

RELEASING.md
```diff
@@ -0,0 +1,60 @@
...
+1. [composition-js/CHANGELOG.md](composition-js/CHANGELOG.md)
+1. [subgraph-js/CHANGELOG.md](subgraph-js/CHANGELOG.md)
+1. [query-planner-js/CHANGELOG.md](query-planner-js/CHANGELOG.md)
+1. [internals-js/CHANGELOG.md](internals-js/CHANGELOG.md)
+1. [query-graphs-js/CHANGELOG.md](query-graphs-js/CHANGELOG.md)
+
+## `lerna version`
+
+Learn more about the [`lerna version` command](https://github.com/lerna/lerna/tree/main/commands/version). For our purposes, the important part is that `lerna` will create a tag for each package (e.g. `@apollo/federation-internals@2.0.0-preview.7`).
```
trevor-scheer, line 48: And update all of the `package*.json` files, yeah?

RELEASING.md
```diff
@@ -0,0 +1,60 @@
...
+
+#### Why does `package-lock.json` get updated?
+
+`npm version` is getting called due to [lerna version lifecycle hooks](https://github.com/lerna/lerna/tree/main/commands/version#lifecycle-scripts), which calls `npm i --package-lock-only` causing the package lock to get regenerated (this is a good thing).
+
+## Publishing
+
+The important part is creating a `publish/*` tag, which starts a job in CircleCI to start the publishing process.
```
trevor-scheer, line 56: It's a little confusing to read this, maybe worth mentioning this is handled by the `release:start-ci-publish` script and not something that needs to be done manually.

package.json
```diff
@@ -8,8 +8,7 @@
     "compile": "tsc --build tsconfig.build.json",
     "compile:clean": "npm run compile -- --clean",
     "watch": "npm run compile && tsc --build tsconfig.json --watch",
-    "version": "npm i --package-lock-only && node ./prompt-update-changelogs.js",
-    "release:version-bump": "lerna version",
+    "version": "npm i --package-lock-only && git add package-lock.json",
```
trevor-scheer, line 11: I added this prompt awhile back. I totally don't mind seeing it be removed, but the main advantage here was landing CHANGELOG updates inside the same release commit that lerna made which I found to be nice. This can only be done during the `version` step (as you've noticed, Lerna calls this script as part of the lifecycle).

It never was a big deal - we did tons of releases with them being separate commits, but just pointing it out in case you think "oh yeah I actually prefer that".

package.json
```diff
@@ -8,8 +8,7 @@
     "compile": "tsc --build tsconfig.build.json",
     "compile:clean": "npm run compile -- --clean",
     "watch": "npm run compile && tsc --build tsconfig.json --watch",
-    "version": "npm i --package-lock-only && node ./prompt-update-changelogs.js",
-    "release:version-bump": "lerna version",
```
trevor-scheer, line 12: This might seem redundant or extraneous but generally speaking I think npm scripts tell a lot of things about how to work within any particular repo. Having this here is a form of documentation that I consider to be valuable.

trevor-scheer (approved):
Generally fine with the direction this is headed. Most of my feedback within is dismissible - either adding some context or sharing an opinion which you're welcome to disagree with!

I don't think it's mandatory but it would be nice to keep the process aligned in the `version-0.x` branch as well.

RELEASING.md
benweatherman: Absosmurfly

RELEASING.md
benweatherman: I appreciate the extra context. I don't feel strongly about it.

I would say that it made it easy to think about and deploy if everything is at the same version. It felt kinda awkward to say we're getting ready for a `preview.3` release, even though some of the components were at some other version.

Though I think the noise aspect could get overwhelming downstream. Happy to take out the `force-publish` and see how that goes.

RELEASING.md
benweatherman: Are you suggesting taking it out? Or adding the other packages in there to also check? Or something else? I agree that using the `@next` dist tag would be a good check.

Doing _something_ here was helpful for me as a very new deployer to make sure it did what I wanted.

RELEASING.md
benweatherman: You're right, this was an overly broad generalization to `DEVELOPMENT.md`.

Though it seems like that doc says we should especially be rolling up `subgraph-js`

https://github.com/apollographql/federation/blob/0fb2212b83d46110afc5cff8f82be65458e5742d/DEVELOPMENT.md?plain=1#L35-L37

RELEASING.md
benweatherman: Oh yes, great point. I was very much in the "how does the publishing magic work?" mindset. I'll clean this up.

RELEASING.md
benweatherman: Good call. I should make this a footnote or add more context here.

package.json
benweatherman: I do like landing those changelogs in the same commit. I think I was a little confused being prompted to make CHANGELOG changes after the instructions in `DEVELOPMENT.md` said to open a PR with the CHANGELOG changes.

I don't feel strongly about it, as long as we can keep the instructions fairly straightforward.

package.json
benweatherman: Good point, I'll add it back.

For context, I initially removed this because I ran this command and hit `ctrl-c` when being prompted to commit changelogs and it still finished the release. It was a footgun that I wanted to remove, but I believe it might've been my misunderstanding how the `prompt-update-changelogs.js` script worked (and maybe that script would need to account for `ctrl-c`).

trevor-scheer: Oh, this comment was directed at the `release:version-bump` script. I think it's easy to say "yeah you just run `npx lerna version`" but it's nice to have that written out plainly in the scripts which to me says "this is a regular thing we do often".

package.json
trevor-scheer: Ah, that's on me for not updating the instructions then. Very open to other approaches here as well if you've got anything in mind. My only real motivation here was to afford the user some time during the `version` hook where the CHANGELOGs could be updated.

RELEASING.md
trevor-scheer: This one's on me. `@apollo/subgraph` was once `@apollo/federation` which included both subgraph and composition related bits. It's the composition bits which used to be a part of the gateway (though that's also barely true anymore). I think this note can be removed and I'd standby `@apollo/subgraph` being independent of the others. With fed2 GA it might even make sense to separate this package from the monorepo...maybe.

RELEASING.md
trevor-scheer: Sorry for the ambiguity, I'm not sure I have a preference. I agree something should be here, I just dislike all the options equally so I'd be equally disappointed to see it:
* as-is
* `npm view @apollo/<package-name>@next`
* with all of them listed

I think the pasteability aspect might be worth preserving which would rule out my second option.

RELEASING.md
benweatherman: To add to the lock-step version bumping camp, it seems like it was mostly already discussed https://apollographql.quip.com/BzkQAXFw2niK/045-Supergraph-Versioning-for-Fed-2-PreviewGA. The way I read that, is that we'll use the same version across components.

I also like the thought of potentially consolidating a few packages.

RELEASING.md
benweatherman: All roads lead to Trevor's sadness, noted 😸

RELEASING.md
benweatherman: ...but I'll do it as part of another set of changes https://github.com/apollographql/federation/issues/1642

Lemme know if that seems OK to you

RELEASING.md
benweatherman: Made this into a footnote, lemme know if that makes it better

package.json
benweatherman: I removed the prompt for now, we'll see how that goes

package.json
benweatherman: Added back `release:version-bump`

RELEASING.md
benweatherman: Lemme know what you think of the updated info in a footnote

RELEASING.md
benweatherman: Added a few options, to ensure your complete misery

benweatherman:
> I don't think it's mandatory but it would be nice to keep the process aligned in the `version-0.x` branch as well.

Good idea! https://github.com/apollographql/federation/issues/1643

RELEASING.md
```diff
@@ -0,0 +1,63 @@
...
+1. `git add ':/*CHANGELOG.md'` to add only changelogs to the commit
+1. `git commit --message "Update CHANGELOGs for v$FEDERATION_RELEASE_VERSION release"`
+1. `git push -u origin "release-$FEDERATION_RELEASE_VERSION"`
+1. `open "https://github.com/apollographql/federation/compare/main...release-$FEDERATION_RELEASE_VERSION?quick_pull=1&title=Release+$FEDERATION_RELEASE_VERSION&template=APOLLO_RELEASE_TEMPLATE.md"` to open PR in Github
+
+#### Bump version and tag release
+
+1. `npx lerna version --no-push --force-publish=\* "$FEDERATION_RELEASE_VERSION"`[^lerna-version]
+    - `git show --pretty="" --name-only HEAD` to see the files in the commit
+    - `git --no-pager tag --points-at HEAD` to see the tags that were created
```
abernix, line 20: This is helpful debug information!

RELEASING.md
```diff
@@ -0,0 +1,63 @@
...
+1. `git checkout -b "release-$FEDERATION_RELEASE_VERSION" main`
+1. Update `CHANGELOG.md` in each repo ([more info](#changelogs))
+1. `git add ':/*CHANGELOG.md'` to add only changelogs to the commit
+1. `git commit --message "Update CHANGELOGs for v$FEDERATION_RELEASE_VERSION release"`
+1. `git push -u origin "release-$FEDERATION_RELEASE_VERSION"`
+1. `open "https://github.com/apollographql/federation/compare/main...release-$FEDERATION_RELEASE_VERSION?quick_pull=1&title=Release+$FEDERATION_RELEASE_VERSION&template=APOLLO_RELEASE_TEMPLATE.md"` to open PR in Github
+
+#### Bump version and tag release
+
+1. `npx lerna version --no-push --force-publish=\* "$FEDERATION_RELEASE_VERSION"`[^lerna-version]
```
abernix, line 18: I'm wondering if we wanted to switch the `lerna` command (here, and in a few other places in the content) back to the `release:version-bump` command in the `package.json`, now that it's being re-introduced though?

Going closer to the bare-metal commands seems a bit backward to the spirit of whittling down to me though I understand it's uncovering true mechanisms.    And maybe this is more compatible with replacing this all entirely?

RELEASING.md
```diff
@@ -0,0 +1,63 @@
...
+    - `git show --pretty="" --name-only HEAD` to see the files in the commit
+    - `git --no-pager tag --points-at HEAD` to see the tags that were created
+1. `git push --follow-tags origin` to push the version bumps & tags created by lerna in the previous step
```
abernix, line 21: Why is this **now** necessary? 🤔

RELEASING.md
```diff
@@ -0,0 +1,63 @@
...
+1. `APOLLO_DIST_TAG=next npm run release:start-ci-publish`[^publishing]
+1. `open https://app.circleci.com/pipelines/github/apollographql/federation?filter=mine` to approve publishing to NPM
```
abernix, line 23: Same `open` comment

abernix (commented):
On the whole, this looks great! I just found I had this open still when going through some tabs! I have a few comments throughout — which I'm submitting as merely that with no 👍  or 👎  and no objections and then I'm going to be 🌴 !

.github/PULL_REQUEST_TEMPLATE/APOLLO_RELEASE_TEMPLATE.md
```diff
@@ -1,6 +1,4 @@
-Release @apollo/PACKAGE1@X.Y.Z [, @apollo/PACKAGE2@X.Y.Z]
-
-As with [release PRs in the past](https://github.com/apollographql/federation/issues?q=label%3A%22:label:%20release%22+is%3Aclosed), this is a PR tracking a `release-x.y.z` branch for an upcoming release. 🙌   The version in the title of this PR should correspond to the appropriate branch.
```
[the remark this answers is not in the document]
benweatherman: I appreciate the extra context 🙏 

I think it makes sense to add a `gh` command in the instructions. I'll likely keep the web-page link version as well, since I'd like to keep the deployer required software to a minimum to encourage more people to create releases.

internals-js/CHANGELOG.md
```diff
@@ -1,11 +1,5 @@
 # CHANGELOG for `@apollo/federation-internals`
 
-## vNEXT
-
```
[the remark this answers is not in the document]
benweatherman: Nope, great catch!

RELEASING.md
benweatherman: `echo` makes sense, as well as adding in some commands for `gh`

RELEASING.md
benweatherman: I added the `--no-push` flag above, which would allow us to catch weird things that `lerna` does or doesn't do. `lerna` does some kinda dangerous things and it seemed less anxiety-inducing to have more control over when those things are done.

RELEASING.md
benweatherman: My thoughts being new to everything: To me, it was much easier to figure out and change the process once I understood the song+dance between npm and lerna. There were a lot of things that felt like I had to uncover, and making it obvious we were using lerna was helpful.

My inclination is def based on a test size of 1, happy to change these to whatever will make deploys easiest for everyone.

package.json
```diff
@@ -8,7 +8,7 @@
     "compile": "tsc --build tsconfig.build.json",
     "compile:clean": "npm run compile -- --clean",
     "watch": "npm run compile && tsc --build tsconfig.json --watch",
-    "version": "npm i --package-lock-only && node ./prompt-update-changelogs.js",
+    "version": "npm i --package-lock-only && git add package-lock.json",
```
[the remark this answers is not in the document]
benweatherman: I'm a little stumped as well. Reading through the docs, I would've been surprised if it adds the root `package-lock.json` because lerna is not expecting any changes in `package.json` since we use independent versioning. There are several lerna issues that seems like adding `package-lock.json` for the root could've been considered a bug, so I wonder if this was unintended behavior that got fixed along the way.

.github/PULL_REQUEST_TEMPLATE/APOLLO_RELEASE_TEMPLATE.md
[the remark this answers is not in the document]
benweatherman: Added in https://github.com/apollographql/federation/pull/1614/commits/15288bfcd7d4e06f8466a8edc71d331e35c639e8. Lemme know what you think

subgraph-js/CHANGELOG.md
```diff
@@ -2,10 +2,6 @@
 
 This CHANGELOG pertains only to Apollo Federation packages in the 2.x range. The Federation v0.x equivalent for this package can be found [here](https://github.com/apollographql/federation/blob/version-0.x/subgraph-js/CHANGELOG.md) on the `version-0.x` branch of this repo.
 
-## vNEXT
```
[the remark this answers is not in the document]
benweatherman: Makes sense, I added them back

PR #1615, opened 2022-03-16
Improve validation and error messages around @link and federation directives

pcmanus:
Validation is currently pretty light around the new `@link(import:)` argument and it's easy to run into unhelpful error is you try to pass invalid values. There is also no validation of the imported name, so if you try to import a directive/name that doesn't exists, you get no errors and it's very easy to miss a typo for instance. But also, users coming from fed1 where name imports are not a thing are likely to forgot to add the directive they want to use to `import` a bunch of time initially, and (before this patch) you get a fairly laconic error message like `Unknown directive "@key"` (and if you're brand new to `@link`, this may not click as "I forgot to import" right away).

Anyway, this commit tries to improve the user experience around all this by:
1. providing suggestions if you typo a directive name
2. providing additional guidance is you try to use a fed2 directive but it's a fed1 schema, or you haven't imported it.
3. properly validating the `@link(import:)` argument
4. for known features (mostly matter for the `federation` one right now, but the code handles all "known" features), it ensures the imported names are known name for that feature.

### DIFF
--- a/internals-js/src/buildSchema.ts
+++ b/internals-js/src/buildSchema.ts
@@ -46,7 +46,8 @@ import {
   UnionType,
   InputObjectType,
   EnumType,
-  Extension
+  Extension,
+  ErrGraphQLValidationFailed,
 } from "./definitions";
 
 function buildValue(value?: ValueNode): any {
@@ -66,6 +67,7 @@ export function buildSchemaFromAST(
   documentNode: DocumentNode,
   options?: BuildSchemaOptions,
 ): Schema {
+  const errors: GraphQLError[] = [];
   const schema = new Schema(options?.blueprint);
   // We do a first pass to add all empty types and directives definition. This ensure any reference on one of
   // those can be resolved in the 2nd pass, regardless of the order of the definitions in the AST.
@@ -77,13 +79,13 @@ export function buildSchemaFromAST(
   // populated (and at this point, we don't really know the name of the `@core` directive since it can be renamed, so
   // we just handle all directives).
   for (const directiveDefinitionNode of directiveDefinitions) {
-    buildDirectiveDefinitionInner(directiveDefinitionNode, schema.directive(directiveDefinitionNode.name.value)!);
+    buildDirectiveDefinitionInner(directiveDefinitionNode, schema.directive(directiveDefinitionNode.name.value)!, errors);
   }
   for (const schemaDefinition of schemaDefinitions) {
-    buildSchemaDefinitionInner(schemaDefinition, schema.schemaDefinition);
+    buildSchemaDefinitionInner(schemaDefinition, schema.schemaDefinition, errors);
   }
   for (const schemaExtension of schemaExtensions) {
-    buildSchemaDefinitionInner(schemaExtension, schema.schemaDefinition, schema.schemaDefinition.newExtension());
+    buildSchemaDefinitionInner(schemaExtension, schema.schemaDefinition, errors, schema.schemaDefinition.newExtension());
   }
 
   schema.blueprint.onDirectiveDefinitionAndSchemaParsed(schema);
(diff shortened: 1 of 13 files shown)


internals-js/src/buildSchema.ts
```diff
@@ -312,41 +341,47 @@ function buildNamedTypeInner(
...
+    buildArgumentDefinitionInner(inputValueDef, field.addArgument(inputValueDef.name.value), errors);
   }
-  buildAppliedDirectives(fieldNode, field);
+  buildAppliedDirectives(fieldNode, field, errors);
   field.description = fieldNode.description?.value;
   field.sourceAST = fieldNode;
 }
 
-function ensureOutputType(type: Type, what: string, node: ASTNode): OutputType {
+function ensureOutputType(type: Type, what: string, node: ASTNode, errors: GraphQLError[]): OutputType | undefined {
```
benweatherman, line 370: **observation** Personally, I prefer the `ensureYYYY` methods that throw, since it feels what I'd expect to happen as a caller. I don't love having to add an `error` input to everything. I'd expect the caller to do the error array aggregation, but maybe that's a bit too cumbersome for these cases?

internals-js/src/coreSpec.ts
```diff
@@ -144,33 +149,129 @@ export type CoreImport = {
...
+  // either a string or a valid `CoreImport` value.
+  const { imports, errors } = validateCoreFeatureImports(url, directive, args.import);
   for (const i of imports) {
     if (!i.as) {
       continue;
     }
     if (i.name.charAt(0) === '@' && i.as.charAt(0) !== '@') {
-      throw ERRORS.INVALID_LINK_DIRECTIVE_USAGE.err({
-        message: `Invalid @link import renaming: directive ${i.name} imported name should starts with a '@' character, but got "${i.as}"`,
+      errors.push(ERRORS.INVALID_LINK_DIRECTIVE_USAGE.err({
```
benweatherman, line 166: **question** Is there a reason these checks are in this method vs `validateCoreFeatureImports`?

benweatherman (approved):
🐐 

Seems like a step in the right direction to surface better errors to users. I don't love passing `errors` around to so many places, though I'm not sure how big of a consideration that is at this stage of development.

internals-js/src/buildSchema.ts
pcmanus: I'd preface by saying that I agree this probably won't and shouldn't win any code beauty contest.

That said, pragmatically, I'll point out that it's all private methods and that having throw-catch everywhere would imo be less readable here (and more generally, I feel throw-catch doesn't always land itself super well to error aggregations due to interrupting the control flow while error aggregation is about collecting errors but preserving the normal flow; but of course, in practice, "it depends").

I'd agree though that as we evolve such file, there might come a point where we'd want to refactor things a bit more, but it felt a bit more effort than is worth at this stage. Definitively a judgment call :).

internals-js/src/coreSpec.ts
pcmanus: Yeah, not really. I initially split `validateCoreFeatureImport` to only do the "syntax" validation, but it ended up doing more than that and so I agree having those validation at a different place is now just weird. I've merge those 2 methods together (once you move those checks in `validateCoreFeatureImports`, not much logic remains so ...).

internals-js/src/buildSchema.ts
benweatherman: That makes sense to me, thanks for writing that out.

I'm starting to see that `try`/`catch` doesn't make sense for these cases. Mostly I wanted to point it out as @queerviolet had the same need to aggregate errors without interrupting control flow. I was hoping we could start to establish some patterns for that case, since I think y'all are solving it in slightly different ways (though everyone is agreeing to stay away from `try`/`catch` 😄 ).

PR #1617, opened 2022-03-16
Update the code block tags with proper syntax

smyrick:
This might be from some large docs migrations but the triple backtick syntax needs to just have ` ```graphql ` with no extra text after it. 

I don't think this was used for any linking in the document but if I am incorrect and need to add the text `{2}` somewhere, let me know.

You can see live on the docs now the in-proper and proper syntax highlighting

https://www.apollographql.com/docs/federation/v2/entities-advanced#advanced-keys

Screen Shot 2022-03-16 at 11 41 14 AM

### DIFF
--- a/docs/source/entities-advanced.mdx
+++ b/docs/source/entities-advanced.mdx
@@ -14,7 +14,7 @@ You can define more than one `@key` for an entity, when applicable.
 
 In this example, a `Product` entity can be uniquely identified by either its `id` _or_ its `sku`:
 
-```graphql{2}
+```graphql
 # Products subgraph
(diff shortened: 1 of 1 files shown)


trevor-scheer:
cc @blades @StephenBarlow I believe this was for code highlighting and presenting a file name. Possibly something broken in the plugin we're using?

trevorblades:
We just need to add a space between the language tag and the line highlighting curly braces! I documented this change in the docs readme here: https://github.com/apollographql/docs#code-block-titles-and-line-highlighting

Hope that helps clear things up!

PR #1627, opened 2022-03-21
Update the subgraph spec to be more explicit about `_service.sdl`

martinbonnin:
I'm not sure what branch to target ? I came here from https://www.apollographql.com/docs/federation/federation-spec/

### DIFF
--- a/docs/source/federation-spec.md
+++ b/docs/source/federation-spec.md
@@ -51,7 +51,7 @@ Schema composition at the gateway requires having each service's schema, annotat
 
 > Note that the `_service` field is not exposed by the gateway; it is solely for internal use.
 
-The `_service` resolver should return the `_Service` type which has a single field called `sdl`. This SDL (schema definition language) is a printed version of the service's schema **including** the annotations of federation directives. This SDL does **not** include the additions of the federation spec above. Given an input like this:
+The `_service` resolver should return the `_Service` type which has a single field called `sdl`. This SDL (schema definition language) is a printed version of the service's schema **including** the federation directives usages (like `@key` or `@external` usages). This SDL **must not** include the federation directive definitions (like  `directive @external on FIELD_DEFINITION`), type definitions (like the `_Entity` or `_Any` type definitions) or federation fields (like `_service` or `_entities`). Given an input like this:
 
 ```graphql
 extend type Query {


docs/source/federation-spec.md
```diff
@@ -51,7 +51,7 @@ Schema composition at the gateway requires having each service's schema, annotat
 
 > Note that the `_service` field is not exposed by the gateway; it is solely for internal use.
 
-The `_service` resolver should return the `_Service` type which has a single field called `sdl`. This SDL (schema definition language) is a printed version of the service's schema **including** the annotations of federation directives. This SDL does **not** include the additions of the federation spec above. Given an input like this:
+The `_service` resolver should return the `_Service` type which has a single field called `sdl`. This SDL (schema definition language) is a printed version of the service's schema **including** the federation directives usages (like `@key` or `@external` usages). This SDL **must not** include the federation directive definitions (like  `directive @external on FIELD_DEFINITION`), type definitions (like the `_Entity` or `_Any` type definitions) or federation fields (like `_service` or `_entities`). Given an input like this:
```
pcmanus, line 54: I do agree that the spec could use being a bit more precise, but I'm also conflicted about the changes, because in reality, at least for federation 2 (but I think it's mostly also true for fed 1), the composition code (which I think is the main consumer of this output?) is totally happy if that SDL contains federation directive definitions and type definitions and/or the `_service` and `_entities` fields. And it actually feel *better* to have subgraphs sends the full SDL with none of those filtered because:
1. I suspect it's actually _easier_ for most implementations to include those defs/fields than to filter them out (it feels like you're forced to have special printing code to filter them, while you can just use normal graphQL printing on your actual subgraph SDL otherwise).
2. (relatedly) the SDL _without_ those defs/fields is guaranteed invalid graphQL while it may just be valid graphQL if you keep them ("may" because in fed1 most subgraph will use type extensions in a way that is not strictly valid graphQL). And I'm not sure why we would _insist_ on `_service.sdl` having to be invalid graphQL.

So my own inclination would be to make the spec a bit more lenient and let subgraphs send the full SDL if that's more convenient, and even maybe encourage that?

But with all that said, I'll note that there was probably a reason why this was initially done the way it is and I don't know those reasons, so it could be those initial reason simply don't apply anymore, but it also could be that there is usages of `_service.sdl` I don't know about.

Of course, there is also the fact that `_service.sdl` probably has a limited shelf life at this point (afaict, it's not really used for managed federation, and we'll probably stop needing it entirely once graphQL introspection has a way to surface directive applications), so offering those though but I'm happy to not overthink this. It just bother me to make subgraph server authors harder than it needs to be.

@martijnwalraven: opinions? You might have some insights into the history of this in particular?

martinbonnin: > So my own inclination would be to make the spec a bit more lenient and let subgraphs send the full SDL if that's more convenient, and even maybe encourage that?

I'm all for that 🙌 . Havid valid GraphQL sounds better if composition code can deal with it.

martijnwalraven: @pcmanus I agree with everything you wrote there, and I honestly have no idea why we filtered those out in the first place…

trevor-scheer: I'll just note that `_service.sdl` probably shouldn't go away entirely, though maybe it becomes optional. I agree its usefulness is pretty limited within our own tooling since we use SDL throughout, but for code-first subgraphs we do need a way to extract the SDL at runtime. On the other hand, maybe it doesn't need to be in the spec and is instead "here's one approach commonly taken for extracting your SDL at runtime" like a documented recipe.

martinbonnin: @trevor-scheer I think what @pcmanus was pointing at is that ultimately `_service.sdl` could be supported by introspection. We can't today because introspection would just remove all `@keys` but once (if?) introspection exposes directive usages (https://github.com/graphql/graphql-spec/issues/300) then we won't need `_service.sdl` any more?

martinbonnin:
2025 cleanup

PR #1635, opened 2022-03-22
Expand fed2 opt-in snippet

StephenBarlow:
EDU folks pointed out that we need a couple more directives imported in this `@link` for current fed2 completeness, yeah?

### DIFF
--- a/docs/source/federated-types/overview.mdx
+++ b/docs/source/federated-types/overview.mdx
@@ -49,7 +49,7 @@ type User @key(fields: "id") {
 # Federation 2 features.)
 extend schema
     @link(url: "https://specs.apollo.dev/federation/v2.0",
-          import: ["@key", "@requires", "@shareable"])
+          import: ["@key", "@requires", "@shareable", "@provides", "@external"])
 ```
 
 ```graphql title="Products"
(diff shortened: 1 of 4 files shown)


pcmanus (approved):
The change looks good and happy to get it merged as is.

But let me add some context around this and suggest a potential future evolution.
The list of names we have to those `import` is starting to grow. And contract users would actually want to add `@tag` in there as well. And soon enough, we'll add `@inaccessible` and `@override` to that list. Which get uncomfortably long.

And as the name implies, I think users should think of the `import` argument much like you think about imports in programming language. And in other languages, importing stuff you don't use is usually frown upon, so it feels weird to me to recommend it here. Also, some of those directives are more advanced than others, so in term of the new-user-documentation-experience, it feels listing the directives only as they are introduced would avoid distracting with stuffs you don't need to worry about yet.

So my proposition would be to move toward recommending user only list the directives they actually use, and to do the same in the documentation. The only slight risk doing so is users forgetting to import some directive they use and wondering why it doesn't work, but https://github.com/apollographql/federation/pull/1615 hopefully fixes that risk by ensuring there is a pretty clear error message in those cases.

If we do so, we may want to have a few sentence of explanation around how `import` should be used however.

StephenBarlow:
Great point @pcmanus. I'm going to merge this for now and create a dedicated article on "federated directives" that explains what all of these do in one place, and explains the importing methodology at the top.

PR #1637, opened 2022-03-23
Fix merging of renamed @tag directive ...

pcmanus:
... but requires that it is renamed consistently in all subgraphs.


composition-js/src/merging/merge.ts
```diff
@@ -161,25 +163,15 @@ function isGraphQLBuiltInDirective(def: DirectiveDefinition): boolean {
...
-    // don't need to be defined.
-    return false;
-  }
-  return definition.locations.some(loc => executableDirectiveLocations.includes(loc));
+type MergedDirectiveInfo = {
+  specInSupergraph: FeatureDefinition,
+  definitionInSubgraph: (subgraph: Subgraph) => DirectiveDefinition,
 }
+
+const MERGED_FEDERATION_DIRECTIVES: MergedDirectiveInfo[] = [
```
pcmanus, line 171: If someone wonders why this patch is kind of over-generic despite only have 1 merged directive, that's because `@inaccessible` will be add to this list shortly.

clenfest (approved):
I question a bit whether this needs to be an error in the long term or if we can just do something smart, but I guess it could lead to confusion when examinging the supergraph if naming is inconsistent. Otherwise LGTM.

pcmanus:
> I question a bit whether this needs to be an error in the long term or if we can just do something smart

And for the record, I agree with that questioning. This PR is meant to be a stop-gap until we figure out the full rules we want for merging core directives.

PR #1651, opened 2022-03-25
Docs additions for GA

StephenBarlow:
Following up from the last PR that expanded `extend schema` snippets, this should help clarify things


docs/source/federated-types/federated-directives.mdx
```diff
@@ -0,0 +1,240 @@
...
+<tr class="required">
+<td>
+
+##### `fields`
+
+`String`
+</td>
+<td>
+
+**Required.** A comma-separated list of field names that contribute to the entity's primary key.
```
pcmanus, line 66: That's incorrect. Most importantly because it's *not* comma-separated. But there is also the fact that you  can have sub-selections.

Technically, `fields` is syntactically parsed like a graphQL [selection set](https://spec.graphql.org/draft/#sec-Selection-Sets) but without the top-level braces (even if in practice there is limitations to what is accepted), so that might the most concise-and-correct way to describe it.

docs/source/federated-types/federated-directives.mdx
```diff
@@ -0,0 +1,240 @@
...
+<tr class="required">
+<td>
+
+##### `fields`
+
+`String`
+</td>
+<td>
+
+**Required.** A comma-separated list of object field names that the subgraph can resolve at this query path.
```
pcmanus, line 165: Same remarks as for `@key`. And I'll note that there is even less restrictions than in the `@key` case as for instance, this is valid:
```graphql
type Query {
  q: U @provides(fields: "... on T1 { a }")
}

union U = T1 | T2

... // definitions for T1 and T2
```

docs/source/federated-types/federated-directives.mdx
```diff
@@ -0,0 +1,240 @@
...
+
+</td>
+</tr>
+
+</tbody>
+</table>
+
+## `@requires`
+
+Indicates that a particular object field's resolver requires the values of _other_ object fields that are resolved by _other_ subgraphs.
```
pcmanus, line 175: Here too, I'd just say "_other_ fields" and drop the "object" as you can require interface fields.

StephenBarlow:
Thanks for the review @pcmanus! I pushed up a followup commit that hopefully addresses all of your feedback. If a rendered preview doesn't eventually appear in the PR's checks list, I'll check in on getting that added.

docs/source/entities-advanced.mdx
```diff
@@ -152,14 +152,12 @@ We're done! `Bill` is now resolved in a new subgraph, and it was resolvable duri
 
 <ExpansionPanel title="With managed federation">
 
-> During the Federation 2 preview, this process uses the `@shareable` directive to make sure your supergraph schema composes during each step of the migration process. The full release of Federation 2 will include a different mechanism _specifically_ designed for migrating fields.
-
-1. In the Payments subgraph's schema, add the `@shareable` directive to the `Bill` entity's definition:
+1. In the Billing subgraph, define the `Bill` entity exactly as it's defined in the Payments subgraph:
```
StephenBarlow, line 155: Changes to managed federation field migration to use `@override`

docs/source/federated-types/federated-directives.mdx
```diff
@@ -0,0 +1,302 @@
...
+
+</td>
+</tr>
+
+</tbody>
+</table>
+
+### `@override`
+
+Indicates that a non-`@shareable` entity field is now resolved by _this_ subgraph instead of the _one other_ subgraph where it's currently defined.
```
pcmanus, line 108: You can remove `non-@shareable`: it's possible to override a shareable field if one wants to. In that case, the specifically pointed overridden field will never be resolved anymore but both the newly added one (the one with the `@override` and any other unrelated definitions of that field may be resolved, at the discretion of the query planner (as is true for normal shareable entity fields).

Essentially, `@override` just says "pretend that other definition I'm pointing to doesn't exists", but there is no specific other constraint. In fact, you can use `@override` on a non-entity field if you want, and that could actually be useful if you want to move a top-level operation (a `Query` field basically) between subgraphs.

docs/source/federated-types/federated-directives.mdx
```diff
@@ -0,0 +1,302 @@
...
+<tr class="required">
+<td>
+
+##### `fields`
+
+`String`
+</td>
+<td>
+
+**Required.** A comma-separated list of `@external` object fields that this field requires.
```
pcmanus, line 296: Like for `@key`/`@provides` we should probably describe it as "A GraphQL selection set of field names" rather than "A comma-separated list" (nesting is supported for instance).

Regarding the "of `@external` object fields", I'm fine saying that but I'll note that:
1. `@provides` are the exact same restrictions than `@requires` on that one, so if we say it here, we should say it for `@provides` too.
2. it is technically true that a subgraph can only require/provide stuff it doesn't resolve by itself, but nesting makes it so that it's not 100% equivalent to having _all_ the fields in a require/provide be `@external`. By which I mean that if you have something like `@provides(fields: "f { x }")` (or, equivalently, `@requires(fields: "f { x }")`), the one of `f` or `x` must be `@external`, and both can be, but it's not necessary that both _are_. Anyway, unsure it's worth getting into that kind of details here, but just wanted to provide context here so we at least avoid language that is too prohibitive. That is, "`@provides`/`@requires` can only be used to provide/require fields that the subgraph does not resolve itself" is maybe better than "all fields in the a `@provides`/`@requires` must be `@external`"?.

docs/source/federated-types/sharing-types.mdx
```diff
@@ -189,7 +189,74 @@ This definition works for querying Subgraph A, because Subgraph A's definition i
 
 > Note that Subgraph A's actual subgraph schema is _not_ modified. Within Subgraph A, `x` and `y` remain non-nullable.
 
-#### Omitting fields
+### Arguments
+
+Arguments for a shared field can differ between subgraphs in certain ways:
+
+* If an argument is _required_ in at least one subgraph, it can be _optional_ in other subgraphs. **It cannot be omitted.**
+* If an argument is _optional in every subgraph where it's defined_, it can be _omitted_ in other subgraphs.
```
pcmanus, line 197: A big caveat that should added here for "always optional" arguments is that while they can be omitted in some subgraphs, they will _not_ be included in the supergraph. That is, an argument is included in the supergraph only if it appears in all subgraphs (that define the shareable field).

In other words, an optionally argument will automatically behave "as if" it was `@inaccessible` if it's not consistently defined in all subgraphs.

docs/source/federated-types/sharing-types.mdx
```diff
@@ -270,6 +335,122 @@ And here's the problem: if Subgraph B _doesn't_ define `Query.currentPosition`,
...
+```graphql title="Subgraph B"
+type Position @shareable {
+  x: Int!
+  y: Int!
+}
+```
+
+</CodeColumns>
+
+> Even if `Position.z` is defined in _multiple_ subgraphs, you only need to apply `@inaccessible` in _one_ subgraph to omit it. In fact, you usually _should_ only apply it in one subgraph to simplify removing it later.
```
pcmanus, line 403: Nit: I'd personally replace that `usually _should_` by `might want to`: to me, this move from telling you what to do, to explaining you how to be clever about it, which, I don't know, might engage the reader brain a bit more :).

docs/source/federated-types/sharing-types.mdx
```diff
@@ -270,6 +335,122 @@ And here's the problem: if Subgraph B _doesn't_ define `Query.currentPosition`,
...
+
+With this syntax, composition ignores `Position.z`, and the resulting `Position` type includes only `x` and `y` fields.
```
pcmanus, line 405: Nit: technically, composition doesn't "ignore" `Position.z`: if you have 2 definitions of an inaccessible field that have incompatible, composition will still actually error out. It's that because `z` is not exposed in the supergraph API, it cannot be queried and so composition won't complain that "if it was queried, it couldn't be reached in the other subgraphs".

docs/source/federation-2/moving-to-federation-2.mdx
```diff
@@ -122,11 +122,25 @@ First, add the following definition to your subgraph schema:
...
 
-This definition identifies a schema as a true Federation 2 schema. Without this definition, the composition process assumes a subgraph schema is a Federation 1 schema, and it applies intelligent defaults to support backward compatibility.
+This definition identifies a schema as a true Federation 2 schema, and it `import`s any federation-specific directives that the schema uses.
+
+> **Note:** Most subgraph schemas use the `@key` and `@shareable` directives, so we've included them in the snippet above. Depending on your schema, you might need to add other federated directives to the `import` array, such as:
+>
+> * `@inaccessible`
+> * `@external`
+> * `@provides`
+> * `@requires`
```
pcmanus, line 136: Nit: We might want to add `@override` and `@tag`? Otherwise, if we'd rather avoid having to maintain/update that list so it includes all directives, then I think we might want to avoid having a list at all, and instead just link to the specific page listing them all. But as is, I worry this might confuse some (and to be clear, I do understand the phrasing doesn't actually pretend that this list is complete, but I still suspect some will read this a bit too fast to get the nuance).

pcmanus (approved):
Left a few additional comments, but this great overall and I don't mind if some of those comments are addressed later (or not at all for most of them).

docs/source/federated-types/federated-directives.mdx
```diff
@@ -0,0 +1,302 @@
...
+  </thead>
+
+<tbody>
+
+<tr class="required">
+<td>
+
+##### `fields`
+
+`String`
```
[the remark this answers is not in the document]
pcmanus: > My favorite solution would be to host the actual (GraphQL) specification at https://specs.apollo.dev/federation/v2.0

It'll be hosted there eventually, it's just not ready yet (but we'll certainly try to have "something" available there for GA or immediately afterwards).

> I get that this might be too much detail in a high level doc but it's also confusing for someone reading another definition from somewhere else.

I do both agree with the risk of confusion and worry that using `FieldSet` might unnecessarily distract here.

At the same time, while I do think we should link to the spec when it's available, I like that we describe each argument with a minimum of details here and that the page stand on its own. I wouldn't want that page to be reduced to a list of examples.

A possible cope-out could be to have this doc describe the "values" the argument accept, but without mentioning the type per-se. That is, instead of writing "`String`" here (and other similar places of this doc), we could write "a string value".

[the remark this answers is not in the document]
martinbonnin: Docs are hard 😅 . I think I'd just leave out the "type" and let the description speak for itself but having something like "accepted values" works too 👍

docs/source/entities-advanced.mdx
```diff
@@ -169,20 +167,21 @@ We're done! `Bill` is now resolved in a new subgraph, and it was resolvable duri
...
+      id: ID!
+      amount: Int!
+    }
+    ```
 
-    * When the gateway fetches its new supergraph schema, its behavior doesn't change. Even though the `Bill` entity is now `@shareable`, the Payments subgraph is still the only subgraph that defines it.
+    </CodeColumns>
 
-3. In the Billing subgraph's schema, define the `Bill` entity exactly as it's now defined in the Payments subgraph:
+2. In the Billing subgraph, apply the `@override` directive to every field of `Bill` that _isn't_ included in its `@key` (in this case, just `amount`):
```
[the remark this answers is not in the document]
StephenBarlow: Yeah, it does make the instructions more straightforward if they say "add `@override` to every field", so I'll simplify accordingly

docs/source/federated-types/federated-directives.mdx
StephenBarlow: Cool thanks! Let's say I have a `@shareable` field and I want to `@override` two _different_ subgraphs. Can I apply two `@override` directives to the same field? Or do I somehow include both subgraph names in one `from` argument? Or is this not possible currently?

docs/source/federated-types/sharing-types.mdx
StephenBarlow: Ooh that is a big one! I'll add.

docs/source/federation-2/moving-to-federation-2.mdx
StephenBarlow: Cool, I've trimmed this to only include a couple directives that are also in Fed 1 so they look familiar, and make it seem less like we're telling folks exactly what to add.

docs/source/federated-types/federated-directives.mdx
StephenBarlow: I believe I've cleaned up the language here correctly, but feel free to take another look to confirm!

docs/source/federated-types/federated-directives.mdx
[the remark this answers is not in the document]
StephenBarlow: I've cut the `String` type from each table and refer to each value as "A GraphQL field selection set (provided as a string) blah blah"

Is that roughly what we're going for?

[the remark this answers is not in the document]
martinbonnin: LGTM! Thanks!

docs/source/federated-types/federated-directives.mdx
pcmanus: Currently no, it's not possible: `@override` is not repeatable and `from` cannot have multiple options. I considered it, but decided to leave that out initially. I couldn't come up with a realistic scenario where it would be that useful, and I didn't wanted that to introduce new unforeseen problem just because we haven't taken the time to think it through.

But I "think" it's pretty trivial to add once we want to by allowing `@override` to be repeatable.

There is a few other limits we currently impose to make it simply to reason about, like you cannot "chain" overrides (overriding a field that itself override another subgraph) currently, but in theory I think we could lift that limitation if good use cases for it comes.

docs/source/federated-types/federated-directives.mdx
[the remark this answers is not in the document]
pcmanus: Same, looks great to me.

StephenBarlow:
@martinbonnin You mean the "Federated directives" article in the left nav? I do see it from that article; could be your browser cache was funky at the time? That happens to me all the time with a docs push

martinbonnin:
Right, "empty caches and hard reload made the trick. For some reason this is what I got before 

Screenshot 2022-04-13 at 10 18 22

In all cases, all good now, thanks!

PR #1652, opened 2022-03-28
Add support for tag v0.2

sachindshinde:
This PR introduces tag v0.2, which adds support for `@tag` in more locations (arguments, scalars, enums, enum values, input objects, input object fields).

### DIFF
--- a/subgraph-js/src/directives.ts
+++ b/subgraph-js/src/directives.ts
@@ -4,7 +4,6 @@ import {
   GraphQLNonNull,
   GraphQLString,
   GraphQLNamedType,
-  isInputObjectType,
   DirectiveNode,
   GraphQLField,
   FieldDefinitionNode,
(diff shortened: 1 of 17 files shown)


subgraph-js/src/directives.ts
```diff
@@ -173,7 +189,7 @@ export function typeIncludesDirective(
   directiveName: string,
 ): boolean {
   if (isInputObjectType(type)) return false;
-  const directives = gatherDirectives(type as GraphQLNamedTypeWithDirectives);
```
pcmanus, line 176: We probably want to remove the special case for `isInputObjectType` just before this line? (I'll note that this method is only used internally in the case of object types, so doesn't matter so far for our internal usage, but since it's exported ...).

pcmanus (approved):

PR #1653, opened 2022-03-28
Optimise indirect path computations

pcmanus:
This commit implements 2 optimizations for the computation of "indirect
paths" (think "following all possible key edges"):
1. it adds some pre-computation of edges-following-an-edge that skips
   filters some trivially unecessary options. Essentially, the
   precomputaton is relatively cheap and avoid repeatedly considering
   (and utlimately eliminating) those unecessary options.
2. query planning already has a mechanism by which "the computation of
   indirect paths folowing a given path" is memoized to save
   recomputation. This wasn't done for composition validation however
   and this commit correct that.

Those optimizations are particularly potent when composing a large number
of subgraphs with many entities spanning lots of subgraphs. On a fairly
large example (100+ subgraphs), this reduce composition time from 4+
"minutes" to less than 8 seconds ("on my machine").

### DIFF
--- a/query-graphs-js/src/graphPath.ts
+++ b/query-graphs-js/src/graphPath.ts
@@ -387,7 +387,10 @@ export class GraphPath<TTrigger, RV extends Vertex = Vertex, TNullEdge extends n
    * The set of edges that may legally continue this path.
    */
   nextEdges(): readonly Edge[] {
-    return this.graph.outEdges(this.tail);
+    const tailEdge = this.edgeToTail;
+    return tailEdge
+      ? this.graph.nonTrivialFollowupEdges(tailEdge)
+      : this.graph.outEdges(this.tail);;
   }
 
   /**
@@ -659,7 +662,7 @@ export class Unadvanceables {
   }
 }
 
-export function isUnadvanceable<V extends Vertex>(result: GraphPath<Transition, V>[] | Unadvanceables): result is Unadvanceables {
+export function isUnadvanceable(result: any[] | Unadvanceables): result is Unadvanceables {
   return result instanceof Unadvanceables;
 }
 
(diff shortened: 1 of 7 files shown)


query-graphs-js/src/graphPath.ts
```diff
@@ -941,11 +1004,6 @@ function advancePathWithNonCollectingAndTypePreservingTransitions<TTrigger, V ex
       }
       excludedEdges = addEdgeExclusion(excludedEdges, edge);
 
-      if (isConditionExcluded(edge.conditions, excludedConditions)) {
```
pcmanus, line 944: Note to reviewer: this condition hasn't been removed, just moved down. The reason I did that is that this condition is a tad more expansive to check than the following ones and they can be swapped easily, so having that one later is probably a tad more efficient. In reality though, this probably don't make much of a difference, especially after the other optimisations of this commit, but I had made that change early and figured "why change back now?".

query-graphs-js/src/graphPath.ts
```diff
@@ -387,7 +387,10 @@ export class GraphPath<TTrigger, RV extends Vertex = Vertex, TNullEdge extends n
    * The set of edges that may legally continue this path.
    */
   nextEdges(): readonly Edge[] {
-    return this.graph.outEdges(this.tail);
+    const tailEdge = this.edgeToTail;
+    return tailEdge
```
clenfest, line 391: If this is an instance of `TNullEdge`, is it guaranteed to be false here? I don't know how to interpret a type that extends null.

query-graphs-js/src/graphPath.ts
```diff
@@ -387,7 +387,10 @@ export class GraphPath<TTrigger, RV extends Vertex = Vertex, TNullEdge extends n
...
+      ? this.graph.nonTrivialFollowupEdges(tailEdge)
+      : this.graph.outEdges(this.tail);;
```
clenfest, line 393: extra semicolon

query-graphs-js/src/graphPath.ts
pcmanus: My understanding/intent (and hopefully I didn't miss something) is that the only type extending `null` is `null` itself, so this is really a way to say that `tailEdge` can be only be one of an `Edge`, `null`, or `undefined`, which some specific instantiation making the `null` part guaranteed to never happen.

Here, we'll default to `outEdges()` if `tailEdge` is `null`, which is fine. That is, it's always safe to return `outEdges()` from this method, with `nonTrivialFollowupEdges()` being simply a potentially more efficient alternative, and `null` edges in a `GraphPath` are super rare (they can only happen during query planning, not validation composition, and only when the input query has some essentially useless type-condition that we keep in the path only because there is some execution directives associated we don't want to lose) so I really don't think it's worth doing extra work to try to optimise them in this case.

Anyway, I did added a comment that may help a bit with was is going on there.

clenfest (approved):

query-graphs-js/src/nonTrivialEdgePrecomputing.ts
```diff
@@ -0,0 +1,48 @@
...
+
+function computeNonTrivialFollowups(edge: Edge, allFollowups: readonly Edge[]): readonly Edge[] {
+  switch (edge.transition.kind) {
+    case 'KeyResolution':
+      // After taking a key from subgraph A to B, there is no point of following that up with another key
+      // to subgraph C if that key has _the same_ conditions. This is because, due to the way key edges
+      // are created, if we have a key (with some conditions X) from B to C, then we are guaranteed to
+      // also have a key (with the same conditions X) from A to C, and so it's that later key we
+      // should be using in the first place. In other words, it's never better to do 2 hops rather than 1.
+      return allFollowups.filter((followup) => followup.transition.kind !== 'KeyResolution' || !sameConditions(edge, followup));
```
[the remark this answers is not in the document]
pcmanus: > So if I'm reading this correctly (and going back to the diagram you showed us), if A->B uses key 'x', but A->C uses key 'y', a transition from B->C with key 'y' would not be filtered?

It's correct that it wouldn't be filtered by this method. But also, that case is fairly different because at the moment we add edges to the query graph without looking if their condition is "satisfiable" (it's the algorithm that checks it later) . That is, we may have a key 'y' between A->C but 'y' not being "reachable"/"collectable" from 'A'. Let's take an example:
```graphql
# subgraph A
type Query {
  t: T
}

type T {
  x: Int
}
```
```graphql
# subgraph B
type T @key(fields: "x") {
  x: Int
  y: Int
}
```
```graphql
# subgraph C
type T @key(fields: "y") {
  y: Int
  z: Int
}
```
In that example, the query graph will technically have the key edge "y" from A to C, but you cannot use directly because A doesn't not have "y". So to get "z", you genuinely have to got B then C. If anything, it's the A -> C edge that should be filtered.

And so the next question would be: why are adding edges like A->B for key "y" in the query graph is A doesn't have "y"? And we "might" (I haven't though about it deeply) be able to do something like that, but it's far from trivial because checking if a key "condition" can be satisfied can be pretty complex: it's trivial in the example above, but keys can have many field, including nested ones, and some of those field can well be external or have requires, etc..., so checking the satisfiability of a key condition essentially means running the validation algorithm on that condition. Again, it might be possible to:
1. create a query graph as we do today (so, with all key edges added regardless of the satisfiability of their condition).
2. run some "optimisation" step on that query graph that look at each key edge, see if it is ever satisfiable (essentially running a version of the validation algorithm on their condition), and remove them if they aren't.

We could them run the validation algorithm/query planning on that now "optimized" query graph. But I'll note that:
1. this would require quite a refactor of the current code. Definitively high effort.
2. I doubt it would speed up composition validation, because I don't think we would save work in that case, we'd just split the work in more steps, but it would sum up to the same work. That said ...
3. ... it would be likely beneficial to query planning, where the "optimized" query graph might be use for computing many query plans. How much it'll help in practice is unclear however.

PR #1657, opened 2022-03-29
[repro] composition fails with core subgraphs

queerviolet:


queerviolet:
it looks like this is because they have bootstrap `@link`s.

adding a definition for `@link` fails with the failures we're seeing in #1554 (`Directive @link already exists in this schema`)

pcmanus:
There is indeed some issue where the code gets confused and adds definitions that are already there. I pushed #1662 to fix this/add more testing around this.

That said, I'll note that the repro here uses definitions that are not the one currently used and running it with the aforementioned PR yields:
```
Expected composition to succeed but got errors:
[Subgraph1] Invalid definition of directive "@key": argument "fields" should have type "federation__FieldSet!" but found type "federation__FieldSet"

[Subgraph2] Invalid definition of directive "@link": argument "url" should have type "String!" but found type "link__Url!"

[Subgraph2] Invalid definition of directive "@link": argument "import" should have type "[link__Import]" but found type "link__Imports"
```
which, as of the current "preview" code, is correct.

PR #1662, opened 2022-03-30
Fix handling of core/link when definitions are provided or partially so

pcmanus:
This fixes the issue described on #1657 but also fix some related handling of errors (that were not properly surfaced previously). I made a separate PR because the issue was not so much with composition than with subgraph validations, so I preferred moving tests in the existing test file dedicated to those subgraph validations. But also, the repro from #1657 uses some definitions that are not exactly the ones that have been used so far and so it doesn't  pass for those reasons.

### DIFF
--- a/internals-js/src/coreSpec.ts
+++ b/internals-js/src/coreSpec.ts
@@ -1,6 +1,6 @@
 import { ASTNode, DirectiveLocation, GraphQLError, StringValueNode } from "graphql";
 import { URL } from "url";
-import { CoreFeature, Directive, DirectiveDefinition, EnumType, ErrGraphQLValidationFailed, ListType, NamedType, NonNullType, ScalarType, Schema, SchemaDefinition } from "./definitions";
+import { CoreFeature, Directive, DirectiveDefinition, EnumType, ErrGraphQLValidationFailed, InputType, ListType, NamedType, NonNullType, ScalarType, Schema, SchemaDefinition } from "./definitions";
 import { sameType } from "./types";
 import { err } from '@apollo/core-schema';
 import { assert } from './utils';
@@ -8,6 +8,7 @@ import { ERRORS } from "./error";
 import { valueToString } from "./values";
 import { coreFeatureDefinitionIfKnown, registerKnownFeature } from "./knownCoreFeatures";
 import { didYouMean, suggestionList } from "./suggestions";
+import { ArgumentSpecification, createDirectiveSpecification, createEnumTypeSpecification, createScalarTypeSpecification, DirectiveSpecification, TypeSpecification } from "./directiveAndTypeSpecification";
 
 export const coreIdentity = 'https://specs.apollo.dev/core';
 export const linkIdentity = 'https://specs.apollo.dev/link';
@@ -64,7 +65,7 @@ export abstract class FeatureDefinition {
     return nameInSchema != undefined && (directive.name === nameInSchema || directive.name.startsWith(`${nameInSchema}__`));
   }
 
-  abstract addElementsToSchema(schema: Schema): void;
+  abstract addElementsToSchema(schema: Schema): GraphQLError[];
 
   abstract allElementNames(): string[];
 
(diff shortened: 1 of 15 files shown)


queerviolet (commented):
(just a quick drive-by comment, will play with this and offer more shortly. thanks for the quick turn-around!)

internals-js/src/definitions.ts
```diff
@@ -1499,6 +1509,10 @@ export class SchemaDefinition extends SchemaElement<SchemaDefinition, Schema>  {
       const imports = extractCoreFeatureImports(url, schemaDirective);
       const core = new CoreFeature(url, args.as ?? url.name, schemaDirective, imports, args.for);
       Schema.prototype['markAsCoreSchema'].call(schema, core);
+      // We also any core features that may have been added before we saw the @link for link itself
```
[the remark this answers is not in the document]
pcmanus: Yeah, that's the an artifact of how the code works right now. Essentially, because the abstraction is mutable, the way incomplete schema are expanded is done by taking the input and adding what's missing. So when we get an input that only has a `@link` to the federation spec, but no `@link`-to-link-itself, we end up adding that `@link`-to-link-itself to a document hat already has another existing `@link` and this code ensure that pre-existing feature is still tracker correctly.

To be clear, the resulting "final" document still end up with the `@link`-to-link as the first one on the `SchemaDefinition`, the code is just need due to way incomplete schema are auto-completed.

>  the `@core` and `@link` specs both require that the bootstrap `@link` is the first one in the document

As a minor aside, this made me notice that the current core/v0.2 spec (https://specs.apollo.dev/core/v0.2/) only ever talk about `SchemaDefinition` in the bootstrap/feature collection descriptions. Since we've making extensive doc of putting `@link` on `SchemaExtension` in our docs/examples, we should make sure to clarify their handling.

pcmanus:
> can we call the spec we're using here `https://specs.apollo.dev/link/v0.3`?

I'm afraid it's not that simple because we've had a few `preview` released with `link/v1.0`, and the last of those release even use that in supergraphs, so backpedaling to `link/v0.3` now would mean the next release cannot read the supergraphs of the previous release. Of course, we're in "preview", so breaking change are not completely out of question, but I know we've been a lot more careful with it for preview.

Fwiw, my personal opinion is that since `@link` is used as an integral part of user schema in fed2, then it essentially becomes GA as fed2 gets GA, and I think calling it 1.0 make sense, especially in terms of backward-compatibility guarantees associated with being out of the 0.x phase.

internals-js/src/coreSpec.ts
```diff
@@ -288,43 +297,61 @@ export function isCoreSpecDirectiveApplication(directive: Directive<SchemaDefini
...
+      name,
+      locations: [DirectiveLocation.SCHEMA],
+      repeatable: true,
+      argumentFct: (schema, nameInSchema) => this.createDefinitionArgumentSpecifications(schema, nameInSchema),
+    });
+  }
+
+  private createDefinitionArgumentSpecifications(schema: Schema, nameInSchema?: string): { args: ArgumentSpecification[], errors: GraphQLError[] } {
+    const args: ArgumentSpecification[] = [
+      { name: this.urlArgName(), type: new NonNullType(schema.stringType()) },
```
queerviolet, line 322: can we weaken this to just `String` for the moment? i'd like v1.0 to maintain universality. there are a few mechanisms to do this, but `@link(url: null, import: "looks__namespacedButIsnt")` is one way to handle it, and i'd like us not to have painted ourselves into a corner by shipping it like this.

queerviolet (approved):
yeah, i figured that was probably the case. it's a bummer we pinned 1.0 semantics without spec review.

there's one specific change i'd like us to make if possible, but we can always do it in another pr.

internals-js/src/coreSpec.ts
pcmanus: Make sense. And was simple-enough, so made the change in the PR (with a test to double-check it's a safe change, that is that we were still correctly reading previously generated definition having the non-nullable type).

PR #1665, opened 2022-04-01
Remove `[reviews]` from the `Products` subgraph

dero:
AFAICT, to ensure separation of concerns, the `Products` subgraph should not use the `Review` type.


StephenBarlow (approved):

StephenBarlow:
@dero You are absolutely right, thank you! 🤦

PR #1679, opened 2022-04-05
Small spelling changes in error messages

benweatherman:
I noticed a few spelling changes we could make from apollographql/federation#1672


pcmanus (approved):
The hints test needs to be updated to reflect the change in the hint message, but lgtm otherwise. Thanks.

PR #1680, opened 2022-04-05
Adding @sharable and @override to the federation spec

clenfest:
Adding @sharable and @override to the federation spec

### DIFF
--- a/docs/source/federation-spec.md
+++ b/docs/source/federation-spec.md
@@ -339,7 +339,9 @@ type User @key(fields: "email") @shareable {
 directive @override(from: String!) on FIELD_DEFINITION
 ```
 
-The `@override` directive is used to annotate that the field in the current subgraph is taking responsibility for providing the value of the field away from the subgraph specified in the `from` argument. The following example will result in all query plans made to resolve `User.name` to be directed to `SubgraphB`.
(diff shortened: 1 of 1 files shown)


benweatherman (approved):
🐐

Left some minor suggestions to use `shareable` consistently (and you already made some improvements to that topic, which were helpful).

docs/source/federation-spec.md
```diff
@@ -309,3 +309,48 @@ extend type User @key(fields: "email") {
...
+}
+```
+
+### `@override`
+
+```graphql
+directive @override(from: String!) on FIELD_DEFINITION
+```
+
+The `@override` directive is used to annotate that the field in the current subgraph is taking responsibility for providing the value of the field away from the subgraph specified in the `from` argument. The following example will result in all query plans made to resolve `User.name` to be directed to `SubgraphB`.
```
rkoron007, line 342: I tweaked the first sentence and would suggest separating these two sentences, like so:
```suggestion
The `@override` directive is used to indicate that the current subgraph is taking responsibility for resolving the marked field _away_ from the subgraph specified in the `from` argument. 

The following example will result in all query plans made to resolve `User.name` to be directed to SubgraphB.
```

rkoron007 (commented):
Feel free to disagree to push back on any of these suggestions!

Issue #1659, opened 2022-03-29
Some bugs/typos in Fed 2 composition hints


lennyburdette:
1. `InconsistentObjectValueTypeField`:
     The `elementCoordinate` is the type name, while the `mainElementDescription` indicates that it should be the field:
     
     > Product: the field that is inconsistently declared between subgraphs

2. `InconsistentArgumentPresence`:
      Looks like some values are missing: 

      > Argument "Inconsistent.missingArg(a:)" will not be added to "missingArg(a: undefined): undefined" in the supergraph as it does not appear in all subgraphs: it is defined in subgraph "a" but not in subgraph "b".

3. Nitpick: "executable directive" instead of "execution directive" — the spec called them "ExecutableDirectiveLocations".
4. `NoExecutionDirectiveLocationsIntersection`:
     > Execution directive "@bar" has no location that is common to all subgraphs: it will not appear in the **~subgraph~ supergraph** as there no intersection between location "FIELD" in subgraphs "a" and "c".
5. `InconsistentEnumValue`:
      > MyEnum: the **~union~ enum** type which has an inconsistent member
6. I was able to get redundant `InconsistentUnionMember` and `InconsistentEnumValue` hints with three subgraphs. This hint showed up twice:
      > Member type "Inconsistent" in union type "MyUnion" is only defined in a subset of subgraphs defining "MyUnion" (but can always be resolved from these subgraphs): "Inconsistent" is defined in subgraph "b" but not in subgraphs "a" and "c".
7. I couldn’t get *InconsistentExecutionDirectiveLocations* to appear. I thought that this would do it:
      ```
      directive @baz on FIELD | OBJECT
      # and
      directive @baz on FIELD | INTERFACE
      ```

pcmanus:
I've pushed #1683 which does a bunch of changes to hints and should all the points here.  But to expand on a few of those points more directly:

> The `elementCoordinate` is the type name

I've removed the `elementCoordinate` part of hints for now. The rational being that we want hints to be behave essentially similarly to errors (so they can be handled consistently), and errors don't have that. I don't exclude adding this back at some point, but the truth is that this part hadn't been though thru too much so I'm reluctant to commit to it for GA before we give a bit more thought.

> I couldn’t get InconsistentExecutionDirectiveLocations to appear.

[...]

PR #1683, opened 2022-04-05
Hints improvements

pcmanus:
That PR contains a number of improvements for hints, but the main goals are to:
1. introduce a level for hints (currently `WARN`, `INFO` or `DEBUG`).
2. improve consistency between hints and errors since those are fairly similar beasts.

It also include a number of hints related minor fixes, including those pointed in https://github.com/apollographql/federation/issues/1659, and a few rewording of messages (admittedly, many messages could be improved further, but I think this can done post-GA).


pcmanus:
I'll note that the PR uses a generated documentation for hints, like we have for errors, but it's currently a different document. I've be thinking that it might be cleaner to merge the 2 documents together however, so I might try to get a commit for that soon. But I don't think this should hold back review.

package.json
```diff
@@ -22,7 +22,9 @@
     "codegen:check": "npm run codegen && git diff --exit-code",
     "lint": "eslint . --ext .ts",
     "error-code-doc": "node ./internals-js/dist/genErrorCodeDoc.js > ./docs/source/errors.md",
-    "error-code-doc:check": "npm run error-code-doc && git diff --exit-code"
+    "error-code-doc:check": "npm run error-code-doc && git diff --exit-code",
+    "hints-doc": "node ./composition-js/dist/genHintDoc.js > ./docs/source/hints.md",
+    "hints-doc:check": "npm run hints-doc && git diff --exit-code"
```
benweatherman, line 27: **suggestion** Should we add this to `.circleci/config.yaml` so we run checks for the docs and the hints? (obviously pending any refactors to combine errors/hints docs)

```
npm run error-code-doc:check hints-doc:check
```

composition-js/src/genHintDoc.ts
```diff
@@ -0,0 +1,88 @@
...
+const header = `---
+title: Federation Hints
+sidebar_title: Hints
+---
+
+When Apollo Gateway successfully **compose** the schemas provided by your [subgraphs](./subgraphs/) into a **supergraph schema**, it may generate a number of "hints" that provide additional information the composition.
+
+Hints are categorized under the following levels:
+1. WARN: indicates a situation that may expected but is usually temporary and should be double-checked. Typically, composition might have had to ignore some elements from some subgraph when creating the supergraph.
+2. INFO: information that may hint at some improvements or highlight noteworthy resolution made by composition, but that can otherwise be ignored.
```
benweatherman, line 13: **nitpick**

```suggestion
2. INFO: information that may hint at some improvements or highlight noteworthy resolution made by composition but can otherwise be ignored.
```

composition-js/src/genHintDoc.ts
```diff
@@ -0,0 +1,88 @@
...
+
+const header = `---
+title: Federation Hints
+sidebar_title: Hints
+---
+
+When Apollo Gateway successfully **compose** the schemas provided by your [subgraphs](./subgraphs/) into a **supergraph schema**, it may generate a number of "hints" that provide additional information the composition.
+
+Hints are categorized under the following levels:
+1. WARN: indicates a situation that may expected but is usually temporary and should be double-checked. Typically, composition might have had to ignore some elements from some subgraph when creating the supergraph.
```
benweatherman, line 12: **nitpick**

```suggestion
1. WARN: indicates a situation that may be expected but is usually temporary and should be double-checked. Typically, composition might have had to ignore some elements from some subgraph when creating the supergraph.
```

benweatherman (approved):
🐐 

Excellent addition!

composition-js/src/hints.ts
```diff
@@ -1,164 +1,186 @@
...
+export enum HintLevel {
+  WARN = 60,
+  INFO = 40,
+  DEBUG = 20,
+}
 
-  toString(): string {
-    return `${this.code}: ${this.description}`;
-  }
+export type HintCodeDefinition = {
```
clenfest, line 10: I think that you'll find that we want to be able to get the english readable level from the HintCodeDefinition. I'd probably go with something like this, though it would involve changing the HintCodeDefinition back to a class and using a constructor

```typescript
const levels = {
  fatal: 60,
  error: 50,
  warn: 40,
  info: 30,
  debug: 20,
  trace: 10
};

type HintLevel = keyof typeof levels;

export class HintCodeDefinition {
  levelNum: (level: HintLevel) => number = (level) => levels[level];
  constructor(readonly code: string, readonly level: HintLevel, readonly description: string) {}
}
 
const INCONSISTENT_BUT_COMPATIBLE_FIELD_TYPE = new HintCodeDefinition(
  'INCONSISTENT_BUT_COMPATIBLE_FIELD_TYPE',
  'info',
  'Indicates that a field does not have the exact same types in all subgraphs, but that the types are "compatible"'
    + ' (2 types are compatible if one is a non-nullable version of the other, a list version, a subtype, or a'
    + ' combination of the former).',
);
```

composition-js/src/hints.ts
```diff
@@ -1,164 +1,186 @@
...
+  code: string,
+  level: HintLevel,
+  description: string,
```
clenfest, line 13: We probably should consider that at some point in the future we may want to merge errors and hints into a single class as I think errors are really just hints that have a higher alerting level. Not sure if there's a case for keeping them separate.

package.json
cpeacock: I'd say write a ticket for that, just in case there is weirdness ahead of release and it becomes a flaky test for the next two days

composition-js/src/hints.ts
pcmanus: >  english readable level from the HintCodeDefinition

Isn't it what `HintLevel[def.code]` does? I'm probably brain-farting this but I don't see which advantage the solution above brings (not that I'm against classes :D)?

composition-js/src/hints.ts
pcmanus: I generally agree, though there is a bunch of technical considerations to take into account when we do so (probably hard to unplug errors from `GraphQLError`, but also unclear using `GraphQLError` for debug level message is exactly "clean"; also, errors are use somewhat more pervasively than hints: maybe that gets in the way, maybe not, but worth looking). So I would prefer leaving any of that to future us for now.

As I mentioned, I do would want to try to at least get the doc merged, so "externally" it looks like those a different levels of the same thing, but I'm gonna call for this to be a followup too at this point.

composition-js/src/hints.ts
clenfest: Sorry, I'm not seeing how you go from the value to the name (and specifically how rover will do it). If there is a way that I'm not seeing, feel free to ignore.

pcmanus: Oh, I see what you mean. Forgot we were "blindly" serializing JS object to json for rover.

I'll deal with that. But thinking of maybe have level be a `{ value: HintLevel, name: string }`. Kind of made sense to me to talk of the level as being fundamentally an integer value, but some human-readable name attached for convenience.

pcmanus: Pushed a change for this. Kept the enum cause it kind of feel appropriate for this to me.

Side-note: I noticed your snippet above included more levels than I have in the patch. I do prefer avoid to add too many levels until it genuinely adds value, but there is of course plenty of space to add more later if the need does arise.

package.json
pcmanus: With permission from @cpeacock, I did added the check, since it's been pretty useful for the error doc and hasn't been flaky, so I'm somewhat confident about.

I did added it as a separate check, mostly because
```
npm run error-code-doc:check hints-doc:check
```
somehow didn't work locally for me, but each worked separately, so I didn't wanted to risk screwup up something.

Having a different check is admittedly a tad wasteful, though it's also a tad more informative (we'll probably merge both doc eventually and end up with a single check anyway).

composition-js/src/hints.ts
clenfest: On the extra levels thing, I copied some code from `pino`, wasn't suggesting that we actually add all those levels now.

clenfest (approved):

PR #1688, opened 2022-04-05
Switch Logger usages away from `apollo-server-types` to `@apollo/utils.logger`

trevor-scheer:
This is just an interface which I'm migrating out of `apollo-server-types`, so breaking gateway's dependency on that one in favor of the one published by `@apollo/utils.logger`.


benweatherman (approved):
🐐

PR #1692, opened 2022-04-07
Fix default value validation for input fields

pcmanus:
The validation of default values was called for field and directive arguments, but not for input fields, and this commit adds it.

This commits further adds testing for this default value validation and fix a number of issue in that value validation.

### DIFF
--- a/internals-js/src/validate.ts
+++ b/internals-js/src/validate.ts
@@ -254,6 +254,12 @@ class Validator {
           sourceASTs(field.appliedDirectivesOf('deprecated')[0], field)
         ));
       }
+      if (field.defaultValue !== undefined && !isValidValue(field.defaultValue, field, new VariableDefinitions())) {
+        this.errors.push(new GraphQLError(
+          `Invalid default value (got: ${valueToString(field.defaultValue)}) provided for input field ${field.coordinate} of type ${field.type}.`,
+          sourceASTs(field)
+        ));
+      }
     }
   }
 
--- a/internals-js/src/values.ts
+++ b/internals-js/src/values.ts
@@ -1,5 +1,6 @@
 import {
   ArgumentDefinition,
+  InputFieldDefinition,
   InputObjectType,
   InputType,
   isBooleanType,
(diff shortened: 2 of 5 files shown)


internals-js/src/values.ts
```diff
@@ -38,11 +39,24 @@ import { assert, assertUnreachable } from './utils';
 const MAX_INT = 2147483647;
 const MIN_INT = -2147483648;
 
+/**
+ * Converts a graphQL value into it's textual representation.
+ *
+ * @param v - the value to convert/display. This method assumes that it is a value graphQL
```
benweatherman, line 45: **nitpick**

```suggestion
 * @param v - the value to convert/display. This method assumes that it is a valid graphQL
```

internals-js/src/values.ts
```diff
@@ -38,11 +39,24 @@ import { assert, assertUnreachable } from './utils';
...
+ *   value (essentially, one that could have been produced by `valueFromAST`/`valueFormASTUntyped`).
+ *   If this is not the case, the behaviour is unspecified, and in particular this method may
+ *   throw or produce an output that is not valid graphQL syntax.
+ * @param expectedType - the type of the value being converted. This is optional is only used to
```
benweatherman, line 49: **nitpick**

```suggestion
 * @param expectedType - the type of the value being converted. This is optional and only used to
```

internals-js/src/values.ts
```diff
@@ -38,11 +39,24 @@ import { assert, assertUnreachable } from './utils';
...
+ *   ensure enum values are displayed as such and not as strings. In other words, the type of
+ *   the value should be provided when possible (when the value is known to be of a ype) but
```
benweatherman, line 51: **nitpick**

```suggestion
 *   the value should be provided when possible (when the value is known to be of a type) but
```

benweatherman (approved):
🐐

internals-js/src/values.ts
```diff
@@ -61,18 +75,26 @@ export function valueToString(v: any, expectedType?: InputType): string {
...
       elementsType = expectedType.ofType;
     }
     return '[' + v.map(e => valueToString(e, elementsType)).join(', ') + ']';
   }
 
+  // We know the value is not a list/array. But if the type is a list, we still want to print
+  // the value correctly, at least as long as it's a valid value for the element type, since
+  // list input coercions may allow this.
+  if (expectedType && isListType(expectedType)) {
+    expectedType = expectedType.ofType;
```
sachindshinde, line 91: Note you need to unwrap multiple times for nesting (e.g. `[[[Foo!]!]!]`), so I think this might need to be:

```suggestion
    expectedType = expectedType.ofType;
    return valueToString(v, expectedType);
```

sachindshinde (changes requested):

internals-js/src/values.ts
pcmanus: Oh yeah, good point. Going fix and add a test.

sachindshinde (approved):

internals-js/src/validate.ts
```diff
@@ -254,6 +254,12 @@ class Validator {
           sourceASTs(field.appliedDirectivesOf('deprecated')[0], field)
         ));
       }
+      if (field.defaultValue !== undefined && !isValidValue(field.defaultValue, field, new VariableDefinitions())) {
```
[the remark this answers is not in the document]
pcmanus: `undefined` means "there is not default value" so it felt right to not validate. And in fact, `isValidValue` currently deal with `undefined` as a value like it does for `null` and so it would fail for a non-nullable type. To be fair, we could probably change the handling of `undefined` in `isValidValue` so it works, but I think it's more explicit to have the code say "there is no default, no validation to do".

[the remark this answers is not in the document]
clenfest: got it. Thanks.

PR #1694, opened 2022-04-07
docs: fix v2 redirects

trevorblades:
This PR removes the `/federation-2/*` redirect and removes the `/v2` path prefix infront of all v2 redirects.


docs/source/_redirects
```diff
@@ -1,15 +1,13 @@
-/federation-2/*  /docs/federation/v2/federation-2/:splat
+/federation-2/coming-to-federation-2/ /docs/federation/v2/federation-2/under-consideration/
```
StephenBarlow, line 1: We can actually cut these first two (I've done so in a separate PR that hasn't landed yet)

StephenBarlow (approved):
Approved with one optional note 🤷

docs/source/_redirects
trevorblades: Let's just let your PR remove them separately for hopefully fewer merge conflicts.

PR #1697, opened 2022-04-08
Try implementing subgraph-js using the federation internals

pcmanus:

### DIFF
--- a/subgraph-js/src/types.ts
+++ b/subgraph-js/src/types.ts
@@ -10,6 +10,7 @@ import {
   GraphQLNamedType,
   isNamedType,
   isObjectType,
+  GraphQLResolveInfo,
 } from 'graphql';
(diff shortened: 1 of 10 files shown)


subgraph-js/src/types.ts
```diff
@@ -60,6 +61,58 @@ function addTypeNameToPossibleReturn<T>(
   return maybeObject as null | (T & { __typename: string });
 }
 
+export function entitiesResolver({
```
clenfest, line 64: Is there any functional change in this file that I'm not seeing or is it just a non-function refactor?

clenfest (approved):

subgraph-js/src/types.ts
pcmanus: There is not change to the resolver code, but it extract it as a method because it's called in `buildSubgraphSchema` directly now.

PR #1702, opened 2022-04-11
Additional end-to-end tests

pcmanus:
Our existing end-to-end testing is a bit low and the goal of this PR is to ever-so-slightly improve this.


pcmanus:
Calling this a draft PR currently as I'd to add a few most tests, but pushing it to show that unfortunately simple end-to-end with fed1 schema is currently broken because https://github.com/apollographql/federation/pull/1554 changed the generation for fed1 schema from using type `_FieldSet` to using `federation__FieldSet` and composition has not been updated to accepted/handle that (and fwiw, I would argue that it would better to fix the generation side to output `_FieldSet` again since it's what fed1 schema has always used, but I suppose it's debatable, and I don't mind making the fed1 upgrade code more resilient to such things).

pcmanus:
Rebased on `main` and tests are passing. Also added a test for `@inaccessible` since previously. I'm also include the test that repo the failure of #1703 to show that it is now passing and ensure no future similar regressions.

clenfest (approved):

PR #1715, opened 2022-04-12
bugfix: Use `SECURITY` core purpose for merged inaccessible spec

sachindshinde:
In #1638 , as part of the ability to use `@inaccessible` in subgraphs, composition will add the inaccessible spec to the supergraph when it's used in a subgraph.

However, we currently don't specify a core purpose. We would like to use `for: SECURITY` in the core/link directive application for inaccessible. This PR makes this change via introducing `defaultCorePurpose` on `FeatureDefinition`.


benweatherman (approved):
🐐

PR #1718, opened 2022-04-13
Federation subgraph library compatibility updates

hwillson:
Updated Federation subgraph library compatibility test results, including Federation 2 changes. Let me know if the formatting should change (it's a dump of the markdown that comes from the testing process). Thanks!


StephenBarlow (approved):
Looks good! Now that the page is much longer it might be good to figure out an in-page nav for quick-jumping to at least a particular language, but I can investigate that

hwillson:
Sounds awesome @StephenBarlow - thanks very much! I was also thinking we could split the table into several tables, having one table per language. Food for thought anyways. Thanks again!

PR #1723, opened 2022-04-13
created a new precomposition rule to detect fed2 schemas

clenfest:
Please make sure you agree with the appoach and that I've got the version numbers correct.

### DIFF
--- a/federation-js/src/composition/validate/preComposition/noFed2Subgraphs.ts
+++ b/federation-js/src/composition/validate/preComposition/noFed2Subgraphs.ts
@@ -1,10 +1,10 @@
 import { visit, GraphQLError, Kind } from 'graphql';
 import { ServiceDefinition } from '../../types';
 
-import { errorWithCode } from '../../utils';
+import { errorWithCode, findDirectivesOnNode } from '../../utils';
 
 /**
- * - There are no fields with @external on base type definitions
+ * - There are no subgraphs that @link to a version that is unsupported by 1.0 composition.
  */
 export const noFed2Subgraphs = ({
   name: serviceName,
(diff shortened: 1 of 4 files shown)


federation-js/src/composition/validate/preComposition/noFed2Subgraphs.ts
```diff
@@ -0,0 +1,50 @@
...
+ */
+export const noFed2Subgraphs = ({
+  name: serviceName,
+  typeDefs,
+}: ServiceDefinition) => {
+  const errors: GraphQLError[] = [];
+  visit(typeDefs, {
+    SchemaExtension(schemaExtensionNode) {
+      const directives = schemaExtensionNode.directives?.filter(directive => {
+        if (directive.name.value === 'link') {
```
trevor-scheer, line 17: Nit: you can use `findDirectivesOnNode` for this, but it's really just equivalent / sugar for what you're doing.

federation-js/src/composition/validate/preComposition/noFed2Subgraphs.ts
```diff
@@ -0,0 +1,50 @@
+import { visit, GraphQLError, Kind } from 'graphql';
+import { ServiceDefinition } from '../../types';
+
+import { errorWithCode } from '../../utils';
+
+/**
+ * - There are no fields with @external on base type definitions
+ */
```
trevor-scheer, line 8: Update or remove

federation-js/src/composition/validate/preComposition/noFed2Subgraphs.ts
```diff
@@ -0,0 +1,50 @@
...
+              const version = parseFloat(versionRaw);
+              return (
+                (spec === 'federation' && version >= 2.0) ||
+                (spec === 'link' && version >= 1.0)
+              );
+            }
+          }
+        }
+        return false;
+      }) ?? [];
```
trevor-scheer, line 37: I think the `?? []` is extraneous

trevor-scheer (approved):
We should add a changelog entry and update the error codes doc with the code we're introducing.

federation-js/src/composition/validate/preComposition/noFed2Subgraphs.ts
clenfest: It was necessary when `schemaExtensionNode.directives` was potentially undefined, but since I switched to `findDirectivesOnNode` it's no longer necessary.

PR #1730, opened 2022-04-14
Exclude v0 packages from the grouped patch/minor update group

trevor-scheer:
See: https://github.com/renovatebot/renovate/discussions/15093

Right now our renovate config will group version 0.x updates into the automerged minor/patch group. This isn't really desirable since any update to a v0 package can be a breaking change. This update to the minor/patch group excludes all packages with `0.y.z`, `~0.y.z`, and `^0.y.z` from the group, forcing them into their own PR which won't be automerged (like all other major version bumps).

### DIFF
--- a/renovate.json5
+++ b/renovate.json5
@@ -80,6 +80,10 @@
     {
       groupName: "all non-major dependencies",
       matchUpdateTypes: ["patch", "minor"],
+      // Exclude all v0 dependencies from the patch/minor group. Changes to v0 packages
+      // should be treated as major updates, but renovate treats all patches and minors
+      // the same.
+      matchCurrentVersion: "!/^[\^~]?0/",
       groupSlug: "all-npm-minor-patch",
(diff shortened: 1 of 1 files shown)


benweatherman (approved):
🐐

abernix:
Late-stage reply after being OOO: This appears like it should work as is, just thought I'd reference [the technique we used to do something _similar_ in Router-land](https://github.com/apollographql/router/blob/5985321c3b5f6910d2d53443acd5748dee771bbf/renovate.json5#L13-L26).

In that case, we've decided that typically in Rust-land "pre-1.0 patch" updates are typically "minor" whereas "pre-1.0 minor" are usually "major".  We used the `separateMinorPatch` option to Renovate with `minor` and `patch` objects to represent that (and do grouping).

Anyhow, now asking for a change, just prior art!  Thanks for doing this!

PR #1732, opened 2022-04-14
Improve merging of groups during @require handling

### LINKED ISSUE #1725: Fed2 query planning regression with @requires
Reproduction: https://stackblitz.com/edit/basic-federation-o5kzav?file=readme.md

The issue is triggered by including the `computed2` field—without it, the query plans are identical in fed1 and fed2.

This was reported by a customer.

pcmanus:
Query planning handles `@requires` by, when it needs to handle a field
with requires, computing a sub-plan to gather the requirements
(so-called "fetch groups") and then checking if those groups can be
merged into earlier groups to avoid needless fetches.

The test to check if we could merge into earlier groups was a bit
conservative in 2 small ways:
1. as part of the process, we were looking if we had created a useless
   fetch group, one where all the selections were already parts of
   its inputs. It was however possible to have a selection not strictly
   included in the inputs, but just because of a few additional
   `__typename`. And those shouldn't prevent optimizing since we can
   always move those to the parent of whichever group were were trying
   to optimize away (`__typename` is essentially always resolved by all
   subgraphs).
2. further down, we were only merging groups where "merge path" were
   matching, but this was limiting merging when some nesting was
   involved. The patch fixes this by always merging but ensuring we
   properly concatenate path as needed to handle nesting.

Fixes #1725

### DIFF
--- a/query-planner-js/src/buildPlan.ts
+++ b/query-planner-js/src/buildPlan.ts
@@ -38,6 +38,7 @@ import {
   FederationMetadata,
   federationMetadata,
   entitiesFieldName,
+  concatOperationPaths,
 } from "@apollo/federation-internals";
 import {
   advanceSimultaneousPathsWithOperation,
@@ -1366,7 +1367,10 @@ class FetchDependencyGraph {
     return rootNodes;
   }
 
-  dumpOnConsole() {
+  dumpOnConsole(msg?: string) {
+    if (msg) {
+      console.log(msg);
+    }
     console.log('Groups:');
     for (const group of this.groups) {
       console.log(`  ${group}`);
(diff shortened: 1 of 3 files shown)


query-planner-js/src/buildPlan.ts
```diff
@@ -1366,7 +1367,10 @@ class FetchDependencyGraph {
     return rootNodes;
   }
 
-  dumpOnConsole() {
+  dumpOnConsole(msg?: string) {
```
pcmanus, line 1370: Nit: that method isn't used, but I keep it around as it helps when debugging. Adding the message here just save some `console.log` when doing so.

clenfest (approved):

PR #1735, opened 2022-04-14
Improvements for fed1 schema support

pcmanus:
This PR adds 2 small leniency for fed1 schema (when read by fed2):
1. it allows `@tag` on an external field if the same tag is on a non-external definition of that field.
2. it allows definitions for federation directives that are not quite what they should but, but are close enough.

The reason in both case is that federation 1 allowed those things, and while we don't  accept those in federation 2 schema intentionally, we want the most users possible to be able to use fed2 on fed1 schema and to delay changes to schema for when use migrate to actual fed2 schema as often as reasonable. This provide a slightly better experience since fed1 schema is, after all, a backward compatibility feature.


benweatherman (approved):
🐐 

Before: 981 failed
After: 615 failed

PR #1744, opened 2022-04-19
Update supergraphSDL to be a string

jsegaran:
We expect `supergraphSdl` to be a string in the gateway ([code](https://github.com/apollographql/federation/blob/7331cbdb824af20f70ca500f1826859c92edda49/gateway-js/src/config.ts#L281-L286)). 

If `supergraphSdl` is a type document the gateway throws with 
```
Error: When a manual configuration is not provided, gateway requires an Apollo configuration. 
See https://www.apollographql.com/docs/apollo-server/federation/managed-federation/ for more information.
Manual configuration options include: `serviceList`, `supergraphSdl`, and `experimental_updateServiceDefinitions`
```


trevor-scheer (approved):
🚀

PR #1751, opened 2022-04-20
Add gateway version to schema extensions

### LINKED ISSUE #1749: Add schema extension to gateway schemas to differentiate federated schemas
In fed1, [schema extensions are added](https://github.com/apollographql/federation/blob/version-0.x/federation-js/src/composition/compose.ts#L703) to allow detecting whether or not a schema was coming from a gateway. We should replicate that in fed2. This is needed so Apollo Server plugins can Do The Right Thing™ when schemas are updated. Specifically, this is needed for https://github.com/apollographql/apollo-server/pull/6317.

### Considerations
- Would it be useful to expose the gateway version?

### Implementation notes
- We need to add extensions in `ApolloGateway.load` and `ApolloGateway.updateWithSchemaAndNotify`

benweatherman:
Adds gateway version as a schema extension

## TODO
- [x] Add typing info similar to the `__resolveReference` PR
- [x] Add tests
- [x] Figure out why we're still getting [type errors during tests](https://app.circleci.com/pipelines/github/apollographql/federation/8155/workflows/fb35072a-d359-4506-b7e5-b7b57e9907fb/jobs/49512?invite=true#step-107-150) [this was happening because types were being excluded from compilation, fixed with 697bf88b9b2255853d4111420cc897ac2bf21f91]
- ~Change uses of `require("package.json")` to the newly generated version file~ Decided against this because we would still need to import `package.json` to get the name

Fixes #1749


trevor-scheer (approved):
LGTM. I do think having to add the typings file to the test tsconfig is curious and likely symptomatic of some bad config somewhere, but I'm not sure what and I'm perfectly comfortable moving forward with this as-is.

Worth noting this doesn't need a backport because fed1 gateway schema objects are identifiable whereas fed2 gateway schema objects have no metadata (that I'm aware of) to make them identifiable.

PR #1758, opened 2022-04-21
Various tidying in the federation docset

StephenBarlow:


benweatherman (approved):
🐐

StephenBarlow:
Some of these edits run a bit contrary to some voice stuff we're going to be changing in the near future, so converting to draft for now

PR #1764, opened 2022-04-22
Update documentation for merging of input/output types

### LINKED ISSUE #1763: Only common values are being considered when merging enum or input types
I am using federation v2. I have defined one `enum` type that is shared in two subgraphs. Now according to [docs](https://www.apollographql.com/docs/federation/federated-types/sharing-types/#enums-unions-and-interfaces) the composition of supergraph sdl should merge this enum from both subgraphs and it should contain all the values provided in subgraphs. But supergraph schema is having only common values between the two.

Input

subgraph schema 1
```
enum Color {
  RED
  GREEN
  BLUE
}
```

subgraph schema 2
```
enum Color {
  CYAN
  MAGENTA
  YELLOW
  RED
}
```

supergraph schema generated
```
enum {
  RED
}
```

supergraph schema expected
```
enum Color {
  CYAN
  MAGENTA
  YELLOW
  RED
  GREEN
  BLUE
}
```

And if there is no common value between the two then composition throws error like
```
None of the values of enum type "Color" are defined consistently in all the subgraphs defining that type. As only values common to all subgraphs are merged, this would result in an empty type.
```

Same is happening with `input` types also.

I am using `InterospectAndCompose` API for composing supergraph sdl.

Dependencies
```
"dependencies": {
    "@apollo/gateway": "^2.0.1",
    "apollo-server": "^3.6.7",
    "apollo-server-types": "^3.5.2",
    "express": "^4.17.1",
    "express-graphql": "^0.12.0",
    "graphql": "^16.0.1"
  }
```

clenfest:
Updating documentation for feature that was changed in https://github.com/apollographql/federation/pull/1672

Fixes #1763

### DIFF
--- a/docs/source/federated-types/sharing-types.mdx
+++ b/docs/source/federated-types/sharing-types.mdx
@@ -480,6 +480,7 @@ interface User {
 union Media = Book | Podcast
 
 enum Color {
+  RED
   CYAN
   MAGENTA
   YELLOW
@@ -495,7 +496,7 @@ interface User {
 
 </CodeColumns>
 
-Compositional logic merges these definitions in your supergraph schema:
+Compositional logic merges these definitions differently in your supergraph schema depending on whether they are used as an input or output type. Output types are merged by "union", and input types are merged by "intersection". If the above are output types, the composed schema looks like this:
 
 ```graphql title="Supergraph schema"
 union Media = Book | Movie | Podcast
@@ -516,5 +517,23 @@ interface User {
   age: Int!
 }
 ```
+
+Whereas if they are input types, this would be the result:
+
+```graphql title="Supergraph schema"
+union Media = Book
+
+enum Color {
+  RED
+}
+
+# The object types that implement this interface are
+# responsible for resolving these fields.
+interface User {
+  name: String!
+}
+```
+
+For output types, the supergraph schema lists all values that may be returned, but for input types we restrict values that are allowed in the shared type so that we can ensure all subgraphs can process the value provided.
 
 This can be useful when different subgraphs are responsible for different subsets of a particular set of related types or values.


pcmanus:
As mentioned in my previous comment, we should distinguish a bit between interface/union, that are always only output types, and enums, which can be used as output type only, input type only, or both (the later of which isn't mentioned in the PR but probably should be explained). 

I'll also note that input object types don't seem to be mentioned by the doc at all, and that could be a good place to add them. They can only be input types (hence the name) and use an "intersection" merge strategy.

Last point of context: field arguments, which are also "inputs", are also merged as "intersection" and the symmetry with input types is very much on purpose so it might be worth to higlight in the documentation. And to clarify, what I mean by this is that if you take:
```graphql
# Subgraph A
type T @shareble {
  f(a: String, b: String, d: string = "default"): Int
  g(i: I!): Int
}

input I {
  a: String
  b: String
  d: String = "default"
}
```
```graphql
# Subgraph B
type T @shareable {
  f(a: String, b: String, c: String = "other default"): Int
  g(i: I!): Int
}

input I {
  a: String
  b: String
  c: String = "other default"
}
```
there there isn't a ton of conceptual difference with the arguments for `f` ang `g`: they both essentially take the same arguments, and each option can be slightly preferable depending on context, but they don't differ fundamentally. And so you expect their merging to be similiar, and it is. They merge into:
```graphql
type T {
  f(a: String, b: String): Int
  g(i: I!): Int
}

input I {
  a: String
  b: String
}
```
that is both both the arguments of `f` and the fields of `I` are intersected.

So I wonder if all this isn't a hint that a slightly bigger re-organisation is needed to explain it well. 

The way I think of this, and thus how I would organize it (but not trying to mandate anything) is that I'd start with the general concept of output types being merged by "union" and input types merged by "intersection". This can probably be illustrated using object types as posterchild for an output type and input object types for input types. It could be nice to quickly justify the reasoning behind those strategy: essentially, for output types, it's ok that different subgraph return different "parts" of a type as long as the gateway knows how to gather the various parts from the various subgraphs (which composition validates) as it can merge those various parts at runtime. But we cannot, in general, pick-and-choose to which subgraph an input value will be send to and so can only expose the parts of an input type that all subgraph know about.

And then I would one sub-section per type, with a simple example of how this apply to that specific type plus any specifics. In particular, the enum section could get into how enum can be used in either only inputs, only output, or both and explain the merging in each case. And the input object type section
could draw the parallel with argument merging.

benweatherman:
Closing since this was fixed in #1839

PR #1765, opened 2022-04-22
Migrate to `@apollo/utils` packages for `createSHA` and `isNodeLike`

trevor-scheer:
These functions are common logic between the server and federation monorepos so it's been moved into the utils monorepo. Most importantly, this removes a deep import from `'apollo-server-core/dist/utils/createSHA'` which is a dependency the server team is hoping to remove over time.


package.json
```diff
@@ -41,6 +41,7 @@
   },
   "devDependencies": {
     "@apollo/client": "3.5.10",
+    "@apollo/utils.createhash": "1.0.0",
```
benweatherman, line 44: **question** Do we need this in the root package?

benweatherman (approved):
🐐

package.json
trevor-scheer: Good q. Technically no, but since I'm also using it in tests, I added it here as well. If it were ever removed as a runtime dependency of the gateway then the tests would still have a copy of this package to run.

benweatherman: If it got removed as a runtime dependency, wouldn't the tests fail? So then the dev would know to add a different dependency. It feels a little like it'd be a gateway devDependency if that happened.

(Wanna be clear, I'm totally cool with this in there, I wanna make sure I follow these best practices since I've not worked in a monorepo before)

trevor-scheer: It's a fair point and I don't feel strongly about it. I wouldn't call this a best practice necessarily, just a consideration that came to mind.

In fact, maybe getting those test failures could be helpful since you might not be testing a good thing anymore after removing the runtime dependency. The failure would at least point you to some related things that might want an update!

trevor-scheer: You brought me to the other side, whether it was your intention or not.
[6e1c2e9](https://github.com/apollographql/federation/pull/1765/commits/6e1c2e9c27acb737991c88fcd8dc13b4bd862b12)

benweatherman: break on through

PR #1767, opened 2022-04-23
add application/graphql+json mimeType support to gateway

eformat:
Add support for application/graphql+json to gateway.

I know this is not universally accepted yet, and there are related fixes in progress

https://github.com/graphql/graphql-over-http/issues/31
https://github.com/apollographql/apollo-server/pull/6295

This issue comes up when using apollo gateway with popular Java microservices frameworks that set application/graphql+json e.g. https://github.com/smallrye/smallrye-graphql is one such framework

### DIFF
--- a/gateway-js/src/datasources/RemoteGraphQLDataSource.ts
+++ b/gateway-js/src/datasources/RemoteGraphQLDataSource.ts
@@ -279,7 +279,7 @@ export class RemoteGraphQLDataSource<
     _context?: TContext,
   ): Promise<object | string> {
     const contentType = fetchResponse.headers.get('Content-Type');
-    if (contentType && contentType.startsWith('application/json')) {
+    if (contentType && (contentType.startsWith('application/json') || contentType.startsWith("application/graphql+json"))) {
       return fetchResponse.json();
     } else {
       return fetchResponse.text();
--- a/gateway-js/src/datasources/RemoteGraphQLDataSource.ts
+++ b/gateway-js/src/datasources/RemoteGraphQLDataSource.ts
@@ -279,7 +279,7 @@ export class RemoteGraphQLDataSource<
     _context?: TContext,
   ): Promise<object | string> {
     const contentType = fetchResponse.headers.get('Content-Type');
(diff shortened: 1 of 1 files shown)


eformat:
> @eformat: Thank you for submitting a pull request! Before we can merge it, you'll need to sign the Apollo Contributor License Agreement here: https://contribute.apollographql.com/

Done. Thanks.

pcmanus:
Thanks for the contribution.

I'll admit that I'm personally just a little bit unsure about adding `application/graphql+json` explicitely because:
1. the whole [graphql-over-http](https://github.com/graphql/graphql-over-http) effort is still Stage 0, and the [stages](https://github.com/graphql/graphql-over-http/blob/main/ROADMAP.md#stages) description kind of suggest that it's "reasonable to start implementations" only at stage 1.
2. I haven't had time to really follow that proposal, and afacit most of the main maintainers of this repo are currently on that same boat, so while I get the general idea, I want to be cautious about accepting something I don't understand all that well.

But to be very clear, I understand that this is very far from a full implementation, and is in fact a super trivial change with relatively little "strings attached". So I'm not trying to be contrarian on principle. Rather, I'm trying to justify my position, which is that while I'm happy to facilitate early experiments with that graphql-over-http proposal if it's easy, I also think that it's way too early for federation/the gateway to "officially" support it, and so I'd like to be careful with managing expectations.

All this to say that I'd presonally prefer to change the content type detection for `json` to be something along the lines of `(contentType.startsWith('application/') && contentType.includes('json'))` (or maybe `contentType.endsWith('json')` for the latter? Which I'll note was actually an offline idea from @abernix). I feel that it's what you'd roughtly expect, and it avoids hardcoding the currently unofficial/un-registered `application/graphql+json` (and to my earlier rambling, may more strongly suggest some kind of official support).

eformat:
@pcmanus - yeah, look i hear you. adjuted to use `contentType.includes('json')` which seems sane.

dariuszkuc:
Thanks for the PR. Closing this old PR with the same reason as on https://github.com/apollographql/federation/pull/2769#issuecomment-1706891804

PR #1782, opened 2022-04-26
Fix Schema.clone when directive application happens before definition

### LINKED ISSUE #1768: Schema.clone throws when directive applied before definition
### Actual behavior

This code throws an exception because `@bar` is applied before it's defined:

```ts
import { buildSchema } from "@apollo/federation-internals";
buildSchema(`    
  directive @foo(arg: String @bar) on FIELD
  
  directive @bar on ARGUMENT_DEFINITION
  
  type Query {
    foo: String!
  }
`).clone();
```

```ts
/code/apollographql/federation/internals-js/dist/definitions.js:403
                throw this.schema().blueprint.onGraphQLJSValidationError(this.schema(), new graphql_1.GraphQLError(`Unknown directive "@${nameOrDef}".`));
                ^

Uncaught GraphQLError: Unknown directive "@bar".
    at ArgumentDefinition.applyDirective (/code/apollographql/federation/internals-js/src/definitions.ts:530:11)
    at copyAppliedDirective (/code/apollographql/federation/internals-js/src/definitions.ts:3314:20)
    at /code/apollographql/federation/internals-js/src/definitions.ts:3310:43
    at Array.forEach (<anonymous>)
    at copyAppliedDirectives (/code/apollographql/federation/internals-js/src/definitions.ts:3310:28)
    at copyArgumentDefinitionInner (/code/apollographql/federation/internals-js/src/definitions.ts:3358:3)
    at copyDirectiveDefinitionInner (/code/apollographql/federation/internals-js/src/definitions.ts:3366:5)
    at copy (/code/apollographql/federation/internals-js/src/definitions.ts:3217:5)
    at Schema.clone (/code/apollographql/federation/internals-js/src/definitions.ts:1423:5)
    at /code/apollographql/federation/<repl>.ts:2:184 {
  path: undefined,
  locations: undefined,
  extensions: [Object: null prototype] {}
}
```

### Expected behavior

Calling `.clone` on a schema with a directive that's applied before it's defined works. Specifically, `Schema.clone` should copy directive definitions before calling `copyDirectiveDefinitionInner`.

### Version information

```sh
➜ npm ls @apollo/federation-internals
apollo-federation-monorepo@ /Users/weatherman/code/apollographql/federation
├─┬ @apollo/composition@2.0.1 -> ./composition-js
│ └── @apollo/federation-internals@2.0.1 deduped -> ./internals-js
├── @apollo/federation-internals@2.0.1 -> ./internals-js
├─┬ @apollo/gateway@2.0.1 -> ./gateway-js
│ └── @apollo/federation-internals@2.0.1 deduped -> ./internals-js
├─┬ @apollo/query-graphs@2.0.1 -> ./query-graphs-js
│ └── @apollo/federation-internals@2.0.1 deduped -> ./internals-js
├─┬ @apollo/query-planner@2.0.1 -> ./query-planner-js
│ └── @apollo/federation-internals@2.0.1 deduped -> ./internals-js
└─┬ @apollo/subgraph@2.0.1 -> ./subgraph-js
  └── @apollo/federation-internals@2.0.1 deduped -> ./internals-js
```

benweatherman:
Naively copying directives means that directives that use other directives could appear in the wrong order. This changes the order directives are copied such that dependent directives are copied first.

For example, the directive order in this schema will be `@bar` then `@foo` because `@foo` uses `@bar`.

```graphql
directive @foo(arg: String @bar) on FIELD_DEFINITION
directive @bar on ARGUMENT_DEFINITION
```

Fixes #1768


internals-js/src/definitions.ts
```diff
@@ -3201,13 +3201,48 @@ function *typesToCopy(source: Schema, dest: Schema): Generator<NamedType, void,
...
+ *
+ * For example, the directive order in this schema will be `@bar` then `@foo`
+ * because `@foo` uses `@bar`.
+ *
+ * ```graphql
+ * directive @foo(arg: String @bar) on FIELD_DEFINITION
+ * directive @bar on ARGUMENT_DEFINITION
+ * ```
+ */
+function *walkDirectives(schema: Schema): Generator<DirectiveDefinition, void, undefined> {
```
benweatherman, line 3216: I don't love this method here, but wasn't sure of a better place for it. Does it belong as a method on `Schema`?

internals-js/src/__tests__/definitions.test.ts
```diff
@@ -590,6 +590,34 @@ test('default arguments for directives', () => {
...
+      directive @wizz(arg: String @fuzz(arg: "wizz")) on ARGUMENT_DEFINITION
+      directive @fuzz(arg: String @buzz(arg: "fuzz")) on ARGUMENT_DEFINITION
+      directive @buzz(arg: String @baz(arg: "buzz")) on ARGUMENT_DEFINITION
+      directive @baz(arg: String @bar) on ARGUMENT_DEFINITION
+      directive @bar on ARGUMENT_DEFINITION
+
+      type Query {
+        foo: String! @foo(arg: "query")
+      }
+    `).clone();
```
benweatherman, line 606: **blocker** Test to ensure the directives are all in the schema

pcmanus:
I think this solution almost work but would still throw on cyclic dependency, that if you have:
```graphql
directive @foo(a: Int @bar) on ARGUMENT_DEFINITION
directive @bar(b: Int @foo) on ARGUMENT_DEFINITION
```
Now, according to the spec this should be invalid (end of https://spec.graphql.org/June2018/#sec-Type-System.Directives, "While defining a directive, it must not reference itself directly or indirectly") but:
1. I just tested it and graphQL-js does not reject this in practice, and nor does our code.
2. even if we added proper validation, validation happens separately (and after) cloning, there would still be a small risk of running into a non-descriptive exception in the cloning code instead of getting a proper validation error, so it could be worth be a bit defensive and handle such cycles.

I'll note that we have do deal with such cycles with types too, where cycles are genuinely allowed, and so I might suggest we might use the same solution here. That is, what we do for type is to do a "shallow" copy of all type definition first, and then copy their inner elements in a 2nd pass when we know all references will resolve properly. More precisely, my suggestion for this would be https://github.com/pcmanus/federation/commit/1779cefdda0e5c8e1ea9316edf90099818a5368e (which only shows the fix but of course we should reuse your tests and maybe add one for the cycles above, since it's currently not rejected).

internals-js/src/definitions.ts
benweatherman: Even better, this method is totally unnecessary

PR #1783, opened 2022-04-26
Avoid modifying query graph edge conditions for @requires

### LINKED ISSUE #1750: Plans of one query incorrectly alters the plans of other queries when there are requires with nested fields
I am noticing that the query plan for one operation would leak to the plans of other queries when there are nested required fields. 

Reproduce:
```javascript
const supergraphSdl = `
    schema
    {
        query: Query
    }
    directive @join__field(graph: join__Graph!, requires: join__FieldSet, provides: join__FieldSet, type: String, external: Boolean, override: String, usedOverridden: Boolean) repeatable on FIELD_DEFINITION | INPUT_FIELD_DEFINITION
    directive @join__implements(graph: join__Graph!, interface: String!) repeatable on OBJECT | INTERFACE
    directive @join__type(graph: join__Graph!, key: join__FieldSet, extension: Boolean! = false, resolvable: Boolean! = true) repeatable on OBJECT | INTERFACE | UNION | ENUM | INPUT_OBJECT | SCALAR
    scalar join__FieldSet
    enum join__Graph {
        GRAPH1 @join__graph(name: "graph1", url: "http://localhost:4000/graphql")
        GRAPH2 @join__graph(name: "graph2", url: "http://localhost:4000/graphql")
    }
    scalar link__Import
    enum link__Purpose {
        SECURITY
        EXECUTION
    }

    type Query
        @join__type(graph: GRAPH1)
        @join__type(graph: GRAPH2)
    {
        entity: Entity @join__field(graph: GRAPH1)
    }

    type Entity
        @join__type(graph: GRAPH1, key: "entityKey")
        @join__type(graph: GRAPH2, key: "entityKey")
    {
        entityKey: String
        entityField1: ValueType @join__field(graph: GRAPH1) @join__field(graph: GRAPH2, external: true)
        entityField2: String @join__field(graph: GRAPH2, requires: "entityField1 { valueTypeField1 }")
        entityField3: String @join__field(graph: GRAPH2, requires: "entityField1 { valueTypeField2 }")
    }

    type ValueType
        @join__type(graph: GRAPH1)
        @join__type(graph: GRAPH2)
    {
        valueTypeField1: String
        valueTypeField2: String
    }
`

const op1 = `
    {
        entity {
            entityField2
            entityField3
        }
    }
`

const op2 = `
    {
        entity {
            entityField3
        }
    }
`

const { QueryPlanner, serializeQueryPlan } = require('@apollo/query-planner');
const { buildSchema , parseOperation } = require('@apollo/federation-internals');

const schema = buildSchema(supergraphSdl);
const planner = new QueryPlanner(schema);

// plan op2
const op2QueryPlan = planner.buildQueryPlan(parseOperation(schema, op2));

// plan op1
planner.buildQueryPlan(parseOperation(schema, op1));

// plan op2 again
const op2QueryPlan2 = planner.buildQueryPlan(parseOperation(schema, op2));

// prints false
console.log(serializeQueryPlan(op2QueryPlan) === serializeQueryPlan(op2QueryPlan2));
```

The expected behavior is this script would print true since they are query plans for the same query. However, it prints false. The query plan of `op2QueryPlan2` looks like following
```
QueryPlan {
  Sequence {
    Fetch(service: "graph1") {
      {
        entity {
          __typename
          entityKey
          entityField1 {
            valueTypeField2
            valueTypeField1
          }
        }
      }
    },
    Flatten(path: "entity") {
      Fetch(service: "graph2") {
        {
          ... on Entity {
            __typename
            entityKey
            entityField1 {
              valueTypeField2
              valueTypeField1
            }
          }
        } =>
        {
          ... on Entity {
            entityField3
          }
        }
      },
    },
  },
}
```
Note how `valueTypeField1` is being fetched and passed to graph2 even though it's not a requirement for `entityField3`. It only happens after the same query planner generates a plan for op1 and it seems the extra required fields are leaked from the plan for `op1` which does need `valueTypeField1` to be fetched.

I am able to reproduce this behavior on federation version > 2.0.0.

pcmanus:
The problem this fixes is that `SelectionSet` is mutable, but query
graphs needs to be immutable post-construction since they are reused
between query plans, and the edge of those use `SelectionSet` for the
conditions. As a result, it was a bit too easy to inadvertently
modify an edge conditions without meaning to, which is exactly what
was happens in the handling of requires.

Note that we could manually fix the exact issue in the requires-handling
code by manually cloning the edge conditions we weren't properly
cloning, but this commit take a slightly more systemic approach
that attempts to make it harder for such issue to popup in another
place later.

That is, this adds the possibility for "freezing" a `SelectionSet`,
which makes it effectively immutable from that point on. This should
allow to find unexpected modification to edge conditions much sooner
during testing. But this also allow to automatically clone frozen
selection sets when they would otherwise be embeded/aliased in other
non-frozen (and thus mutable) sets, which is actually the issue
from #1750.

Fixes #1750

### DIFF
--- a/gateway-js/CHANGELOG.md
+++ b/gateway-js/CHANGELOG.md
@@ -9,6 +9,7 @@ This CHANGELOG pertains only to Apollo Federation packages in the 2.x range. The
 - More helpful error message for errors encountered while reading supergraphs generated pre-federation 2 [PR #1796](https://github.com/apollographql/federation/pull/1796)
 - Fix handling of @require "chains" (a @require whose fields have @require themselves) [PR #1790](https://github.com/apollographql/federation/pull/1790)
 - Fix bug applying an imported federation directive on another directive definition [PR #1797](https://github.com/apollographql/federation/pull/1797).
+- Fix bug where planning a query with `@require` impacts the plans of followup queries [PR #1783](https://github.com/apollographql/federation/pull/1783).
 
 ## v2.0.2-alpha.0
 
--- a/query-graphs-js/CHANGELOG.md
+++ b/query-graphs-js/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG for `@apollo/query-graphs`
 
+## vNEXT
+
+- Fix bug where planning a query with `@require` impacts the plans of followup queries [PR #1783](https://github.com/apollographql/federation/pull/1783).
+
 ## v2.0.1
 
 - Released in sync with other federation packages but no changes to this package.


internals-js/src/__tests__/operations.test.ts
```diff
@@ -141,3 +141,105 @@ test('fragments optimization of selection sets', () => {
...
+    // Control: check we can add to the selection set while not yet frozen
+    expect(s.isFrozen()).toBeFalsy();
+    expect(() => s.mergeIn(s1)).not.toThrow();
+
+    s.freeze();
+    expect(s.isFrozen()).toBeTruthy();
+    expect(() => s.mergeIn(s2)).toThrowError('Cannot add to frozen selection: { t { a } }');
+  });
+
+  it('it does not clone non-frozen selections when adding to another one', () => {
```
clenfest, line 174: Should you have a test that checks that `updateForAddingTo` will clone if necessary or is that covered by one of these cases?

internals-js/src/operations.ts
```diff
@@ -653,18 +654,84 @@ export class SelectionSet {
...
+   * Adds the provided selection to this selection, merging it to any existing selection of this set as appropriate.
+   *
+   * Please note that by default, the input selection may (or may not) be directly referenced by this selection
+   * set after this method return. That is, future modification of this selection set may end up modifying the input
+   * selection due to direct aliasing. If direct aliasing should be prevented, the input selection should be frozen
+   * (see `freeze` for details).
+   */
   add(selection: Selection): Selection {
+    // It's a but to try to add to a frozen selection set
+    assert(!this.isFrozen(), () => `Cannot add to frozen selection: ${this}`);
```
clenfest, line 733: We don't need to add this assertion to other methods? i.e `addPath`, etc?

query-graphs-js/src/querygraph.ts
```diff
@@ -189,10 +197,15 @@ export class Edge {
...
   addToConditions(newConditions: SelectionSet) {
-    if (!this._conditions) {
-      this._conditions = new SelectionSet(this.head.type as CompositeType);
-    }
+    // As mentioned in the ctor, we freeze the conditions to avoid unexpected modifications once a query
+    // graph has bee fully built (this method is called _during_ the building, so can still mutate the
+    // edge freely). Which means we need to clone any existing conditions (so we can modify them), and need
+    // to re-freeze the result afterwards.
+    this._conditions = this._conditions
+      ? this._conditions.clone()
```
clenfest, line 205: Is this clone() necessary? Aren't we already cloning in the constructor?

internals-js/src/__tests__/operations.test.ts
pcmanus: The test using `mergeIn` indirectly test it, as it's the cloning in this method that end up ensuring `SelectionSet::add` and `SelectionSet::mergeIn` do a clone. In fact, `updateForAddingTo` is not really meant to be use externally, it's only public because there is no simple way to define a method in `Selection` that only `SelectionSet` can use (well, I can I can make it private and cheat the type-system with `prototype`, but didn't wanted).

That said, I'll admit I didn't overdo testing here, and I could probably add at least a few test to exercise fragment selections, and why not direct test for both `add` and `updateForAddingTo`.

internals-js/src/operations.ts
pcmanus: The code make sure all modifications are done by this method, so that one should be sufficient in practice.

query-graphs-js/src/querygraph.ts
pcmanus: We're also freezing in the constructor, and here we're trying to mutate the conditions to add new ones. So that `clone` is essentially about "unfreezing" the selections (but in theory, we could add an `unfreeze` method that just flip back the `_isFrozen` boolean, and that would be enough for that one specific case; but an `unfreeze` method makes the frozen story a bit dodgy and I don't think cloning here matters performance wise so ...).

internals-js/src/operations.ts
```diff
@@ -542,6 +542,7 @@ export class SelectionSet {
   private readonly _selections = new MultiMap<string, Selection>();
```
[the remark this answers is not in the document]
pcmanus: I added an abstract class to save duplication when possible.

pcmanus:
Took the liberty to rebase because there was some conflict in `buildPlan.test.ts` given recent commits to `main`, but kept the review-changes separate. Along with the new `Freezable` abstraction, I added a few additional unit tests. I'll admit that I haven't covered all possible permutations that could be tested, but knowing how the code is organised, I felt that I was starting to hit diminishing returns (and I'm eager to get this fix released).

clenfest (approved):

PR #1796, opened 2022-04-28
Triggers user friendly message on supergraph extracting errors more o…

pcmanus:
…ften

When encountering errors while extract subgraphs from fed1 supergraph,
which can happen due to stricter fed2 validations, we were providing
an error message suggesting that user try fed2 composition, but that
message was only triggered for error thrown when validating the
extracted subgraph, but some error may be triggered earlier in the
extraction process.

This commit ensure that the try-catch we use to trigger that more
helpful error message is widenned to enclose all of the extraction
code.


internals-js/src/__tests__/extractSubgraphsFromSupergraph.test.ts
```diff
@@ -597,3 +597,86 @@ test('throw meaningful error for invalid federation directive fieldSet', () => {
...
+    enum join__Graph {
+      SERVICEA @join__graph(name: "serviceA" url: "")
+      SERVICEB @join__graph(name: "serviceB" url: "")
+    }
+  `;
+
+  const schema = buildSupergraphSchema(supergraph)[0];
+  expect(() => extractSubgraphsFromSupergraph(schema)).toThrow(
+    'Error extracting subgraphs from the supergraph: this might be due to errors in subgraphs that were mistakenly ignored by federation 0.x versions but are rejected by federation 2.\n'
+    + 'Please try composing your subgraphs with federation 2: this should help precisely pinpoint the problems and, once fixed, generate a correct federation 2 supergraph.\n'
```
benweatherman, line 677: **observation** I wonder if we'll eventually be able to make this a suggestion the user will be able to act on (i.e. `Add a @link to the federation 2 schema`, etc). This message is definitely a good direction, but I might be a little lost as a user. Maybe if a tool like rover sees this message, it could try to recompose with fed2 automagically.

Anyway, nothing to do here but thought I'd mention it while I was thinking about it.

benweatherman (approved):
🐐

internals-js/src/__tests__/extractSubgraphsFromSupergraph.test.ts
pcmanus: The thing is that user are going to run into this error mostly only when they run a gateway/router with a fed1 supergraph. So `rover` is probably not in the picture here (that is, `rover` does composition, but if you compose with the fed 2 code base, you never get a fed1 supergraph and so that doesn't apply), and the `gateway`/`router` usually don't really have any access to the subgraphs, so I'm not sure how much more can be done.

I mean, I'm obviously open to further improvements, it's just not obvious what more we can do (well, unless we can to consider pretty complex stuff like having the `gateway`/`router` reach to studio somehow and do some magic, but I'm not even sure that would desirable).

> i.e. `Add a @link to the federation 2 schema`

I'll note that this isn't the issue here. It's perfectly for the user running into this to keep its schema as a fed1 schema initially; the problem is that there is bugs/dodgy stuff in their fed1 schema which fed 1 didn't detect and that's what they should fix. In fact, my advise would be that they _first_ fix the problem in the fed1 schema and only _then_ consider migrating the schema into a fed 2 schema.

benweatherman: Got it, that makes sense. I appreciate the explanation.

PR #1799, opened 2022-04-28
Remove `node` types from `@apollo/composition` tsconfig

trevor-scheer:
The `@apollo/composition` library is also compiled for `rover`, to be used in deno where the Node API is unavailable. To prevent accidental usage of Node builtins, this removes the typically-available `node` types from the `tsconfig.json` of this package in particular.

For example, `console` is no longer available globally. Packages like `crypto` and `util` (among many Node standard libs) are also no longer available and their use will result in TS errors.

These errors may be surprising to a developer. This constraint isn't something most people keep at the forefront of their mind (which is why I think this change is very helpful in preventing a future problem). Any feedback on how to document this well would be very welcome!

### DIFF
--- a/composition-js/scripts/genHintDoc.ts
+++ b/composition-js/scripts/genHintDoc.ts
@@ -1,5 +1,5 @@
 import { assert, MultiMap } from '@apollo/federation-internals';
-import { HintCodeDefinition, HintLevel, HINTS } from './hints';
+import { HintCodeDefinition, HintLevel, HINTS } from '../src/hints';
 
 const header = `---
 title: Composition hints
--- a/composition-js/scripts/tsconfig.json
+++ b/composition-js/scripts/tsconfig.json
@@ -1,0 +1,6 @@
+{
+  "extends": "../tsconfig",
+  "compilerOptions": {
+    "types": ["node"],
+  },
+}
--- a/composition-js/tsconfig.json
+++ b/composition-js/tsconfig.json
@@ -2,7 +2,8 @@
   "extends": "../tsconfig.base",
   "compilerOptions": {
     "rootDir": "./src",
-    "outDir": "./dist"
+    "outDir": "./dist",
+    "types": [],
   },
   "include": ["src/**/*"],
   "exclude": ["**/__tests__"],
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "lint": "eslint . --ext .ts",
     "error-code-doc": "ts-node ./internals-js/src/genErrorCodeDoc.ts > ./docs/source/errors.md",
     "error-code-doc:check": "npm run error-code-doc && git diff --exit-code",
-    "hints-doc": "ts-node ./composition-js/src/genHintDoc.ts > ./docs/source/hints.md",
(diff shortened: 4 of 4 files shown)


pcmanus:
> For example, `console` is no longer available

😭 . How am I supposed to debug stuffs now ?!@#

Some kidding aside, I know that I will need a simple way to disable this for when debugging, and while I can probably work something locally on my end, that might be something more generally and I'd appreciate adding something here.

Also, `@apollo/composition` depends on other packages, isn't it a concern for those too? I'm thinking of `@apollo/federation-internals` in particular, which include some specific baked-in debugging capability that rely on `console` (and _are_ used by composition, though only when you setup some specific env variable).

trevor-scheer:
Good points @pcmanus. I suspect there's a good way to:
* Pinpoint / target _only_ files used by `rover`
* Continue to allow things that are already polyfilled by `rover` (like `console`!)

And yes, not pulling in `query-graphs` and `federation-internals` is an oversight (and really makes the bullets above more important since the other packages do need to follow these rules as well, so limiting the impact is even more preferable).

Going to mark this as WIP. I welcome anyone who feels nerd-sniped to collaborate on this problem with me, else I might get to it in some free time if I feel inspired. I don't think there's any immediate concern here.

A couple other things:
* `// @ts-ignore` would suffice for debugging (though less ergonomic, a snippet would be handy for this)
* Happy to share an ergonomic step debugging approach with you (if you're in VS Code for TS development). It's fantastic, I usually just skip console debugging altogether now.

abernix:
I agree that `// @ts-ignore` and native debugging integrations could be viable paths here.  As another option, if `console` is the biggest point of contention, we could just add a local `global` TS declaration that itself declares the `Console` types in the way that they are compatible with our usage of them.

Basically, we could copy this and whittle it down to what we need:

- [The `Console` from the `dom` libs](https://github.com/microsoft/TypeScript/blob/15f7b6f68791487c059d41d5fe99117f79e1f1a2/lib/lib.dom.d.ts#L17727-L17748)
- [Something like what `@types/node` does](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/console.d.ts#L58-L61)

Overall, this still solves what could be an eventual "surprise" factor in our release pipeline between `federation` and `rover` so the consideration is still viable and I think this was a creative solution.  It could also be solved with linting or something.

That said, there is a lot of other stuff to do right now so if the team is at least conscious of their usage of Node.js standard lib usage and actively avoids introducing that in their implementations and reviews of those implementations, I don't think this is urgent.

I'm going to (finally) remove myself as a reviewer on this because I get pinged on it every day, but those are my thoughts.  Happy to discuss.

trevor-scheer:
@abernix agree re: urgency and I appreciate the alternatives! I've removed the others as reviewers as well. I'll leave this open since I think there are some good options here and I do still think this potential pitfall should be patched up. I won't be tackling this though, so I'll leave it to the team to decide on the future of this (closing is fine by me if that's the eventual decision).

trevor-scheer:
Closing in favor of a more thorough approach theorized in https://github.com/apollographql/federation/issues/2453

PR #1800, opened 2022-04-28
change _FieldSet to FieldSet

queerviolet:
while collecting definitions for apollographql/specs#21, @pcmanus noted that `_FieldSet` is now `FieldSet` in fed 2.0


benweatherman (approved):
🐐

PR #1812, opened 2022-04-30
Add callback option when polling Uplink fails

### LINKED ISSUE #1801: Uplink Fallback Function
When there are Uplink failures, whether it's an internal outbound networking issue or Uplink failing to resolve, there's a lot of value in being able to provide a fallback function that returns the last working schema. Although [additional Uplink endpoints can be provided](https://www.apollographql.com/docs/federation/managed-federation/uplink/#environment-variable), it adds to the list of round-robin candidate, where we don't always want it to be. Providing a way to provide a fallback function would give us the flexibility to fallback to the last healthy schema from a local source such as a cache, or even other up and running gateway pods in the same cluster.

Alternatively, @lennyburdette's [resiliency solution](https://github.com/apollosolutions/managed-federation-resiliency/pull/5/files#diff-c7c29c4fdc9870447c7b403009bbfc2bd35ca20a232ab6b1919b7226e5722ad6) is another option where a fallback schema is provided via the constructor. Although this works, we want to prioritize Uplink, over having to hit our fallback to populate the constructor on Gateway startup.

benweatherman:
When polling Uplink fails, it's nice to have a fallback to load the supergraph from somewhere else.

This change exposes `UplinkSupergraphManager` (the artist formerly known as `UplinkFetcher`). It also adds config options for `onFailureToFetchSupergraphSdlDuringInit` and `onFailureToFetchSupergraphSdlAfterInit` that will be called if fetching from Uplink fails. It can return a string that will be used to update the supergraph schema. `onFailureToFetchSupergraphSdlAfterInit` can return `null` to signal the gateway should continue to use the existing schema.

Basic usage looks like the following:

```ts
import { promises } from 'fs';
import * as path from 'path';
import { ApolloServer } from 'apollo-server';
import { ApolloGateway, UplinkSupergraphManager } from '@apollo/gateway';

const gateway = new ApolloGateway({
  debug: true,
  supergraphSdl: new UplinkSupergraphManager({
    apiKey: process.env.APOLLO_KEY + "x", // Adding a "x" to artificially ensure we have an invalid key
    graphRef: process.env.APOLLO_GRAPH_REF,
    initialMaxRetries: 0, // Don't retry during initialization phase, so fallback is called sooner
    async onFailureToFetchSupergraphSdlDuringInit({ error, graphRef, logger }) {
      const filePath = path.resolve(__dirname, 'supergraph.graphql');
      logger.info(`Falling back to cached graph for ${graphRef} during init due to ${error}, loading from ${filePath}`);
      return promises.readFile(filePath, {encoding: 'utf8'});
    },
    async onFailureToFetchSupergraphSdlAfterInit({ error, graphRef, logger, fetchCount }) {
      const filePath = path.resolve(__dirname, 'supergraph.graphql');
      logger.info(`Falling back to cached graph for ${graphRef} after init due to ${error}, loading from ${filePath} (fetch count: ${fetchCount})`);
      return null;
    }
  }),
});
const server = new ApolloServer({ gateway });

server
  .listen()
  .then(({ url }) => {
    console.log(`🚀 Gateway ready at ${url}`);
  })
  .catch((err) => {
    console.error(err);
  });

```

Fixes #1801

## TODO

- [x] Unit tests
- [x] Ensure good errors if callback provides invalid schema
- [x] Make it easier to use `UplinkSupergraphManager` defaults

## Things that aren't here

1. I started refactoring `UplinkSupergraphManager` into its own package so we could make big changes without needing to make a major version bump in the gateway. That requires making another package that holds just the types, so it can be used by the new uplink supergraph manager package and the gateway. That complicates the release slightly because we'd potentially want to version those separately (since the whole reason for breaking it out is to have the option of making breaking changes). It doesn't seem like the benefit is worth the effort and ongoing maintenance.
2. I thought about making `onFailureToUpdateSupergraphSdl` a function on gateway (in line with `onSchemaLoadOrUpdate`), but that seemed like a larger refactor and can still be added in the future. That also has a slightly different mechanism than `onFailureToFetchSupergraphSdl`.


benweatherman:
Moved this back to a draft while I add some unit tests

gateway-js/src/supergraphManagers/UplinkFetcher/index.ts
```diff
@@ -80,28 +97,45 @@ export class UplinkFetcher implements SupergraphManager {
...
+    } catch (e) {
+      if (!this.config.updateSupergraphSdlFailureCallback) {
+        throw e;
+      }
+
+      this.config.logger?.debug('Error fetching supergraphSdl from Uplink, using updateSupergraphSdlFailureCallback');
+      supergraphSdl = await this.config.updateSupergraphSdlFailureCallback.call(this, { error: e });
+      if (!supergraphSdl) {
+        throw new UplinkFetcherError('updateSupergraphSdlFailureCallback returned invalid supergraphSdl');
+      }
```
sMorm, line 133: Hi Ben! This callback is pretty much what I had envisioned as a fallback. My only question is that should we do any sort of validation on the response here to ensure that the response here is a functional schema? As of right now, we expect a string. Unless this will get caught further down the line, I'm good with that. Just wanted to make sure we see meaningful errors in case an invalid string/schema was returned.

benweatherman: Great suggestion! I still need to add unit tests and I'll make sure we expose a good error in case of an invalid schema.

benweatherman:
@lennyburdette great suggestion! I'll incorporate that as well.

lennyburdette:
any progress on this? i'd love to update the [fallback demos](https://github.com/apollosolutions/managed-federation-resiliency) to use this approach!

gateway-js/src/supergraphManagers/UplinkFetcher/index.ts
benweatherman: Right now, the error messages leave a little to be desired, depending on how invalid the schema is. I've added a test here and seeing if we can make things any nicer https://github.com/apollographql/federation/pull/1812/files#diff-39988a715de21379d4ae848a801bc735d1f45ce72a819e23c279466e5ce37c14R85-R110

benweatherman:
@lennyburdette Thanks for the ping! I just finished up some tests and refactoring how the config works. I started down the path of adding a `uplinkFetcherDefaults` function, but it seemed better for everyone to implement better defaults in the `UplinkFetcher` constructor. Thoughts on that approach?

Also, the idea came up of publishing `UplinkFetcher` as a separate package so we can iterate on features a bit faster if folks are gonna be using it directly. Any idea how useful that will be?

Finally, I _could_ see this being useful across other supergraph managers, so I'm wondering if we want a more generic mechanism that only adding it to `UplinkFetcher`. This would probably be along the lines of `onSchemaLoadOrUpdate`.

gateway-js/src/supergraphManagers/UplinkFetcher/index.ts
```diff
@@ -80,28 +130,51 @@ export class UplinkFetcher implements SupergraphManager {
...
+      ({supergraphSdl, minDelaySeconds} = result);
+    } catch (e) {
+      if (!this.onFailureToUpdateSupergraphSdl) {
+        throw e;
+      }
+
+      this.logger.debug('Error fetching supergraphSdl from Uplink, calling updateSupergraphSdlFailureCallback');
+      supergraphSdl = await this.onFailureToUpdateSupergraphSdl.call(this, { error: e });
+      // if (!supergraphSdl) {
+      //   throw new UplinkFetcherError('updateSupergraphSdlFailureCallback returned invalid supergraphSdl');
```
lennyburdette, line 170: intentionally left commented?

lennyburdette:
Moving the default values to the UplinkFetcher itself makes sense! Hopefully makes the ApolloGateway class a little simpler too.

gateway-js/src/supergraphManagers/UplinkFetcher/index.ts
benweatherman: This goes back to what kind of checking and error message we want https://github.com/apollographql/federation/pull/1812#discussion_r862858907

Only checking `!supergraphSdl` doesn't do enough. So I think we're better off not doing any checks here and letting the errors bubble up from `Gateway.updateWithSupergraphSdl`. We _could_ export some of the validation stuff from `Gateway.createSchemaFromSupergraphSdl`, but I'm not sure it's helpful to get the same error at a different part of the call chain.

**TL;DR** This will likely be removed and possibly replaced with more rigorous schema checking.

gateway-js/src/index.ts
```diff
@@ -451,11 +415,6 @@ export class ApolloGateway implements GraphQLService {
         healthCheck: this.externalSubgraphHealthCheckCallback.bind(this),
         getDataSource: this.externalGetDataSourceCallback.bind(this),
       });
-      if (!result?.supergraphSdl) {
```
benweatherman, line 454: **note** This is somewhat of a silly check, since much worse schemas could be passed, and are caught later (e.g. `" "` won't fail here but will fail when the schema is parsed later).

So I decided to move this check later in the stack, and make it a GraphQL error https://github.com/apollographql/federation/pull/1812/files#diff-1a556fe2fd6a545a37d601c8185534fec07e931e2599b32d71c767cd149df20eR447, since it's an invalid schema (oddly, our parse code doesn't handle that case and throws an obscure error...I should file an issue about that).

gateway-js/src/__tests__/integration/configuration.test.ts
```diff
@@ -320,46 +317,6 @@ describe('gateway config / env behavior', () => {
       );
     });
   });
-
-  describe('schema config delivery endpoint configuration', () => {
```
benweatherman, line 324: **note** These were moved to be part of integration tests, since they really should be tested as part of gateway construction

https://github.com/apollographql/federation/pull/1812/files#diff-9351bdb42a25121eac9a7bf90afbca18146a69c980ed56acbd158bcda26b80b7R88-R107

gateway-js/src/config.ts
```diff
@@ -325,6 +326,7 @@ export function isManagedConfig(
     'schemaConfigDeliveryEndpoint' in config ||
     'uplinkEndpoints' in config ||
     'fallbackPollIntervalInMs' in config ||
+    (isSupergraphManagerConfig(config) && config.supergraphSdl instanceof UplinkSupergraphManager) ||
```
benweatherman, line 329: **note** This change is needed so we can treat a user-constructed `UplinkSupergraphManager` as a managed config when deciding how to behave (i.e. implicit managed config should be the same as passing in your own `UplinkSupergraphManager`)

gateway-js/src/index.ts
```diff
@@ -1008,9 +972,6 @@ export class ApolloGateway implements GraphQLService {
     switch (this.state.phase) {
       case 'initialized':
       case 'failed to load':
-        throw Error(
```
benweatherman, line 1011: **note** It's safe to call `gateway.stop`, even if the gateway is only `initialized` or `failed to load`. There's no public interface to know if the gateway is in a good state to stop anyway.

gateway-js/src/supergraphManagers/UplinkFetcher/index.ts
benweatherman: Removed this bit, since checking is done further down the path in https://github.com/apollographql/federation/pull/1812/files#diff-1a556fe2fd6a545a37d601c8185534fec07e931e2599b32d71c767cd149df20eR447

gateway-js/src/supergraphManagers/UplinkFetcher/index.ts
benweatherman: We'll get `GraphQL` errors when the gateway tries to call `externalSupergraphUpdateCallback` (at load, gateway calls that explicitly; `UplinkSupergraphManager` calls that during polling). Depending on how "bad" the schema is, the error contents should help diagnose why the schema is invalid. The empty string case continues to be handled in the gateway https://github.com/apollographql/federation/pull/1812/files#diff-1a556fe2fd6a545a37d601c8185534fec07e931e2599b32d71c767cd149df20eR447.

Though I do wanna call out there's no fallback to the fallback. Meaning if there is an invalid schema, it'll either crash (on load) or continue using the previous schema (on update).

gateway-js/src/index.ts
trevor-scheer: Does `trim()`ming in the meantime help with that or does parse throw funky on empty string too?

gateway-js/src/supergraphManagers/UplinkSupergraphManager/__tests__/UplinkSupergraphManager.test.ts
```diff
@@ -0,0 +1,65 @@
+import mockedEnv from 'mocked-env';
+
+import { UplinkSupergraphManager } from '@apollo/gateway';
+
+import { DEFAULT_UPLINK_ENDPOINTS } from '../index';
```
trevor-scheer, line 5: In case you're unaware, this is fine too. If what you have is a preference then by all means, please keep it as is.
```suggestion
import { DEFAULT_UPLINK_ENDPOINTS } from '..';
```

gateway-js/src/supergraphManagers/UplinkSupergraphManager/loadSupergraphSdlFromStorage.ts
```diff
@@ -135,6 +142,7 @@ export async function loadSupergraphSdlFromStorage({
   const startTime = new Date();
   let result: FetcherResponse;
   try {
+    logger?.debug(`🔧 Fetching supergraph schema from ${endpoint}`);
```
trevor-scheer, line 145: I'd be very happy to see these logs expected in tests somewhere

trevor-scheer: This is not to say I was ever good about this in the past.

gateway-js/src/index.ts
```diff
@@ -194,23 +191,8 @@ export class ApolloGateway implements GraphQLService {
...
-    // And also support the `debug` option, if it's truthy.
-    if (this.config.debug === true) {
-      loglevelLogger.setLevel(loglevelLogger.levels.DEBUG);
-    } else {
-      loglevelLogger.setLevel(loglevelLogger.levels.WARN);
-    }
-
-    return loglevelLogger;
+  public get supergraphManager(): SupergraphManager | undefined {
+    return this._supergraphManager;
```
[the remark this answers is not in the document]
benweatherman: I like that thinking. If we make this a `public readonly supergraphManager` on the class, we can't set it during `Gateway.load`. It was nice to have for testing, but it's a bit of a leaky abstraction.

Maybe it's more appropriate to include in `__testing`

gateway-js/src/supergraphManagers/UplinkSupergraphManager/loadSupergraphSdlFromStorage.ts
benweatherman: I tend to think of logging as an after-effect for most things, so long as there's a good test for the actual functionality we're logging about (and there is for this stuff at least). So I tend to leave it out of tests. Though I don't feel strongly and happy to add them if there's even a slight preference.

gateway-js/src/index.ts
```diff
@@ -381,19 +366,21 @@ export class ApolloGateway implements GraphQLService {
...
-          maxRetries:
-            this.config.uplinkMaxRetries ?? uplinkEndpoints.length * 3 - 1, // -1 for the initial request
-          subgraphHealthCheck: this.config.serviceHealthCheck,
-          fetcher: this.fetcher,
+          shouldRunSubgraphHealthcheck: this.config.serviceHealthCheck,
+          uplinkEndpoints:
+            this.config.uplinkEndpoints ?? schemaDeliveryEndpoints,
+          maxRetries: this.config.uplinkMaxRetries,
           logger: this.logger,
-          fallbackPollIntervalInMs: this.pollIntervalInMs ?? 10000,
```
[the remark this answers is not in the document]
benweatherman: Good call, I'll fall back to `fallbackPollIntervalInMs`. It's _kinda_ sensical for this case where Uplink fails and we need _something_ to tell us how long to wait. We expose in as a Gateway config so we should probably continue to support it here. Though maybe it's better to support a `overridePollIntervalMs` that is always used if set?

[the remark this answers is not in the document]
trevor-scheer: Yeah, in the case that uplink is failing, then I suppose this config pairs nicely with your new functionality :)

I think we explicitly _don't_ want an override option. My understanding of the Uplink-controlled interval is that it was introduced for the safety of our infra; i.e., in the event of an outage we can reduce the frequency (as much as is in our control) at which a ton of gateways suddenly coming online are polling for a supergraph until things stabilize. This is a good @Y-Guo or @pcarrier question if you want a real answer.

gateway-js/src/supergraphManagers/UplinkSupergraphManager/loadSupergraphSdlFromStorage.ts
trevor-scheer: With or without is fine by me. My preference towards testing them is just to ensure the code I wrote does what I expect and someone doesn't change that without being intentional about it. I do agree that logs (especially debug) are considerably lower consequence though, and wouldn't block on this.

gateway-js/src/supergraphManagers/UplinkSupergraphManager/index.ts
```diff
@@ -179,9 +179,7 @@ export class UplinkSupergraphManager implements SupergraphManager {
       this.logger.debug(
         'Error fetching supergraphSdl from Uplink, calling updateSupergraphSdlFailureCallback',
       );
-      supergraphSdl = await this.onFailureToFetchSupergraphSdl.call(this, {
-        error: e,
-      });
+      supergraphSdl = await this.onFailureToFetchSupergraphSdl({ error: e });
```
trevor-scheer, line 176: I can't leave the comment where I wanted to, but I expected a `.bind(this)` call on L96 now:

```ts
this.onFailureToFetchSupergraphSdl = onFailureToFetchSupergraphSdl.bind(this);
```

Since this is a user-provided callback, by default the `this` within their function would depend on the context (in the general sense) of where and how they wrote the function.

The typing you've provided:
```ts
export type FailureToFetchSupergraphSdlFunction = (
  this: UplinkSupergraphManager,
  // ...
)
```
...indicates to me that you want them to have access to _the `UplinkSupergraphManager` instance_ which seems reasonable to me, though I'd be interested to know more about the expected use case is these. It might not be _that_ helpful (but I'd be happy to be wrong).

I digress. Let's assume they need a handle on that instance. We've established that the default `this` within the callback currently depends on the context of where that callback was written. But really we want it to be the class instance. This is where `.bind()` becomes useful. We can decide what `this` is from inside of the function that the _user_ handed off to us. Whatever we provide as the first argument to bind is what `this` is from inside the function.

That leaves one open question: how do we tell the function what it's _overridden_ `this` type actually is? Exactly what you have above - we provide a `this: UplinkSupergraphManager` as the "first" argument to a function. At runtime it's actually _not_ the first argument, so this is a special way to indicate what the expected `this` is for a function type.

If you already knew all this stuff, well, don't mind me. If you're unsure of anything I'm happy to discuss!

benweatherman: Ha, I thought your previous comment was saying the `.call` was unnecessary since passing the callback is all done within the context of `new UplinkSupergraphManager` constructor and would already know the right `this` (which will probably _usually_ be the case, but will depend on how users construct it).

Were you more saying we should call `.bind` once in the constructor instead of `.call` each time it's called?

trevor-scheer: I had you all wrong (my recollection of `.call()` was incorrect!). What you had to begin with was fine after all, and what I'm suggesting turns out to be just a minor nitpick that accomplishes the same. Sorry for the detour.

gateway-js/src/index.ts
benweatherman: Well, after some digging, the cause of the error was we ended up passing `null` as the supergraph schema if we got a falsy supergraph schema during load (i.e. we'd make a `""` into a `null`). So that was the cause of the weird parsing errors.

I've fixed those up and added tests for the types of errors in https://github.com/apollographql/federation/pull/1812/commits/c62b4a04c8c977f31a7154bbe4421e5854e1ca34

gateway-js/src/__tests__/integration/managed.test.ts
```diff
@@ -0,0 +1,176 @@
...
+        async onFailureToFetchSupergraphSdl(
+          this: UplinkSupergraphManager,
+          { error }: { error: Error },
+        ) {
+          this.logger.info(error);
+          return supergraphSchema;
+        },
+      }),
+    });
+
```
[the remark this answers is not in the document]
benweatherman: Should be fixed in the latest

gateway-js/src/supergraphManagers/UplinkSupergraphManager/index.ts
benweatherman: All good! As we discussed IRL, doing the 1x `.bind` at construction time is probably better than `.call` each time. Updated in the latest.

gateway-js/src/supergraphManagers/UplinkSupergraphManager/loadSupergraphSdlFromStorage.ts
benweatherman: I'm gonna leave out of the tests for now

gateway-js/src/__tests__/integration/managed.test.ts
```diff
@@ -0,0 +1,176 @@
...
+        logger,
+        supergraphSdl: new UplinkSupergraphManager({
+          apiKey,
+          graphRef,
+          logger,
+          maxRetries: 0,
+          pollIntervalInMs: 0,
+          async onFailureToFetchSupergraphSdl(
+            this: UplinkSupergraphManager,
+            { error: _error }: { error: Error },
```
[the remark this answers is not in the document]
benweatherman: Updated in the latest

gateway-js/src/index.ts
```diff
@@ -295,7 +277,10 @@ export class ApolloGateway implements GraphQLService {
       const { key, keyHash, graphRef, graphId, graphVariant } = options.apollo;
       this.apolloConfig = {
         key,
-        keyHash,
+        keyHash:
```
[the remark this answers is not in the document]
benweatherman: This is holdover from my early work (I was getting a warning about a config conflict, but the updated logic to `isManagedConfig` is the right fix). Reverted this change in the latest commits.

gateway-js/src/supergraphManagers/UplinkSupergraphManager/index.ts
```diff
@@ -0,0 +1,225 @@
...
+      ({ supergraphSdl, minDelaySeconds } = result);
+    } catch (e) {
+      if (!this.onFailureToFetchSupergraphSdl) {
+        throw e;
+      }
+
+      this.logger.debug(
+        'Error fetching supergraphSdl from Uplink, calling updateSupergraphSdlFailureCallback',
+      );
+      supergraphSdl = await this.onFailureToFetchSupergraphSdl({ error: e });
```
lennyburdette, line 176: Is there a way to tell if `onFailureToFetchSupergraphSdl` is called on startup, versus on a subsequent poll?

Most of the time, we want to provide a fallback supergraph _only on startup_. If Uplink is unreachable during a deploy or restart, the pod will crash. But it's fine if Uplink is unavailable for some period of time after the gateway is already running. 

I'm not sure I'd want to fallback to (possibly older) cached supergraph on momentary blips. I just want to make sure the pods can start up.

lennyburdette: Actually, now that I think about it — should there be a way for the `onFailureToFetchSupergraphSdl` to "no-op"? What happens if it doesn't return a string? I'd like a way to just let things poll after the gateway has started.

```js
onFailureToFetchSupergraphSdl({ error, hasStarted, graphRef }) {
  logger.warn(error);

  if (!hasStarted) {
    return fetchSupergraphFromS3(graphRef);
  }

  return null; // let the gateway continue polling until Uplink is available again
}
```

Or maybe the type of error could help indicate the state of the gateway?

```js
switch (error.type) {
  case 'UplinkPollFailed':
  case 'GatewayStartedFailed':
}
```

lennyburdette: lol jinx

benweatherman: Awesome feedback, thank you!

I added callbacks for `onFailureToFetchSupergraphSdlDuringInit` and `onFailureToFetchSupergraphSdlAfterInit`, with the biggest difference being the latter can `return null` to signify to the gateway to reuse the existing schema (other than being called at different times).

I made these the params to the callbacks, but happy to expose other info that's needed/makes sense

```ts
export type FailureToFetchSupergraphSdlFunctionParams = {
  error: Error;
  graphRef: string;
  logger: Logger;
  fetchCount: number;
};
```

lennyburdette: sounds great! 👍🏻

gateway-js/src/supergraphManagers/UplinkSupergraphManager/index.ts
```diff
@@ -0,0 +1,294 @@
...
+        roundRobinSeed: this.fetchCount++,
+        earliestFetchTime: this.earliestFetchTime,
+        logger: this.logger,
+      });
+
+      if (!result) {
+        return null;
+      }
+
+      this.compositionId = result.id;
```
clenfest, line 205: What is this used for?

gateway-js/src/supergraphManagers/UplinkSupergraphManager/index.ts
```diff
@@ -0,0 +1,294 @@
...
+
+    if (this.state.phase !== 'polling') {
+      return;
+    }
+
+    this.state.nextFetchPromise = resolvable();
+
+    this.timerRef = setTimeout(async () => {
+      if (this.state.phase === 'polling') {
+        let pollingPromise = resolvable();
```
clenfest, line 266: use const here?

gateway-js/src/logger.ts
```diff
@@ -0,0 +1,11 @@
+import loglevel from 'loglevel';
+import type { Logger } from '@apollo/utils.logger';
+
+export function getDefaultLogger(debug: Boolean = true): Logger {
```
clenfest, line 4: use `boolean` instead of `Boolean`

gateway-js/src/supergraphManagers/UplinkSupergraphManager/index.ts
```diff
@@ -0,0 +1,294 @@
...
+    this.onFailureToFetchSupergraphSdlAfterInit =
+      onFailureToFetchSupergraphSdlAfterInit;
+
+    this.state = { phase: 'constructed' };
+  }
+
+  public async initialize({ update, healthCheck }: SupergraphSdlHookOptions) {
+    this.update = update;
+
+    if (this.shouldRunSubgraphHealthcheck) {
```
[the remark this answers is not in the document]
benweatherman: The `initialize` function is called by the gateway, so we'll always get the function in the callback, and the types will make sure the compiler barks if we forget to pass it in https://github.com/apollographql/federation/blob/fb2bbc9f6a43278226a177f873ec8076b5158348/gateway-js/src/index.ts#L409-L413

gateway-js/src/supergraphManagers/UplinkSupergraphManager/index.ts
```diff
@@ -0,0 +1,294 @@
...
+      shouldRunSubgraphHealthcheck ?? this.shouldRunSubgraphHealthcheck;
+    this.onFailureToFetchSupergraphSdlDuringInit =
+      onFailureToFetchSupergraphSdlDuringInit;
+    this.onFailureToFetchSupergraphSdlAfterInit =
+      onFailureToFetchSupergraphSdlAfterInit;
+
+    this.state = { phase: 'constructed' };
+  }
+
+  public async initialize({ update, healthCheck }: SupergraphSdlHookOptions) {
```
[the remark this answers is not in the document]
benweatherman: It's not a bad idea, but it'd be tough to change now because we've published this type already

gateway-js/src/supergraphManagers/UplinkSupergraphManager/index.ts
benweatherman: It's passed into subsequent requests so we only get updates after this `compositionId` https://github.com/apollographql/federation/blob/cc7b6dabc54a261d7071d9331f06d1250cda8357/gateway-js/src/supergraphManagers/UplinkSupergraphManager/loadSupergraphSdlFromStorage.ts#L122-L129

gateway-js/src/supergraphManagers/UplinkSupergraphManager/index.ts
```diff
@@ -0,0 +1,294 @@
...
+    let initialSupergraphSdl: string | undefined = undefined;
+    try {
+      const result = await this.updateSupergraphSdl();
+      if (!result) {
+        throw new Error(
+          'Invalid supergraph schema supplied during initialization.',
+        );
+      }
+      initialSupergraphSdl = result.supergraphSdl;
+      if (result?.minDelaySeconds) {
```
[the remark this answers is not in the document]
benweatherman: If `this.minDelaySeconds` is falsey, it'll use `this.fallbackPollIntervalMs` https://github.com/apollographql/federation/blob/e61c633ddbeba251f791943cc7733319cba1aa80/gateway-js/src/supergraphManagers/UplinkSupergraphManager/index.ts#L254-L256.

Maybe it would be more clear to use `result.minDelaySeconds || this.fallbackPollIntervalMs` here?

lleadbet:
As a small comment/request- would it be possible to get the timestamp of the cachedSDL's last update for the `onFailureToFetchSupergraphSdlAfterInit`? To @lennyburdette's earlier point, serving outdated schemas would be bad, but if users could detect when it was last updated, they may be able to then confirm a new schema may be available via fallback.

gateway-js/src/supergraphManagers/UplinkSupergraphManager/index.ts
```diff
@@ -0,0 +1,294 @@
...
+    }
+    return this.state.nextFetchPromise;
+  }
+
+  private async updateSupergraphSdl(): Promise<{
+    supergraphSdl: string;
+    minDelaySeconds?: number;
+  } | null> {
+    let supergraphSdl;
+    let minDelaySeconds: number | undefined =
```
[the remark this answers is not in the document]
benweatherman: Had to rejigger some types, but it's doable. Thanks for the nudge.

gateway-js/src/supergraphManagers/UplinkSupergraphManager/index.ts
benweatherman: Gonna stick with the "pure function" approach. Happy to revisit if that seems like a path of sadness.

gateway-js/src/index.ts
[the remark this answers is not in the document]
benweatherman: Gonna keep this here. It seems potentially valuable and can't be a property due to it not being created in the constructor.

gateway-js/src/supergraphManagers/UplinkSupergraphManager/index.ts
```diff
@@ -0,0 +1,294 @@
...
+    shouldRunSubgraphHealthcheck?: boolean;
+    onFailureToFetchSupergraphSdlDuringInit?: FailureToFetchSupergraphSdlDuringInit;
+    onFailureToFetchSupergraphSdlAfterInit?: FailureToFetchSupergraphSdlAfterInit;
+  }) {
+    this.apiKey = apiKey;
+    this.graphRef = graphRef;
+    this.logger = logger ?? getDefaultLogger(debug);
+
+    this.uplinkEndpoints = uplinkEndpoints ?? this.uplinkEndpoints;
+    this.maxRetries = maxRetries ?? this.uplinkEndpoints.length * 3 - 1;
```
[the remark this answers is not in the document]
benweatherman: Added comment https://github.com/apollographql/federation/blob/0434ecd969029ed8000a72e7b48a1b6e97e334e8/gateway-js/src/supergraphManagers/UplinkSupergraphManager/index.ts#L119-L121

gateway-js/src/logger.ts
benweatherman: Updated in the latest

benweatherman:
@lleadbet Thanks for the suggestion! I added a `mostRecentSuccessfulFetchAt` param. Does that give you the info you're looking for?

gateway-js/src/supergraphManagers/UplinkSupergraphManager/index.ts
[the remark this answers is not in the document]
benweatherman: Since this was existing code, I'm inclined to leave it as-is. We can refactor to clear this up sometime down the line. Lemme know if that seems shortsighted.

gateway-js/src/supergraphManagers/UplinkSupergraphManager/index.ts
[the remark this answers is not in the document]
benweatherman: Lemme know if you think there's more we can do to clear this up

gateway-js/src/supergraphManagers/UplinkSupergraphManager/index.ts
benweatherman: Lemme know if there's anything else unclear around this

glasser:
Was it intentional that this PR changed the default log level to DEBUG?

benweatherman:
Nope, good catch! Documented and fixed as part of this issue https://github.com/apollographql/federation/issues/2047

PR #1817, opened 2022-05-02
export checkFeatureSupport

Geal:
this is needed to implement the unsupported features check in router-bridge: https://github.com/apollographql/federation-rs/pull/111


Geal:
not needed anymore, thanks

PR #1818, opened 2022-05-02
Remove superfluous commits in release process

benweatherman:
Previously, we were creating a commit for changelogs and for version bumping. It's much nicer to have those show up as a single commit for a clean commit history.

Now, we'll wait to commit the changelogs until lerna bumps the version. This simplifies the release process and makes the commit history cleaner.

### Other detritus

- Added instructions for setting the right dist tag for non-release vs release publishing
- Added a script for checking published versions

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -2,36 +2,35 @@
 
 This is a quick and opinionated set of commands for building and releasing a new version of the packages in this monorepo.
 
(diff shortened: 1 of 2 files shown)


pcmanus (approved):
Haven't actually tested, but I like the changes and didn't spot anything that looked wrong.

PR #1821, opened 2022-05-03
Fix bug with type extension of empty type definition

pcmanus:
It is valid graphQL to do:
```graphql
type Foo

extend type Foo {
  bar: Int
}
```
and while that's uncommon, this could be the convenient when mechanically concatenating files.

This wasn't handled correctly however as the code was not preserving the empty type definition and was thus considering the schema as just:
```graphql
extend type Foo {
  bar: Int
}
```
which, when trying to compose this, may lead into an error complaining that the type is extended but has no definition (which is obviously incorrect).

### DIFF
--- a/internals-js/src/__tests__/subgraphValidation.test.ts
+++ b/internals-js/src/__tests__/subgraphValidation.test.ts
@@ -878,6 +878,30 @@ describe('@core/@link handling', () => {
     // Just making sure this don't error out.
     buildAndValidate(doc);
   });
+
+  it('allows defining a repeatable directive as non-repeatable but validates usages', () => {
+    const doc = gql`
+      type T @key(fields: "k1") @key(fields: "k2") {
+        k1: ID!
+        k2: ID!
+      }
+
+      directive @key(fields: String!) on OBJECT
+    `;
+
+
+    // Test for fed2 (with @key being @link-ed)
+    expect(buildForErrors(doc)).toStrictEqual([[
+      'INVALID_GRAPHQL',
+      '[S] The directive "@key" can only be used once at this location.',
+    ]]);
+
+    // Test for fed1
+    expect(buildForErrors(doc, { asFed2: false })).toStrictEqual([[
+      'INVALID_GRAPHQL',
+      '[S] The directive "@key" can only be used once at this location.',
+    ]]);
+  });
 });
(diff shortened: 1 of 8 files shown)


internals-js/src/__tests__/subgraphValidation.test.ts
```diff
@@ -878,6 +878,30 @@ describe('@core/@link handling', () => {
     // Just making sure this don't error out.
     buildAndValidate(doc);
   });
+
+  it('allows defining a repeatable directive as non-repeatable but validates usages', () => {
```
pcmanus, line 882: Reviewer notes: this test is unrelated to the issue and it wasn't meant to be here, but I just wrote it earlier (to prove to myself that the behaviour of the test does work as the test shows) and didn't noticed it was there. But seeing it now, I'm thinking: "it's a nice test to have and is it really worth opening a separate PR for this?". Happy to remove though if it bothers someone.

internals-js/src/buildSchema.ts
```diff
@@ -171,17 +171,46 @@ function buildNamedTypeAndDirectivesShallow(documentNode: DocumentNode, schema:
...
+        let type = schema.type(definitionNode.name.value);
+        // Note that the type may already exists due to an extension having been processed first, but we know we
+        // have seen 2 definitions (which is invalid) if the definition has `preserverEmptyDefnition` already set
+        // since it's only set for definitions, not extensions. 
+        // Also note that we allow to redefine built-ins.
+        if (!type || type.isBuiltIn) {
+          type = schema.addType(newNamedType(withoutTrailingDefinition(definitionNode.kind), definitionNode.name.value));
+        } else if (type.preserveEmptyDefinition)  {
+          // Note: we reuse the same error message than graphQL-js would output
+          throw new GraphQLError(`There can be only one type named "${definitionNode.name.value}"`);
```
pcmanus, line 183: Reviewer note: writing this, I realised that we weren't validating this properly. That is:
```graphql
type Foo {
  bar: Int
}

type Foo {
  baz: Int
}
```
was not erroring out, even though it should as it's not valid graphQL. So took the liberty to fix as well (and there is a test for it).

benweatherman (approved):
🐐

PR #1830, opened 2022-05-05
Allow `RemoteGraphQLDataSource` to use custom fetcher

clenfest:
- Allow RemoteGraphQLDataSource to use a provided fetcher
- Use the Gateway's fetcher in RemoteGraphQLDataSource if a custom `buildService` function is not provided.


clenfest:
This PR is unnecessary because `RemoteGraphQLDataSource` already supports this with the subsequent `Object.assign`

PR #1831, opened 2022-05-05
Fix `printSubgraphSchema` method.

pcmanus:
This PR has 2 main fixes:
1. `buildSubgraphSchema` was failing if a subgraph was defining the `link__Purpose` type but not including the `@link` directive definition, and this with a confusing error message (one that didn't make sense for the user and didn't help you understand what to fix). While fixing the error was an option, we're generally lenient toward a user provided parts of the definitions of a core feature but not others, and so it makes sense to simply make it for for `@link` itself as well.
2. the output of `printSubgraphSchema` in fed2 is currently fairly unhelpful and illogical. It includes some federation directives but not others, skip some federation private parts but include others, etc. Some of it also makes `buildSubgraphSchema` fail on the output of that method. The patch tries to fix `printSubgraphSchema` so it outputs "the schema as the user would usually write it", that is it skip all auto-imported directive definition (and their dependencies) consistently, and additionally skip the auto-added federation semi-private type and fields (`_service` and friends).


benweatherman (approved):
🐐

PR #1835, opened 2022-05-06
Fix issue with `@requires` and conditional queries (`@include` and `@skip`)

pcmanus:
Writing a few unit test around the interaction `@require` and `@include`/`@skip`, I noticed 2 issues/inefficiencies. See the commits for more details.


pcmanus:
Added a 3rd commit to this PR as a 3rd issue around `@requires` and `@skip`/`@include` was found. That later issue is due to multiple `@skip` and/or `@include` being involved and actually leads to inefficient plans where extra obviously-useless fetches are kept instead of being optimised out.

PR #1840, opened 2022-05-10
Warn on merging inconsistent non-repeatable directive applications in…

### LINKED ISSUE #1804: Composing subgraphs with different @deprecated args should succeed with a hint instead of error
## Behavior

Composing subgraphs with different `reason` arguments for `@deprecated` cause a composition error:

```ts
import { composeServices } from '@apollo/composition';

const service1 = {
  name: 'sad-service',
  typeDefs: gql`
    type Query {
      hi: String! @deprecated(reason: "it's sad")
    }
  `,
};
const service2 = {
  name: 'mad-service',
  typeDefs: gql`
    type Query {
      hi: String! @deprecated(reason: "it's mad")
    }
  `,
};
const { supergraphSdl, errors, hints } = composeServices([service1, service2]);
console.error(errors);
console.warn(hints);
console.info(supergraphSdl);
```

### Expected behavior

I'd expect those schemas to compose. A potential output could be the the concatenation of the service names and their respective reasons:

```
sad-service: it's sad
mad-service: it's mad
```

### Version

I'm using federation 2.0.1

pcmanus:
…stead of erroring

When, for a given element, a non-repeatable directive having arguments
appeared with inconsistent arguments in different subgraph, we used to
error out. Instead, this commit picks one of the application for the
supergraph and issue a warning that the other applications were ignored.

The version picked follows the same algorithm that for the similar case
with description: if a given application is repeated in multiple
subgraphs, more than other applications, we pick that version.
Otherwise, we pick the application of the first subgraph in alphabetical
order.

Additionally, this commit ensures that we treat an non-provided argument
with a default the same than that if that argument is provided but uses
exactly the default value.

Fixes #1804


benweatherman (approved):
🐐

PR #1859, opened 2022-05-16
Fix bug when field covariance meet type-explosion

### LINKED ISSUE #1858: Certain queries error when Federation v1 supergraph has covariance in field types
**Describe the bug**
Certain queries trigger a runtime error when a Federation v1 supergraph uses covariance in object/interface field return types.

**To Reproduce**
When using a single Federation v1 subgraph with this schema:
```graphql
type Query {
  dummy: Interface
}

interface Interface {
  field: Interface
}

type Object implements Interface {
  field: Object
}
```
Federation v1 composition yields this supergraph:
```graphql
schema @core(feature: "https://specs.apollo.dev/core/v0.1") @core(feature: "https://specs.apollo.dev/join/v0.1") {
  query: Query
}

directive @core(feature: String!) repeatable on SCHEMA

directive @join__field(graph: join__Graph, requires: join__FieldSet, provides: join__FieldSet) on FIELD_DEFINITION

directive @join__type(graph: join__Graph!, key: join__FieldSet) repeatable on OBJECT | INTERFACE

directive @join__owner(graph: join__Graph!) on OBJECT | INTERFACE

interface Interface {
  field: Interface
}

scalar join__FieldSet

enum join__Graph {
  SUBGRAPH @join__graph(name: "subgraph", url: "http://localhost:4001/")
}

type Object implements Interface {
  field: Object
}

type Query {
  dummy: Interface @join__field(graph: SUBGRAPH)
}
```
When router is configured with this supergraph, the following query causes a runtime error:
```graphql
query {
  dummy {
    field {
      ... on Object {
        field {
          __typename
        }
      }
    }
  }
}
```
The error seen by clients is:
```json
{
  "errors": [
    {
      "message": "value retrieval failed: router bridge error: the deno runtime raised an error: `request: couldn't receive response couldn't deserialize payload `id: fcf80d4b-5630-4fd3-ad3c-deb78d920671`: `deno: couldn't deserialize response : `Error(\"missing field `usageReporting`\", line: 0, column: 0)``.`",
      "locations": [],
      "path": null
    }
  ]
}
```
The router logs show:
```
WARN router_bridge::worker: an error happened in the worker runtime Cannot add selection of field "Object.field" to selection set of parent type "Interface"
ERROR request{method=POST uri=/ version=HTTP/1.1 otel.kind=server otel.status_code= otel.status_code="ERROR"}: tower_http::trace::on_failure: response failed classification=Status code: 500 Internal Server Error latency=18 ms
```

**Expected behavior**
The query above is valid GraphQL, and the router should not fail to parse/validate it.

**Desktop (please complete the following information):**
 - OS: MacOS (Intel)
 - Router Version: v0.9.0

pcmanus:
As we build selection sets during query planning, and due to how query
planning work, we have cases where we try to add an interface field (as
coming from the original query) to a selection set whose parent type is
one of the interface implementation. As we do, the code recognize this
case and make sure to use the proper implementation field instead (so
that the resulting selection remains a well-typed object). But that
code was slightly buggy in 2 small ways (we updated the field added,
but didn't update its sub-selection set accordindly, and
`SelectionSet#add` method was returning the original selection instead
of the added (and potentially updated) one). This fixes those issues.

Fixes #1858

### DIFF
--- a/gateway-js/CHANGELOG.md
+++ b/gateway-js/CHANGELOG.md
@@ -7,6 +7,7 @@ This CHANGELOG pertains only to Apollo Federation packages in the 2.x range. The
 - Fix bug with type extension of empty type definition [PR #1821](https://github.com/apollographql/federation/pull/1821)
 - Fix output of `printSubgraphSchema` method, ensuring it can be read back by composition and `buildSubgraphSchema` [PR #1831](https://github.com/apollographql/federation/pull/1831).
 - Fix issue with `@requires` and conditional queries (`@include`/`@skip`) [1835](https://github.com/apollographql/federation/pull/1835).
+- Fix bug with field covariance when the underlying plan use type-explosion [1859](https://github.com/apollographql/federation/pull/1859).
 
 ## 2.0.2
 
--- a/gateway-js/src/__tests__/executeQueryPlan.test.ts
+++ b/gateway-js/src/__tests__/executeQueryPlan.test.ts
@@ -2993,5 +2993,4 @@ describe('executeQueryPlan', () => {
         `);
     });
   });
-
 });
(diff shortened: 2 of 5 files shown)


clenfest (approved):
LGTM

PR #1865, opened 2022-05-17
Add v2 directive to spec code block

kdawgwilk:
Adds v2 directives to spec code block


benweatherman (approved):
Thanks for making this update @kdawgwilk! 🐐

PR #1871, opened 2022-05-19
docs: update @override docs with multiple usage note

smyrick:
The directive is meant to only be used once on a field in the supergraph, so usage over multiple subgraphs will throw a composition error.

Example of composition errors: https://stackblitz.com/edit/apollo-basic-fed-demo-csza64?file=b.js,a.js

### DIFF
--- a/docs/source/federation-spec.md
+++ b/docs/source/federation-spec.md
@@ -377,7 +377,7 @@ type User @key(fields: "email") @shareable {
 directive @override(from: String!) on FIELD_DEFINITION
 ```
 
-The `@override` directive is used to indicate that the current subgraph is taking responsibility for resolving the marked field _away_ from the subgraph specified in the `from` argument. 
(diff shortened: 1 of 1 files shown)


docs/source/federation-spec.md
```diff
@@ -377,7 +377,7 @@ type User @key(fields: "email") @shareable {
 directive @override(from: String!) on FIELD_DEFINITION
 ```
 
-The `@override` directive is used to indicate that the current subgraph is taking responsibility for resolving the marked field _away_ from the subgraph specified in the `from` argument. 
+The `@override` directive is used to indicate that the current subgraph is taking responsibility for resolving the marked field _away_ from the subgraph specified in the `from` argument. The directive is meant to only be used once on a field in the supergraph, so usage over multiple subgraphs will throw a composition error.
```
StephenBarlow, line 380:
```suggestion
The `@override` directive is used to indicate that the current subgraph is taking responsibility for resolving the marked field _away_ from the subgraph specified in the `from` argument.

> Only one subgraph can `@override` any given field. If multiple subgraphs attempt to `@override` the same field, a composition error occurs.
```

StephenBarlow (approved):
Approved with suggestion

PR #1873, opened 2022-05-20
Fix issue when using a type condition on an inaccessible type in @req…

### LINKED ISSUE #1866: Fed 2 Gateway doesn't provide full selection set to requirer when `@requires` uses an `@inaccessible` object type
**Description**
When an object type is `@inaccessible` and is used in an `@requires`, it can cause Fed 2 gateway to give the wrong selection set to the requiring subgraph (it seems to query the right data from the provider, oddly enough). Note that I'm not sure what the precise conditions are for the selection set to be missing data.

**Example**
Suppose you're running the following two subgraphs, with implementations and schemas shown below:
```typescript
import { ApolloServer, gql } from "apollo-server";
import { buildSubgraphSchema } from "@apollo/subgraph";
import { GraphQLSchemaModule } from "@apollo/subgraph/src/schema-helper/resolverMap";

const typeDefs = gql`
  extend schema
  @link(
    url: "https://specs.apollo.dev/federation/v2.0",
    import: ["@key", "@shareable"]
  )

  type Query @shareable {
    dummy: Entity
  }
  
  type Entity @key(fields: "id") {
    id: ID!
    data: Foo
  }

  interface Foo {
    foo: String!
  }
  
  interface Bar implements Foo {
    foo: String!
    bar: String!
  }
  
  type Baz implements Foo & Bar @shareable {
    foo: String!
    bar: String!
    baz: String!
  }

  type Qux implements Foo & Bar @shareable {
    foo: String!
    bar: String!
    qux: String!
  }
`;

const resolvers: GraphQLSchemaModule["resolvers"] = {
  Query: {
    dummy() {
      return {};
    },
  },
  Entity: {
    __resolveReference() {
      return {};
    },
    id() {
      return "id";
    },
    data() {
      return {
        __typename: "Baz",
        foo: "foo",
        bar: "bar",
        baz: "baz",
      };
    },
  },
};

const server = new ApolloServer({
  schema: buildSubgraphSchema({
    typeDefs,
    resolvers,
  }),
});

server.listen(4002).then(({ url }) => {
  console.log(`Subgraph Data started at ${url}`);
});

```

```typescript
import { ApolloServer, gql } from "apollo-server";
import { buildSubgraphSchema } from "@apollo/subgraph";
import { GraphQLSchemaModule } from "@apollo/subgraph/src/schema-helper/resolverMap";

const typeDefs = gql`
  extend schema
  @link(
    url: "https://specs.apollo.dev/federation/v2.0",
    import: ["@key", "@shareable", "@external", "@requires", "@inaccessible"]
  )

  type Query @shareable {
    dummy: Entity
  }

  type Entity @key(fields: "id") {
    id: ID!
    data: Foo @external
    requirer: String! @requires(fields: "data { foo ... on Bar { bar } ... on Baz { baz } ... on Qux { qux } }")
  }

  interface Foo {
    foo: String!
  }

  interface Bar implements Foo {
    foo: String!
    bar: String!
  }

  type Baz implements Foo & Bar @shareable @inaccessible {
    foo: String!
    bar: String!
    baz: String!
  }

  type Qux implements Foo & Bar @shareable {
    foo: String!
    bar: String!
    qux: String!
  }
`;

const resolvers: GraphQLSchemaModule["resolvers"] = {
  Query: {
    dummy() {
      return {};
    },
  },
  Entity: {
    __resolveReference(representation) {
      console.log(`representation: ${JSON.stringify(representation, null, 2)}`);
      return {};
    },
    id() {
      return "id";
    },
    requirer() {
      return "requirer";
    },
  },
};

const server = new ApolloServer({
  schema: buildSubgraphSchema({
    typeDefs,
    resolvers,
  }),
});

server.listen(4003).then(({ url }) => {
  console.log(`Subgraph Requirer started at ${url}`);
});

```
Running `composeServices()` from `@apollo/composition@2.0.2` yields this supergraph schema:
```graphql
schema
  @link(url: "https://specs.apollo.dev/inaccessible/v0.2", for: SECURITY)
{
  query: Query
}

directive @inaccessible on FIELD_DEFINITION | OBJECT | INTERFACE | UNION | ARGUMENT_DEFINITION | SCALAR | ENUM | ENUM_VALUE | INPUT_OBJECT | INPUT_FIELD_DEFINITION

directive @join__field(graph: join__Graph!, requires: join__FieldSet, provides: join__FieldSet, type: String, external: Boolean, override: String, usedOverridden: Boolean) repeatable on FIELD_DEFINITION | INPUT_FIELD_DEFINITION

directive @join__implements(graph: join__Graph!, interface: String!) repeatable on OBJECT | INTERFACE

directive @join__type(graph: join__Graph!, key: join__FieldSet, extension: Boolean! = false, resolvable: Boolean! = true) repeatable on OBJECT | INTERFACE | UNION | ENUM | INPUT_OBJECT | SCALAR

interface Bar implements Foo
  @join__implements(graph: DATA, interface: "Foo")
  @join__implements(graph: REQUIRER, interface: "Foo")
  @join__type(graph: DATA)
  @join__type(graph: REQUIRER)
{
  foo: String!
  bar: String!
}

type Baz implements Foo & Bar
  @join__implements(graph: DATA, interface: "Foo")
  @join__implements(graph: DATA, interface: "Bar")
  @join__implements(graph: REQUIRER, interface: "Foo")
  @join__implements(graph: REQUIRER, interface: "Bar")
  @join__type(graph: DATA)
  @join__type(graph: REQUIRER)
  @inaccessible
{
  foo: String!
  bar: String!
  baz: String!
}

type Entity
  @join__type(graph: DATA, key: "id")
  @join__type(graph: REQUIRER, key: "id")
{
  id: ID!
  data: Foo @join__field(graph: DATA) @join__field(graph: REQUIRER, external: true)
  requirer: String! @join__field(graph: REQUIRER, requires: "data { foo ... on Bar { bar } ... on Baz { baz } ... on Qux { qux } }")
}

interface Foo
  @join__type(graph: DATA)
  @join__type(graph: REQUIRER)
{
  foo: String!
}

scalar join__FieldSet

enum join__Graph {
  DATA @join__graph(name: "data", url: "http://localhost:4002/")
  REQUIRER @join__graph(name: "requirer", url: "http://localhost:4003/")
}

scalar link__Import

enum link__Purpose {
  """
  `SECURITY` features provide metadata necessary to securely resolve fields.
  """
  SECURITY

  """
  `EXECUTION` features provide metadata necessary for operation execution.
  """
  EXECUTION
}

type Query
  @join__type(graph: DATA)
  @join__type(graph: REQUIRER)
{
  dummy: Entity
}

type Qux implements Foo & Bar
  @join__implements(graph: DATA, interface: "Foo")
  @join__implements(graph: DATA, interface: "Bar")
  @join__implements(graph: REQUIRER, interface: "Foo")
  @join__implements(graph: REQUIRER, interface: "Bar")
  @join__type(graph: DATA)
  @join__type(graph: REQUIRER)
{
  foo: String!
  bar: String!
  qux: String!
}
```
When we run this supergraph schema with `@apollo/gateway@2.0.2` and execute this query:
```graphql
query {
  dummy {
    requirer
  }
}
```
we see this sub-query being made to the `data` subgraph:
```graphql
query ($representations: [_Any!]!) {
  _entities(representations: $representations) {
    ... on Entity {
      data {
        __typename
        foo
        ... on Baz {
          bar
          baz
        }
        ... on Qux {
          bar
          qux
        }
      }
    }
  }
}
```
And the `data` subgraph responds with a valid response:
```json
{
  "_entities": [
    {
      "data": {
        "__typename": "Baz",
        "foo": "foo",
        "bar": "bar",
        "baz": "baz"
      }
    }
  ]
}
```

However, the representation gateway gives to the `requirer` subgraph is erroneously missing `bar`:
```
representation: {
  "__typename": "Entity",
  "data": {
    "foo": "foo",
    "baz": "baz"
  },
  "id": "id"
}
```

pcmanus:
…uire

The `@require(fields:)` argument can use some type conditions to require
data specific to a subtype. And those condition can even be on an
`@inaccessible` type since "inaccessibility is about the federated API
schema, not about subgraphs.

However, this wasn't properly working because when the execution code
was extracting the "requirements" data to pass it to the subgraph having
the `@require`, the (federated) API schema we used, and consequently
when the code tried to get the definition for the type condition in the
`@require`, it didn't find any definition (since the type in question is
`@inaccessible` and so not in said API schema) and as a result the
corresponding fields were not included in the subgraph query.

This commit fixes this issue, ensuring we pass the full supergraph
(which _has_ the `@inaccessible` elements) to the execution code so
that supergraph can be used when extracting required data.

There was also another place within query planning (in `graphPath.ts`
more precisely) where the code was looking up the possible
implementation types, in the supergraph, of the required `@inaccessible`
type, and was using the supergraph API to do so, which was incorrect
for the same reason. This commit fix that problem as well, passing
the full supergraph instead.

Fixes #1866


composition-js/src/validate.ts
```diff
@@ -216,17 +215,40 @@ function generateWitnessValue(type: InputType): any {
...
+  subgraphs: QueryGraph
+): {
+  errors? : ValidationError[]
+} {
+  const errors = new ValidationTraversal(supergraphSchema, supergraphAPI, subgraphs).validate();
   return errors.length > 0 ? {errors} : {};
 }
 
-export function computeSubgraphPaths(supergraphPath: RootPath<Transition>, subgraphs: QueryGraph): {traversal?: ValidationState, isComplete?: boolean, error?: ValidationError} {
+export function computeSubgraphPaths(
```
pcmanus, line 239: Note to reviewer: if you look up, you may notice that this method is not used internally. But it's mean to be a handy tool to help validation at times (it's essentially about walking validation step-by-step).

gateway-js/src/index.ts
```diff
@@ -903,6 +905,7 @@ export class ApolloGateway implements GraphQLService {
             serviceMap,
             requestContext,
             operationContext,
+            this.supergraphSchema!,
```
pcmanus, line 908: Note to reviewer: I know this `!` is a tad "smelly", but this method does plenty of this earlier, with `this.apiSchema!` and `this.queryPlanner!` in particular, which are assigned in the same places than `this.supergraphSchema`, and I didn't exactly wanted to rework the entire lifecycle of the gateway in this patch.

gateway-js/src/executeQueryPlan.ts
```diff
@@ -52,6 +53,7 @@ export async function executeQueryPlan<TContext>(
   serviceMap: ServiceMap,
   requestContext: GraphQLRequestContext<TContext>,
   operationContext: OperationContext,
+  supergraphSchema: GraphQLSchema,
```
benweatherman, line 56: **question** Since we export this function, is adding this required param a breaking change that would require a major version bump?

gateway-js/src/index.ts
```diff
@@ -580,20 +581,21 @@ export class ApolloGateway implements GraphQLService {
...
-      this.updateWithSchemaAndNotify(schema, supergraphSdl);
+      this.updateWithSchemaAndNotify(supergraphSchema, supergraphSdl);
 
       if (this.experimental_didUpdateSupergraph) {
         this.experimental_didUpdateSupergraph(
           {
             compositionId: result.id,
             supergraphSdl,
-            schema: schema.toGraphQLJSSchema(),
+            schema: this.schema!,
```
benweatherman, line 598: **question** Should this be `this.supergraphSchema`? Previously it was "the thing returned by `createSchemaFromSupergraphSdl` with `toGraphQLJSSchema` called", and now it's gonna be "the thing returned by `createSchemaFromSupergraphSdl`"...right?

benweatherman, line 598: **question** Should this be `this.supergraphSchema`? Previously it was "the thing returned by `createSchemaFromSupergraphSdl` with `toGraphQLJSSchema` called", and now it's gonna be "the thing returned by `createSchemaFromSupergraphSdl`"...right?

benweatherman (commented):
Just a question about how we version this, but the functionality looks good

gateway-js/src/executeQueryPlan.ts
pcmanus: A really good question really.

I don't really disagree here, but I'd like to lay out a few reflexions.

First, the federation code base is full of exported methods, most of which are exported more just so we can organize the code internally than to indicate a publically maintained API, and I'm pretty sure we've routinely changed the signature of exported methods in the past without bumping the major version (maybe not in the `@apollo/gateway` module, but certainly in others). I guess my point here is that I'm not convinced "every export method is a public API with backward compatibility guarantees" is a tenable thing in our current reality. By which I really mean that we should probably discuss what exactly constitutes our public API, so 1) it's clear between ourselves and we avoid avoid repeating this kind of conversations too often, and 2) we can document this more clearly for users.

Anyway, in general, I don't think bumping federation to 3.0+ in the short/mid term is a good idea (frankly, I think it'll just create confusion), and it's especially not worth it for a small bug fix like this imo. But at the same time, _not_ fixing this bug feels even more silly and I do think using the supergraph schema is the correct fix here.

To sum up, I'd say our options are:
1. do bump federation to 3.0 short term. As said, I'm really not a fan, but it's an option.
2. just document the change in the changelog but commit this in 2.1 regardless. Pragmatically, I don't see many users calling `executeQueryPlan` manually. I mean, what's the use case? Writing your own gateway? Why would you do that? Don't get me wrong, I'm sure there could be a handful users calling it for some testing or simulation or something, but it's also not the kind of change that could break a running gateway in production. The risk here is annoying a few users that may have to update code they didn't expect to have to for a minor release, and that's not great of course, but insofar as we have to make a trade-off here, it's one to consider imo.
3. we work-around the backward incompatibility by making the new argument optional. It migth be a decent middle ground and I pushed a commit to do that. I don't think it's ideal though, and I'm a bit uncomfortable with it, because really, use *should* pass the argument or they'll run into the bugs we fix here, so making the argument option feels a big misleading, and in a sense, error prone.

Anyway, it's a lot of words, but that's because outside of this particular issue, I'd like for us to have the discussion around what our actual public API. In my mind, our public API is primarily the federation semantic, not the actual code API.

Focusing on this particular issue, I'm ok going with option 2 or 3 above. I still kind of prefer 2, but I understand the concern and have no qualm going with 3.

gateway-js/src/index.ts
pcmanus: Yeah good question as well.

This still pass a `GraphQLSchema` object, but it essentially switch from passing the supergraph to passing the API schema. The reason I did that is that the prior code stroke me as inconsistent. That is, `experimental_didUpdateSupergraph` takes 2 arguments of the form `new stuff, old stuff`, but the pre-existing code was passing the new supergraph schema for the new stuff but the API schema for the old stuff, so one of them was wrong in my book.

Now, I didn't find any place that would document which one it should definitively be, so it's everyone guess what users might actually expect. But from the POV of what is publicly exposed, `ApolloGateway.schema` is the API schema, so it kind of made a bit more sense to me that `CompositionInfo.schema` also be the API schema (add to that the fact that we also separately pass `supergraphSdl`, so if you care about the supergraph schema, you can just parse that).

But anyway, it is a small change of behaviour and I should have mentioned it, sorry. If we've uncomfortable doing it here, that's ok, I can switch back to the old (imo wrong but...) behaviour.

Fwiw, I suspect this doesn't matter too much. I think few users probably use `experimental_didUpdateSupergraph`, and those that do likely only care about the API schema (I mean, the supergraph schema is really mostly just the API schema + additional join spec directives that are not well documented, so I'm not sure how users could rely on the latter). It's also called "experimental" :).

Anyway, that was my reasoning here and I definitively should have been more explicit about it upfront. And I don't mind "reverting" the change if you prefer?

gateway-js/src/executeQueryPlan.ts
benweatherman: thanks for laying that out so thoroughly. i agree with all you wrote, including proceeding with option 2.

gateway-js/src/index.ts
benweatherman: as always, i appreciate the detailed response.

the previous code kinda made sense to me that we'd be getting both the supergraph SDL text as well as the object that represents the parsed version of said text so someone wouldn't need to re-parse if you care about the supergraph. though i suspect you're right that what users really want is the API schema.

another point toward keeping this change: i think the prevailing sentiment is that we'd deprecate this function and point users toward `onSchemaLoadOrUpdateListeners` (which btw does exactly what your change does e.g. provides the API schema as an object and the supergraph SDL as text).

**TL;DR** Let's change it!

gateway-js/src/executeQueryPlan.ts
pcmanus: Ok, thanks. I've reverted the commit for option 3 (thus sticking to option 2) and added a changelog entry that warns about the potential for breaking change for direct callers of `executeQueryPlan`.

benweatherman (approved):
🐐

PR #1878, opened 2022-05-21
feat: Expose sub-query DocumentNode to RemoteGraphQLDataSource#willSendRequest hook

### LINKED ISSUE #1352: Expose sub-query DocumentNode to RemoteGraphQLDataSource#willSendRequest hook
Currently sub-query generated for each `RemoteGraphQLDataSource` needed to complete the query is passed to `RemoteGraphQLDataSource` as a string, which limits possible interactions with the query on [#willSendRequest](https://www.apollographql.com/docs/federation/api/apollo-gateway/#willsendrequest) level, e.g. introducing custom permission checking on field/type level.

Generated `DocumentNode` is immediately serialized into string, see [buildPlan.ts#L803](https://github.com/apollographql/federation/blob/950eb931e38746bb7cfed05382d6970a22e43cc4/query-planner-js/src/buildPlan.ts#L803), and exposing it won't cause additional performance toll.

In order to achieve this I propose:
1. Updating interface [FetchNode](https://github.com/apollographql/federation/blob/1ffecef9ef52f6dc127939cdc375f0465774f093/query-planner-js/src/QueryPlan.ts#L32) to have `operation` field of type `DocumentNode` instead of `string`.
2.  Change [sendOperation](https://github.com/apollographql/federation/blob/950eb931e38746bb7cfed05382d6970a22e43cc4/gateway-js/src/executeQueryPlan.ts#L373) to also accept query as `DocumentNode` instead of `string`.
3. Change type `GraphQLDataSourceProcessOptions` to have `DocumentNode` included:
```
export type GraphQLDataSourceProcessOptions<
  TContext extends Record<string, any> = Record<string, any>,
> = {
  /**
   * The request to send to the subgraph.
   */
  request: GraphQLRequestContext<TContext>['request'];
} & {
  document: GraphQLRequestContext<TContext>['document'];
} & ( ...
```
4. Finally, change `GraphQLDataSource#process` calls to fit new `GraphQLRequestContext` definition:
```
    const response = await service.process({
      kind: GraphQLDataSourceRequestKind.INCOMING_OPERATION,
      request: {
        query: stripIgnoredCharacters(print(source)), // updated line
        variables,
        http,
      },
      document: source, // new line
      incomingRequestContext: context.requestContext,
      context: context.requestContext.context,
    });
```

theJC:
Fixes #1352

Initial proposed change to provide access to the AST so that customer code processing a subgraph query no longer has to waste cycles re-parsing it back into a document when needed.

At the moment there are failing tests that will need to be updated to include the additional operationDocumentNode in the returned but I did not want to undergo that effort until we were sure that this was an agreeable go-forward direction to support this capability.

CC: @lennyburdette - here is the MR per your recommendation


theJC:
Adjusted.

Whats your thoughts on the best way to approach updating testing so it passes?  Are we best served by adding the representation of the documentNode into all the query plan results in build-query-plan.feature?

https://app.circleci.com/pipelines/github/apollographql/federation/8610/workflows/bf18dcef-c4cf-45e4-8ec0-b592a80f8e24/jobs/52912?invite=true#step-107-8795

pcmanus:
@theJC As the patch stand,  I don't know of other option than adding the representations all over in `build-query-plan.feature` (but to be honest, my knowledge of cucumber, used by those test, is weak). I do get it's annoying, both now and frankly moving forward since any future change will now have to be done in 2 places; and the value is unclear.

That said, following on my previous suggestion, one option might be to do "add an option to disable this" now. This would mean adding some new `options` argument to the `QueryPlanner` ctor, which currently has no options (but we'll likely have need for other options soon enough, so that's fine imo). If populating the new field can be disabled, then it could be used that to disable it for all the cucumber tests. It'd make sense in that case to have a dedicated non-cucumber unit-test that target this new addition, but at least the existing cucumber tests won't have to be updated.

Personally, I kind of prefer that later option because as said before, I suspect we'll have some use case at some point where generated this somewhat redundant info is not desirable. But I won't object to just updating all the cucumber tests either.

theJC:
@pcmanus I believe I've addressed your concerns, let me know if there is anything else needed to make this PR mergeable.

Thanks again for your input,
Jon

pcmanus (approved):
Sorry the delay. Patch looks great, thanks, though I'm afraid it needs rebasing (one of the reason is that the `buildPlan.ts` file has been reorganised a bit, which confuses the diff algorithm). It'd be awesome if you could:
1. rebase it.
2. add a quick line in `query-planner/CHANGELOG.md`, and copy-paste it also in `gateway-js/CHANGELOG.md`.

and then we can ship it.

theJC:
Should be all set @pcmanus

pcmanus:
Perfect, thanks a lot of the contribution!

pcmanus:
Hi @theJC. I wanted to let you know that we're planning to switch this option to `false` by default in the upcoming 2.2 release (see #2227/#2230), so you would have to manually opt-in as you upgrade if you still rely on this. Sorry for the inconvenience.

PR #1882, opened 2022-05-23
Bug when all root operation are in schema extension

### LINKED ISSUE #1875: `extend schema @link(...) { query: Query }` causes error in buildSubgraph
Failing test: https://github.com/apollographql/federation/compare/main...lennyburdette:extend-schema-parse-error?expand=1

A customer encountered this when their sdl merging script (based on `@graphql-tools/merge`) resulted in:
```graphql
extend schema
  @contact(...)
  @link(url: "https://specs.apollo.dev/federation/v2.0" import: ["@key", "@shareable", "@tag"]) {
  query: Query
  mutation: Mutation
}
```

The exception thrown is 
```
GraphQLError: Syntax Error: Expected "{", found Name "extend".
```

I printed out the intermediary SDL in `Schema.toAST` and got:

```graphql
schema

extend schema
  @link(url: "https://specs.apollo.dev/federation/v2.0", import: ["@tag"])
{
  query: Query
}

directive @federation__key(fields: federation__FieldSet!, resolvable: Boolean = true) repeatable on OBJECT | INTERFACE

directive @federation__requires(fields: federation__FieldSet!) on FIELD_DEFINITION

directive @federation__provides(fields: federation__FieldSet!) on FIELD_DEFINITION

directive @federation__external(reason: String) on OBJECT | FIELD_DEFINITION

directive @tag(name: String!) repeatable on FIELD_DEFINITION | OBJECT | INTERFACE | UNION | ARGUMENT_DEFINITION | SCALAR | ENUM | ENUM_VALUE | INPUT_OBJECT | INPUT_FIELD_DEFINITION

directive @federation__extends on OBJECT | INTERFACE

directive @federation__shareable on OBJECT | FIELD_DEFINITION

directive @federation__inaccessible on FIELD_DEFINITION | OBJECT | INTERFACE | UNION | ARGUMENT_DEFINITION | SCALAR | ENUM | ENUM_VALUE | INPUT_OBJECT | INPUT_FIELD_DEFINITION

directive @federation__override(from: String!) on FIELD_DEFINITION

type Query {
  hello: String
  _service: _Service!
}

enum link__Purpose {
  """
  `SECURITY` features provide metadata necessary to securely resolve fields.
  """
  SECURITY

  """
  `EXECUTION` features provide metadata necessary for operation execution.
  """
  EXECUTION
}

scalar link__Import

scalar federation__FieldSet

scalar _Any

type _Service {
  sdl: String
}
```

pcmanus:
The code that was "expanding" subgraphs was always adding the
`@link` to link-itself to the schema _definition_. However,
if all the root operation of the schema were defined in a schema
_extension_ (which is technically illegal according to the graphQL
spec, but de-facto accepted because graphQL-js has never validated
that schema extension were illegal unless a corresponding definition
was present), then there was no way to print that schema properly
and indeed the code was generating invalid syntax.

This commit makes the code a bit more lenient to that case (where
there is a schema extension defining the query root operation but no
corresponding definition) by putting the `@link`-to-link on the
schema extension.

Fixes #1875

### DIFF
--- a/internals-js/src/coreSpec.ts
+++ b/internals-js/src/coreSpec.ts
@@ -3,7 +3,7 @@ import { URL } from "url";
 import { CoreFeature, Directive, DirectiveDefinition, EnumType, ErrGraphQLAPISchemaValidationFailed, ErrGraphQLValidationFailed, InputType, ListType, NamedType, NonNullType, ScalarType, Schema, SchemaDefinition, SchemaElement } from "./definitions";
 import { sameType } from "./types";
 import { err } from '@apollo/core-schema';
-import { assert } from './utils';
+import { assert, firstOf } from './utils';
 import { ERRORS } from "./error";
 import { valueToString } from "./values";
 import { coreFeatureDefinitionIfKnown, registerKnownFeature } from "./knownCoreFeatures";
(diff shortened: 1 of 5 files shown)


internals-js/src/coreSpec.ts
```diff
@@ -373,7 +373,42 @@ export class CoreSpecDefinition extends FeatureDefinition {
...
+    // 2. earlier versions (pre-#1875) were always putting that directive on the definition, and we wanted
+    //   to avoid suprising users by changing that for not reason.
+    //
+    // So instead, we put the directive on the schema definition unless some extensions exists but no
+    // definition does (that is, no non-extension elements are populated).
+    const schemaDef =  schema.schemaDefinition;
+    // Side-note: this test must be done _before_ we call `applyDirective`, otherwise it would take it into
+    // account.
+    const hasDefinition = schemaDef.hasNonExtensionElements();
+    const directive = schemaDef.applyDirective(as ?? this.url.name, args, true);
```
clenfest, line 406: I found this code a bit confusing for a second because of the fact that I was reading `as` as the typescript casting keyword. I'd probably change the variable name for clarity.

PR #1884, opened 2022-05-23
Expand support for node@18

### LINKED ISSUE #1752: Add node18 support
- node18 has officially been released! We should officially support that.
- node12 is out of LTS at the end of April ~, so official support could be deprecated.~. It would be a major version bump to deprecate node12, so we'll continue supporting that for now.

trevor-scheer:
Fixes #1752

We can't drop support for Node 12 without a major version bump.
Aside from it being EOL, I don't see much incentive to drop it in any hurry, either. i.e. we're no longer polyfilling anything (which was an incentive to drop v10).

@benweatherman if you'd like me to leave your issue open anyway, I'll remove the `Fixes #1752` and just reference the issue instead.

According to the comment in the Orb we use, the "install specific npm version" should no longer be necessary (as of a very long time ago). I removed it since it was causing issues with the node 18 tests and everything seems to be working as expected.
https://github.com/apollographql/CircleCI-Orbs/blob/main/src/oss-ci-cd-tooling/orb.yml#L41-L46

### TODO
- [x] Update quick start guide to also include node18 (it currently says node 14 or 16 is needed https://github.com/apollographql/federation/pull/1916)
  - This will be handled as part of https://github.com/apollographql/federation/issues/1925, since we should do that closer to release


benweatherman:
💝  Thanks @trevor-scheer! I updated the issue to only include the node18 update, totally agree with waiting on node12 deprecation. Changes look great, once we figure out which version to land this in.

benweatherman (approved):
🐐

PR #1891, opened 2022-05-25
Monkeying around with memory and performance

benweatherman:
Monkeying around with memory and performance. Adding this draft PR so I can NPM install from code sandbox (I always seem to have fits npm installing from the local build).


benweatherman:
Closing this PR, since it was only exploratory. The one good idea we should implement is a pluggable cache.

PR #1897, opened 2022-06-01
Compose user directives

clenfest:
Adds ability to compose custom directives by passing in an optional `exposeDirectives` parameter to the `composeServices` function. This PR intentionally muddies the waters somewhat between executable and type system directives. The type system portion of the directive (e.g. type system locations) are composed by union, whereas the executable portion will continue to compose as it has done.

Add exposeDirectives option to IntrospectAndCompose

### DIFF
--- a/composition-js/src/__tests__/compose.custom.test.ts
+++ b/composition-js/src/__tests__/compose.custom.test.ts
@@ -1,0 +1,431 @@
+import { composeServices } from '../compose';
+import {
+  Schema,
+} from '@apollo/federation-internals';
+import gql from 'graphql-tag';
+import './matchers';
+
+const expectDirectiveOnElement = (schema: Schema, location: string, directiveName: string, props?: { [key: string]: any }) => {
+  const elem = schema.elementByCoordinate(location);
+  expect(elem).toBeDefined();
+  const directive = elem?.appliedDirectives.find(d => {
+    if (d.name !== directiveName) {
+      return false;
+    }
+    if (props) {
+      for (const prop in props) {
+        if (d.arguments()[prop] !== props[prop]) {
+          return false;
+        }
+      }
+    }
+    return true;
+  });
+  expect(directive).toBeDefined();
+};
+
+const expectNoDirectiveOnElement = (schema: Schema, location: string, directiveName: string) => {
+  const elem = schema.elementByCoordinate(location);
+  expect(elem).toBeDefined();
+  const directive = elem?.appliedDirectives.find(d => d.name === directiveName);
+  expect(directive).toBeUndefined();
+};
+
+describe('composing custom directives', () => {
+  it('executable directive successful merge, not explicitly exposed', () => {
+    const subgraphA = {
+      name: 'subgraphA',
+      typeDefs: gql`
+        directive @foo(name: String!) on QUERY | FIELD_DEFINITION
+        type Query {
+          a: User
+        }
+
+        type User @key(fields: "id") {
+          id: Int
+          name: String @foo(name: "graphA")
+        }
+      `,
+    };
+
+    const subgraphB = {
+      name: 'subgraphB',
+      typeDefs: gql`
+        directive @foo(name: String!) on QUERY | FIELD_DEFINITION
+        type User @key(fields: "id") {
+          id: Int
+          description: String @foo(name: "graphB")
+        }
+      `,
+    };
+    const result = composeServices([subgraphA, subgraphB]);
+    expect(result.errors).toBeUndefined();
+    expect(result.hints?.length).toBe(0);
+    const { schema } = result;
+    expect(schema).toBeDefined();
+    if (schema) {
+      expectNoDirectiveOnElement(schema, 'User.name', 'foo');
+      expect(schema.directive('foo')?.name).toBe('foo');
+      expect(schema.directive('foo')?.locations).toEqual(['QUERY']);
+    }
+  });
+
+  it('executable directive successful merge, explicitly exposed', () => {
+    const subgraphA = {
+      name: 'subgraphA',
+      typeDefs: gql`
+        directive @foo(name: String!) on QUERY | FIELD_DEFINITION
+        type Query {
+          a: User
+        }
+
+        type User @key(fields: "id") {
+          id: Int
+          name: String @foo(name: "graphA")
+        }
+      `,
+    };
+
+    const subgraphB = {
+      name: 'subgraphB',
+      typeDefs: gql`
+        directive @foo(name: String!) on QUERY | FIELD_DEFINITION
+        type User @key(fields: "id") {
+          id: Int
+          description: String @foo(name: "graphB")
+        }
+      `,
+    };
+    const result = composeServices([subgraphA, subgraphB], { exposeDirectives: ['@foo']});
+    expect(result.errors).toBeUndefined();
+    expect(result.hints?.length).toBe(0);
+    const { schema } = result;
+    expect(schema).toBeDefined();
+    if (schema) {
+      expectDirectiveOnElement(schema, 'User.name', 'foo', { name: 'graphA'});
+      expect(schema.directive('foo')?.name).toBe('foo');
+      expect(schema.directive('foo')?.locations).toEqual(['QUERY', 'FIELD_DEFINITION']);
+    }
+  });
+
+  it('executable directive, conflicting definitions non-explicit', () => {
+    const subgraphA = {
+      name: 'subgraphA',
+      typeDefs: gql`
+        directive @foo(name: String!) on QUERY
+        type Query {
+          a: User
+        }
+
+        type User @key(fields: "id") {
+          id: Int
+          name: String
+        }
+      `,
+    };
+
+    const subgraphB = {
+      name: 'subgraphB',
+      typeDefs: gql`
+        directive @foo(name: String!) on QUERY | FIELD_DEFINITION | OBJECT
+        type User @key(fields: "id") @foo(name: "objectB") {
+          id: Int
+          description: String @foo(name: "graphB")
+        }
+      `,
+    };
+    const result = composeServices([subgraphA, subgraphB]);
+    expect(result.errors).toBeUndefined();
+    expect(result.hints?.length).toBe(0);
+    const { schema } = result;
+    expect(schema).toBeDefined();
+    if (schema) {
+      expectNoDirectiveOnElement(schema, 'User.name', 'foo');
+      expect(schema.directive('foo')?.name).toBe('foo');
+      expect(schema.directive('foo')?.locations).toEqual(['QUERY']);
+    }
+  });
+
+  it('executable directive, conflicting definitions explicit', () => {
+    const subgraphA = {
+      name: 'subgraphA',
+      typeDefs: gql`
+        directive @foo(name: String!) on QUERY
+        type Query {
+          a: User
+        }
+
+        type User @key(fields: "id") {
+          id: Int
+          name: String
+        }
+      `,
+    };
+
+    const subgraphB = {
+      name: 'subgraphB',
+      typeDefs: gql`
+        directive @foo(name: String!) on QUERY | FIELD_DEFINITION | OBJECT
+        type User @key(fields: "id") @foo(name: "objectB") {
+          id: Int
+          description: String @foo(name: "graphB")
+        }
+      `,
+    };
+    const result = composeServices([subgraphA, subgraphB], { exposeDirectives: ['@foo']});
+    expect(result.errors).toBeUndefined();
+    expect(result.hints?.length).toBe(0);
+    const { schema } = result;
+    expect(schema).toBeDefined();
+    if (schema) {
+      expectDirectiveOnElement(schema, 'User.description', 'foo', { name: 'graphB'});
+      expectDirectiveOnElement(schema, 'User', 'foo', { name: 'objectB'});
+      expect(schema.directive('foo')?.name).toBe('foo');
+      expect(schema.directive('foo')?.locations).toEqual(['QUERY', 'FIELD_DEFINITION', 'OBJECT']);
+    }
+  });
+
+  it('type-system directive, not exposed', () => {
+    const subgraphA = {
+      name: 'subgraphA',
+      typeDefs: gql`
+        directive @foo(name: String!) on FIELD_DEFINITION
+        type Query {
+          a: User
+        }
+
+        type User @key(fields: "id") {
+          id: Int
+          name: String @foo(name: "graphA")
+        }
+      `,
+    };
+
+    const subgraphB = {
+      name: 'subgraphB',
+      typeDefs: gql`
+        directive @foo(name: String!) on FIELD_DEFINITION
+        type User @key(fields: "id") {
+          id: Int
+          description: String @foo(name: "graphB")
+        }
+      `,
+    };
+    const result = composeServices([subgraphA, subgraphB]);
+    expect(result.errors).toBeUndefined();
+    expect(result.hints?.length).toBe(0);
+    const { schema } = result;
+    expect(schema).toBeDefined();
+    if (schema) {
+      expectNoDirectiveOnElement(schema, 'User.name', 'foo');
+    }
+  });
+
+  it('type-system directive, explicitly exposed', () => {
+    const subgraphA = {
+      name: 'subgraphA',
+      typeDefs: gql`
+        directive @foo(name: String!) on FIELD_DEFINITION
+        type Query {
+          a: User
+        }
+
+        type User @key(fields: "id") {
+          id: Int
+          name: String @foo(name: "graphA")
+        }
+      `,
+    };
+
+    const subgraphB = {
+      name: 'subgraphB',
+      typeDefs: gql`
+        directive @foo(name: String!) on FIELD_DEFINITION
+        type User @key(fields: "id") {
+          id: Int
+          description: String @foo(name: "graphB")
+        }
+      `,
+    };
+    const result = composeServices([subgraphA, subgraphB], { exposeDirectives: ['@foo']});
+    expect(result.errors).toBeUndefined();
+    expect(result.hints?.length).toBe(0);
+    const { schema } = result;
+    expect(schema).toBeDefined();
+    if (schema) {
+      expectDirectiveOnElement(schema, 'User.name', 'foo', { name: 'graphA'});
+    }
+  });
+
+  it('type-system directive, repeatable sometimes', () => {
+    const subgraphA = {
+      name: 'subgraphA',
+      typeDefs: gql`
+        directive @foo(name: String!) repeatable on FIELD_DEFINITION
+        type Query {
+          a: User
+        }
+
+        type User @key(fields: "id") {
+          id: Int
+          name: String @foo(name: "graphA")
+        }
+      `,
+    };
+
+    const subgraphB = {
+      name: 'subgraphB',
+      typeDefs: gql`
+        directive @foo(name: String!) on FIELD_DEFINITION
+        type User @key(fields: "id") {
+          id: Int
+          description: String @foo(name: "graphB")
+        }
+      `,
+    };
+    const result = composeServices([subgraphA, subgraphB], { exposeDirectives: ['@foo']});
+    expect(result.errors).toBeUndefined();
+    expect(result.hints?.length).toBe(1);
+    expect(result.hints?.[0].definition.code).toBe('INCONSISTENT_TYPE_SYSTEM_DIRECTIVE_REPEATABLE');
+    const { schema } = result;
+    expect(schema).toBeDefined();
+    if (schema) {
+      expectDirectiveOnElement(schema, 'User.name', 'foo', { name: 'graphA'});
+    }
+  });
+
+  it('type-system directive, different locations', () => {
+    const subgraphA = {
+      name: 'subgraphA',
+      typeDefs: gql`
+        directive @foo(name: String!) on FIELD_DEFINITION | OBJECT
+        type Query {
+          a: User
+        }
+
+        type User @key(fields: "id") @foo(name: "objectA") {
+          id: Int
+          name: String @foo(name: "graphA")
+        }
+      `,
+    };
+
+    const subgraphB = {
+      name: 'subgraphB',
+      typeDefs: gql`
+        directive @foo(name: String!) on FIELD_DEFINITION
+        type User @key(fields: "id") {
+          id: Int
+          description: String @foo(name: "graphB")
+        }
+      `,
+    };
+    const result = composeServices([subgraphA, subgraphB], { exposeDirectives: ['@foo']});
+    expect(result.errors).toBeUndefined();
+    expect(result.hints?.length).toBe(0);
+    const { schema } = result;
+    expect(schema).toBeDefined();
+    if (schema) {
+      expectDirectiveOnElement(schema, 'User.name', 'foo', { name: 'graphA'});
+      expectDirectiveOnElement(schema, 'User', 'foo', { name: 'objectA'});
+    }
+  });
+
+  it('type-system directive, core feature, not exposed', () => {
+    const subgraphA = {
+      name: 'subgraphA',
+      typeDefs: gql`
+      extend schema
+        @link(url: "https://specs.apollo.dev/link/v1.0")
+        @link(url: "https://specs.apollo.dev/join/v0.2", for: EXECUTION)
+        @link(url: "https://specs.apollo.dev/foo/v1.0", import: ["@foo"])
+
+        directive @foo(name: String!) on FIELD_DEFINITION
+
+        type Query {
+          a: User
+        }
+
+        type User @key(fields: "id") {
+          id: Int
+          name: String @foo(name: "graphA")
+        }
+      `,
+    };
+
+    const subgraphB = {
+      name: 'subgraphB',
+      typeDefs: gql`
+      extend schema
+        @link(url: "https://specs.apollo.dev/link/v1.0")
+        @link(url: "https://specs.apollo.dev/join/v0.2", for: EXECUTION)
+        @link(url: "https://specs.apollo.dev/foo/v1.0", import: ["@foo"])
+
+        directive @foo(name: String!) on FIELD_DEFINITION
+
+        type User @key(fields: "id") {
+          id: Int
+          description: String @foo(name: "graphB")
+        }
+      `,
+    };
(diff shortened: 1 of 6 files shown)


composition-js/src/__tests__/compose.custom.test.ts
```diff
@@ -0,0 +1,496 @@
...
+    expect(result.errors).toBeUndefined();
+    expect(result.hints?.length).toBe(0);
+    const { schema } = result;
+    expect(schema).toBeDefined();
+    if (schema) {
+      expectDirectiveOnElement(schema, 'User.name', 'foo', { name: 'graphA'});
+    }
+  });
+
+  it('type-system directive, repeatable sometimes', () => {
```
pcmanus, line 325: That test might be a tad more interesting if the subgraph in which the directive is repeatable was repeating it too, and we were asserting the result of that.

composition-js/src/__tests__/compose.custom.test.ts
```diff
@@ -0,0 +1,496 @@
...
+        directive @foo(name: String!) on FIELD_DEFINITION
+        type User @key(fields: "id") {
+          id: Int
+          description: String @foo(name: "graphB")
+        }
+      `,
+    };
+    const result = composeServices([subgraphA, subgraphB], { exposeDirectives: ['@foo']});
+    expect(result.errors).toBeUndefined();
+    expect(result.hints?.length).toBe(0);
```
pcmanus, line 390: Surprised we have an hint for "inconsistent on repeatable" but not "inconsistent on locations". We should probably fix that (?).

composition-js/src/__tests__/compose.custom.test.ts
```diff
@@ -0,0 +1,496 @@
...
+    const { schema } = result;
+    expect(schema).toBeDefined();
+    if (schema) {
+      expectDirectiveOnElement(schema, 'User.name', 'foo', { name: 'graphA'});
+    }
+  });
+
+  it.todo('expose builtin directive');
```
pcmanus, line 495: Worst adding that one :). Especially to be clear on what behaviour we want: should it be essentially ignored (since built-ins are merged by default) or should it error out (to make it clear the user is doing something not useful).

As I've hinted a bunch of time, I tend to like defaulting on erroring when something appears to be a user "mistake". My reasoning being that:
1. as a user, if I do something  not useful, I'm usually more happy than mad to learn about it, even if it's not harmful per-se.
2. I usually feel that if it's a choice between maybe preventing a mistake and maybe getting in the way, I'd rather risk getting in the way but prevent a mistake, because the cost of mistake tends to be much higher than small get-in-the-ways. It's a balance of course but ...
3. it's easier to remove a restriction than add one later.

composition-js/src/merging/merge.ts
```diff
@@ -2003,7 +2034,7 @@ class Merger {
     for (const source of sources) {
       if (source) {
         for (const directive of source.appliedDirectives) {
-          if (this.isMergedDirective(directive)) {
+          if (this.isMergedDirective(directive) || this.exposedDirectives().includes(directive.name)) {
```
pcmanus, line 2037: It would have made more sense to me to move the "is exposed" check inside `isMergedirective` (at least that was the original intend of that `isMergedDirective` method: to capture the full extend of when a directive was merged or not). Or was there a specific reason not too?

composition-js/src/merging/merge.ts
```diff
@@ -2286,4 +2317,11 @@ class Merger {
         : err;
     });
   }
+
+  /**
+   * Get rid of leading '@' if present and return the list of directives passed to the Merger object
```
pcmanus, line 2322: In the same spirit of another comment, I'd rather we validate the user option and completely reject the case where a name doesn't start with a `@` rather than be flexible about it. I'd really worry we may regret that flexibility later.

Similarly, I'd have personally opted for, in that validation, maybe erroring out when one of the provided argument is simply not present in any subgraph.

composition-js/src/types.ts
```diff
@@ -0,0 +1,8 @@
+import {
```
pcmanus, line 1: Nite: I'm note sure why this file is called `types`. What's the reasoning?

composition-js/src/merging/merge.ts
```diff
@@ -353,6 +369,7 @@ class Merger {
     // supergraph. This allow to be able to reference those from that point on.
     this.addTypesShallow();
     this.addDirectivesShallow();
+    this.addCustomTypeSystemDirectives();
```
pcmanus, line 372: I'm running a bit out of time to dig into this method more deeply today so this may turn out to be a stupid remark, but it's not obvious to me a priori why we need that method, and why `addDirectivesShallow` doesn't handle this directly. It certainly was the initial intent that it would.

I'm also not sure why we need to do something specific about the locations at that stages. Isn't that handled by ` mergeTypeSystemDirectiveDefinition`?

pcmanus (commented):
Left some remarks/questions for a first pass.

Aside from those, I do think we should have that discussion about core directives. Afaict, the code may currently merge 2 different major versions of a core directive, and I'm of the opinion that it's not something we should do, so if we allow it now, changing it later would be a breaking change. More generally, merging core directives without their corresponding links (so making them non-code in the supergraph) is imo of limited interest, in the sense that if you bother making a directive a core one, it's probably that you care about what core features provide, and that whichever code process that directive expects it to be a core directive, and merging it as a non-core one defeats that.

That is, I don't strongly object to punting on core directives initially if we want to do so, but if we do, I think we should hard error out on exposing a core directive. Anything else runs the risk of blocking us later once we decide what we do want to do if it turns out to not be compatible. Of course, the alternative is to not punt on core directives and I'm ok with that too, but we should then discuss what we want :).

composition-js/src/__tests__/compose.custom.test.ts
clenfest: Agreed. Trying to promote builtin directives will error. That was the outstanding work I had.

composition-js/src/__tests__/compose.custom.test.ts
```diff
@@ -0,0 +1,496 @@
...
+        }
+      `,
+    };
+    const result = composeServices([subgraphA, subgraphB]);
+    expect(result.errors).toBeUndefined();
+    expect(result.hints?.length).toBe(0);
+    const { schema } = result;
+    expect(schema).toBeDefined();
+    if (schema) {
+      expectNoDirectiveOnElement(schema, 'User.name', 'foo');
```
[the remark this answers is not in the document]
clenfest: This one may have been a copy/paste issue. Nice catch.

composition-js/src/__tests__/compose.custom.test.ts
```diff
@@ -0,0 +1,496 @@
...
+    const { schema } = result;
+    expect(schema).toBeDefined();
+    if (schema) {
+      expectDirectiveOnElement(schema, 'User.name', 'foo', { name: 'graphA'});
+      expect(schema.directive('foo')?.name).toBe('foo');
+      expect(schema.directive('foo')?.locations).toEqual(['QUERY', 'FIELD_DEFINITION']);
+    }
+  });
+
+  it('executable directive, conflicting definitions non-explicit', () => {
```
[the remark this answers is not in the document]
clenfest: I think originally I was intending it to mean explicitly exposed, but there is no implicit version, so I agree that it's confusing.

composition-js/src/merging/merge.ts
clenfest: The reason that it's outside of it is because `isMergedDirective` returns true if the directive is executable, regardless of whether or not we want to expose it. I'll see what I can do to refactor.

pcmanus: > The reason that it's outside of it is because `isMergedDirective` returns true if the directive is executable, regardless of whether or not we want to expose it

But juste to make sure we're on the same page, I don't think we want to change that anyway. That is, `isMergedDirective` returns true for directive _definition_ having executive locations, but that should still be true regardless of what is exposed.

Looking at it, I believe just adding:
```typescript
if (this.exposedDirectives().includes(definition.name)) {
  return true;
}
```
at the very beginning of `isMergedDirective` should be the correct thing to do. But if there is a reason this doesn't work I'm not seeing, it might be worth some commenting.

composition-js/src/types.ts
clenfest: I guess I was thinking that it could be a general place for putting types that might otherwise resolve in a circular dependency, but I'll name it to be more specfic.

composition-js/src/merging/merge.ts
```diff
@@ -471,16 +490,35 @@ class Merger {
...
     }
+
+    // now for all custom directives, make sure locations are properly added
+    customDirectiveMap.forEach((def, directiveName) => {
+      const definitions: DirectiveDefinition[] = this.subgraphsSchema
+        .map(schema => schema.directive(directiveName))
+        .filter((def): def is DirectiveDefinition => def !== undefined);
+      // if any directive definition has a executable location, we'll just add all locations
+      // but if it's a type-system only directive, we can attempt to merge the directives together
+      if (definitions.some(d => d.locations.some(loc => executableDirectiveLocations.includes(loc)))) {
```
pcmanus, line 515: I'm not sure the new logic regarding executable-only  and "mixed" (executable and type system locations) directives.

In particular, it's confusing that we assign locations to the merged definition here, but do it again in `mergeExecutableDirectiveDefinition`. And here, we seems to assign any locations found in any subgraph, which is wrong for executable locations (we should only merge a given executable location if it is in _all_ subgraph; because we cannot forbid a user to include the directive in a query anywhere the merged locations permit, but if one of those application reach a subgraph that does not have the location, that subgraph query will break). So as far as I can tell, this break the proper merging of executable directives, and the fact that no unit test broke means we need to fix that coverage. 

More generally, the merging logic is that executable-only directives are merged by intersection, while type-system-only ones are merged by union. But I think that means mixed ones needs a middle ground strategy. More precisely, I think that for mixed directives:
* we should intersect executable locations but union type-system locations. As said above, not intersecting executable locations would break subgraph queries down the line, and not unioning type-system locations would result in invalid supergraphs.
* we should hard error out if `repeatable` is not consistent. If it's inconsistent, all choices in the subgraph will create issues: if we mark it repeatable, then subgraph query will break a query with the directive repeatable reach a subgraph where the definition is not repeatable, but if we mark it non-repeatable, we may have some repeated applications in some subgraph and upon merging, the supergraph will be invalid graphQL.

In other words, I don't think we can treat mixed directives the same as either executable-only or type-system only  ones.

composition-js/src/merging/merge.ts
```diff
@@ -1810,79 +1848,54 @@ class Merger {
...
       this.mergeExecutableDirectiveDefinition(sources, dest);
     }
   }
 
   // Note: as far as directive definition goes, we currently only merge directive having execution location, and only for
-  // thos locations. Any type system directive definition that propagates to the supergraph (graphQL built-ins and `@tag`)
+  // those locations. Any type system directive definition that propagates to the supergraph (graphQL built-ins and `@tag`)
   // is currently handled in an hard-coded way. This will change very soon however so keeping this code around to be
   // re-enabled by a future commit.
-  //private mergeTypeSystemDirectiveDefinition(sources: (DirectiveDefinition | undefined)[], dest: DirectiveDefinition) {
```
pcmanus, line 1822: As we re-enable this method, it would be nice to re-enable the currently skipped related tests in `hints.test.ts`.

composition-js/src/merging/merge.ts
```diff
@@ -1810,79 +1848,54 @@ class Merger {
...
-  //  }
-  //}
+  private mergeTypeSystemDirectiveDefinition(sources: DirectiveDefinition[], dest: DirectiveDefinition) {
+    this.addArgumentsShallow(sources, dest);
+    for (const destArg of dest.arguments()) {
+      const subgraphArgs = sources.map(f => f?.argument(destArg.name));
+      this.mergeArgument(subgraphArgs, destArg);
+    }
+
+    const repeatable = sources[0]?.repeatable ?? false;
```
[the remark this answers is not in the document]
clenfest: I think I must have made a change here. The type of `sources` is `DirectiveDefinition[]`, so `sources[0].repeatable` should just be the first value in a subgraph that has that directive definition (the undefined ones will be skipped). So I think this is ok, yes?

composition-js/src/merging/merge.ts
clenfest: I've checked in the code for the first half of this. Will come back to the second part tomorrow.

composition-js/src/merging/merge.ts
clenfest: I don't think that the tests as written are applicable since we only do this detection when the directive is explicitly exposed.

composition-js/src/__tests__/compose.custom.test.ts
```diff
@@ -0,0 +1,572 @@
...
+    expect(result.errors?.[0].message).toBe(`Directive "${directiveName}" cannot be specified in \"mergeDirectives\" argument because it is a built in directive`);
+  });
+
+  it.each(['@key', '@link', '@customTag'])('fed 2 directives are rejected. Even if they are aliased', (directive) => {
+    const result = composeServices([subgraphA, subgraphB], { mergeDirectives: [directive]});
+    expect(result.errors?.length).toBe(1);
+    expect(result.errors?.[0].message).toBe(`Directive "${directive}" cannot be specified in "mergeDirectives" argument because it is linked via a core feature in at least one subgraph`);
+  });
+
+  it('custom @tag still composes as long as federation version is renamed', () => {
```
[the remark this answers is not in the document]
clenfest: I'm not quite sure I understand this. Would my subgraph look like this?

```
    const subgraphC = {
      name: 'subgraphC',
      typeDefs: gql`
        extend schema
          @link(url: "https://specs.apollo.dev/federation/v2.0")

        directive @tag(name: String!) on QUERY | FIELD_DEFINITION
        type User @federation__key(fields: "id") {
          id: Int
          description: String @tag(name: "graphB")
        }
      `,
    };
```

If so I'm getting an unexpected hint when composing telling me that `@key` is not defined in all subgraphs. Am I doing something wrong?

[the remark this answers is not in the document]
clenfest: Maybe a dumb question, but when I use the fully qualified `@key`, I get a hint indicating that the type no longer seems to be an entity.

```
Field "User.name" of non-entity object type "User" is defined in some but not all subgraphs that define "User": "User.name" is defined in subgraph "subgraphC" but not in subgraph "subgraphD".```

```
  it('custom @tag still composes with no import', () => {
    const subgraphC = {
      name: 'subgraphC',
      typeDefs: gql`
        extend schema
          @link(url: "https://specs.apollo.dev/federation/v2.0")

        directive @tag(name: String!) on QUERY | FIELD_DEFINITION
        type Query {
          a: User
        }

        type User @federation__key(fields: "id") {
          id: Int
          name: String @tag(name: "graphA")
        }
      `,
    };

    const subgraphD = {
      name: 'subgraphD',
      typeDefs: gql`
        extend schema
          @link(url: "https://specs.apollo.dev/federation/v2.0")

        directive @tag(name: String!) on QUERY | FIELD_DEFINITION
        type User @federation__key(fields: "id") {
          id: Int
          description: String @tag(name: "graphB")
        }
      `,
    };

    const result = composeServices([subgraphC, subgraphD], { mergeDirectives: ['@tag']});
    const schema = expectNoErrorsOrHints(result);
    expectDirectiveOnElement(schema, 'User.name', 'tag', { name: 'graphA'});
    expectDirectiveOnElement(schema, 'User.description', 'tag', { name: 'graphB'});
    expect(schema.directive('tag')?.name).toBe('tag');
    expect(schema.directive('tag')?.locations).toEqual(['QUERY', 'FIELD_DEFINITION']);
  });```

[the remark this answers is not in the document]
clenfest: Seems like this is a bug in the existing code. I'm going to fix it here.

composition-js/src/merging/merge.ts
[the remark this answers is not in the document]
pcmanus: > So I think this is ok, yes?

Hum. I agree the code itself is consistent with itself now. But we're still not on the page on the implemented logic. To quote back my previous comment, my opinion is that "the logic we want is that the merged directive is repeatable as soon as any subgraph has it repeatable". By which I implied, but that was maybe not clear, that we would still compose for inconsistencies in repeatable (as was done in previously commented code) and not hard error (but hint, of course).

What I initially didn't saw is that the patch now hard error on inconsistent repeatable, and that's the part I think we should change. Hard erroring, as usual, will in particular make it really hard for users to migrate a non-repeatable directive to be repeatable in an incremental way, so I think we should be lenient here.

composition-js/src/merging/merge.ts
pcmanus: While it's fair that we cannot just remove the `skip` on those test and them work, the spirit of my comment that:
1. there is no point in keeping those skipped test around as skipped after this patch.
2. we really should have test for the hints enabled by this patch.

I'd personally just reuse those tests (I'll note that you've removed one of them, but see my other comment for why I think it should be re-added) and modify them ever so slightly to use a manually exposed random directive, but any alternatives that achieve the 2 points above wfm too.

composition-js/src/merging/merge.ts
```diff
@@ -826,20 +859,34 @@ class Merger {
...
       }
-      if (source.hasAppliedDirective('key')) {
+      const keyDirectiveName = federationMetadata(source.schema())?.keyDirective().name ?? 'key';
+      if (source.hasAppliedDirective(keyDirectiveName)) {
         sourceAsEntity.push(source);
       } else {
         sourceAsNonEntity.push(source);
       }
     }
+    sources.forEach((source, idx) => {
```
pcmanus, line 869: I believe this block of code does the same than the one before that. I assume you decided to rewrote it and forgot to remove the old version? I do like the new version better.

composition-js/src/merging/merge.ts
```diff
@@ -1810,79 +1857,64 @@ class Merger {
...
+
+    const definedSources = sources.filter((src): src is DirectiveDefinition => src !== undefined);
+    if (definedSources.some(s => s.locations.some(loc => executableDirectiveLocations.includes(loc)))) {
+      if (this.mergedDirectives().includes(dest.name)) {
+        const { locations } = getLocationsFromDirectiveDefs(definedSources);
+        dest.addLocations(...locations);
+      }
       this.mergeExecutableDirectiveDefinition(sources, dest);
+    } else {
+      if (this.mergedDirectives().includes(dest.name)) {
```
pcmanus, line 1869: Earlier `isMergedDirective` calls should have ensured that this test is not necessary? I believe the same test in the previous branch is also unnecessary, though for the slightly different reason that `getLocationsFromDirectiveDefs` should be safe to call for executable-only directives.

 In general, I'd find it "reassuring" if `this.mergedDirectives()` was checked only in `isMergedDirective`.

composition-js/src/merging/merge.ts
```diff
@@ -1810,79 +1857,64 @@ class Merger {
     //   definition is the intersection of all definitions (meaning that if there divergence in
     //   locations, we only expose locations that are common everywhere).
     this.mergeDescription(sources, dest);
-    if (sources.some((s) => s && this.isMergedDirective(s))) {
+
+    const definedSources = sources.filter((src): src is DirectiveDefinition => src !== undefined);
+    if (definedSources.some(s => s.locations.some(loc => executableDirectiveLocations.includes(loc)))) {
+      if (this.mergedDirectives().includes(dest.name)) {
+        const { locations } = getLocationsFromDirectiveDefs(definedSources);
+        dest.addLocations(...locations);
```
pcmanus, line 1865: A little strange to set `dest`'s location here and to have it set _again_ in `mergeExecutableDirectiveDefinition`. I would simply move `getLocationsFromDirectiveDefs` within `mergeExecutableDirectiveDefinition`, same as `mergeTypeSystemDirectiveDefinition` does.

composition-js/src/merging/merge.ts
```diff
@@ -248,6 +243,40 @@ function locationString(locations: DirectiveLocation[]): string {
   return (locations.length === 1 ? 'location ' : 'locations ') + '"' + locations.join(', ') + '"';
 }
 
+/**
+ * Given a list of directive sources, get the set of locations, removing duplicates.
+ * Executable locations are merged by union while non-executable locations are merged by intersection
```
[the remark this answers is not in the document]
clenfest: I think I caught myself saying it backwards about 3 times in comments / docs, but I missed one.

composition-js/src/merging/merge.ts
```diff
@@ -333,19 +362,23 @@ class Merger {
     // If it's a directive application, then we skip it unless it's a graphQL built-in
     // (even if the definition itself allows executable locations, this particular
     // application is an type-system element and we don't want to merge it).
+    const isMergedDirective = this.mergedDirectives().includes(definition.name);
```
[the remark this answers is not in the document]
clenfest: I didn't extract it that way because of the `isGraphQLBuiltInDirective` case below, but we don't allow that case, so I'll change it.

composition-js/src/merging/merge.ts
clenfest: yup, exactly right.

composition-js/src/merging/merge.ts
clenfest: This check does have to be here because the federation directives already have their arguments added, so removing it completely would blow up. I did change it to call `isMergeDirective` rather than repeating the same logic, which I think is a good change.

composition-js/src/merging/merge.ts
```diff
@@ -1810,85 +1863,86 @@ class Merger {
     //   definition is the intersection of all definitions (meaning that if there divergence in
     //   locations, we only expose locations that are common everywhere).
     this.mergeDescription(sources, dest);
-    if (sources.some((s) => s && this.isMergedDirective(s))) {
+
+    const definedSources = sources.filter((src): src is DirectiveDefinition => src !== undefined);
+    if (definedSources.some(s => s.locations.some(loc => executableDirectiveLocations.includes(loc)))) {
       this.mergeExecutableDirectiveDefinition(sources, dest);
+    } else {
+      if (this.isMergedDirective(dest)) {
```
benweatherman, line 1871: **question** Is there a reason to keep this as an `if` nested under `else` or could this be an `else if`?

composition-js/src/merging/merge.ts
```diff
@@ -248,6 +242,56 @@ function locationString(locations: DirectiveLocation[]): string {
   return (locations.length === 1 ? 'location ' : 'locations ') + '"' + locations.join(', ') + '"';
 }
 
+/**
+ * Given a list of directive sources, get the set of locations, removing duplicates.
+ * Executable locations are merged by intersection while non-executable locations are merged by union
+ */
+const getLocationsFromDirectiveDefs = (sources: (DirectiveDefinition | undefined)[]) => {
```
benweatherman, line 249: **question** Is there a reason to use function expressions instead of function declarations? I see a few of the latter style already in this file, but not sure if it's already mixed. FWIW, I don't feel strongly.

benweatherman (commented):
Added a few non-blocking comments & questions, do with them as you wish. Mostly good job iterating through this one.

composition-js/src/__tests__/compose.custom.test.ts
```diff
@@ -0,0 +1,706 @@
...
+
+const expectNoDirectiveOnElement = (schema: Schema, location: string, directiveName: string) => {
+  const elem = schema.elementByCoordinate(location);
+  expect(elem).toBeDefined();
+  expect(elem?.hasAppliedDirective(directiveName)).toBe(false);
+};
+
+const expectNoErrorsOrHints = (result: CompositionResult): Schema => {
+  expect(result.errors).toBeUndefined();
+  expect(result.hints?.length).toBe(0);
```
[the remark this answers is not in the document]
clenfest: Nice. If I knew about this I forgot.

composition-js/src/compositionOptions.ts
```diff
@@ -0,0 +1,60 @@
...
+    } else {
+      const directiveNameWithoutAt = directiveName.slice(1);
+
+      // for the directive specified, get the DirectiveDefinition for each subgraph it appears in
+      const subgraphDirectives = subgraphs
+        .map(sg => sg.schema.directive(directiveNameWithoutAt))
+        .filter((directive): directive is DirectiveDefinition => directive !== undefined);
+
+      if (subgraphDirectives.length === 0) {
+        // If the directive does not appear in any subgraph, throw an error. Provide a suggestion if we think it's a typo.
```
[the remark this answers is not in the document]
pcmanus: Agreed. Some `IGNORED_MERGED_DIRECTIVES_ARGUMENT` "warning" hint could make sense, both for this one, but also the next 2 cases (built-ins and only-executable). In all 3 cases, we just want to say that the provided value is essentially not gonna do anything, but a warning is arguably probably even a bit better than a hard error.

Tbc, I'd still keep a hard error for the lacking of `@` (it's a syntax error, no need to be lenient) and the `no core stuff`, because it's meant to be supported eventually and so ignoring with a warning feels super dodgy for future upgrades.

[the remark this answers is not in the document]
clenfest: Unless you strongly object, I think I'll leave the hard error for the next two since it seems best to just force them to get the arguments correct.

composition-js/src/merging/merge.ts
```diff
@@ -248,6 +242,56 @@ function locationString(locations: DirectiveLocation[]): string {
...
+    .forEach((src, idx) => {
+      const prevLength = locationSet.size;
+      const nonExecutableLocations = src.locations.filter(loc => !executableDirectiveLocations.includes(loc));
+      nonExecutableLocations.forEach(locationSet.add, locationSet);
+      if (idx > 0 && (prevLength !== locationSet.size || prevLength !== nonExecutableLocations.length)) {
+        consistentLocations = false;
+      }
+    });
+  return {
+    consistentLocations,
```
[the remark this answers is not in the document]
clenfest: I had consistentLocations first and then added the executable version, but didn't change the name.

nateql:
@jeffjakub what is going on with this ticket?  It is marked done on your team board but it looks incomplete here.

clenfest:
Closing this ticket, as this is supported via core schemas. If we decide that we need to do this later on, we'll need to substantially rewrite in any case.

PR #1900, opened 2022-06-02
Fix issue generating plan for a "diamond-shaped" dependency

pcmanus:
The core of this commit is to fix the handling of merging groups when
we handle requires, which wasn't done probably in particular due to
a sub-optimal tracking of the "path in parents" for groups (without
which we cannot do merging properly). The surrounding code was however
starting to be difficult to read/follow, so this commit is by and large
a refactor the dependency graph code to make things clearer. Amongst
other things, it changes the naming from talking about a group
dependencies/depedents to talking of childs/parents: this is the same
thing, but feels much easier to read/reason about.

A few additional optimisation are needed to ensure the generated plan
is optimal. One of them ensure identical fetches (same group/same
inputs) always get merged together, and that fix doesn't only include
@require. In fact, we had an existing test in `abstract-types.test.ts`
that was generating an inefficient with such duplicate fetch, and
the test had a comment explaining that the plan was inefficient and
needed to be improved. This commit fix that known-inefficient test
as a consequence.

### DIFF
--- a/query-planner-js/src/buildPlan.ts
+++ b/query-planner-js/src/buildPlan.ts
@@ -41,6 +41,7 @@ import {
   Directive,
   directiveApplicationsSubstraction,
   conditionalDirectivesInOperationPath,
+  MultiMap,
 } from "@apollo/federation-internals";
 import {
   advanceSimultaneousPathsWithOperation,
@@ -85,13 +86,32 @@ const debug = newDebugLogger('plan');
 // Note: exported so we can have a test that explicitly requires more than this number.
 export const MAX_COMPUTED_PLANS = 10000;
 
-function mapOptionsToSelections<RV extends Vertex>(
-  selectionSet: SelectionSet,
-  options: SimultaneousPathsWithLazyIndirectPaths<RV>[]
-): [Selection, SimultaneousPathsWithLazyIndirectPaths<RV>[]][]  {
-  // We reverse the selections because we're going to pop from `openPaths` and this ensure we end up handling things in the query order.
-  return selectionSet.selections(true).map(node => [node, options]);
-}
+type CostFunction = FetchGroupProcessor<number, number[], number>;
+
+const fetchCost = 10000;
+const pipeliningCost = 100;
+const sameLevelFetchCost = 100;
+
+const defaultCostFunction: CostFunction = {
+  onFetchGroup: (group: FetchGroup) =>  selectionCost(group.selection),
+  reduceParallel: (values: number[]) => values,
+  // That math goes the following way:
+  // - we add the costs in a sequence (the `acc + ...`)
+  // - within a stage of the sequence, the groups are done in parallel, hence the `Math.max(...)` (but still, we prefer querying less services if
+  //   we can help it, hence the `+ (valueArray.length - 1) * sameLevelFetchCost`).
+  // - but each group in a stage require a fetch, so we add a cost proportional to how many we have
+  // - each group within a stage has its own cost plus a flat cost associated to doing that fetch (`fetchCost + s`).
+  // - lastly, we also want to minimize the number of steps in the pipeline, so later stages are more costly (`idx * pipelineCost`)
+  reduceSequence: (values: (number[] | number)[]) =>
+    values.reduceRight(
+      (acc: number, value, idx) => {
+        const valueArray = Array.isArray(value) ? value : [value];
+        return acc + ((idx + 1) * pipeliningCost) * (fetchCost * valueArray.length) * (Math.max(...valueArray) + (valueArray.length - 1) * sameLevelFetchCost)
+      },
+      0
+    ),
+  finalize: (roots: number[], rootsAreParallel: boolean) => roots.length === 0 ? 0 : (rootsAreParallel ? (Math.max(...roots) + (roots.length - 1) * sameLevelFetchCost) : sum(roots))
+};
 
 class QueryPlanningTaversal<RV extends Vertex> {
   // The stack contains all states that aren't terminal.
@@ -350,11 +370,7 @@ class QueryPlanningTaversal<RV extends Vertex> {
     let totalCombinations = 1;
     for (let i = 0; i < others.length; ++i) {
       const eltSize = others[i].length;
-      assert(eltSize, "Got empty option: this shouldn't have happened");
-      if(!eltSize) {
-        totalCombinations = 0;
-        break;
-      }
+      assert(eltSize > 0, "Got empty option: this shouldn't have happened");
       eltIndexes[i] = 0;
       totalCombinations *= eltSize;
     }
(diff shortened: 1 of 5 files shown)


query-planner-js/src/buildPlan.ts
```diff
@@ -788,22 +851,129 @@ class FetchGroup {
...
+    this.mergeInInternal(child, childPathInThis);
+  }
+
+  /**
+   * Merges the provided sibling (shares a common parent) of `this` group into it.
+   *
+   * @param sibling - a sibling group of `this`. Both `this` and `sibling` must share a parent but it should also be
+   * their _only_ parent. Further `this` and `sibling` must be on the same subgraph and have the same `mergeAt`.
+   */
+  mergeSiblingIn(sibling: FetchGroup) {
```
clenfest, line 882: I notice that for merging grandchildren, you've created the helper `canMergeGrandChildIn`. What are your thoughts on creating a similar helper for siblings to protect against calling this function inadvertently?

query-planner-js/src/buildPlan.ts
```diff
@@ -969,20 +1140,46 @@ function sameMergeAt(m1: ResponsePath | undefined, m2: ResponsePath | undefined)
...
+ *   is maintained for the case where we want/need to merge groups into each other. One can roughly think of
+ *   this as similar to a `mergeAt`, but that is relative to the start of `group`. 
+ */
+type ParentRelation = {
+  group: FetchGroup,
+  path?: OperationPath,
+}
+
+function concatPathsInParents(first: OperationPath | undefined, second: OperationPath | undefined): OperationPath | undefined  {
+  return first && second ? concatOperationPaths(first, second) : undefined;
```
clenfest, line 1157: If `first` is defined, but `second` is undefined, you want to return undefined and not `first`?

query-planner-js/src/buildPlan.ts
```diff
@@ -969,20 +1140,46 @@ function sameMergeAt(m1: ResponsePath | undefined, m2: ResponsePath | undefined)
...
-    // For each groups, an optional path in its "unique" parent. If a group has more than one parent, then
-    // this will be undefined. Even if the group has a unique parent, it's not guaranteed to be set.
-    private readonly pathsInParents: (OperationPath | undefined)[]
-  ) {}
+    private readonly edges: {
+      outs: FetchGroup[],
+      ins: ParentRelation[],
+    }[],
+  ) {
+    assert(groups.length === edges.length, () => `Invalid arguments: ${groups.length} groups provided, but the edges array is only of size ${edges.length}`);
```
clenfest, line 1181: There is a relationship here between the groups and the edges? Are you sure you don't want the type to be:

```
items: {
  group: FetchGroup,
  edge: {
    outs: FetchGroup[],
    ins: ParentRelation[],
  }
}[],
```

query-planner-js/src/buildPlan.ts
```diff
@@ -1063,191 +1282,220 @@ class FetchDependencyGraph {
...
   }
 
-  addDependency(dependentGroup: FetchGroup, dependentOn: FetchGroup | FetchGroup[], pathInDependentOn?: OperationPath) {
-    this.onModification();
-    const groups = Array.isArray(dependentOn) ? dependentOn : [ dependentOn ];
-    for (const group of groups) {
-      this.addEdge(group.index, dependentGroup.index, pathInDependentOn);
-    }
+  childs(group: FetchGroup): readonly FetchGroup[] {
+    return this.edges[group.index].outs;
```
clenfest, line 1376: Is it really necessary that the FetchGroup has to know it's own index? This might be necessary for other reasons, but if it's just to be able to match the groups with the edges there is probably a better way.

query-planner-js/src/buildPlan.ts
```diff
@@ -763,12 +764,74 @@ class FetchGroup {
...
+   * Returns whether this group is both a child of `maybeParent` but also if we can show that the
+   * dependency between the group is "artificial" in the sense that this group inputs do not truly
+   * depend on anything `maybeParent` fetches.
+   */
+  isChildOfWithArtificialDependency(maybeParent: FetchGroup): boolean {
+    const relation =  this.parentRelation(maybeParent);
+    // To be a child with an artificial dependency, it needs to be a child first,
+    // and the "path in parent" should be know to be empty (which essentially
+    // means the groups are on the same entity at the same level).
+    if (!relation || relation.path?.length !== 0) {
```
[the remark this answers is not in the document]
pcmanus: > Is that the behavior you want?

It is. And I was hoping that the comment was clear enough that this is what we want as I'm not sure how to explain it much better. But to rephrase that comment, we want to restrict ourselves to the case where the child and parent fetches are on the same entity/same level (it could be we can relax that restriction, but I'm not sure if it's even useful and haven't though deeply about what that involves), and `path === 0` is what tells us that. `path > 0` would tell us that the child is fetching things at a deeper level than the parent, while `path === undefined` essentially means "don't make any assumption" (that last path may be where the confusion is coming from? If so, I did added precision in the comment on `ParentRelation`).

query-planner-js/src/buildPlan.ts
```diff
@@ -763,12 +764,74 @@ class FetchGroup {
...
+
+  parents(): readonly ParentRelation[] {
+    return this.dependencyGraph.parents(this);
+  }
+
+  parentGroups(): readonly FetchGroup[] {
+    return this.parents().map((p) => p.group);
+  }
+
+  childs(): readonly FetchGroup[] {
```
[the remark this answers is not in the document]
pcmanus: Right. I could have swore `childs` was a real alternative, but I must have made it up. Renamed.

query-planner-js/src/buildPlan.ts
pcmanus: Good point, that's probably cleaner, and it's more consistent with the other methods, so added.

query-planner-js/src/buildPlan.ts
pcmanus: Correct. `undefined` for "path in parents" means that it's "unknown and don't make any assumption on it", not that it's empty. I tried to specify that a bit better in the comment about `path` in `ParentRelation`.

query-planner-js/src/buildPlan.ts
pcmanus: Yeah, almost did that and worried it would make the diff for this issue even bigger for not that much added value. But agreed we can push the refactor a tad more and so went ahead. That said, instead of what's above, I moved the `outs` and `ins` directly into `FetchGroup` as I feel it's even a bit cleaner since:
1. the suggestion above would only help when iterating on all groups, which we almost never do, but when you have a group and want to know it's edges, you'd still have to index into `items`. Having the edges directly in the group avoids that.
2. almost all the methods on parents and children we were using were on `FetchGroup` anyway, so having the whole logic here save a bit of indirection of method we were having.

query-planner-js/src/buildPlan.ts
pcmanus: > there is probably a better way.

Isn't there always :).

More seriously, with the refactor of the previous comment, that index is mostly not used. Still, it's convenient to have it for a few reasons:
1. it's nice for debugging (`dumpOnConsole`) to associate a simple short number to each group, and index is perfect for that.
2. it's convenient when cloning a graph which we need to do because it gives us a trivial way to associate a group to its clone, which we kind of need.

Now, we can certainly achieve those things in a different way:  we could assign to each group a simple numerical id uncorrelated to the index for the first point for instance, and create some temporary map for the 2nd point. But using the `index` directly just feel easier/simpler to me at that point honestly, or at least, not worst (the only small "difficulty" of the index is that we need to adjust some of it on group removal, but it's literally 2 lines of code).

query-planner-js/src/buildPlan.ts
```diff
@@ -1256,65 +1153,52 @@ class FetchDependencyGraph {
...
-            // ... and then merge g2 into that.
-            if (g2.inputs) {
-              merged.addInputs(g2.inputs);
-            }
-            merged.addSelections(g2.selection);
-            this.onMergedIn(merged, g2);
+  private mergeChildFetchesForSameSubgraphAndPath(group: FetchGroup) {
+    const children = group.children();
+    if (children.length > 1) {
+      for (const g1 of children) {
```
clenfest, line 1197: This code makes me pretty nervous in that we seem to be iterating over an array that we're performing destructive operations on. Can we maybe move to a while loop with pointers?

query-planner-js/src/buildPlan.ts
```diff
@@ -1326,183 +1210,583 @@ class FetchDependencyGraph {
...
+    return [indent + group.subgraphName + ' <- ' + children.map((child) => child.subgraphName).join(', ')]
+      .concat(children
+        .flatMap(g => g.children().length == 0
+          ? []
+          : this.toStringInternal(g, indent + "  ")))
+      .join('\n');
+  }
+}
+
+interface FetchGroupProcessor<G, P, F> {
```
clenfest, line 1451: Can you name the templatized types to convey more meaning?

query-planner-js/src/buildPlan.ts
```diff
@@ -1326,183 +1210,583 @@ class FetchDependencyGraph {
...
+): [Selection, SimultaneousPathsWithLazyIndirectPaths<RV>[]][]  {
+  // We reverse the selections because we're going to pop from `openPaths` and this ensure we end up handling things in the query order.
+  return selectionSet.selections(true).map(node => [node, options]);
+}
+
+function possiblePlans(closedBranches: SimultaneousPaths<any>[][]): number {
+  let totalCombinations = 1;
+  for (let i = 0; i < closedBranches.length; ++i){
+    const eltSize = closedBranches[i].length;
+    if(!eltSize) {
```
clenfest, line 1508: nit: Since the type is a number, I'd use === 0 here for clarity.

query-planner-js/src/buildPlan.ts
```diff
@@ -1326,183 +1210,583 @@ class FetchDependencyGraph {
...
+      totalCombinations = 0;
```
clenfest, line 1509: Is this right? If any branch has a length of 0, we want the totalCombinations to be 0? If that's true, maybe just return 0 here?

query-planner-js/src/buildPlan.ts
```diff
@@ -1326,183 +1210,583 @@ class FetchDependencyGraph {
...
-    this.reduce();
+  if (sameType(type, fragmentType) && neededDirectives.length === 0) {
+    // we can completely skip this fragment and recurse.
+    return removeRedundantFragmentsOfSet(selection.selectionSet, type, unneededDirectives);
+  } else if (neededDirectives.length === fragment.appliedDirectives.length) {
+    // This means we need all of the directives of the fragment, and so just return it.
+    return selection;
+  } else {
+    // We need the fragment, but we can skip some of its directive.
+    const updatedFragement = new FragmentElement(type, fragment.typeCondition);
```
clenfest, line 1749: spelling: should be `updatedFragment`

query-planner-js/src/buildPlan.ts
pcmanus: I hoped the `return` at the end of `if` would alleviate concerns, but the code was being inefficient anyway (restarting the iteration every time we merged, which is unnecessary), so switched to looping with indexes (not sure if that's what you meant by "while loop with pointers" but ...).

query-planner-js/src/buildPlan.ts
pcmanus: I did, though had a hard to come up with very short names, so hopefully it's an improvement.

query-planner-js/src/buildPlan.ts
pcmanus: It's right. It's basically saying that if we find no path whatsoever for a queried field, then well, we cannot plan the overall query at all. Of course, in practice, this shouldn't happen if composition validation is not buggy and has run successfully, since it's the whole point of composition validation to ensure we don't get this during query planning.

Anyway, switched to `return 0` and added a comment.

clenfest (approved):

PR #1908, opened 2022-06-07
Generates __typename edge on interfaces for fed1 supergraphs

### LINKED ISSUE #1907: Error when querying parent field's sibling type in a fragment with fed1 schema
Given the following subgraph and operation, the following error is returned when running the query

```
We should have an edge from Fruit(fruit-subgraph) for __typename
```

It _seems_ like the issue arises because the fragment is selecting a sibling type to `Apple`. Basically the `... on Banana` part of the fragment could never happen because an apple will never be a banana. If you were to write that directly instead of via fragment, you'd get an error.

Note that this works correctly when using fed2 composition. It's only an issue running a fed2 gateway with a fed1 supergraph.

### Subgraph Schema

```graphql
  interface Fruit {
    edible: Boolean!
  }
  
  type Banana implements Fruit {
    edible: Boolean!
    inBunch: Boolean!
  }
  
  type Apple implements Fruit {
    edible: Boolean!
    hasStem: Boolean!
  }
  
  type Query {
    fruit: Fruit!
  } 
```

### Operation

```graphql
fragment OrangeYouGladIDidntSayBanana on Fruit {
  ... on Banana {
    inBunch
  }
  ... on Apple {
    hasStem
  }
}

query Fruitiness {
  fruit {
    ... on Apple {
      ...OrangeYouGladIDidntSayBanana
    }
  }
}
```

### Source

Here's the bit of code that's triggering the error https://github.com/apollographql/federation/blob/a0c734348eadb153848911008f3ffcaa160b38e8/query-graphs-js/src/graphPath.ts#L584-L614

## Working example

An example that reproduces the error can be found at https://stackblitz.com/edit/apolloeg-supergraph-error-qbodxl

pcmanus:
Because fed1 supergraph don't have proper interface information, we
"force type explosion" for interfaces, which effectively means skipping
adding edges for interface fields to the query graph. However, as we
did so, we also end up skipping the edge for `__typename` on the
interface, and that edge is necessary for the query planner to handle
some specific cases where parts of the user query has unsatisfiable
type conditions (intersection of fragment conditions with no runtime
implementations). This patches ensure that `__typename` edge is always
added.

Fixes #1907

### DIFF
--- a/query-graphs-js/src/graphPath.ts
+++ b/query-graphs-js/src/graphPath.ts
@@ -582,7 +582,6 @@ export function isRootPath(path: OpGraphPath<any>): path is OpRootPath {
 }
 
 export function terminateWithNonRequestedTypenameField<V extends Vertex>(path: OpGraphPath<V>): OpGraphPath<V> {
-
   // If the last step of the path was a fragment/type-condition, we want to remove it before we get __typename.
   // The reason is that this avoid cases where this method would make us build plans like:
   // {
--- a/query-graphs-js/src/querygraph.ts
+++ b/query-graphs-js/src/querygraph.ts
@@ -981,8 +981,14 @@ class GraphBuilderFromSchema extends GraphBuilder {
       // "provides" a particular interface field locally *for all the supergraph interfaces implementations* (in other words, we
       // know we can always ask the field to that subgraph directly on the interface and will never miss anything), then we can
       // add a direct edge to the field for the interface in that subgraph (which avoids unnecessary type exploding in practice).
-      if (this.isFederatedSubgraph && !this.forceTypeExplosion) {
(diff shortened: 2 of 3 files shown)


query-graphs-js/src/graphPath.ts
```diff
@@ -582,7 +582,6 @@ export function isRootPath(path: OpGraphPath<any>): path is OpRootPath {
 }
 
 export function terminateWithNonRequestedTypenameField<V extends Vertex>(path: OpGraphPath<V>): OpGraphPath<V> {
-
```
benweatherman, line 585: 👏

benweatherman (approved):
🐐 🍎

PR #1911, opened 2022-06-09
Fix fragment reuse in subgraph fetches

### LINKED ISSUE #1894: Fragments are not reused in query plans creating large request sizes to subgraphs
As part of generating a query plan, fragment selections are not used and expanded so that all the locations where previously a fragment selection was defined, now contains all the fields in that fragment.

This means that if you have repeated and/or nested fragments your operations sent to the gateway can be of a decent size because of the reuse of fragments, but the generated query plans and operations sent to subgraphs are multipled in size. This can lead to the query plan cache using lots more memory and subgraphs having to parse, validate, and cache their operations that are much larger than what the client specified.

### Example

Client sends an operation that has fragments reused multiple times. If this fragment `FooSelect` is really long, say 100 lines, this selection is redeclared every time so the query plan would be ~n*100 in size.

```graphql
query ExampleQuery {
  a {
    ...on A1 {
      ...FooSelect
    }
    ...on A2 {
      ...FooSelect
    }
    ...on A3 {
      ...FooSelect
    }
    ...on A4 {
      ...FooSelect
    }
    ...on A5 {
      ...FooSelect
    }
  }
}
```

You can view a repo and runtime behavior here: https://stackblitz.com/edit/apolloquery-plan-does-not-reuse-fragments?file=README.md

---------------

This is especially noticeable when you have an operation that only does one fetch to a single subgraph and the client sends its operation to the gateway, but then subgraphs are seeing a much larger and different operation, when in theory the gateway could just forward the operation as is. This enhancement is separate from this issue and is discussed over here: https://github.com/apollographql/federation/issues/660

The harder problem to solve is going to be how to solve creating new fragment selections to send to subgraphs when the fragments span across different schemas, the query plan would have to take some time to create new fragments. It could even do this for operations that don't declare fragments to start but adds them for the query plans to reduce the request made to subgraphs. Because it would be changing the operation, we might consider this being an opt-in feature or something that users can trigger based on certain params of an operation

pcmanus:
The query planner has code originally meant to reuse fragments from the
user query in subgraph fetches when it made sense. Unfortunately that
code wasn't triggered at all due to the query planner not properly
capturing those fragments. This ticket fixes this issue, adds more
testing and improves a few related issues.

Fixes #1894

### DIFF
--- a/internals-js/src/operations.ts
+++ b/internals-js/src/operations.ts
@@ -39,13 +39,13 @@ import {
   variableDefinitionsFromAST,
   CompositeType,
   typenameFieldName,
-  NamedType,
   sameDirectiveApplications,
   isConditionalDirective,
   isDirectiveApplicationsSubset,
+  isAbstractType,
 } from "./definitions";
 import { ERRORS } from "./error";
-import { sameType } from "./types";
+import { isDirectSubtype, sameType } from "./types";
 import { assert, mapEntries, MapWithCachedArrays, MultiMap } from "./utils";
 import { argumentsEquals, argumentsFromAST, isValidValue, valueToAST, valueToString } from "./values";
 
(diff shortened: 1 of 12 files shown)


internals-js/src/operations.ts
```diff
@@ -360,30 +360,37 @@ export class Operation {
...
-      optimizedSelection = optimizedSelection.expandFragments(toDeoptimize);
     }
+
+    // We re-expand any fragments that is used less than our minimum. Optimizing all fragments to potentially
+    // re-expand some is not entirely optimal, but it's simple and probably don't matter too much in practice
+    // (we only call this optimization on the final computed query plan, so not a very hot path; plus in most
+    // cases we won't even reach that point either because there is no fragment, or none will have been
+    // optimized away so we'll exit above). We can optimize later if this show up in profiling though.
+    //
+    // Also note `toDeoptimize` will always contains the unused fragments, which will allow `expandFragments`
```
pcmanus, line 387: Note to reviewers: this comment is the reason of the change in this method. Essentially, it removes the prior `minUsagesToOptimize > ` condition so `expandFragments` can "clean up" unused fragments in all cases.

internals-js/src/operations.ts
```diff
@@ -648,10 +676,6 @@ export class SelectionSet extends Freezable<SelectionSet> {
   }
 
   collectUsedFragmentNames(collector: Map<string, number>) {
-    if (!this.fragments) {
```
pcmanus, line 651: Note to reviewers: that condition (and the same below) was currently incorrect because `this.fragments` is not always set at all the levels of a `SelectionSet`. We could probably change this and using `this.fragments` is set at all the levels, but that's a bit easy to get wrong so it feels that not relying on that was less fragile.

query-planner-js/src/buildPlan.ts
```diff
@@ -859,6 +883,100 @@ class FetchGroup {
...
+      original: fragment,
+      expandedSelectionSet,
+      dependentsOn: Array.from(otherFragmentsUsages.keys()),
+    });
+  }
+
+  const optimizedFragments = new NamedFragments();
+  while (fragmentsMap.size > 0) {
+    for (const [name, info] of fragmentsMap) {
+      // Note that graphQL specifies that named fragments cannot have cycles (https://spec.graphql.org/draft/#sec-Fragment-spreads-must-not-form-cycles)
```
benweatherman, line 966: **question** Does this mean that if a user _does_ try to send a query like this that it'll be caught somewhere else?

benweatherman (commented):
Using some real-world queries and schemas, I see a pretty drastic slowdown of query planning. I know we've talked previously that we don't necessarily wanna optimize for "query plans per second" or anything like that, but I wonder if we should do some more digging on this before we're ready to merge. I don't have a good sense for "yeah this will slow down query planning, but it's still better because the request size to subgraphs is much smaller" so thought I'd start the convo.

EDIT: I have some granular results for said queries & schemas; would it be helpful to look through those results to see if there are some common patterns? Maybe it would also be helpful to look at query plan size?

query-planner-js/src/index.ts
```diff
@@ -8,17 +8,31 @@ import { Schema, Operation } from '@apollo/federation-internals';
...
 
-// There isn't much in this class yet, and I didn't want to make too many
-// changes at once, but since we were already storing a pointer to a
-// Rust query planner instance in the gateway, I think it makes sense to retain
-// that structure. I suspect having a class instead of a stand-alone function
-// will come in handy to encapsulate schema-derived data that is used during
-// planning but isn't operation specific. The next step is likely to be to
-// convert `buildQueryPlan` into a method.
+
+export type QPOptions = {
```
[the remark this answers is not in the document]
pcmanus: Addressed through rebasing since it was added by another commit as `QueryPlannerConfig`.

query-planner-js/src/buildPlan.ts
pcmanus: I assume it does :). That is, queries are validated though the graphQL-js validations so unless graphQL-js forgot to validate that one, we're good. I'll admit I didn't double-checked, but I'd be super surprised if it was an oversight of graphQL-js given how bad this could blow up any server.

benweatherman: Gotcha. I had to test it to see how it works for myself 😈 `apollo-server-core` calls off to `graphql-js` to do validation and throws an error and we end up with something like this `GraphQLError: Cannot spread fragment \"edibleFrag\" within itself via \"hasStemFrag\"`.

It does look like `parseOperation` called on a fragment with a named cycle will crash with a maximum recursion limit, but that's probably not worth dealing with since we don't expect people to call that directly.

benweatherman (approved):
🐐

PR #1914, opened 2022-06-15
Cleanup error related code

pcmanus:
This PR started with wanting to remove our usages of the now deprecated `GraphQLError` that takes all arguments positionally (it's been deprecated by graphql-js in favour of a version that take a new `GraphQLErrorOptions` type).

Looking at this however, I noticed a few related cleanup and improvements that are in that PR. In particular:
- we had our own `GraphQLErrorParams` in `error.ts` which was _almost_ the same than `GraphQLErrorOptions`. The only difference being that we included the `message` in our own, while graphql-js decide to keep the message separate. It didn't felt meaningful to keep our own version just for what, so the first commit switch to extracting the `message` as it's own argument in our own error code and reusing `GraphQLErrorOptions`.
- there was a number of cases where we where using `new GraphQLError(...)` but should really have been using a proper error code.

I'll note that this later point means that this patch introduce a few new error codes. As the error code doc is trying to track when code have been introduced, I've tentatively marked those as "adding: 2.1.0" but don't see this as any commitment: if the review of this slips by 2.1.0, I'll just update those pre-commit (once I know for sure where it lands).


internals-js/src/federation.ts
```diff
@@ -1116,34 +1140,23 @@ export function parseFieldSetArgument({
...
+    throw handleFieldSetValidationError(
+      directive,
+      e,
+      (msg: string) => {
+        if (msg.startsWith('Cannot query field')) {
+          if (msg.endsWith('.')) {
+            msg = msg.slice(0, msg.length - 1);
+          }
+          if (directive.name === keyDirectiveSpec.name) {
+            msg = msg + ' (the field should be either be added to this subgraph or, if it should not be resolved by this subgraph, you need to add it to this subgraph with @external).';
```
benweatherman, line 1152: **nitpick** I know this was already there, but since I see it

```suggestion
            msg = msg + ' (the field should either be added to this subgraph or, if it should not be resolved by this subgraph, you need to add it to this subgraph with @external).';
```

benweatherman (approved):
As always, 🐐

PR #1923, opened 2022-06-23
update other servers with latest results

dariuszkuc:
Update matching format introduced in https://github.com/apollographql/apollo-federation-subgraph-compatibility/pull/109.


hwillson (approved):
Looks great - thanks @dariuszkuc!

benweatherman:
There was some side discussion that the custom styling may be getting removed by gatsby

PR #1929, opened 2022-06-24
Remove type chicanery to support Apollo Server 2 types

benweatherman:
it should be impossible to run AS2 AND GW2 because they have no graphql peer dep overlap. GW2 requires 16 at a minimum, which AS2 doesn’t support at all. So we can assume `overallCachePolicy` will exist.


trevor-scheer (approved):
LGTM. Worth mentioning that the lack of overlap situation we have with v2 is sadly _not_ the case for v0.x, so the backcompat change might actually be necessary over there.

benweatherman:
Great clarification @trevor-scheer  🙏

PR #1933, opened 2022-06-27
Add a BREAKING label for the change related to the getDefaultFetcher …

mNalon:
…in the 2.0.2 release

Considering this use case:

```javascript
const fetcher = PROXY_URL
  ? getDefaultFetcher().defaults({ proxy: PROXY_URL })
  : getDefaultFetcher()
```

it should be pointed out as a breaking change since the code would not work after updating to this version.


benweatherman (approved):
Thanks for calling this out @mNalon 🙏  Are you OK with signing the CLA so we can get this merged?

PR #1937, opened 2022-06-28
Fix issue computing query plan costs

pcmanus:
The query plan cost computation, which allows the query planner to pick
the best plans amongst all the possible ones, wasn't always precise
enough in the case of parallel branches. More precisely, the cost of a
number of parallel branches was only based on the number of branches
involved and the max cost amongst the branches cost. But this mean in
particular that the exact cost of the other branches (other than the one
with max cost) were ignored.

Concretely, this could lead to a case where we were comparing 2 plans,
one with 2 parallel fetches of cost, say, 10 and 10, and another one
with also parallel fetches of cost 10 and 1, and ended up giving the
same cost to both option (each have 2 parallel branch and their max
is 10 in both case). But clearly, we'd much rather pick the latter
option in that case.

This commit fixes this issue by ensuring we always take all the costs
of parallel branches into account, and overall simplify a bit the
query plan cost computation.


benweatherman (approved):
🐐

PR #1955, opened 2022-07-03
Update entities.mdx

sharshuv-quotient:
If I understood correctly, in this example, fetchProductByID is a JavaScript _function_, not a GraphQL _field_


pcmanus (approved):
You're right. Thanks for catching that and the PR!

PR #1970, opened 2022-07-09
Add runtime dependency on node-fetch

### LINKED ISSUE #1961: node-fetch runtime dependency is missing in @apollo/gateway v2.0.5
We are currently upgrading our application to Federation v2. After upgrading the necessary packages, we found out that node-fetch is a runtime dependency of @apollo/gateway (current, v2.0.5) and is missing in its package.json.

We therefore get a runtime error:
```bash
Error: Cannot find module 'node-fetch'
Require stack:
- /var/task/node_modules/@apollo/gateway/dist/executeQueryPlan.js
- /var/task/node_modules/@apollo/gateway/dist/index.js 
```

Strange enough the only node-fetch related thing in the package.json are the typings of node-fetch `@types/node-fetch` which is sorted into the `dependencies`, although typing packages typically are  listed in the `devDependencies`-section.

See https://github.com/apollographql/federation/blob/%40apollo/gateway%402.0.5/gateway-js/package.json

Adding node-fetch@2.6.7 as a dependency in our projects `package.json` fixed this problem for us, but is considered a workaround, because our codebase does not use `node-fetch`.

benweatherman:
We use `node-fetch` during runtime for some of the public methods in `RemoteGraphQLDataSource`. Using `node-fetch@2` because v3 is ESM-only. There's already a renovate rule to keep things from going to v3.

We should probably break the interfaces that are using `node-fetch`'s classes, since they aren't used by the default implementation after switching to `make-fetch-happen`.

### Other detritus

- **Remove `@types/node-fetch`** This was originally included in https://github.com/apollographql/apollo-server/pull/3546 as a fix for https://github.com/apollographql/apollo-server/issues/3471.
- **Change some imports to use `type`** so there's no runtime dependency for things that are just using types.
- **Remove `pretty-format`** since that's not a runtime thing

Fixes #1961

## TODO
- [x] Investigate using plain-ol-javascript-object headers instead of `Headers` object
- [x] Add CHANGELOG entry


abernix (approved):
Agree the dependency appears to be required currently because we literally use `new Headers` to pass in the headers to the fetch implementation.

That said,  am curious if we could opt to just pass a plain object as  `headers` rather than a `Headers`.  It's my understanding that most `fetch` implementations will accept merely an object, and the [Fetch API] specification suggests that it's completely fine to do so.  Is that worth a consideration?

Anyhow, just a thought.  LGTM, but also like the idea of not including the dependency at all if we can avoid it.

benweatherman:
> curious if we could opt to just pass a plain object as `headers`

I like this line of thinking and I tried but got hung up on some type conflicts that were buried a bit too far. Lemme dig those up so we can talk specifics, instead of this hand-wavy response I've got so far.

Issue #1921, opened 2022-06-21
Deliver: @defer support in the query planner


pcmanus:
We can implement some support for the upcoming [`@defer` graphQL directive](https://github.com/graphql/graphql-spec/pull/742) in the query planner by using entities keys. This will only work for a `@defer` that applies at the level of an entity, but this feel plenty useful: it sounds likely that a fair number of uses for `@defer` are going to align well with entities.

This ticket is for tracking the implementation of such entity-based `@defer` support _in the query planner_. Additional support for `@defer` could be later provided by passing the remaining non-entity-based defers down to subgraphs supporting it, but this should be considered out of scope for this issue. It also make sense that support for `@stream` be added along the same logic, but this need not be in the scope of this issue either.

pcmanus:
Forgot to updates this earlier, but the bulk of this work was merged (and released) with #1958 so closing this issue.

PR #1975, opened 2022-07-12
Reject directives within `fields` arg of `@key`, `@provides` and `@requires`

pcmanus:
Noticed that we accepting (parsing) directive applications with a `@key`, `@provide` or `@requires`, but were silently ignoring it. In other words:
```graphql
type T @key(fields: "x ... @skip(if: true) { y }") {
  x: Int
  y: Int
}
```
was accepted, but the `@skip` was completely ignored. Of course, it's unclear why would anyone write this, but it still imo qualify as a bug that we silently ignore this since the code essentially do not do what is written. It's certainly an oversight.

More generally, as we don't anything with any such directive applications, and for user/custom directives, there is no way to provide any processing whatsoever, so overall, I think we should just reject any directive applications to avoid any risk of confusion, and this is what the patch does.

I'll further note that:
1. the example above with `@skip` is a bit arbitrary when you need a hard-coded `true` or `false` for the `if`: why would you ever write this? But the context here is actually #1921: someone could want to try to see what a `@defer` does within a `fields` value, and we should be clear it just doesn't work.
2. I'm not trying to say that it will never make sense to allow directives in such values. There absolutely could be future use case for it, but we can just relax this validation when that happens. The point is that we shouldn't accept something we don't actually support.
3. one could make the argument that, insofar as this reject things that were not rejected before, this could be considered as a backward incompatible change. My argument here is that the existing behaviour is imo a bug and so I have no qualm putting this in a minor release (I mean, fixing bugs always modify behaviour after all). But of course, we'll still put a word about it in the changelog.

### DIFF
--- a/docs/source/errors.md
+++ b/docs/source/errors.md
@@ -49,6 +49,7 @@ The following errors might be raised during composition:
 | `INVALID_LINK_DIRECTIVE_USAGE` | An application of the @link directive is invalid/does not respect the specification. | 2.0.0 |  |
 | `INVALID_LINK_IDENTIFIER` | A url/version for a @link feature is invalid/does not respect the specification. | 2.1.0 |  |
 | `INVALID_SUBGRAPH_NAME` | A subgraph name is invalid (subgraph names cannot be a single underscore ("_")). | 2.0.0 |  |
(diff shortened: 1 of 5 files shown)


benweatherman (approved):
🐐

Issue #1971, opened 2022-07-11
Update federation specs for @inaccessible


jeffjakub:
Acceptance Criteria
- update documentation
- Add note about composed directive names being consistent across subgraphs
- link this issue to apollo-federation-subgraph-compatibility test changes.

PR #1992, opened 2022-07-18
1971 inaccessible spec

jeffjakub:
Creating a draft PR to iterate on the `@inaccessible` directive spec. 
This PR relates to: https://github.com/apollographql/federation/issues/1971


docs/source/federation-spec.md
```diff
@@ -413,3 +413,25 @@ type User @key(fields: "id") {
   name: String @override(from: "SubgraphA")
 }
 ```
+### `@inaccessible`
+```graphql
+directive @inaccessible(from: String!) on FIELD_DEFINITION | INTERFACE | OBJECT | UNION | ARGUMENT_DEFINITION | SCALAR | ENUM | ENUM_VALUE | INPUT_OBJECT | INPUT_FIELD_DEFINITION
+```
+
+The `@inaccessible` directive is used to indicate that a location within the schema is inaccessible. Inaccessible types and fields are available internally but are not exposed publically.
```
dariuszkuc, line 421: maybe add a note why you would want to do that -> e.g. something along the lines `Inaccessible can be used to incrementally add schema elements (e.g. fields) to multiple subgraphs without breaking composition.` or maybe something about being useful for surfacing additional data for local debugging, some other use cases?

dariuszkuc, line 421: Lets also call out that location within schema will be inaccessible from the GraphQL Gateway as long as **any** of the subgraphs marks that location as `@inacessible`.

benweatherman: Does this part of the documentation center around _subgraph library authors_ since it's in the `Subgraph specification` part of the docs? We do call out the incremental usecase in the `Federated directives` section https://www.apollographql.com/docs/federation/federated-types/federated-directives/#inaccessible.

Not saying we shouldn't add the note, but also wanted to bring this up (as it was a little confusing for me too).

dariuszkuc: Personally I'd include duplicate information just so folks don't have to jump around multiple pages to get complete info.

docs/source/federation-spec.md
```diff
@@ -413,3 +413,25 @@ type User @key(fields: "id") {
...
 ```
+### `@inaccessible`
+```graphql
+directive @inaccessible(from: String!) on FIELD_DEFINITION | INTERFACE | OBJECT | UNION | ARGUMENT_DEFINITION | SCALAR | ENUM | ENUM_VALUE | INPUT_OBJECT | INPUT_FIELD_DEFINITION
+```
+
+The `@inaccessible` directive is used to indicate that a location within the schema is inaccessible. Inaccessible types and fields are available internally but are not exposed publically. The @inaccessible directive is helpful for adding a new field to a shared value type. Often when you add a value type field in one subgraph, composition fails because that field isn't resolvable in other subgraphs. This directive enables you to preserve composition while adding the field to your remaining subgraphs. When the rollout is complete, you can remove the @inaccessible directive and begin using the field.
+
+```graphql
+extend schema
```
prasek, line 424: @jeffjakub this is the supergraph usage of `@link`.

the [subgraph usage of `@inaccessible` uses `import` from the federation/v2.0 spec](https://github.com/apollographql/supergraph-demo-fed2/blob/0876cd37626c3f582054386901d983ecf2d87777/subgraphs/products/products.graphql#L2).

docs/source/federation-spec.md
```diff
@@ -413,3 +413,25 @@ type User @key(fields: "id") {
   name: String @override(from: "SubgraphA")
 }
 ```
+### `@inaccessible`
+```graphql
+directive @inaccessible(from: String!) on FIELD_DEFINITION | INTERFACE | OBJECT | UNION | ARGUMENT_DEFINITION | SCALAR | ENUM | ENUM_VALUE | INPUT_OBJECT | INPUT_FIELD_DEFINITION
+```
+
+The `@inaccessible` directive is used to indicate that a location within the schema is inaccessible. Inaccessible types and fields are available internally but are not exposed publically. The @inaccessible directive is helpful for adding a new field to a shared value type. Often when you add a value type field in one subgraph, composition fails because that field isn't resolvable in other subgraphs. This directive enables you to preserve composition while adding the field to your remaining subgraphs. When the rollout is complete, you can remove the @inaccessible directive and begin using the field.
```
prasek, line 421: @jeffjakub `@inaccessible` usages in subgraphs are just composed up into the supergraph.  There is NO runtime enforcement of `@inaccessible` in the subgraphs, it's only composed into the supergraph and `@inaccessible` is always enforced at runtime in the supergraph router.

prasek (changes requested):
we need to show the subgraph usage of `@inaccessible` not the supergraph usage.

PR #2021, opened 2022-07-29
Revamp quickstart and gateway articles to prioritize router

StephenBarlow:


docs/source/quickstart/setup.mdx
```diff
@@ -151,65 +151,70 @@ iwr 'https://rover.apollo.dev/win/latest' | iex
...
 ```
-Error: When a manual configuration is not provided, gateway requires an Apollo configuration.
-```
 
-That's because we aren't currently providing a supergraph schema to our gateway! We'll fix that soon.
+</ExpansionPanel>
+
+That's because we aren't currently providing a supergraph schema to the router! We'll fix that soon.
 
 ## 3. Obtain your subgraph schemas
```
rkoron007, line 212: While setting this up, it did feel like I was jumping around from tool to tool in every section, making it harder to understand the cohesion between sections. 

So here is my wild suggestion, what if the order of this article was more like this:
1. Install Rover
2. Create a Studio Account
3. Authenticate Rover with Studio
4. Create a router project directory
5. Obtain your subgraph schemas

So the user starts by setting up Rover, then signing up for Apollo Studio, then authenticating Rover, before installing the Router and creating their `.graphql` files. The user directly transitions from setting up their subgraph files to learning about managed federation in the next quickstart step.

docs/source/quickstart/local-composition.mdx
```diff
@@ -14,7 +14,7 @@ Like Apollo Studio, the Rover CLI needs the following information about each of
```
rkoron007, line 14: On the line above this, a "gateway" mention slipped in!

rkoron007 (commented):
Added a few suggestions :)

docs/source/quickstart/setup.mdx
```diff
@@ -135,7 +135,7 @@ type SubmitReviewResponse {
 
 ## 1. Install the Rover CLI
```
rkoron007, line 136: Add a note for folks to check their Rover version after running the `curl` command. :)

docs/source/quickstart/setup.mdx
```diff
@@ -5,32 +5,32 @@ description: Part 1 - Project setup
 
 > 📣 **Federation 2 demo:** In addition to completing this tutorial, you can [clone our Federation 2 demo on GitHub](https://github.com/apollographql/supergraph-demo-fed2). The demo is distinct from the project you create in this tutorial, and it helps demonstrate many concepts of Federation 2.
```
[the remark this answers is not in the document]
StephenBarlow: Wonderful point! I just totally glossed over this callout and didn't consider that we had a shiny new thing to include in it.

docs/source/quickstart/setup.mdx
StephenBarlow: Great suggestion! I've updated the order and lumped all the tool setup steps under a single high-level step.

docs/source/gateway.mdx
```diff
@@ -1,25 +1,46 @@
...
-## Setup
+Apollo actively supports two options for your graph router:
 
-First, let's install the necessary packages:
+* **The Apollo Router (recommended):** This is a high-performance, precompiled Rust binary.
+    * [Go to the Apollo Router docs.](/router/quickstart/)
+    * [See the federation quickstart](./quickstart/setup/) (which uses the Apollo Router).
+* **Apollo Server:** Apollo Server can act as your graph router via the [`@apollo/gateway`](https://www.npmjs.com/package/@apollo/gateway) extension library.
+
+**We recommend starting with the Apollo Router.** It's faster to configure, it's more performant (especially with high request loads), and it rarely requires writing custom code.
```
jeffjakub, line 21: No change requested, simply identifying this looks like a good place to provide a link to the new RHAI scripting available for the Router. This will enable folks to write custom code without having to know Rust.

jeffjakub (approved):

docs/source/gateway.mdx
StephenBarlow: Good call, I'll add a link!

PR #2025, opened 2022-08-01
Fix bug in mergeDescriptions

clenfest:
Fix bug in mergeDescriptions, and remove type assertion from `otherElementsPrinter`. 

Issue was when merging empty and non-empty descriptions, we'd pass undefined.


benweatherman (approved):
🐐

Issue #1938, opened 2022-06-29
opentelemetry: rename `service` to `gateway.service`


mac2000:
If we rename `service` to something like `gateway.service` [here](https://github.com/apollographql/federation/blob/main/gateway-js/src/executeQueryPlan.ts#L264)

```js
return tracer.startActiveSpan(OpenTelemetrySpanNames.FETCH, {attributes:{service:fetch.serviceName}}, async span => {
```

we may use this attribute in span metrics processors of collectors to monitor the whole ecosystem (just imagine dashboards where you have gateway and underlying services) aka

<img width="466" alt="image" src="https://user-images.githubusercontent.com/88868/176382920-de40b599-a27d-4ab5-9c88-fd516aa0a203.png">

**Why we need this change:** 

- `service` itself is a well-known span attribute, so trying to add it one more time in processor makes no sence
- idiomatic naming is something like `http.host`, `http.status`, also in apollo there is already everything prefixed with `gateway.`

**Alternative:** if for some reason this one may break something, we can just add one more attribute with `gateway.service` name and the same value

**Things to consider:** did not find a way to change/add attributes on the fly, probably it will be an easiest workaround but it is so deep inside of the gateway that seems to be unreal

PR #2032, opened 2022-08-02
Separate package name and version in the OpenTelemetry scope

mmanciop:
Currently the OpenTelemetry tracer created by the Apollo Gateway is creating a `Scope` like:

```
"scope": {
  "name": "@apollo/gateway/0.49.0"
}
```

Embedding the version in the scope name makes it harder and less efficient to match the spans in the OpenTelemetry Collector, e.g., with the transform [1] processor: instead of using a `strict` match to select spans from all `@apollo/gateway` versions, it requires the use of a `regexp` matcher.

For reference, this is how the scope looks like from one of the OpenTelemetry JS instrumentation packages:

```
"scope": {
  "name": "opentelemetry-instrumentation-express",
  "version": "0.28.0"
}
```

This commit changes the way the tracer is instantiated, so that the resulting scope carries the version of the `@apollo/gateway` package in the `version` field:

```
"scope": {
  "name": "@apollo/gateway",
  "version": "0.49.0"
}
```

**Note:** This change will make life easier down the line when earlier versions are retired. For now, it _may_ make existing matching rules miss. Maybe an option would be to introduce this change with the next major release.

[1] https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/transformprocessor


benweatherman:
Thanks for the change @mmanciop !

We're having a similar debate about when to release a similar change, since it _may_ be breaking https://github.com/apollographql/federation/issues/1938. Lemme ask around a bit and see when we should release this, and possibly release this other "breaking" OpenTelemetry change at the same time.

mmanciop:
Hi @benweatherman, have you come to a decision?

pcmanus:
Hi @mmanciop. Really sorry about the lack of progress/news here. Ben had his attention go to other places since his last comments, and we've let this fall to the wayside. Now, your proposed change certainly look more than reasonable to me, but I know next to nothing about open-telemetry and our integration of it at the moment, so I'd like to get some opinions from a more knowledgeable people regarding this (and how big of a deal it is for backward compat), but said people are a bit overbooked this week.

So I guess this is me saying that this may unfortunately still take us a week or two to get to this seriously, but we're not forgetting about it. Again, sorry it's taking this long.

PR #2038, opened 2022-08-04
update federation subgraph info with latest results

dariuszkuc:
Adds NestJS and StepZen integration results.


StephenBarlow (approved):
Looks good to me if this is the right support info! 😄

PR #2041, opened 2022-08-05
fix @composeDirective corner case

clenfest:
@composeDirective wasn't working when @link of the link directive was not present.

The reason this is happening is that although we are waiting to process core schemas for think that aren't the `link` schema, we aren't doing the same for other directives. This fixes this with a slightly hacky approach of catching all exceptions that occur, and then replaying them once `completeSubgraphSchema` has been called, letting them still throw if they threw originally.

### DIFF
--- a/internals-js/src/definitions.ts
+++ b/internals-js/src/definitions.ts
@@ -17,7 +17,8 @@ import {
   VariableNode,
   TypeSystemDefinitionNode,
   SchemaDefinitionNode,
-  TypeDefinitionNode
+  TypeDefinitionNode,
+  DirectiveNode
 } from "graphql";
 import {
   CoreImport,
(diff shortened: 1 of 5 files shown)


internals-js/src/definitions.ts
```diff
@@ -495,11 +496,37 @@ export class Extension<TElement extends ExtendableElement> {
...
+    });
+  }
+
+  processUnappliedDirectives() {
+    for (const { nameOrDef, args, extension, directive } of this._unappliedDirectives) {
+      const d = this.applyDirective(nameOrDef, args);
+      d.setOfExtension(extension);
+      d.sourceAST = directive;
+    }
+  }
```
benweatherman, line 528: **thought** We _may_ want to clear the `_unappliedDirectives` after they've been processed

benweatherman (approved):
🐐

PR #2042, opened 2022-08-05
Add more print schema ordering options

id-ilych:
Hello :wave:

I noticed that `printSchema` only have options to order type and directives definitions. But smaller elements like fields, and types in unions have unspecified order which on major upgrades creates quite chaotic diff in the dumped `.graphql` file.
I added options for sub elements so that it is more stable.

I also added default `localeCompare`-based implementations to `orderPrintedDefinitions` helper and updated affected tests.

My main two concerns:
* Is it ok that I updated the `orderPrintedDefinitions` to give more stable output?
* Is it ok not to create dedicated test for that (there are no tests for `print.ts` file) considering that update of the `orderPrintedDefinitions` is observable in other tests?

### DIFF
--- a/composition-js/src/compose.ts
+++ b/composition-js/src/compose.ts
@@ -3,7 +3,8 @@ import {
   Schema,
   Subgraphs,
   defaultPrintOptions,
-  orderPrintedDefinitions,
+  shallowOrderPrintedDefinitions,
+  PrintOptions,
   ServiceDefinition,
   subgraphsFromServiceList,
   upgradeSubgraphsIfNecessary,
@@ -30,7 +31,15 @@ export interface CompositionSuccess {
   errors?: undefined;
 }
 
-export function compose(subgraphs: Subgraphs): CompositionResult {
+export interface ComposeOptions {
+  sdlPrintOptions: (options: PrintOptions) => PrintOptions;
+}
+
+export const defaultComposeOptions: ComposeOptions = {
+  sdlPrintOptions: shallowOrderPrintedDefinitions
+}
+
+export function compose(subgraphs: Subgraphs, options: ComposeOptions = defaultComposeOptions): CompositionResult {
   const upgradeResult = upgradeSubgraphsIfNecessary(subgraphs);
   if (upgradeResult.errors) {
     return { errors: upgradeResult.errors };
@@ -60,7 +69,7 @@ export function compose(subgraphs: Subgraphs): CompositionResult {
   try {
     supergraphSdl = printSchema(
       supergraphSchema,
-      orderPrintedDefinitions(defaultPrintOptions)
+      options.sdlPrintOptions(defaultPrintOptions)
     );
   } catch (err) {
     return { errors: [err] };
@@ -73,7 +82,7 @@ export function compose(subgraphs: Subgraphs): CompositionResult {
   };
 }
 
-export function composeServices(services: ServiceDefinition[]): CompositionResult  {
+export function composeServices(services: ServiceDefinition[], options: ComposeOptions = defaultComposeOptions): CompositionResult  {
   const subgraphs = subgraphsFromServiceList(services);
   if (Array.isArray(subgraphs)) {
     // Errors in subgraphs are not truly "composition" errors, but it's probably still the best place
@@ -81,5 +90,5 @@ export function composeServices(services: ServiceDefinition[]): CompositionResul
     // include the subgraph name in their message.
     return { errors: subgraphs };
   }
-  return compose(subgraphs);
+  return compose(subgraphs, options);
 }
--- a/composition-js/src/compose.ts
+++ b/composition-js/src/compose.ts
@@ -8,6 +8,8 @@ import {
   ServiceDefinition,
   subgraphsFromServiceList,
   upgradeSubgraphsIfNecessary,
+  SubtypingRule,
+  assert,
 } from "@apollo/federation-internals";
 import { GraphQLError } from "graphql";
 import { buildFederatedQueryGraph, buildSupergraphAPIQueryGraph } from "@apollo/query-graphs";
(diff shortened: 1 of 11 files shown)


benweatherman:
howdy @id-ilych  and thanks for the PR! Would you mind updating with the latest from `main` (apologies, we just merged something that required updating a few of the same tests). Also, would you mind adding a note at the top of the `composition-js` and `internals-js` `CHANGELOG`s?

> Is it ok that I updated the orderPrintedDefinitions to give more stable output?

> Is it ok not to create dedicated test for that (there are no tests for print.ts file) considering that update of the orderPrintedDefinitions is observable in other tests?

Yes to both of these!

id-ilych:
Thanks @benweatherman. I did as you suggested.

benweatherman:
@id-ilych thanks for the updates! for some reason, our CI isn't running fully on your commits. Have you signed the apollo CLA previously? is there any chance you're using CircleCI's follow feature to follow your fork of the code?

id-ilych:
Hey @benweatherman :wave: I have signed CLA on 23 Jun 2021. But I did so again, just in case. Regarding the CircleCI, I did not do any configuration. Should I?

benweatherman:
No need to do any special config. Following the fork in CircleCI was the internal working theory as to why the tests weren't running. Will dig into this a bit more.

pcmanus:
I'll admit that I'm on the fence about changing the current output of composition.

I'm totally on board about adding the option to order everything, and in principle I don't disagree that ordering everything can be desirable. But changing the order has very visible consequences, and doing so "by default" might lead to some painful/scary initial upgrades.

Such change to the default could be something we do in a major release, but I think we trying hard to not do one of those any time soon, and I'm less confortable doing it in a minor release (but open to other opinions).

That said, even if we keep the current default, I'm again very on board with adding ordering options. That does imply adding some option to the `compose` method though, and begs the question of how such options gets passed down to that method by its existing callers.

id-ilych:
Hello @pcmanus :wave: What do you think about adding a separate method `shallowOrderPrintedDefinitions` with an "old" behavior and use it in the composite by default, so that default behavior is preserved?

pcmanus:
First, sorry for the delay in my answers here.

Since my last comment, I realised that things weren't exactly as I though they were around this issue. Let me explain.

I "though" that the ordering composition is using in federation 2 (which orders types/directives but fields within types) was the same one that in federation 1 and the one composition had always used. Which is why I was a bit worried about changing it now.

Fwiw, that's why I initially didn't understood why you said that "on major upgrades creates quite chaotic diff".

Anyway, the reason I held that belief is that I implemented the order for federation 2 and when I did so, I very much was careful to match federation 1 behaviour (I probably wouldn't have sorted only some of the elements otherwise).

But apparently there was a communication issue somewhere a year ago, because I missed that https://github.com/apollographql/federation/pull/824 was committed to federation 1 to change the ordering so everything gets sorted. This was unfortunately done after the work on federation 2 had started, and I never realised the behaviour had changed in federation 1.

Tl;dr, the fact that the ordering changed in the major upgrade between federation 1 and federation 2 is very much an oversight/accident. Not sure how that could have slipped all this time but well ...

Anyway, this does change a bit my position here. Because:
1. I do agree that the current ordering in federation 2 is inconsistent.
2. I probably overestimated before how big of a deal it is to change the ordering in a non-major release since, well, we did it with #824 and while I'm sure it was noticed, there also wasn't too much complaints.
3. we did inadvertently change the ordering in federation 2 upgrades.

So I think my new preferred position here is to decide what ordering we actually want long term (at least by default), and switch to that even if it mean one more change of ordering, but then make sure we're better at keeping that consistent moving forward.

Long story short, maybe we just get back to your original suggestion @id-ilych after all, but I'd like to consult a few people that may opinions on this before completely decided here.

> What do you think about adding a separate method shallowOrderPrintedDefinitions with an "old" behaviour

Whatever we end up deciding of the default, I'm totally ok with adding more methods that implement various option in the code, so +1 on that part.

id-ilych:
Hello @pcmanus :wave: Any updates on that?

id-ilych:
Hello @benweatherman @pcmanus :wave: A friendly reminder about pending PR.

id-ilych:
Hello @benweatherman @pcmanus 👋 A friendly reminder about pending PR.

benweatherman:
howdy @id-ilych! Unfortunately I no longer have the commit-bit privileges for this repo, so someone within the apollo team will need to weigh in on getting this merged.

From what I recall before I left: This is a good change, to be sure. I imagine the apollo team is wrestling with which version in which to release this type of change. It's valuable, but could cause a lot of downstream headaches looking at schema diffs that might only be reordering things.

@jeffjakub might also be able to help with this

id-ilych:
Hello @jeffjakub :wave: Can you please take a look at clarify the status? What needs to be done from my side to get this merged? This PR is stuck for 5 months already :(

pcmanus:
I'm very sorry on the progress on this issue. I certainly understand that this is frustrating.

To be transparent, I've been having a bit of analysis paralysis on the question of whether we should change the default composition ordering, and if yes, to what. I was hoping to get some larger conversation going on that question to gather some consensus, but didn't got much engagement so far, either because it's not very top of mind for most folks or because the status quo is good enough (but I'll admit too that I have had other priorities myself and I haven't pushed that much).

That said, I do realise that the original ask was only to add more options in the code, and not so much about changing the default for composition, and it's entirely on me for distracting from that original purpose.

Anyway, I'm still iffy about changing the default ordering of composition at the moment (until some greater consensus emerges about that question), but maybe we can just go back to your earlier suggestions of:
> What do you think about adding a separate method shallowOrderPrintedDefinitions with an "old" behavior and use it in the composite by default, so that default behavior is preserved?

If the patch is updated to do so, with no change to the default behaviour, then I'll commit to getting it in quickly.

Again, my apologies for my poor handling of this issue.

composition-js/src/compose.ts
```diff
@@ -30,7 +31,15 @@ export interface CompositionSuccess {
   errors?: undefined;
 }
 
-export function compose(subgraphs: Subgraphs): CompositionResult {
+export interface ComposeOptions {
```
id-ilych, line 34: I would prefer to call it `CompositionOptions` but this name is already exported from `merge` :\

I decided not to rename the existing interface to use it for the new thing to avoid breaking somebody's code in a confusing way: same name, totally different usage and meaning.

id-ilych:
Hey @pcmanus :wave:

Please check if that's what you meant.

pcmanus (commented):
Thank you, this looks great. I left a small comment regarding `ComposeOptions`, and we'll need to rollback the changelog manual edits, but happy to commit after that (and happy to deal with those minor "administrative" details myself if you prefer, just let me know)

id-ilych:
Hello @pcmanus :wave:

I am not sure I fully got your idea bout options. Would you mind finishing it on your own to avoid communication overhead?

pcmanus:
Of course. I've pushed a few commits to the branch (and took the liberty to rebase on `main`). The 2 first one just remove the changelog entries and add a `changesets` one instead. The last one is what I meant for the composition options. I did took the liberty of switching the printing option from a function (`(PrintOptions) => PrintOptions`) to just the option object (`PrintOptions`), as this felt a tad more logical to me.

pcmanus (approved):

PR #2045, opened 2022-08-07
Content edit to Fed 2 subgraph spec

StephenBarlow:


docs/source/federation-spec.mdx
```diff
@@ -0,0 +1,623 @@
...
+
+> This content is provided for developers adding federated subgraph support to a GraphQL server library, and for anyone curious about the inner workings of federation. You do _not_ need to read this if you're building a supergraph with existing [subgraph-compatible libraries](./supported-subgraphs/), such as Apollo Server.
+>
+> Servers that are partially or fully compatible with this specification are tracked in Apollo's [subgraph compatibility repository](https://github.com/apollographql/apollo-federation-subgraph-compatibility).
+
+For a GraphQL service to operate as an Apollo Federation 2 subgraph, it must do all of the following:
+
+- Automatically extend its schema with all definitions listed in the [subgraph schema specification](#subgraph-schema-specification)
+- Correctly resolve the `Query._service` [enhanced introspection field](#fetch-service-capabilities)
+- Correctly generate [stubs for externally referenced types](#generating-stub-types)
```
StephenBarlow, line 14: Not sure this section needs to remain

docs/source/federation-spec.mdx
```diff
@@ -0,0 +1,623 @@
...
+```graphql {2}
+type Query {
+  _entities(representations: [_Any!]!): [_Entity]!
+}
+```
+
+Every subgraph must also automatically define the resolver for this field. The  logic for this resolver is as follows:
+
+1. For each entity representation included in the `representations` list, obtain the `__typename` from that representation.
+2. Execute the **reference resolver** for the corresponding `__typename` and provide the full representation to that resolver.
```
StephenBarlow, line 335: I suppose a "reference resolver" is how `@apollo/subgraph` happens to handle resolving representations, but it isn't the mandatory solution for other libraries, right? So I think we need to make this section a little more implementation-agnostic

docs/source/federation-spec.mdx
```diff
@@ -0,0 +1,623 @@
...
+  for: link__Purpose,
+  import: [link__Import]
+) repeatable on SCHEMA
+```
+
+The `@link` directive links definitions within the document to external schemas.
+
+External schemas are identified by their `url`, which optionally ends with a name and version with the following format: `{NAME}`/v`{MAJOR}`.`{MINOR}`
+
+The presence of a `@link` directive makes a document a [core schema](https://specs.apollo.dev/#def-core-schema).
```
StephenBarlow, line 492: Is this outdated?

docs/source/federation-spec.mdx
```diff
@@ -0,0 +1,625 @@
...
+
+```graphql {1-4}
+# ⚠️ This definition must be created dynamically! The union
+#   must include every object type in the schema that uses
+#   the @key directive (i.e., all federated entities).
+union _Entity
+
+scalar _Any
+scalar FieldSet
+scalar link__Import
```
dariuszkuc, line 31: technically `link__` namespace could be skipped IFF those elements are imported through `@link` directive, similarly how we skip the default `federation__` namespace for all other directives.

If elements are not imported through `@link` they need to be namespaced either with default value from the linked spec OR based on the specific namespace value (`@link(url: <spec url>, as: "myNamespace")`)

docs/source/federation-spec.mdx
```diff
@@ -0,0 +1,625 @@
...
+directive @inaccessible on FIELD_DEFINITION | OBJECT | INTERFACE | UNION | ARGUMENT_DEFINITION | SCALAR | ENUM | ENUM_VALUE | INPUT_OBJECT | INPUT_FIELD_DEFINITION
+directive @override(from: String!) on FIELD_DEFINITION
+
+# This is an optional directive (see the glossary below)
+directive @extends on OBJECT | INTERFACE
+```
+
+## Enhanced introspection with `Query._service`
+
+A federated graph router (also known as a gateway) can compose its supergraph schema dynamically at runtime. To do so, it first executes the following enhanced introspection query on each of its subgraphs to obtain all subgraph schemas:
```
dariuszkuc, line 69: AFAIK `_service { sdl }` query is only used for local composition (i.e. running in dev mode). Once gateway is deployed to test/prod we would use rover to publish the schemas.

docs/source/federation-spec.mdx
```diff
@@ -0,0 +1,625 @@
...
+- `Query._service`
+- `_Service.sdl`
+
+`Query._service` returns a `_Service` object, which in turn has a single field, `sdl` (short for schema definition language). The `sdl` field returns a string representation of the subgraph's schema.
+
+The returned `sdl` string has the following requirements:
+
+* It must **include** all uses of all federation-specific directives, such as `@key`.
+    * For a list of all federation-specific subgraph directives, see the [glossary](#glossary-of-schema-additions).
+* It must **omit** all automatically added definitions from the [subgraph schema specification](#subgraph-schema-specification), such as `Query._service` and `_Service.sdl`!
```
dariuszkuc, line 106: this is true for fed v1, it is fine to include all the information in fed v2 schema

docs/source/federation-spec.mdx
```diff
@@ -0,0 +1,625 @@
...
+    - The graph router requires these federation-specific directives to perform composition successfully.
+-  If a subgraph server "disables introspection", the enhanced introspection query is still available.
+
+> Note that the `_service` field is _not_ included in the composed supergraph schema. For security reasons, it's intended solely for use by the graph router.
+
+### Required resolvers
+
+To support the enhanced introspection query, a subgraph service must define resolvers for the following fields:
+
+- `Query._service`
```
dariuszkuc, line 97: instead maybe provide this info as GraphQL SDL (as otherwise you will need to specify the nullability)?

```
extend type Query {
  _service: _Service!
}

type _Service {
  sdl: String!
}
```

dariuszkuc: technically the only required resolver would be for `_entities` query if there is a federated type (type with `@key`) in the schema

docs/source/federation-spec.mdx
```diff
@@ -0,0 +1,625 @@
...
+type Product @key(fields: "upc") {
+  upc: String!
+  weight: Int! @external
+  shippingCost: Int! @requires("weight")
+}
+```
+
+In the above example, the Shipping subgraph defines a `shippingCost` field for the `Product` entity. The `shippingCost` field `@requires` the value of the `weight` field, which is a field that's resolved by a _different_ subgraph. To indicate that `weight` is resolved only by _other_ subgraphs, the Shipping subgraph _also_ defines the field, but it applies the `@external` directive to it.
+
+This allows us to be explicit about dependencies on another service, and service composition will verify that an `@external` field matches up with the original field definition, which can catch mistakes or migration issues (when the original field changes its type for example). `@external` fields also give an individual service the type information it needs to validate and decode incoming representations (this is especially important for custom scalars), without requiring the composed graph schema to be available at runtime in each service.
```
dariuszkuc, line 142: nit: in fed v1 `@external` was required on any field referenced from `@key/@requires/@provides`, fed v2 relaxed this requirement and now it is only needed for fields referenced by `@requires` directive and in some cases for `@provides` (only needed if field is resolved locally in some but not all paths)

docs/source/federation-spec.mdx
```diff
@@ -0,0 +1,625 @@
...
+}
+```
+
+The value returned for the `sdl` field should include all of this information, including directives (excessive whitespace can be removed).
+
+## Generating stub types
+
+Each subgraph in a supergraph should be able to run successfully in isolation, regardless of whether any _other_ subgraphs are running. Many subgraphs run successfully in isolation with no additional logic required. However, subgraph fields _can_ depend on types and fields that are defined in _other_ subgraphs. Whenever this is the case, a subgraph needs to define its own **stubs** for these external types and fields.
+
+To indicate a dependency on another subgraph's type or field, a subgraph schema field uses either the `@requires` or `@provides` directive:
```
dariuszkuc, line 130: maybe we should also add an example for `@provides`?

docs/source/federation-spec.mdx
```diff
@@ -0,0 +1,625 @@
...
+
+> If a subgraph doesn't define any entity types, then it should _not_ define the `Query._entities` field.
+
+The graph router uses this entry point to directly fetch fields of entity objects. It combines those fields with _other_ fields of the same entity that are returned by _other_ subgraphs.
+
+The `Query._entities` field takes a required `representations` argument, which is a list of entity **representations**. A representation is an object that contains all of an entity's `@key` fields, plus its `__typename` field. These are the fields that a subgraph can use to uniquely identify a particular instance of an entity.
+
+Each item in the `representations` list is an `_Any` scalar. This is a federation-specific scalar defined in the [subgraph schema specification](#subgraph-schema-specification). This generic scalar enables the graph router to include representations of _different_ entities in the _same_ query, all of which can have a different shape.
+
+The `Query._entities` field returns a list of objects that belong to the `_Entity` union type, which is why this union must include _all_ entities that the subgraph defines.
```
dariuszkuc, line 238: lets call out the order -> `_entities` return a list of objects that are in the same order as the passed in representations (there was a question about it in the past)

docs/source/federation-spec.mdx
```diff
@@ -0,0 +1,625 @@
...
+type Query {
+  _entities(representations: [_Any!]!): [_Entity]!
+}
+```
+
+> If a subgraph doesn't define any entity types, then it should _not_ define the `Query._entities` field.
+
+The graph router uses this entry point to directly fetch fields of entity objects. It combines those fields with _other_ fields of the same entity that are returned by _other_ subgraphs.
+
+The `Query._entities` field takes a required `representations` argument, which is a list of entity **representations**. A representation is an object that contains all of an entity's `@key` fields, plus its `__typename` field. These are the fields that a subgraph can use to uniquely identify a particular instance of an entity.
```
dariuszkuc, line 234: nit: entity can specify multiple `@key`s (it is a repeatable directive) so it will contain one of field sets that identity the target object

docs/source/federation-spec.mdx
dariuszkuc: yes, libraries have to implement this logic

docs/source/federation-spec.mdx
```diff
@@ -0,0 +1,625 @@
...
+
+For details, see [Resolving entity fields with `Query._entities`](#resolving-entity-fields-with-query_entities).
+
+### == Added types ==
+
+### `type _Service`
+
+This object type must have an `sdl: String!` field, which returns the SDL of the subgraph schema as a string.
+
+The returned schema string _must_ include all uses of federation-specific directives (`@key`, `@requires`, etc.), and it _must not_ include any definitions from the [subgraph schema specification](#subgraph-schema-specification).
```
dariuszkuc, line 381: again this is true for fed v1, fed v2 can include full SDL definition with all the types

docs/source/federation-spec.mdx
```diff
@@ -0,0 +1,625 @@
...
+
+**⚠️ This union type is generated dynamically based on the input subgraph schema!**
+
+This union's possible types must include all entities that the subgraph defines. It's the return type of the `Query._entities` field, which the graph router uses to directly access a subgraph's entity fields.
+
+For details, see [Defining the `_Entity` union](#defining-the-_entity-union).
+
+### `scalar _Any`
+
+This scalar is the type used for entity **representations** that the graph router passes to the `Query._entities` field. An `_Any` scalar is validated by matching its `__typename` and `@key` fields against entities defined in the subgraph schema.
```
dariuszkuc, line 395: do we need to call out this is basically a map of field names/field value pairs (values could be scalars or objects)?

docs/source/federation-spec.mdx
```diff
@@ -0,0 +1,625 @@
...
+
+### `scalar FieldSet`
+
+This string-serialized scalar represents a set of fields that's passed to a federated directive, such as `@requires` or `@provides`.
```
dariuszkuc, line 399: I'd call out `@key` as that is the primary use case

docs/source/federation-spec.mdx
```diff
@@ -0,0 +1,625 @@
...
+
+### == Added directives ==
+
+### `@external`
+
+```graphql
+directive @external on FIELD_DEFINITION
+```
+
+The `@external` directive indicates that a particular subgraph _cannot_ resolve a particular entity field. This enables a subgraph to reference a _different_ subgraph's field _without_ also resolving that field.
```
dariuszkuc, line 411: this is not completely true

```
type Foo @key("id") {
  id: ID!
  localBar: Bar @provides("baz") # resolves field locally even though it is marked as external
  remoteBar: Bar # resolves field remotely
}

extend type Bar @key("id) {
  id: ID!
  baz: String @external
}
```

docs/source/federation-spec.mdx
```diff
@@ -0,0 +1,625 @@
...
+
+In the example above, the Shipping subgraph adds the `shippingCost` field to the `Product` entity. This field requires the value of the `weight` field, which _cannot_ be resolved by the Shipping subgraph. Therefore, the Shipping subgraph defines the `weight` field but also marks it as `@external`.
+
+### `@inaccessible`
+
+```graphql
+directive @inaccessible(from: String!) on FIELD_DEFINITION | INTERFACE | OBJECT | UNION | ARGUMENT_DEFINITION | SCALAR | ENUM | ENUM_VALUE | INPUT_OBJECT | INPUT_FIELD_DEFINITION
+```
+
+The `@inaccessible` directive indicates that a location in the subgraph schema should _not_ be included in the [API schema](./federated-types/overview/#api-schema) that clients use to execute operations against the graph router.
```
dariuszkuc, line 431: lets add a primary use case -> it can be used for defining private keys that cannot be queried by the api schema but will be used by gateway for composition

docs/source/federation-spec.mdx
```diff
@@ -0,0 +1,625 @@
...
+  createdBy: User
+  hidden: String @inaccessible
+}
+```
+
+
+### `@key`
+
+```graphql
+directive @key(fields: FieldSet!, resolvable: Boolean = true) repeatable on OBJECT | INTERFACE
```
dariuszkuc, line 455: lets add some docs about the `resolvable = false`

docs/source/federation-spec.mdx
```diff
@@ -0,0 +1,625 @@
...
+
+In the example above, we import various directives from `federation/v2.0` into our namespace. We also rename one of them, bringing in federation's `@tag` as `@mytag` to distinguish it from a different `@tag` directive already in the schema.
+
+### `@override`
+
+```graphql
+directive @override(from: String!) on FIELD_DEFINITION
+```
+
+The `@override` directive indicates that the current subgraph takes responsibility for resolving a field _away_ from the subgraph specified in the `from` argument.
```
dariuszkuc, line 523: do we need to call out that the subgraph name has to match the one in the studio?

docs/source/federation-spec.mdx
```diff
@@ -0,0 +1,625 @@
...
+  product: Product @provides(fields: "name")
+}
+
+extend type Product @key(fields: "upc") {
+  upc: String @external
+  name: String @external
+}
+```
+
+By default, this subgraph _can't_ resolve the `Product.upc` or `Product.name` fields, so it marks those fields as `@external`. However, this subgraph _can_ resolve the `Product.name` field specifically when it's a subfield of `Review.product`. So, it applies the `@provides` directive to `Review.product` to indicate this fact to the graph router.
```
dariuszkuc, line 568: technically federation v2 does not require `@external` on keys anymore, similarly if that is the only query path that `Product.name` can be resolved then there is no need for `@external` neither

docs/source/federation-spec.mdx
dariuszkuc: in fed v2 `@external` is only required for `@required` fields and in some cases (see above) `@provides` fields

docs/source/federation-spec.mdx
```diff
@@ -0,0 +1,625 @@
...
+directive @external on FIELD_DEFINITION
+directive @requires(fields: FieldSet!) on FIELD_DEFINITION
+directive @provides(fields: FieldSet!) on FIELD_DEFINITION
+directive @key(fields: FieldSet!, resolvable: Boolean = true) repeatable on OBJECT | INTERFACE
+directive @link(url: String, as: String, for: link__Purpose, import: [link__Import]) repeatable on SCHEMA
+directive @shareable on OBJECT | FIELD_DEFINITION
+directive @inaccessible on FIELD_DEFINITION | OBJECT | INTERFACE | UNION | ARGUMENT_DEFINITION | SCALAR | ENUM | ENUM_VALUE | INPUT_OBJECT | INPUT_FIELD_DEFINITION
+directive @override(from: String!) on FIELD_DEFINITION
+
+# This is an optional directive (see the glossary below)
```
dariuszkuc, line 63: where is the glossary?

docs/source/federation-spec.mdx
```diff
@@ -0,0 +1,625 @@
...
+
+This scalar is the type used for entity **representations** that the graph router passes to the `Query._entities` field. An `_Any` scalar is validated by matching its `__typename` and `@key` fields against entities defined in the subgraph schema.
+
+### `scalar FieldSet`
+
+This string-serialized scalar represents a set of fields that's passed to a federated directive, such as `@requires` or `@provides`.
+
+Grammatically, a `FieldSet` is a [selection set](http://spec.graphql.org/draft/#sec-Selection-Sets) minus the outermost curly braces. It can represent a single field (`"upc"`), multiple fields (`"id countryCode"`), and even nested selection sets (`"id organization { id }"`).
+
+### == Added directives ==
```
dariuszkuc, line 403: should we also list `@extends`? it is used by some libs as a workaround for lack of the `extend type` syntax

docs/source/federation-spec.mdx
```diff
@@ -0,0 +1,625 @@
...
+}
+
+directive @external on FIELD_DEFINITION
+directive @requires(fields: FieldSet!) on FIELD_DEFINITION
+directive @provides(fields: FieldSet!) on FIELD_DEFINITION
+directive @key(fields: FieldSet!, resolvable: Boolean = true) repeatable on OBJECT | INTERFACE
+directive @link(url: String, as: String, for: link__Purpose, import: [link__Import]) repeatable on SCHEMA
+directive @shareable on OBJECT | FIELD_DEFINITION
+directive @inaccessible on FIELD_DEFINITION | OBJECT | INTERFACE | UNION | ARGUMENT_DEFINITION | SCALAR | ENUM | ENUM_VALUE | INPUT_OBJECT | INPUT_FIELD_DEFINITION
+directive @override(from: String!) on FIELD_DEFINITION
```
patrick91, line 61: My understanding was that printing/writing the directive definition inside schema was optional, is that not the case?

docs/source/federation-spec.mdx
patrick91: What's our recommendation? Personally, I've left `_service` in strawberry, but maybe I should hide it?

What about `_entities`? Should that be removed as well?

docs/source/federation-spec.mdx
```diff
@@ -0,0 +1,625 @@
...
+
+As a reminder, here's the definition of the `Query._entities` field that every subgraph must automatically define (unless that subgraph defines zero entities):
+
+```graphql {2}
+type Query {
+  _entities(representations: [_Any!]!): [_Entity]!
+}
+```
+
+Every subgraph must also automatically define the resolver for this field. The  logic for this resolver is as follows:
```
patrick91, line 334:
```suggestion
Every subgraph must also automatically define the resolver for this field. The logic for this resolver is as follows:
```

docs/source/federation-spec.mdx
patrick91: maybe, the way link is declared in that spec is slightly different that how it is declared in federation:

```
directive @link(
  url: String!,  # url is non-nullable here
  as: String,
  import: [Import],
  for: Purpose
) repeatable on SCHEMA
```

docs/source/federation-spec.mdx
```diff
@@ -0,0 +1,625 @@
...
+) repeatable on SCHEMA
+```
+
+The `@link` directive links definitions within the document to external schemas.
+
+External schemas are identified by their `url`, which optionally ends with a name and version with the following format: `{NAME}`/v`{MAJOR}`.`{MINOR}`
+
+The presence of a `@link` directive makes a document a [core schema](https://specs.apollo.dev/#def-core-schema).
+
+The `for` argument describes the purpose of a `@link`. Currently accepted values are `SECURITY` or `EXECUTION`. Core schema-aware servers such as Apollo Router and Gateway will refuse to operate on schemas that contain `@link`s to unsupported specs which are `for: SECURITY` or `for: EXECUTION`.
```
patrick91, line 496: > which **are not**

?

docs/source/federation-spec.mdx
dariuszkuc: yes including them in printed SDL (returned by `_service { sdl }` query) is optional but I think they have to be defined within the GraphQL server in order to run (at least `graphql-java` requires definitions to be present)

docs/source/federation-spec.mdx
dariuszkuc: I believe we need to exclude it from `_service { sdl }` for fed 1 but it is fine to keep them for fed 2 (simpler as there is no need for custom logic to filter stuff)

patrick91: got it, maybe we should write that since it will simplify the implementations 😊

docs/source/federation-spec.mdx
patrick91: ah, that makes sense for schema first libraries!

docs/source/federation-spec.mdx
StephenBarlow: This is 100% what we _want_ folks to do, and ideally they wouldn't even use dynamic composition in dev mode. I'll add some bolding and prominence to the callout below to clarify just how much this is discouraged, but I suppose at least for now it is a core requirement for a subgraph.

docs/source/federation-spec.mdx
StephenBarlow: Good idea!

docs/source/federation-spec.mdx
StephenBarlow: good call

docs/source/federation-spec.mdx
StephenBarlow: Yes, I'll get this back in—I prematurely cut it without asking purely because `extend` became much less important of a keyword for Fed 2, but there are certainly still applications for it

docs/source/federation-spec.mdx
StephenBarlow: Right, I use better language for this in the other directives article and I'll make sure to review. The point being at least that a subgraph _usually_ can't resolve an `@external` field unless there's a path that allows it via `@provides`

docs/source/federation-spec.mdx
StephenBarlow: Added a clarification that this is required only if supporting Fed 1

docs/source/federation-spec.mdx
StephenBarlow: Replaced this comment to be more standalone and added `@extends` to the federated directives article

docs/source/federation-spec.mdx
StephenBarlow: Added language to clarify that these definitions need to exist within the server via whatever mechanism the library supports (if not via static SDL)

docs/source/federation-spec.mdx
StephenBarlow: This is a cool and powerful use case that is very important and that I had never considered 😮

docs/source/federation-spec.mdx
StephenBarlow: I'm going to leave it to one example in the federated directives article, just because the listings for `@requires` and `@provides` are both very close by and provide their own examples

docs/source/federation-spec.mdx
StephenBarlow: Good call, this content is now cut so no problem here

docs/source/federation-spec.mdx
StephenBarlow: Should the subgraph spec _require_ this definition alongside the other required definitions?

```
extend schema
  @link(url: "https://specs.apollo.dev/link/v1.0/", import: ["@link", "Import", "Purpose"])
```

Not necessarily because it's necessary to un-namespace `Import` and `Purpose`, but because otherwise `@link` isn't being tied to its associated spec

StephenBarlow:
@dariuszkuc @patrick91 Thanks for all the thoughtful feedback! I've updated the PR with one squashed commit-of-commits, so the diff since the last review should all be straightforward here: https://github.com/apollographql/federation/pull/2045/commits/67d774314989ec3ed7481d6a9414e160f9c068f5

Hopefully this properly addresses the majority of outstanding feedback and we can knock out any remaining concerns this week.

docs/source/federated-types/federated-directives.mdx
```diff
@@ -32,13 +32,30 @@ extend schema
 
 This example subgraph schema uses `@primaryKey` for the federated directive usually named [`@key`](#key).
 
-## Managing entities
+### `@link`
```
dariuszkuc, line 35: its somewhat odd that `@link` definition appears after mentioning it above for importing and renaming the directives - maybe it should be moved up?

docs/source/federated-types/federated-directives.mdx
```diff
@@ -48,6 +65,20 @@ type Product @key(fields: "id") {
...
+
+```graphql {1}
+type Product @key(fields: "upc") @key(fields: "sku") {
+  upc: ID!
+  sku: ID!
+  name: String
+}
+```
+
+To check whether your subgraph library supports repeatable directives, see the `@KEY (MULTI)` item in [Federation-compatible subgraph implementations](../supported-subgraphs/).
```
dariuszkuc, line 78: actually we just fixed that in the subgraph repo and added a proper test (previously we were not really checking for it) -> new test is named `repeatable @key` (we still got ~half of implementations to update to the latest schema, once updated we'll fix the docs)

docs/source/federated-types/federated-directives.mdx
```diff
@@ -48,6 +65,20 @@ type Product @key(fields: "id") {
...
+
+> Although it's valid to apply `@key` to an interface definition, this currently has no effect in Apollo Federation.
```
dariuszkuc, line 80: I thought that it would be inherited by the implementing classes? so if you add `@key` to the interface then all implementing classes will have it applied automatically

docs/source/federated-types/federated-directives.mdx
```diff
@@ -102,13 +133,31 @@ The default value is `true`.
...
 
+### `@extends`
+
+```graphql
+directive @extends on OBJECT | INTERFACE
+```
+
+Indicates that an object or interface definition is an extension of _another_ definition of that same type.
+
+> ⚠️ **If you define your subgraph schema in a static text file, do not use this directive!** Instead, use GraphQL's built-in `extend` keyword.
```
dariuszkuc, line 144: unsure if this is true, I think the general rule is -> use `@extends` only if your subgraph library does not support the `extend` keyword syntax

docs/source/federation-spec.mdx
dariuszkuc: I believe the current suggestion is to NOT import the link spec by default as it is automatically added by the router. Devs can import the link spec if they want to avoid name-spacing of `Import` scalar and `Purpose` enum, but should omit it otherwise.

Side note: when importing a spec, it automatically pulls in the corresponding directive matching the spec - so we don't need to explicitly import `@link` in the example above (by importing federation spec we also pull in `@federation` directive).

docs/source/federation-spec.mdx
```diff
@@ -0,0 +1,447 @@
...
+  SECURITY
+
+  """
+  `EXECUTION` features provide metadata necessary for operation execution.
+  """
+  EXECUTION
+}
+
+type _Service {
+  sdl: String
```
dariuszkuc, line 47:
```suggestion
  sdl: String!
```

docs/source/federated-types/federated-directives.mdx
dariuszkuc: also should we add sections about namespacing? i.e. if you don't explicitly import types then they should be namespaced with `@federation__` prefix (unless custom one is provided)

docs/source/federated-types/federated-directives.mdx
StephenBarlow: Cool, tweaked the language here

docs/source/federated-types/federated-directives.mdx
StephenBarlow: Ooh maybe in Fed 2? Last I knew in Fed 1 this did nothing but was valid syntax. I could be out of the loop though

StephenBarlow: Okay, looks like Fed 2 composition currently straight-up _rejects_ `@key` on interfaces despite being valid syntax, so I'll update this accordingly

docs/source/federated-types/federated-directives.mdx
StephenBarlow: I'd prefer to leave the `@link` section below importing and renaming, because those sections are critical setup for making use of any other directive on the page, whereas the `@link` section is more reference like the rest of the directives.

Yes agreed on the namespacing, I'll add that

docs/source/federated-types/federated-directives.mdx
StephenBarlow: Ok! Should I just cut this for now and we can add a corresponding sentence when update is complete?

dariuszkuc: lets change it to the expected `repeatable @key` test as otherwise pretty sure we'll forgot about it (`@key multi` is for checking keys that reference multiple fields). I can partially update the results.

StephenBarlow:
@dariuszkuc Pushed a followup commit that hopefully addresses remaining feedback! https://github.com/apollographql/federation/pull/2045/commits/d150fadf9e607e44e9df2c765d42de8f266d5ee9

It also folds in `@composeDirective`, which was added to the subgraph spec while this PR has been open.

Does this PR look okay in terms of "no regressions from live docs" at this point? If so, might be nice to get it out so any followup changes feel more manageable than this now-hefty PR

dariuszkuc (approved):
Awesome work!

PR #2046, opened 2022-08-08
Fix @composeDirective corner case

clenfest:
@composeDirective wasn't working when @link of the link directive was not present.

The reason this is happening is that although we are waiting to process core schemas for think that aren't the link schema, we aren't doing the same for other directives. This fix determines if the schema blueprint is a federation schema, and if so, will wait to process non @link directives until they've all been processed (and the @link to the link spec has been synthetically inserted if necessary).


benweatherman (approved):
🐐

PR #2065, opened 2022-08-12
Adding federation-spec entry for @composeDirective

clenfest:


docs/source/federation-spec.md
```diff
@@ -435,3 +436,24 @@ interface Product {
...
 }
 ```
+
+### `@composeDirective`
+```graphql
+directive @composeDirective(name: String!) repeatable on SCHEMA
+```
+
+The `@composeDirective` directive is used to indicate to composition that the custom directive specified should be preserved in the supergraph. In order for composition to pass, the following checks must be met:
+  - The directive must be part of a core schema
```
pcmanus, line 446: I'd maybe rephrase as "The directive must be defined within a core feature"? The small confusion I'm trying to avoid here is that, technically, every fed2 subgraph is a core schema, so the sentence as is could make it sound like that any custom directive definition in a fed2 subgraph is ok, while we're explicitly trying to say that no, you need to have the directive defined within a core "feature" (and yes, a "core feature" is kind of just a core schema you're linking to so this is all very confusing and we need to clean up and document our naming but ....).

pcmanus (approved):
Minor phrasing suggestion, but overall lgtm.

PR #2072, opened 2022-08-16
Update IDEA files to work with Webstorm

smyrick:
I usually do not check in my `.idea` directory as this is annoying to manage with everyone's personal settings and updating project directories, but if we are going to keep it we should at least include all the modules as modules


smyrick:
Closing my PR to clean up my open changes. I think think this is valid so if the team would like to reopen or branch from the original, please do

PR #2076, opened 2022-08-17
WIP draft on federation version changelog

StephenBarlow:


docs/source/federation-versions.mdx
```diff
@@ -0,0 +1,411 @@
+---
+title: Federation version changelog
+description: Understand changes between federation versions
```
zionts, line 3: I think it'd be really useful to identify the "upgrade flow" that we suggest that users follow and the semantics of which versions are compatible in which others for how long.

For example, we suggest that users upgrade their router version before upgrading their build pipeline version before updating their subgraph schemas / libraries. We guarantee that minor versions of router will continue to work with supergraph schemas that were built with build versions that are 2 minor below, etc.

Is that documented anywhere yet?

docs/source/federation-versions.mdx
```diff
@@ -0,0 +1,411 @@
...
+directive @link(url: String, as: String, for: link__Purpose, import: [link__Import]) repeatable on SCHEMA
+```
+
+</td>
+</tr>
+
+<tr>
+<td>
+
+##### `@external`, `@requires`, `@provides`, `@extends`
```
zionts, line 126: Should `@tag` be in this list?

docs/source/federation-versions.mdx
```diff
@@ -0,0 +1,411 @@
...
+* Each subgraph can apply any number of `@key` directives to an entity.
+* Subgraphs _must_ no longer apply the `@external` directive to their `@key` fields.
+
+</td>
+</tr>
+
+<tr>
+<td>
+
+Value types
```
zionts, line 170: Is this still a term in Fed 2? I remember there was a whole conversation around this and now I don't remember how it resolved 😅

StephenBarlow: The conversation resolved with it remaining a term! https://www.apollographql.com/docs/federation/federated-types/sharing-types

docs/source/federation-versions.mdx
StephenBarlow: Yuuup

docs/source/federation-versions.mdx
StephenBarlow: I agree and believe the answer is no, and in fact the "Moving to Fed 2" article (https://www.apollographql.com/docs/federation/federation-2/moving-to-federation-2) doesn't really touch on managed federation considerations at all. This will take some extra lift so I'll create a separate issue for it.

PR #2081, opened 2022-08-18
Add type definitions to schema extensions

smyrick:
TS Compiler error

image

Because the type `schema.extensions` is unknown the TS compiler is failing because it doesn't know for sure if `schema.extensions.apollo` is an object. If it is not, we can not perform a spread on it.

### DIFF
--- a/gateway-js/src/schema-helper/__tests__/addExtensions.test.ts
+++ b/gateway-js/src/schema-helper/__tests__/addExtensions.test.ts
@@ -30,4 +30,36 @@ describe('addExtensions', () => {
       }
     });
   });
+
+  it('works with undefined apollo block', async () => {
+    const schema = buildSchema('type Query { hello: String }');
+    expect(schema.extensions).toEqual({});
+    schema.extensions = {
+      apollo: undefined
+    };
+    expect(addExtensions(schema).extensions).toEqual({
+      apollo: {
+        gateway: {
+          version: version
+        }
+      }
+    });
+  });
+
+  it('works with undefined gateway block', async () => {
+    const schema = buildSchema('type Query { hello: String }');
+    expect(schema.extensions).toEqual({});
+    schema.extensions = {
+      apollo: {
+        gateway: undefined
+      }
+    };
+    expect(addExtensions(schema).extensions).toEqual({
+      apollo: {
+        gateway: {
+          version: version
+        }
+      }
+    });
+  });
 });
--- a/gateway-js/src/schema-helper/__tests__/addExtensions.test.ts
+++ b/gateway-js/src/schema-helper/__tests__/addExtensions.test.ts
@@ -14,20 +14,17 @@ describe('addExtensions', () => {
     const schema = buildSchema('type Query { hello: String }');
     expect(schema.extensions).toEqual({});
     schema.extensions = {
-      one: 1,
+      foo: 'bar',
       apollo: {
-        two: 2,
         gateway: {
-          three: 3
+          version: 'hello'
         }
       }
     };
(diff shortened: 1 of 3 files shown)


samuelAndalon (approved):

cpeacock:
@clenfest mind taking a quick look at this at some point?  seems fine to me but you are more knowledgeable

gateway-js/src/schema-helper/addExtensions.ts
```diff
@@ -1,13 +1,15 @@
...
 export function addExtensions(schema: GraphQLSchema): GraphQLSchema {
+  const apolloExtension = schema.extensions.apollo as any ?? {};
   schema.extensions = {
     ...schema.extensions,
     apollo: {
-      ...schema.extensions.apollo,
+      ...apolloExtension,
       gateway: {
-        ...schema.extensions.apollo?.gateway,
+        ...apolloExtension?.gateway,
```
clenfest, line 12: This is unsafe (`gateway` may be undefined and you can't spread undefined) but because you've set the type of `apolloExtension` to undefined, the type checker isn't helping you. I would change line 6 to be:

```
const apolloExtension = schema.extensions.apollo;
```

And then change this line to be:
```
...(apolloExtension?.gateway ?? {})
```

clenfest: Actually I have no idea why the type checker fails to catch because if I modify just line 6 I still don't see an error. Anyway, my advice on how to change it stands.

gateway-js/src/schema-helper/__tests__/addExtensions.test.ts
```diff
@@ -3,9 +3,31 @@ import { addExtensions } from '../addExtensions';
...
 
 describe('addExtensions', () => {
+
   it('adds gateway extensions to a schema', async () => {
     const schema = buildSchema('type Query { hello: String }');
     expect(schema.extensions).toEqual({});
     expect(addExtensions(schema).extensions).toEqual({ apollo: { gateway: { version: version } } });
   });
+
+  it('does not delete existing extensions', async () => {
```
clenfest, line 13: This test passes whether we're running with or without your fix. Did you want another test where no gateway was present?

smyrick: @clenfest I was trying to reproduce a failing test, but I can't set an invalid type to the `schema.extensions.apollo` without causing a TS compile error in our tests.

I am not sure why our TS compile step is not catching the current error in `main` because it is clearly an error if the type is `unknown`

gateway-js/src/schema-helper/addExtensions.ts
smyrick: I updated to extract all objects safely into variables

clenfest (approved):

trevor-scheer:
@smyrick @clenfest This is just a fly-by comment and it's possible I'm misunderstanding something here, but I would've expected something similar to what `subgraph-js` does by extending the extensions type as the outcome:
https://github.com/apollographql/federation/blob/4e3c161e5deec40a80931d78d23b38677cec6974/subgraph-js/src/schemaExtensions.ts#L13-L18

smyrick:
@trevor-scheer There is that type declared already:

https://github.com/apollographql/federation/blob/2946bdfcb13d429a86ab262481ee9e9e10ab4353/gateway-js/src/typings/graphql.ts#L37-L43

And I am using it here:

https://github.com/apollographql/federation/blob/2946bdfcb13d429a86ab262481ee9e9e10ab4353/gateway-js/src/schema-helper/addExtensions.ts#L2

Issue #2085, opened 2022-08-22
Significantly higher memory usage in gateway 2.0.5 and 2.1.0-alpha


lennyburdette:
Reported by a customer using a large Fed1 supergraph (shared in Slack).

```
Node.js heap total at gateway server start:
0.52.1 --> 126 MB
2.0.5 --> 1489 MB
2.1.0-alpha.4 --> 1575 MB
```

This is just on startup before taking any requests, so it's not caused by large query plans.

iuliiasobolevska:
Some more numbers from the same setup/investigation:
* node version: v14.20.0
* starting express app with ApolloServer and ApolloGateway pulled from diff @apollo/gateway package versions
* measuring consumed memory only on startup (no traffic being served, query planner not involved) 
* fed1-variant and fed2-variant have the same schema, the only diff is the 1st being composed using Federation 1, 2nd - Federation 2

| @apollo/gateway version | @apollo/gateway major version | supergraphSdl variant | Federation Version | Heap total (MB) | Heap used (MB) | RSS (MB) | 
| 0.51.0 | 0.5 | fed1-variant | 1 | 114 | 81 | 160 | 
| 0.52.0 | 0.5 | fed1-variant | 1 | 113 | 81 | 160 |
| 0.52.1 | 0.5 | fed1-variant | 1 | 113 | 81 | 160 |
| 2.0.5 | 2 | fed1-variant | 1 | 1391 | 1367 | 1480 | 
| 2.0.5 | 2 | fed2-variant | 2 | 211 | 189 | 279 | 
| 2.0.5 (2nd run) | 2 | fed2-variant | 2 | 211 | 189 | 275 | 
| 2.1.0-alpha.1 | 2 | fed1-variant | 1 | 1525 | 1496 | 1622 |
| 2.1.0-alpha.1 | 2 | fed2-variant | 2 | 218 | 185 | 290 |
| 2.1.0-alpha.2 | 2 | fed1-variant | 1 | 1481 | 1457 | 1578 |
| 2.1.0-alpha.2 | 2 | fed2-variant | 2 | 219 | 186 | 291 |
| 2.1.0-alpha.3 | 2 | fed1-variant | 1 | 1490 | 1461 | 1588 |
| 2.1.0-alpha.3 | 2 | fed2-variant | 2 | 222 | 189 | 288 |
| 2.1.0-alpha.4 | 2 | fed1-variant | 1 | 1511 | 1482 | 1612 |
[...]

PR #2089, opened 2022-08-24
Fix high memory usage when extracting subgraphs for some fed1 supergraphs

pcmanus:
As mentioned on #2085, some fed1 supergraphs can lead to very high memory usage, much larger than if the same subgraphs are recomposed with fed2. And the underlying reason is that fed1 supergraph have not `join__type` information for value types, so the code extracting subgraphs from supergraphs used to put value types into all extracted subgraphs pessimistically on account that while 1) it was simple and 2) it was harmless from a correctness POV since those were just types that ended up being unreachable.

However, in the case where there is a lot of subgraph, and most of those subgraphs define a decent number of value types (typically each subgraph defining it's own set of (non-shared) value types), then it means that every extracted subgraph had all the value types from all subgraphs. And the sheer amount of data each subgraph ended up having led to that high memory usage.

So the main fix of this PR consists in, for fed1 supergraph, doing an additional pass when extracting the subgraphs to compute which types are actually _reachable_ in each subgraph and make sure we then don't add the unreachable types. Doing so ensure the extracted subgraphs are roughly the same as in fed2 and thus the memory consumptions are comparable.

Additionally, this PR contains a few (non-fed1-supergraph-specific) small optimisations for memory that comes from investigating this issue. Those optimisations leads to measurable memory differences on at least the example considered on this issue. The short version of those optimisations are:
1. as we validate schema, we convert them into graphQL-js asts (`DocumentNode`) to be able to re-use graphQL-js validations and we used to cache the resulting ASTs. However, in hindsight, we don't really make use of this caching (in theory some code could make use of it, but I believe that's not the case currently). So not caching those avoid retaining that memory.
2. there were a number of places in our schema abstractions where were allocating empty array, sets or maps that were never populated more often than not. For instance, most graphQL elements can have directive applied to them, but most elements of most schema don't get any. So for large schema, simply avoiding those empty object allocation adds up somewhat.
3. more of a cleanup than anything else, but while looking at heap profiling, I got confusing by some schema object that were coming from the `core-schema-js` and turns out that it's because that library has a `Schema` class too and was allocating some in globals. To be fair, those weren't using much memory, but we don't really use `core-schema-js` at all outside of a super simple error class that is easily replaced, so I took the opportunity to remove the dependency while at it.


docs/source/errors.md
```diff
@@ -91,6 +91,7 @@ The following errors might be raised during composition:
 | `UNKNOWN_FEDERATION_LINK_VERSION` | The version of federation in a @link directive on the schema is unknown. | 2.0.0 |  |
 | `UNKNOWN_LINK_VERSION` | The version of @link set on the schema is unknown. | 2.1.0 |  |
 | `UNSUPPORTED_FEATURE` | Indicates an error due to feature currently unsupported by federation. | 2.1.0 |  |
+| `UNSUPPORTED_LINKED_FEATURE` | Indicates that a feature used in a @link is either unsupported and is used with unsupported options. | 2.0.0 |  |
```
StephenBarlow, line 94:
```suggestion
| `UNSUPPORTED_LINKED_FEATURE` | Indicates that a feature used in a `@link` is either unsupported or is used with unsupported options. | 2.0.0 |  |
```

internals-js/src/definitions.ts
```diff
@@ -2589,7 +2608,7 @@ export class FieldDefinition<TParent extends CompositeType> extends NamedSchemaE
   }
 
   toString(): string {
-    const args = this._args.size == 0
+    const args = !this.hasArguments()
```
clenfest, line 2611: rearrange so that having arguments case is first? i.e. 
```typescript
const args = this.hasArguments() ? ... : ...;
```

internals-js/src/extractSubgraphsFromSupergraph.ts
```diff
@@ -82,6 +83,115 @@ class SubgraphExtractionError {
...
+    const reachableTypes = new Set<string>();
+    collectFieldReachableTypesForSubgraph(
+      supergraph,
+      subgraphName,
+      (t) => reachableTypes.add(t.name),
+      fieldInfoInSubgraph,
+      typeInfoInSubgraph,
+    );
+    reachableTypesBySubgraphs.set(subgraphName, reachableTypes);
+    console.log(`For ${subgraphName}, reachableTypes = [${[...reachableTypes]}]`);
```
clenfest, line 169: leftover console.log

clenfest (approved):
I'm a little concerned about moving from Sets to arrays across the board, but if you're comfortable with it, it's ok with me. I left a few minor comments, but feel free to merge once you've cleaned those up.

internals-js/src/definitions.ts
```diff
@@ -2271,15 +2290,12 @@ export class EnumType extends BaseNamedType<OutputTypeReferencer, EnumType> {
   }
 
   private removeValueInternal(value: EnumValue) {
```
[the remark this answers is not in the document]
pcmanus: While that method is private, it's actually called from another class, side-stepping the privateness through `EnumType.prototype['removeValueInternal'].call()`. Which is admittedly a bit of a hack but the goal is to _not_ expose it publicly yet still allow calling it from `EnumValue`. Bit of a poor-man [`friend`](https://en.cppreference.com/w/cpp/language/friend) emulation.

In any case, in that situation, this method is just meant to abstract direct accesses to the `EnumType._values` field so only `EnumType` has to care what the type of that field. Granted, not extremely important in this situation, and possibly more of a personal preference in that case, but I do prefer it over the alternative (of inline that method, which would require to have something like `(this._parent as any)._values` in `EnumValue` in particular).

pcmanus:
> I'm a little concerned about moving from Sets to arrays across the board

It's a fair concern, and I should have expanded sooner on my reasoning.

First, I'm not suggesting we stop using `Set` altogether. There is still uses of them in the codebase in fact. It's just that "for relatively small collections", they have a non completely negligeable memory overhead over arrays without necessarilly a performance edge.

In the case of `Schema`, given how it's used, I think we need to watch for memory consumption a bit, maybe even trade a bit of pure performance for it, because the gateway/router reads/builds schemas only at startup, but them keep them in memory for all execution (and I'm obviously not saying startup time doesn't matter at all, just that it's not the hottest path performance wise either).

Anyway, the changes from set to array in the PR are for:
1. `_extensions`: I can't imagine an element ever having more than a handful of extensions in practice so those are almost guaranteed to be tiny. Using arrays for those is frankly probaly ever better perfomance wise (not that it matters).
2. `_referencers`: those are also probably pretty small on average, though some element can admittedly have a large-ish amount of referencers, but never more than there is schema elements, so we're not talking tens of millions either. We also _weren't_ exposing those externally like a set, so the only thing sets were saving were the inclusion check on insertion. Overall, it might make the building of some extremly large schema a bit slower, which again only impact "startup", but in exchange probably mostsly just save some memory with no other impact in all other cases. I absolutely could be wrong, but I suspect it's a good tradeoff overall (hence the chance).

PR #2090, opened 2022-08-24
Merge descriptions for executable directives

clenfest:
Not sure why, but this was a regression from 2.0.0


pcmanus (changes requested):
Does look like an oversight.

Implementation wise however, I think we should just reuse the existing `mergeDescriptions` method that is used for (almost) everything else (both for consistency and because it's a bit more capable in some ways). Unless there is a good reason not too that is, but I don't see one a priori.

There is also an existing `preserves descriptions` unit test in `compose.test.ts` so it would be nice to add an executable directive definition within that test to test that new code.

composition-js/src/__tests__/compose.test.ts
```diff
@@ -221,6 +221,43 @@ describe('composition', () => {
...
+          A
+          B
+        }
+      `
+    }
+
+    const result = composeAsFed2Subgraphs([subgraph1, subgraph2]);
+    assertCompositionSuccess(result);
+
+    expect(result.supergraphSdl).toMatchSnapshot();
```
pcmanus, line 258: Not necessarily a blocker as I suppose it's subjective, but not a huge fan of external snapshots: it makes it unnecessarily hard to follow what the test does imo.

But also, I'd really prefer avoiding matching the whole supergraph SDL as done here since this test is not about the overall supergraph underlying format. Matching everything means that unrelated changes (a bump of some spec version, a new directive definition added to the supergraph, etc...) will require modifying this snapshot.

Here we can easily match the API schema instead (as done in the previous test).

composition-js/src/merging/merge.ts
```diff
@@ -65,6 +65,7 @@ import {
   filterTypesOfKind,
   isNonNullType,
   isExecutableDirectiveLocation,
+  isDefined,
```
pcmanus, line 68: This is what bothers CI at the moment.

pcmanus (approved):
Left comment regarding the unit test that I'd be happy if addressed, but change looks otherwise good to me.

PR #2093, opened 2022-08-25
Fix fragment expansion nuking @defer and other minor @defer updates

pcmanus:
The patch from #1911 has a small bug in that it tries to avoid adding useless type conditions (inline fragments) when it expands named fragments, but it forgets to check if there is directives first and so "nuke" any such directives if there is one. Effectively, this means that if there is a `@defer` applied to a fragment spread, it gets lost and ignore.

Unfortunately, the @defer unit tests were all using inline fragments, so added a test to fix that coverage and the one-liner fix for the actual issue.

Took the opportunity to include a few minor tweaks related `@defer` following @glasser comments on #1958 (mostly adding a missing unit test and making sure we match recent spec updates, but no functional changes otherwise).

### DIFF
--- a/internals-js/src/definitions.ts
+++ b/internals-js/src/definitions.ts
@@ -1128,15 +1128,16 @@ const graphQLBuiltInDirectivesSpecifications: readonly DirectiveSpecification[] 
     locations: [DirectiveLocation.SCALAR],
     argumentFct: (schema) => ({ args: [{ name: 'url', type: new NonNullType(schema.stringType()) }], errors: [] })
   }),
-  // TODO: currently inconditionally adding @defer as the list of built-in. It's probably fine, but double check if we want to not do so when @defer-support is
-  // not enabled or something (it would probably be hard to handle it at that point anyway but well...).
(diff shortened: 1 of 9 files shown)


internals-js/src/definitions.ts
```diff
@@ -1128,15 +1128,16 @@ const graphQLBuiltInDirectivesSpecifications: readonly DirectiveSpecification[]
     locations: [DirectiveLocation.SCALAR],
     argumentFct: (schema) => ({ args: [{ name: 'url', type: new NonNullType(schema.stringType()) }], errors: [] })
   }),
-  // TODO: currently inconditionally adding @defer as the list of built-in. It's probably fine, but double check if we want to not do so when @defer-support is
-  // not enabled or something (it would probably be hard to handle it at that point anyway but well...).
+  // Note that @defer and @stream a inconditionally added to `Schema` even if they are technically "optional" built-in. _But_,
```
clenfest, line 1131:
```suggestion
  // Note that @defer and @stream are unconditionally added to `Schema` even if they are technically "optional" built-in. _But_,
```

clenfest (approved):

PR #2114, opened 2022-08-30
add directive definitions to composeDirective example

clenfest:


docs/source/federated-types/federated-directives.mdx
```diff
@@ -577,6 +577,9 @@ extend schema
     # highlight-start
     @composeDirective(name: "@myDirective")
     @composeDirective(name: "@hello")
+
+directive @myDirective(a: String!) on FIELD_DEFINITION
```
StephenBarlow, line 581: I can't tell for sure with the diff but could we make sure all these lines are indented consistently?

StephenBarlow (approved):
One small thing!

PR #2116, opened 2022-08-30
Fix node-fetch runtime build-time regression

### LINKED ISSUE #2113: `@types/node-fetch` is missing in the `@apollo/gateway` dependencies
Since the `2.1.0` version, the `@types/node-fetch` dependency is missing.

Regression introduced in https://github.com/apollographql/federation/pull/1970/files

This error is thrown when we try to build our project:
```
#21 12.69 ../../.yarn/__virtual__/@apollo-gateway-virtual-9de6cdc87b/0/cache/@apollo-gateway-npm-2.1.0-0009d7c27c-a763b86c26.zip/node_modules/@apollo/gateway/dist/config.d.ts:2:34 - error TS7016: Could not find a declaration file for module 'node-fetch'. '/usr/src/app/.yarn/__virtual__/node-fetch-virtual-1391f376ba/0/cache/node-fetch-npm-2.6.7-777aa2a6df-8d816ffd1e.zip/node_modules/node-fetch/lib/index.js' implicitly has an 'any' type.
#21 12.69   Try `npm i --save-dev @types/node-fetch` if it exists or add a new declaration (.d.ts) file containing `declare module 'node-fetch';`
#21 12.69 
#21 12.69 2 import type { HeadersInit } from 'node-fetch';
```

clenfest:
#1970 seems to have caused a regression based on `@types/node-fetch` not being in the gateway's runtime dependencies

Fixes #2113


gateway-js/package.json
```diff
@@ -35,6 +35,7 @@
     "@apollo/utils.logger": "^1.0.0",
     "@josephg/resolvable": "^1.0.1",
     "@opentelemetry/api": "^1.0.1",
+    "@types/node-fetch": "2.6.2",
```
Grmiade, line 38: Why not `^2.6.2`?

Grmiade (commented):
Thanks for the PR, just one question 🙂

trevor-scheer (approved):
I'd clarify in the title that this is only a build-time regression (i.e. non-TS users should be unaffected). The types package is needed in `dependencies` since dev dependencies _aren't_ installed into `node_modules` but the types must be present for the gateway package to build due to the `node-fetch` import in the `.d.ts` file.

edit: compilation is part of the install lifecycle so probably also affecting non-TS users but still not at runtime 😄

gateway-js/package.json
clenfest: Because I copied it from the top level package.json and didn't notice. Thanks for the catch.

PR #2120, opened 2022-08-31
Allow fields with arguments in @requires

pcmanus:


composition-js/src/merging/merge.ts
```diff
@@ -2092,6 +2096,127 @@ class Merger {
...
+
+    // We need to redo some validation for @requires after merge. The reason is that each subgraph validates that its own
+    // @requires are valid, but "requirements" are requested from _other_ subgraphs (by definition of @requires really),
+    // and there is a few situations (see details below) where a validity within the originated subgraph does not entail
+    // validity for all subgraph that would have to provide those "requirements".
+    // Long story short, we need to re-validate every @requires against the supergraph to guarantee it will always work
+    // at runtime.
+    for (const subgraph of this.subgraphs) {
+      for (const requiresApplication of subgraph.metadata().requiresDirective().applications()) {
+        const originalField = requiresApplication.parent as FieldDefinition<CompositeType>;
```
clenfest, line 2108: Do you think it makes sense to assert on `originalField.kind` when we want to cast like this? I know that we do it throughout the code, but I'm wondering if we want extra sanity.

clenfest (approved):

composition-js/src/merging/merge.ts
pcmanus: Yeah, can't hurt. Added one.

PR #2125, opened 2022-09-02
Fix method removing empty branches from selection sets

### LINKED ISSUE #2123: Empty selection set produced with `@defer`'d query
It seems that even with an otherwise valid query plan using `@defer`, the `prettyFormatQueryPlan` function will throw under particular (unknown, as of yet) circumstances.

This can be reproduced with [this schema](https://raw.githubusercontent.com/apollographql/supergraph-demo-fed2/main/supergraph.graphql) and the following operation on the latest 2.1.1 release:

```graphql
fragment MyFragment on DeliveryEstimates {
    estimatedDelivery
    fastestDelivery
}
query Dimensions {
  allProducts {
    delivery {
      ...MyFragment @defer
    }
    sku
    id
  }
}
```

pcmanus:
As the code for `@defer` build the selection set used for "primary parts",
it may start adding "branches" (selections) that end up being empty
because everything within that branch is actually deferred (and thus
not part of this primary part). So once we've finished building
everything, we call a `withoutEmptyBranches` method on the selection set
that can have such (invalid for graphQL) empty branches to remove
them.

Unfortunately, the logic for that method was broken: it was removing
empty branches in simple cases, but the recursion was done the wrong
way so that if removing an empty "leaf" left its parent empty, this
wasn't removed properly. In other words, given
  `{ a b { c {} } }`
the method was returning:
  `{ a b { } }`
instead of the proper:
  `{ a }`

This commit fixes the logic of that method to behave correctly.

Fixes #2123

### DIFF
--- a/internals-js/src/operations.ts
+++ b/internals-js/src/operations.ts
@@ -1200,9 +1200,6 @@ export class SelectionSet extends Freezable<SelectionSet> {
     validate(!this.isEmpty(), () => `Invalid empty selection set`);
     for (const selection of this.selections()) {
       selection.validate();
-      const selectionFragments = selection.namedFragments();
-      // We make this an assertion because this is a programming error. But validate is a convenient place for this in practice.
-      assert(!selectionFragments || selectionFragments === this.fragments, () => `Selection fragments (${selectionFragments}) for ${selection} does not match selection set one (${this.fragments})`);
     }
   }
 
@@ -1435,17 +1432,15 @@ export class FieldSelection extends Freezable<FieldSelection> {
   }
 
   filter(predicate: (selection: Selection) => boolean): FieldSelection | undefined {
-    if (!predicate(this)) {
-      return undefined;
-    }
     if (!this.selectionSet) {
-      return this;
+      return predicate(this) ? this : undefined;
     }
 
     const updatedSelectionSet = this.selectionSet.filter(predicate);
-    return this.selectionSet === updatedSelectionSet
+    const thisWithFilteredSelectionSet = this.selectionSet === updatedSelectionSet
       ? this
       : new FieldSelection(this.field, updatedSelectionSet);
+    return predicate(thisWithFilteredSelectionSet) ? thisWithFilteredSelectionSet : undefined;
   }
 
   protected freezeInternals(): void {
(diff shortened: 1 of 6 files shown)


internals-js/src/operations.ts
```diff
@@ -1932,9 +1926,13 @@ class FragmentSpreadSelection extends FragmentSelection {
 export function operationFromDocument(
   schema: Schema,
   document: DocumentNode,
-  operationName?: string,
+  options?: {
```
pcmanus, line 1929: Note for reviewer: I didn't wanted to add the new `validate` as a nameless argument because that was starting to get ugly, especially when the `operationName` was already optional. But at the same time, `operationFromDocument` is called in a ton of places (particularly tests), almost always with just the 2 first arguments, so switching to _all_ argument being name-based was pretty annoying and would have generated a big diff. Hence why I opted for this compromised of keeping the 2 first main argument nameless and putting everything else in an "options" object.

internals-js/src/operations.ts
```diff
@@ -1435,17 +1432,15 @@ export class FieldSelection extends Freezable<FieldSelection> {
...
-      return undefined;
-    }
     if (!this.selectionSet) {
-      return this;
+      return predicate(this) ? this : undefined;
     }
 
     const updatedSelectionSet = this.selectionSet.filter(predicate);
-    return this.selectionSet === updatedSelectionSet
+    const thisWithFilteredSelectionSet = this.selectionSet === updatedSelectionSet
```
clenfest, line 1440: Are you sure you want === here? It seems a little strange to me that we'd do that type of equality test immediately after filtering, I sould expect some sort of deep equality check, but I'm not looking into it too deeply.

internals-js/src/operations.ts
```diff
@@ -1627,15 +1622,14 @@ export abstract class FragmentSelection extends Freezable<FragmentSelection> {
...
 
   filter(predicate: (selection: Selection) => boolean): FragmentSelection | undefined {
-    if (!predicate(this)) {
-      return undefined;
-    }
     // Note that we essentially expand all fragments as part of this.
     const selectionSet = this.selectionSet;
     const updatedSelectionSet = selectionSet.filter(predicate);
-    return updatedSelectionSet === selectionSet
+    const thisWithFilteredSelectionSet = updatedSelectionSet === selectionSet
```
clenfest, line 1628: Same question here as above.

internals-js/src/operations.ts
pcmanus: Yeah, this code is just an optimisation. The point is to save allocating a new `FieldSelection` object (or, in the case of `SelectionSet.filter`, a `SelectionSet` object) in the case where nothing was filtered. Which is expected to be common, both because we use `filter` in case where we know filtering rarely kicks in, like in that patch with empty branches which happens but not at all in all cases, but also because the method is recursive and while filtering may happen in some branches of a selection set, it probably still won't happen in most of them).

Anyway, using deep equality doesn't make sense in term of such optimisation since we'd lose more in computing the deep equality check in general than what we'd save in allocations.

clenfest (approved):

PR #2136, opened 2022-09-09
Fix repeatable custom directives

### LINKED ISSUE #2134: Incorrect hint (and composition behavior?) for inconsistent repeatable directives
**Description:** When two subgraphs apply a directive to the same type or shareable field with different arguments, *and the directive is repeatable,* the hint says that it's not repeatable and only one application is emitted. 
**Reproduction:** https://stackblitz.com/edit/basic-federation-dpcqjk?file=test.js
**Version:** `@apollo/composition@2.1.1`
**Actual behavior:**

```graphql
# subgraph one
extend schema @composeDirective(name: "@auth")
directive @auth(scope: String!) repeatable on FIELD_DEFINITION
type Query {
  shared: String @shareable @auth(scope: "VIEWER")
}
```
```graphql
# subgraph 2
extend schema @composeDirective(name: "@auth")
directive @auth(scope: String!) repeatable on FIELD_DEFINITION
type Query {
  shared: String @shareable @auth(scope: "ADMIN")
}
```

```js
[
  CompositionHint {
    definition: {
      code: 'INCONSISTEN_NON_REPEATABLE_DIRECTIVE_ARGUMENTS',
      level: [Object],
      description: 'A non-repeatable directive is applied to a schema element in different subgraphs but with arguments that are different.'
    },
    message: 'Non-repeatable directive @auth is applied to "Query.shared" in multiple subgraphs but with incompatible arguments. The supergraph will use arguments {scope: "VIEWER"} (from subgraph "one"), but found arguments {scope: "ADMIN"} in subgraph "two".',
    nodes: [ /* ... */ ]
  }
]
```

**Expected behavior:** 
* The hint shouldn't call the directive "non-repeatable".
* I think composition should include both directive applications in the supergraph, but I'm not sure about that.

clenfest:
Getting a false positive hint when a repeatable custom directive is composed saying that it's not repeatable. This happens because we addDirectivesShallow, but the directive isn't fully created when this check is run. Solved by switching the merge order between building directives and types.

Fixes #2134


pcmanus (changes requested):
I don't think that this quite solve the problem in all cases.

The underlying issue here is that `mergeAppliedDirective` relies on directive definition having been merged, so we need to make sure it's never called before `mergeDirectiveDefinition`.

Moving the merging of type definitions after that of directive definitions does fixes it for directive applications "within type definitions", but that's not the only place where directive applications can appear.

In particular, we can have directive applications both in "schema definition" and on the arguments or directive definition themselves. For the former, moving the merging of directive definitions one more step up (before the call the `mergeSchemaDefinition`) would probably fix it, but the later problem is a bit more tricky.

So my suggestion for solving this would be to split the merging of directive applications separately (for all elements: schema definition, types and directive definitions), and have that done last during merging. It's a bit more code change (though hardly a big change), but it feels like this is easier to reason about and not have break later.

As an aside, I'll note that:
1. the added test does not pass because the directive in the test is not actually marked repeatable (I assume due to copy-pasta from the github issue).
2. there is a change to (the auto-generated) `hints.md` related to fixing the typo to `INCONSISTEN_NON_REPEATABLE_DIRECTIVE_ARGUMENTS` (missing `T`) but the corresponding code change is not in the commit.

Nit: may I suggest taking the habit of checking CI before marking issues ready for review?

clenfest:
> Nit: may I suggest taking the habit of checking CI before marking issues ready for review?

Sorry about that, for some reason I pushed without saving the files.

composition-js/src/__tests__/compose.test.ts
```diff
@@ -1,55 +1,26 @@
...
   printType,
-  Schema,
-  ServiceDefinition,
-  Subgraphs
 } from '@apollo/federation-internals';
-import { CompositionResult, composeServices, CompositionSuccess } from '../compose';
+import { CompositionResult, composeServices } from '../compose';
 import gql from 'graphql-tag';
 import './matchers';
 import { print } from 'graphql';
```
clenfest, line 16: Note to reviewers: I noticed that `compose.test.ts` tests were getting run multiple times because the helper functions were being imported from other files, so I split those out into a separate file.

gateway-js/src/__tests__/gateway/lifecycle-hooks.test.ts
```diff
@@ -147,7 +147,7 @@ describe('lifecycle hooks', () => {
     // the supergraph (even just formatting differences), this ID will change
     // and this test will have to updated.
     expect(secondCall[0]!.compositionId).toEqual(
-      'cc95112b64179c4e549de788b051f44010a02877f568649a42caeeee6a135601',
+      'a7e83d2e958dc8e0f08326443113802c900a4d977a26df6cfda599f138d9bb2f',
```
clenfest: Note to reviewers: This is just a reordering of applied directives in the SDL (see the diff in the .snap file for a similar case).

composition-js/src/merging/merge.ts
```diff
@@ -1906,11 +1914,33 @@ class Merger {
     return source.locations.filter(loc => isExecutableDirectiveLocation(loc));
   }
 
+  // In general, we want to merge applied directives after merging elements, the one exception
+  // is @inaccessible, which is necessary to exist in the supergraph for EnumValues to properly
+  // determine whether the fact that a value is both input / output will matter
```
pcmanus, line 1919: Nit: the naming of that method, especially along side the almost similar `mergeAllAppliedDirectives`, gets maybe a little bit uninformative on first approximation. It's all private so not a huge deal, but maybe this one can be renamed to something longer but more informative (say `recordAppliedDirectivesToMerge`? of course, it does merge `@inaccessible` in practice, but it's more a special case so maybe ok if not reflected in the name).

pcmanus (approved):

PR #2137, opened 2022-09-12
Fix potential inefficient planning due to __typename

pcmanus:
This PR fixes a query planning performance issue discovered with some user data. Namely, planning for some particular queries was taking a very noticeable amount of time (~30 sec on my somewhat fancy laptop for reference).

Upon inspection, it was due to some large amount of plans being built and compared, but the only reason there were many plans considered was due to the consideration of multiple options for the `__typename` field in a number of places. Obviously, since `__typename` can be requested equally from any subgraph (as long as it knows the type), evaluating all those options was particularly wasteful.

The attaching commit avoids that problem by essentially making the query planning "ignore" `__typename` as it computes all the possible choices/plans to consider, and then add it back where it should at the very end when finalising the plan.

On the particular example mentioned above, this reduces the number of plans to consider/evaluate by the query planner to just one, and so query planner is reduced to some ~300ms (again, on my laptop, and it's a fairly large query/plan).

Testing this, I randomly run into a (very) minor bug with aliasing `__typename`, in that if one was to query `__typename` twice _in the same selection_ by using an alias, the aliased version was discarded. This PR also opportunistically fixes this. Tbc, I can't think of a reason why anyone would request such a thing but it's valid graphQL and was definitively not handled properly so...

### DIFF
--- a/internals-js/src/operations.ts
+++ b/internals-js/src/operations.ts
@@ -62,6 +62,8 @@ function haveSameDirectives<TElement extends OperationElement>(op1: TElement, op
 }
 
 abstract class AbstractOperationElement<T extends AbstractOperationElement<T>> extends DirectiveTargetElement<T> {
+  private attachements?: Map<string, string>;
+
   constructor(
     schema: Schema,
     private readonly variablesInElement: Variables
@@ -74,6 +76,25 @@ abstract class AbstractOperationElement<T extends AbstractOperationElement<T>> e
   }
 
   abstract updateForAddingTo(selection: SelectionSet): T;
+
+  addAttachement(key: string, value: string) {
+    if (!this.attachements) {
+      this.attachements = new Map();
+    }
+    this.attachements.set(key, value);
+  }
+
+  getAttachement(key: string): string | undefined {
+    return this.attachements?.get(key);
+  }
+
+  protected copyAttachementsTo(elt: AbstractOperationElement<any>) {
+    if (this.attachements) {
+      for (const [k, v] of this.attachements.entries()) {
+        elt.addAttachement(k, v);
+      }
+    }
+  }
 }
 
 export class Field<TArgs extends {[key: string]: any} = {[key: string]: any}> extends AbstractOperationElement<Field<TArgs>> {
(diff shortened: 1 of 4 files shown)


internals-js/src/operations.ts
```diff
@@ -86,7 +107,6 @@ export class Field<TArgs extends {[key: string]: any} = {[key: string]: any}> ex
     readonly alias?: string
   ) {
     super(definition.schema(), variablesInArguments(args));
-    this.validate();
```
pcmanus, line 89: Note to reviewers: somewhat unrelated minor optimisation. While starting to look at this issue, I fired up profiling to see if anything specific showed up. While nothing egregious showed up, this validation came up higher than it should have, and upon reflexion, doing that validation in the constructor was a bit heavy. And since we have a `validate()` method on `SelectionSet` that we call when needed (but less aggressively), simply made it so this `Field.validate` was called only at the same time (outside of performance, this even feel a bit more expected anyway).

internals-js/src/operations.ts
```diff
@@ -1261,7 +1287,11 @@ export class SelectionSet extends Freezable<SelectionSet> {
   clone(): SelectionSet {
     const cloned = new SelectionSet(this.parentType);
     for (const selection of this.selections()) {
-      cloned.add(selection.clone());
+      const clonedSelection = selection.clone();
```
pcmanus, line 1290: Notes to reviewer: another minor stuff that profiling showed up. I could probably have put those into a separate commit, but I was lazy. Sorry.

query-planner-js/src/buildPlan.ts
```diff
@@ -983,22 +988,22 @@ class FetchGroup {
...
         )
       : operationForQueryFetch(
           this.rootKind,
           this.selection,
           variableDefinitions,
-          fragments,
           operationName,
         );
 
+    operation = operation.optimize(fragments);
```
pcmanus, line 1005: Note to reviewers: a very minor optimistic refactor. The `optimize` method was called in each of `operationForEntitiesFetch` and `operationForQueryFetch` which was both less DRY, but also less readable in the sense that we always want to call that method, and it has no reason to depend on the kind of operation we deal with.

clenfest (approved):

PR #2140, opened 2022-09-13
Fix bug in patch for #2137

pcmanus:
While doing some small late change to the patch for #2137, I introducing an issue that made the optimisation of the patch not kick in many case (or more precisely, the optimisation was applied, but then mistakenly thrown by a mistake during the recursion). Unfortunately, I was testing this optimisation manually and apparently forgot to re-run the test on the last patch version and so this was committed with the issue.

This PR fixes the issue, but also adds a new automatic test to ensure we catch this kind of unintentional regression.

### DIFF
--- a/query-planner-js/src/index.ts
+++ b/query-planner-js/src/index.ts
@@ -6,7 +6,7 @@ import { QueryPlan } from './QueryPlan';
 
 import { Schema, Operation, Concrete } from '@apollo/federation-internals';
 import { buildFederatedQueryGraph, QueryGraph } from "@apollo/query-graphs";
-import { computeQueryPlan } from './buildPlan';
+import { computeQueryPlan, PlanningStatistics } from './buildPlan';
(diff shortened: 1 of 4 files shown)


query-planner-js/src/index.ts
```diff
@@ -28,11 +29,17 @@ export class QueryPlanner {
       return { kind: 'QueryPlan' };
     }
 
-    return computeQueryPlan({
+    const {plan, statistics} = computeQueryPlan({
```
pcmanus, line 32: Note for reviewers: I initially considered to keep returning the result of `computeQueryPlan` from `buildQueryPlan`, which would be admittedly slightly cleaner than having a separate `lastGeneratedPlanStatistics` method. But that `buildQueryPlan` is used in a bunch of places, including the router bridge I believe, and I was worried about create some headache (especially since this fix is meant for a patch release).

clenfest (approved):
LGTM

PR #2142, opened 2022-09-15
rename federation-spec page to subgraph-spec

dariuszkuc:
Currently subgraph specification is under https://www.apollographql.com/docs/federation/federation-spec/ which can be confusing to the end users. Since this is actually a subgraph spec we should fix the URL to make it more explicit.


hwillson (approved):
I'm 👍 with this - I'd love to hear @StephenBarlow's thoughts as well. Thanks!

StephenBarlow (approved):
Looks good to me! Thanks for adding the all-important redirect 😄

PR #2143, opened 2022-09-16
Never merge fragments that have `@defer`

pcmanus:
For efficiency sake, the code merges inline fragments that are "the same" in queries. So something like:
```graphql
{
  t {
    ... on X {
      a
    }
    ... on X {
      a
      b
    }
  }
}
```
is simplified into:
```graphql
{
  t {
    ... on X {
      a
      b
    }
  }
}
```
Which is actually somewhat important because 1) while such query (the first one) is rarely written directly, it can easily happen when you start expanding named fragments, and 2) not doing this could create serious inefficiencies in more complex examples as it could create unecessary fetches.

Anyway, when directives gets involved on fragments, the code still merge fragments but only if the directive applications are the same, which means same name and arguments. So:
```graphql
query foo($if: Boolean) {
  t {
    ... on X @include(if: $if) {
      a
    }
    ... on X @include(if: $if) {
      a
      b
    }
  }
}
```
is simplified into:
```graphql
query foo($if: Boolean) {
  t {
    ... on X @include(if: $if) {
      a
      b
    }
  }
}
```
but:
```graphql
query foo($if1: Boolean, $if2: Boolean) {
  t {
    ... on X @include(if: $if1) {
      a
    }
    ... on X @include(if: $if2) {
      a
      b
    }
  }
}
```
is left as-is.

But those rules meant that:
```graphql
{
  t {
    ... on X @defer {
      a
    }
    ... on X @defer {
      a
      b
    }
  }
}
```
_was_ simplified by merging the 2 branches.

Except that this is kind of wrong for `@defer` as each individual is expected to be it's own deferred chunk.

This patch special cases `@defer` so we don't such merging for it.

### DIFF
--- a/internals-js/src/definitions.ts
+++ b/internals-js/src/definitions.ts
@@ -3197,20 +3197,40 @@ export class Directive<
   }
 }
 
-export function sameDirectiveApplication(application1: Directive<any, any>, application2: Directive<any, any>): boolean {
-  return application1.name === application2.name && argumentsEquals(application1.arguments(), application2.arguments());
+/**
+ * Checks if 2 directive applications should be considered equal.
+ *
+ * By default, 2 directive applications are considered equal if they are for the same directive and are passed the same values to
+ * the same arguments. However, some special directive can be excluded so that no 2 applications are ever consider equal. By default,
+ * this is the case of @defer, as never want to merge @defer applications so that each create its own "deferred block".
+ */
+export function sameDirectiveApplication(
+  application1: Directive<any, any>,
+  application2: Directive<any, any>,
+  directivesNeverEqualToThemselves: string[] = [ 'defer' ],
+): boolean {
+  // Note: we check name equality first because this method is most often called with directive that are simply not the same
+  // name and this ensure we exit cheaply more often than not.
+  return application1.name === application2.name 
+    && !directivesNeverEqualToThemselves.includes(application1.name)
+    && !directivesNeverEqualToThemselves.includes(application2.name)
+    && argumentsEquals(application1.arguments(), application2.arguments());
 }
 
 /**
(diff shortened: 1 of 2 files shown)


internals-js/src/definitions.ts
```diff
@@ -3197,20 +3197,40 @@ export class Directive<
...
+ * this is the case of @defer, as never want to merge @defer applications so that each create its own "deferred block".
+ */
+export function sameDirectiveApplication(
+  application1: Directive<any, any>,
+  application2: Directive<any, any>,
+  directivesNeverEqualToThemselves: string[] = [ 'defer' ],
+): boolean {
+  // Note: we check name equality first because this method is most often called with directive that are simply not the same
+  // name and this ensure we exit cheaply more often than not.
+  return application1.name === application2.name
```
clenfest, line 3214: I think this code gets called both during operations and also during composition. I'm assuming that it's safe to use the `application.name` because in order to compose, directives would need to be named the same in the supergraph? Just want to make sure this won't provide false positives if directives have been renamed.

clenfest (approved):

internals-js/src/definitions.ts
pcmanus: > I'm assuming that it's safe to use the application.name because in order to compose, directives would need to be named the same in the supergraph?

Correct. And if we do ever allow merging directives that have different names in different subgraphs, we will definitively have to change how we handle this. And fwiw, I'll note that this part of the logic is not changed by this patch: all that I'm truly changing here is that we skip saying that 2 `@defer` application are ever "the same", and we can't have `@defer` applications during composition anyway (since it's not allowed on any type system locations), so this patch should be safe on that front.

PR #2157, opened 2022-09-23
Miscellaneous wording cleanup to federation docs

StephenBarlow:
This is _mostly_ replacing uses of "gateway" with "router", with some other wording improvements sprinkled in.


docs/source/config.json
```diff
@@ -25,10 +25,10 @@
       "3️⃣ Local composition": "/quickstart/local-composition",
       "4️⃣ Working with subgraphs": "/quickstart/local-subgraphs"
     },
-    "Implementing a Supergraph": {
+    "Building Your Supergraph": {
       "Subgraphs in Apollo Server": "/subgraphs",
       "Other subgraph libraries": "/supported-subgraphs",
-      "Graph router/gateway": "/gateway"
+      "The graph router/gateway": "/gateway"
```
dariuszkuc, line 31: should we update the link to `/router` as well?

dariuszkuc (approved):

docs/source/config.json
StephenBarlow: ah that's fair, sure I'll make that change

Issue #213, opened 2020-10-08
Support (and/or documentation) for nested `@provides` fields arg


trevor-scheer:
Currently there's neither support or documentation for a nested `@provides` `fields` argument, like so:
`@provides(fields: "parent { child }")`

In its current form, we can see composition errors by running the test introduced by this commit:
https://github.com/apollographql/federation/commit/cf7a8811a59040ac7e34ab152d9389385fbf781b

The validation that reports this error is not permissive enough to allow this composition to happen. Relatedly, the error being reported isn't very helpful in that it doesn't tell the user precisely what they're doing wrong or how to fix it.

Beyond just the composition errors, we would also need to validate that this actually works as intended in query planning.

Until then, documentation about this limitation would be helpful. As a heavier lift, writing a correct validation which at least informs the user of what they're trying to do and that it's unsupported would be even more helpful.

chriskrogh:
@trevor-scheer is this feature being worked on? I noticed  https://github.com/apollographql/federation/issues/815 and https://github.com/apollographql/federation/issues/416 but I'm not sure where the source of truth is for this?

pcmanus:
Closing this issue as this is nested provides are supported as of federation 2.0. There is even a unit test demonstrating it: https://github.com/apollographql/federation/blob/main/query-planner-js/src/__tests__/buildPlan.test.ts#L200.

PR #2170, opened 2022-09-28
Roadmap updates to include Federation 2.1 release

korinne:
This PR makes changes to the roadmap to include what was recently released in Federation 2.1, as well as update what is under consideration for future releases.


prasek (approved):
Looks good!

ROADMAP.md
```diff
@@ -29,28 +29,24 @@ new Federation 2 (on the [main branch](https://github.com/apollographql/federati
...
-### Next Release
-
-* `@tag` export to API schema
-* Compose user-defined directives in subgraphs
-
 ### Under Consideration
 
-* `@defer` support in query planning
-* Subscriptions support in query planning
+* Ability to add fields to one subgraph at a time, allowing those fields to be immediately usable by clients.
```
pcmanus, line 34: Nit: I'd slightly modify to "Ability to add *value type* fields to one subgraph at a time, ...". I think without that context someone may be confused to what problem we're talking about since this is not an issue at all for entity types for instance.

ROADMAP.md
```diff
@@ -29,28 +29,24 @@ new Federation 2 (on the [main branch](https://github.com/apollographql/federati
...
-* `@defer` support in query planning
-* Subscriptions support in query planning
+* Ability to add fields to one subgraph at a time, allowing those fields to be immediately usable by clients. 
+* Subscriptions support in query planning.
 * Entity interfaces can be spread across multiple subgraphs & interface queries with `@interfaceObject` helper.
 * Harmonizing shared value types across subgraphs to a canonical desired state.
-* Advanced caching, auth, demand control, rate limiting, governance, and more!
+* Advanced caching, auth, demand control, rate limiting, governance, and more.
 * Importing shared types into subgraph schemas, to keep things more DRY.
 * Nested `@provides` support beyond what Fed 2 already supports natively.
```
pcmanus, line 40: Fwiw, I've seen this one ("Nested `@provides` support beyond ...") listed in roadmaps for a while, and I still have no clue what this is referring to. I'm not aware of particular limitations of nested `@provides` in fed 2.

ROADMAP.md
korinne: Thank you! Went ahead and added.

ROADMAP.md
korinne: Removed, and we can add later if we find out what's missing :)

prasek: @pcmanus somehow I missed that [you had already fixed this](https://github.com/apollographql/federation/issues/213#issuecomment-1156140565)
> Closing this issue as this is nested provides are supported as of federation 2.0. There is even a unit test demonstrating it: https://github.com/apollographql/federation/blob/main/query-planner-js/src/__tests__/buildPlan.test.ts#L200.

@korinne The example in this test ☝️ is a great example of query path specific `@provides` on a specific Query root field, which the current https://www.apollographql.com/docs/federation/entities-advanced/#using-provides doesn't really highlight. So think we need add some docs that highlight the use of nested `@provides` and then  think we can close out this issue and then list it as fixed in the Fed 2.0 release lower down in the ROADMAP.md.
- https://github.com/apollographql/federation/issues/416

jeffjakub (approved):
Ship it!

Issue #2135, opened 2022-09-08
Shareable directive doesn't behave as expected with extended types


gwardwell:
## Description

When the same subgraph has a `@shareable` value type that is then extended, Rover does not compile the type in an expected way.

## Expected behavior

When I author schema like the following:

```gql
# subgraph-a.graphqls
type Example @shareable {
  foo: String!
}

extend type Example {
  bar: String!
}
```

```gql
# subgraph-b.graphqls
type Example @shareable {
  foo: String!
  bar: String!
}
```

I would expect Rover to successfully generate the following supergraph schema (minus decoration):

```gql
# Supergraph Schema
type Example @shareable {
  foo: String!
  bar: String!
}
```

## Actual behavior

The following schema results in the following Rover error:

```gql
# subgraph-a.graphqls
type Example @shareable {
  foo: String!
}

extend type Example {
  bar: String!
}
```

```gql
# subgraph-b.graphqls
type Example @shareable {
  foo: String!
  bar: String!
}
```

```
composing supergraph with Federation v2.1.1.
error[E029]: Encountered 1 build error while trying to build a supergraph.

Caused by:
    UNKNOWN: Non-shareable field "Example.bar" is resolved from multiple subgraphs: it is resolved from subgraphs "subgraph_a" and "subgraph_b" and defined as non-shareable in subgraph "subgraph_a"
    
        The subgraph schemas you provided are incompatible with each other. See https://www.apollographql.com/docs/federation/errors/ for more information on resolving build errors.
```

If I add `@shareable` to the extended `Example` type, I get the following error:

```gql
# subgraph-a.graphqls
type Example @shareable {
  foo: String!
}

extend type Example @shareable {
  bar: String!
}
```

```gql
# subgraph-b.graphqls
type Example @shareable {
  foo: String!
  bar: String!
}
```

```
composing supergraph with Federation v2.1.1.
error[E029]: Encountered 1 build error while trying to build a supergraph.

Caused by:
    UNKNOWN: [subgraph_a] The directive "@shareable" can only be used once at this location.
    
        The subgraph schemas you provided are incompatible with each other. See https://www.apollographql.com/docs/federation/errors/ for more information on resolving build errors.
```

Only by marking the individual fields as `@shareable` in the extended type will successfully compile:

```gql
# subgraph-a.graphqls
type Example @shareable {
  foo: String!
}

extend type Example {
  bar: String! @shareable
}
```

```gql
# subgraph-b.graphqls
type Example @shareable {
[...]

PR #2175, opened 2022-09-30
Allows `@shareable` to be repeatable (to allow it on both definition and extensions)

pcmanus:
The rational this is explained on #2135 and in particular allows the example in [this comment](https://github.com/apollographql/federation/issues/2135#issuecomment-1248317927) to work.

For backward compatibility concerns, the patch introduces a new version (`federation/v2.2`) of the federation spec and only make `@shareable` repeatable in that version. This means that nothing changes for existing subgraphs on update and you have to "manually" change the subgraph to use the new version to be able to use this. More importantly, this means that subgraph libraries that continue to generate the previous version won't break in any way. They just may want to start generating the new version at some point in the future so their users can make use of the additional flexibility afforded by this change.

On the upgrade side of things, probably stating the obvious, but if a subgraph does upgrade to use the new 2.2 version, then it needs to make sure to use a version of composition that knows it.

Regarding the code itself, marking `@shareable` repeatable is trivial, but more interesting changes are:
1. the added validation discussed on #2135.
2. some refactoring around the code for the federation spec. Essentially, the exact definitions of the federation directives really depends on the spec version, even more so with this change, but that wasn't well abstracted by the code so far, so this patch tries to improve this a bit.

### DIFF
--- a/docs/source/federated-types/sharing-types.mdx
+++ b/docs/source/federated-types/sharing-types.mdx
@@ -95,6 +95,36 @@ Both subgraphs A _and_ B can now resolve the `x` and `y` fields for the `Positio
 * If a type or field is marked `@shareable` in _any_ subgraph, it must be marked either `@shareable` or [`@external`](./federated-directives/#external) in _every_ subgraph that defines it. Otherwise, composition fails.
 * If multiple subgraphs can resolve a field, **make sure each subgraph's resolver for that field behaves identically.** Otherwise, queries might return inconsistent results depending on which subgraph resolves the field.
 
+#### 📝 A note regaring `@shareable` on types
+
+When `@shareable` is put on an object type declaration, it only applies to the fields within that exact declaration. This become relevant if a subgraph defines a type though an [object definition](https://spec.graphql.org/June2018/#ObjectTypeDefinition) and one or more [object extensions](https://spec.graphql.org/June2018/#sec-Object-Extensions). For instance, if a subgraph declares:
+```graphql title="Subgraph A"
+type Position @shareable {
+  x: Int!
+  y: Int!
+}
+
+extend type Position {
+  z: Int!
+}
+```
+then field `z` will _not_ be shareable. To make `z` shareable, you can either mark it with `@shareable` directly or mark the extension itself shareable:
+```graphql title="Subgraph A"
+# This example only works with v2.2+ of the federation specification. In earlier versions, `@shareable` was
+# not marked `repeatable` and so this example would be rejected by graphQL.
+extend schema
+  @link(url: "https://specs.apollo.dev/federation/v2.2", import: ["@shareable"])
+
+type Position @shareable {
+  x: Int!
+  y: Int!
+}
+
+extend type Position @shareable {
+  z: Int!
+}
+```
+
 ## Differing shared fields
 
 Shared fields can differ between subgraphs in specific ways:
--- a/internals-js/src/federation.ts
+++ b/internals-js/src/federation.ts
@@ -7,6 +7,7 @@ import {
   Directive,
   DirectiveDefinition,
   ErrGraphQLValidationFailed,
+  Extension,
   FieldDefinition,
   InputFieldDefinition,
   InterfaceType,
@@ -27,7 +28,7 @@ import {
   sourceASTs,
   UnionType,
 } from "./definitions";
-import { assert, joinStrings, MultiMap, printHumanReadableList, OrderedMap } from "./utils";
+import { assert, joinStrings, MultiMap, printHumanReadableList, OrderedMap, mapValues } from "./utils";
 import { SDLValidationRule } from "graphql/validation/ValidationContext";
 import { specifiedSDLRules } from "graphql/validation/specifiedRules";
 import {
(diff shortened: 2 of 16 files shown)


docs/source/federated-types/sharing-types.mdx
```diff
@@ -95,6 +95,36 @@ Both subgraphs A _and_ B can now resolve the `x` and `y` fields for the `Positio
...
+extend type Position {
+  z: Int!
+}
+```
+then field `z` will _not_ be shareable. To make `z` shareable, you can either mark it with `@shareable` directly or mark the extension itself shareable:
+```graphql title="Subgraph A"
+# This example only works with v2.2+ of the federation specification. In earlier versions, `@shareable` was
+# not marked `repeatable` and so this example would be rejected by graphQL.
+extend schema
+  @link(url: "https://specs.apollo.dev/federation/v2.2", import: ["@shareable"])
```
pcmanus, line 116: Side-note: I noticed that the document has plenty of `@link`, and almost all of them still link to `federation/v2.0` (the one exception is the part about `@composeDirective` with links to `federation/v2.1` because it has to). But I've been wondering if we shouldn't try to have the doc always link to the most recent spec version? I mean, we probably have to at the moment because studio composition is lagging behind, but this doesn't feel like we setup user for the best experience, so it might be worth discussing how we get to a better state. Anyway, certainly an aside...

subgraph-js/src/__tests__/buildSubgraphSchema.test.ts
```diff
@@ -133,12 +133,12 @@ directive @requires(fields: _FieldSet!) on FIELD_DEFINITION
 
 directive @provides(fields: _FieldSet!) on FIELD_DEFINITION
 
-directive @extends on OBJECT | INTERFACE
```
pcmanus, line 136: Note to reviewers: the reason the order of this changed is due to some code "consolidation" of the refactoring done in `federationSpec.ts`/`federation.ts`. More precisely, the `completeFed1SubgraphSchema` method in `federation.ts` used to manually list the directive, and it did so in an ever so slightly different order than how those same directives are listed in the fed2 spec. But the code now use the same list in both case (the `legacyFederationDirectives` list in `federationSpec.ts`) and that list use the existing order of the fed2 spec, but thus slightly change the order when we're "completing" a fed1 schema.

Now, we could of course avoid that order change here by recreating 2 separate lists, but I genuinely think the change of order here doesn't matter (and so is worth consolidating the code a bit). In particular, I don't think this would show up in "diffs" for users, because this here is the output of `buildSubgraphSchema`, and that's only used by subgraph to generate a valid schema that can be run by Apollo Server, but it is not the schema that user upload to studio in particular (use upload the "input" of `buildSubgraphSchema`).

internals-js/src/federation.ts
```diff
@@ -488,6 +480,48 @@ function validateInterfaceRuntimeImplementationFieldsTypes(
   }
 }
 
+function validateShareableNotRepeatedOnSameDeclaration(
+  element: ObjectType | FieldDefinition<ObjectType>,
+  metadata: FederationMetadata,
+  errorCollector: GraphQLError[],
+) {
+  const shareableApplications: Directive<any, {}>[] = element.appliedDirectivesOf(metadata.shareableDirective());
```
clenfest, line 488: Is there a reason you want `Directive<any, {}>[]` instead of `Directive[]`?

internals-js/src/federation.ts
```diff
@@ -488,6 +480,48 @@ function validateInterfaceRuntimeImplementationFieldsTypes(
...
+  if (shareableApplications.length <= 1) {
+    return;
+  }
+
+  type ByExtensions = {
+    without: Directive<any, {}>[],
+    with: MultiMap<Extension<any>, Directive<any, {}>>,
+  };
+  const byExtensions = shareableApplications.reduce<ByExtensions>(
```
clenfest, line 497: Oh neat! Specifying the return type of reduce in the <> is much more readable.

clenfest (approved):
A couple of minor comments, but approved otherwise.

subgraph-js/src/__tests__/buildSubgraphSchema.test.ts
```diff
@@ -133,12 +133,12 @@ directive @requires(fields: _FieldSet!) on FIELD_DEFINITION
...
-
 directive @external(reason: String) on OBJECT | FIELD_DEFINITION
 
 directive @tag(name: String!) repeatable on FIELD_DEFINITION | OBJECT | INTERFACE | UNION | ARGUMENT_DEFINITION | SCALAR | ENUM | ENUM_VALUE | INPUT_OBJECT | INPUT_FIELD_DEFINITION
 
+directive @extends on OBJECT | INTERFACE
```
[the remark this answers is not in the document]
pcmanus: See my comment made above. I explained why that happened and what is imo ok.

internals-js/src/federation.ts
pcmanus: It was more than the former was a bit more precise than the later (making it clear there is not argument to the applications), but certainly doesn't really change anything to that method so happy to shorten it.

PR #2176, opened 2022-09-30
Fix validation of variable on input field not taking default into account

pcmanus:
Pretty obvious oversight: the code just "forgot" to pass the default value for input field during validation so they were not taken into account.


clenfest (approved):
LGTM

PR #2182, opened 2022-10-05
Optimize plan for defer where only keys are fetched

pcmanus:
Router-based defer is based on using an entity key to "resume" the deferred part of a `@defer`, but that means that at least the key(s) needs to be fetched in the non-deferred part. But if only the key were selected in the deferred part, the code was issuing a new deferred fetch to get those key field "again". This commit avoids this by recognizing that those key fields don't need to be fetched again.

### DIFF
--- a/query-planner-js/src/buildPlan.ts
+++ b/query-planner-js/src/buildPlan.ts
@@ -49,6 +49,7 @@ import {
   MultiMap,
   NamedFragmentDefinition,
   typenameFieldName,
+  mapKeys,
 } from "@apollo/federation-internals";
 import {
   advanceSimultaneousPathsWithOperation,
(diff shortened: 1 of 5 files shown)


query-planner-js/src/buildPlan.ts
```diff
@@ -1522,6 +1539,34 @@ class FetchDependencyGraph {
...
+      // cases (other cases will remain non-optimal, but hopefully this handle all the cases
+      // we care about in practice):
+      //   1. if the group has no children. In which case we can just remove it with no ceremony.
+      //   2. if the group has only a single parent and we have a path to that parent.
+      if (group.children().length === 0) {
+        this.remove(group);
+      } else {
+        const parents = group.parents();
+        const parent = parents[0];
+        if (parents.length === 1 && parent.path) {
```
clenfest, line 1563: Just for my understanding, what are the conditions under which a parent wouldn't have a path?

pcmanus: There is some cases where the `path` doesn't really make sense at I explained in [this comment](https://github.com/apollographql/federation/blob/main/query-planner-js/src/buildPlan.ts#L603-L607). But more concretely, it show in code paths at least [here](https://github.com/apollographql/federation/blob/main/query-planner-js/src/buildPlan.ts#L3274) and [here](https://github.com/apollographql/federation/blob/main/query-planner-js/src/buildPlan.ts#L3319).

clenfest (approved):

PR #2184, opened 2022-10-09
Add GraphQL Mesh to supported subgraphs

enisdenjo:
Follow-up task from: https://github.com/apollographql/apollo-federation-subgraph-compatibility/pull/242.


Meschreiber:
Also looks already done :)

PR #2185, opened 2022-10-10
bug in objectEquals() for objects created with Object.create(null)

### LINKED ISSUE #2173: Fed 2 @composeDirective + repeatable composition exception
Reproduction: https://stackblitz.com/edit/basic-federation-sjbdop?file=one.js,two.js
Versions: `@apollo/gateway@2.1.3`
Behavior: Adding `repeateable` to a directive definition and composing the directive using `@composeDirective` results in an exception on composition: 

```
TypeError: b.hasOwnProperty is not a function
    at objectEquals (/home/projects/basic-federation-sjbdop/node_modules/@apollo/federation-internals/dist/values.js:94:42)
    at valueEquals (/home/projects/basic-federation-sjbdop/node_modules/@apollo/federation-internals/dist/values.js:68:41)
    at Merger.sameDirectiveApplication (/home/projects/basic-federation-sjbdop/node_modules/@apollo/composition/dist/merging/merge.js:1339:55)
    at eval (/home/projects/basic-federation-sjbdop/node_modules/@apollo/composition/dist/merging/merge.js:1290:53)
    at Array.filter (<anonymous>)
    at eval (/home/projects/basic-federation-sjbdop/node_modules/@apollo/composition/dist/merging/merge.js:1290:35)
    at Array.map (<anonymous>)
    at Merger.mergeAppliedDirective (/home/projects/basic-federation-sjbdop/node_modules/@apollo/composition/dist/merging/merge.js:1290:22)
    at Merger.mergeAllAppliedDirectives (/home/projects/basic-federation-sjbdop/node_modules/@apollo/composition/dist/merging/merge.js:1252:22)
    at Merger.merge (/home/projects/basic-federation-sjbdop/node_modules/@apollo/composition/dist/merging/merge.js:220:14)
```

clenfest:
If an object was created with Object.create(null), it will not have the `hasOwnProperty` function, which will cause us to throw. Fixes #2173


internals-js/src/values.ts
```diff
@@ -162,7 +162,12 @@ function objectEquals(a: {[key: string]: any}, b: {[key: string]: any}): boolean
     // Beware of false-negative due to getting undefined because the property is not
     // in args2.
     if (v2 === undefined) {
-      return v1 === undefined && b.hasOwnProperty(key);
+      // if objects were created with Object.create(null), they will not have
```
pcmanus, line 165: Not too sure about this option. In particular, this makes the method not always symmetric if the 2 input objects are not created the same way:
```typescript
const v1 = Object.create(null);
v1['foo'] = undefined;
const v2 = { bar: undefined };

// That returns `false`
valueEquals(v1, v2);
// _but_ that returns `true`
valueEquals(v2, v1);
```
Not that this asymmetry is not necessarily a deal breaker by itself, in the sense that inputs to this method should be graphQL input values, and we should theoretically not get properties explicitly set to `undefined` (it doesn't really make sense in graphQL; we can get a property set to `null`, or not set at all, but set to `undefined` is fishy), the behaviour of this shouldn't matter in the first place. But if we're saying that the behaviour is unspecified when the input objects have properties set to `undefined`, then there is no point in checking `hasOwnProperty` at all, and doing it conditionally is even more confusing.

Now, we obviously _do_ end up calling this method with an object having properties explicitly set to `undefined`, or we wouldn't have a bug here. And as said above, that's imo the more important problem to fix here. And I believe this is because the `Directive.arguments(true)` method (in `definition.ts`) has a bug where it does sometimes set properties of the returned object to `undefined`. Namely, in this method, we should change:
```typescript
updated[argDef.name] = withDefaultValues(this._args[argDef.name], argDef);
```
into
```typescript
const value = withDefaultValues(this._args[argDef.name], argDef);
if (value !== undefined) {
  updated[argDef.name] = value;
}
```
And I believe that's enough to fix the bug here.

*That said*, we should probably put `objectEquals` out its current misery nonetheless. And I think there is 2 valid options:
1. we explicitly document that the arguments should *not* have properties explicitly set to `undefined`. In which case we can remove the `if (v2 === undefined)` condition altogether, and maybe replace it by a simple `assert(v1 !== undefined)` to help debug future potential misuse of the method.
2. or we prefer making the method a bit more defensive and work even if there is properties set to `undefined` (even if again, this really shouldn't happen), and I think the simplest option is to replace `b.hasOwnProperty(key)` by `keys2.include(key)`.

pcmanus (changes requested):

internals-js/src/values.ts
clenfest: Happy to do option 2, but I think the 3rd option would be to ask whether if you have two objects:

```typescript
const a = {};
const b = { foo: undefined };
```

Why shouldn't these objects be equal? Obviously, I'm concerned that this is a bigger change that could break things, so it's possible that we wouldn't want to do it for that reason.

internals-js/src/values.ts
```diff
@@ -162,7 +162,12 @@ function objectEquals(a: {[key: string]: any}, b: {[key: string]: any}): boolean
...
+      // a hasOwnProperty function, which makes it impossible to tell if
+      // the property is there explicitly or not
```
pcmanus, line 167: Nit: it would be nice to slightly amend the comment. I think if we get to it in 6 months, it may be slightly confusing as phrased since `hasOwnProperty` is not mentioned anymore.

internals-js/src/values.ts
pcmanus: I'll note that I'm good with the change, but I really think the important fix here is to modify the `Directive.arguments(true)` method so it stop setting properties to `undefined` in the generated object. This is unexpected and so might create other similar issues later.

internals-js/src/definitions.ts
```diff
@@ -3062,7 +3062,11 @@ export class Directive<
     assert(definition, () => `Cannot include default values for arguments: cannot find directive definition for ${this.name}`);
     const updated = Object.create(null);
     for (const argDef of definition.arguments()) {
-      updated[argDef.name] = withDefaultValues(this._args[argDef.name], argDef);
+      const argValue = withDefaultValues(this._args[argDef.name], argDef);
+      // Note that argValue could be '0' or something falsy here, so we must explicitly check === undefined
+      if (argValue === undefined) {
```
pcmanus, line 3067: I believe you meant `!==` here (I believe that's at least one of the reason for test breakage).

pcmanus (approved):

Issue #1917, opened 2022-06-16
Apollo Gateway Query Plan incorrectly handles union/fragment of sharable Interfaces


getglad:
I am running into a problem when running a query of a shared interface that is also a field on `Query`.

Essentially, it appears that something is causing confusion in the query plan where the query plan is defaulting to a particular subgraph. If that subgraph does not have the implementation of the type, it drops the union/fragment.

We have built a plugin with logic that can analyze the `id` and point the query to the correct subgraph, and can get a return on the gateway, but when the gateway returns the the client, the result is `null`. I assume this is because the gateway is attempting to return a type belonging to the "default" subgraph instead of the correct subgraph and something is failing.

I have seen some recent PRs that impact the query plan - was not sure if this may be related.
- https://github.com/apollographql/federation/pull/1911
- https://github.com/apollographql/federation/pull/1859

Also entirely possible that I've misunderstood how to decorate these types/fields, so could be human error.

Happy to provide more details, but didn't want to overwhelm with a wall of output.

### Reproducable example

Subgraph 1
```
extend schema
  @link(url: "https://specs.apollo.dev/federation/v2.0",
        import: ["@key", "@shareable"])

interface Node {
  id: ID!
}

type Query {
  node(id: ID!): Node @shareable
}
```

Subgraph 2
```
extend schema
  @link(url: "https://specs.apollo.dev/federation/v2.0",
        import: ["@key", "@shareable"])

interface Node {
  id: ID!
}

type Foobar implements Node {
  id: ID!

  uuid: String!
  version: Int!
  name: String!
  description: String
  category: [String!]
}

type Query {
  node(id: ID!): Node @shareable
}
```

Supergraph (partial/relevant bits - I can share more if helpful)
```
enum join__Graph {
  SUBGRAPH1 @join__graph(name: "subgraph1", url: "https://path.com")
  SUBGRAPH2 @join__graph(name: "subgraph2", url: "https://otherpath.com")
}

interface Node
  @join__type(graph: SUBGRAPH1)
  @join__type(graph: SUBGRAPH2)
{
  id: ID!
}

type Foobar implements Node
  @join__implements(graph: SUBGRAPH2, interface: "Node")
  @join__type(graph: SUBGRAPH2)
{
  id: ID!
  uuid: String!
  version: Int!
  name: String!
  description: String
  category: [String!]
}

type Query
  @join__type(graph: SUBGRAPH1)
  @join__type(graph: SUBGRAPH2)
{
[...]

PR #2194, opened 2022-10-17
Fix building additional abstract type edges

mpospelov:
This is an attempt to fix the following related issues
https://github.com/apollographql/federation/issues/378
https://github.com/apollographql/federation/issues/361
https://github.com/apollographql/federation/issues/1917

### DIFF
--- a/federation-integration-testsuite-js/src/fixtures/documents.ts
+++ b/federation-integration-testsuite-js/src/fixtures/documents.ts
@@ -1,3 +1,4 @@
+import { GraphQLResolverMap } from '../resolverMap';
 import { fed2gql as gql } from '../utils/fed2gql';
 
 export const name = 'documents';
@@ -16,10 +17,15 @@ export const typeDefs = gql`
     name: String!
   }
 
-  type Image implements NamedObject {
+  interface WebResource {
+    resourceUrl: String
+  }
+
+  type Image implements NamedObject & WebResource @key(fields: "name") {
       name: String!
       # Same as option below but the type is different
       attributes: ImageAttributes!
+      resourceUrl: String
   }
 
   type Text implements NamedObject {
@@ -37,3 +43,15 @@ export const typeDefs = gql`
     text: String
   }
 `;
+
+export const resolvers: GraphQLResolverMap<unknown> = {
+  Image: {
+    __resolveReference: (image: { name: string }) => {
+      return {
+        name: image.name,
+        resourceUrl: `https://images.com/${image.name}`,
+        attributes: { url: 'https://example.com/image.jpg' },
+      };
+    }
+  }
+}
--- a/federation-integration-testsuite-js/src/fixtures/documents.ts
+++ b/federation-integration-testsuite-js/src/fixtures/documents.ts
@@ -21,7 +21,12 @@ export const typeDefs = gql`
     resourceUrl: String
   }
 
-  type Image implements NamedObject & WebResource @key(fields: "name") {
+  interface ReviewSubjectInterface {
+    subjectInterfaceId: ID!
+  }
+
+  type Image implements NamedObject & WebResource & ReviewSubjectInterface @key(fields: "name") {
+      subjectInterfaceId: ID!
       name: String!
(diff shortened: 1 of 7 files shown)


mpospelov:
Hello @pcmanus @jeffjakub! Thanks for your great project and to all Apollo team! 
Could you please check the updated code that fixed mentioned issues? Thanks

pcmanus:
Hi @mpospelov, thanks a lot for the contribution. Before all, I want to apologize for not following up on this much sooner. By sheer coincidence, I happened to have started looking at this this morning.

I agree that this correctly identify and fix the issue described on #361. Fwiw, I'm less sure this should also close #378 and #1917, because some of the conversation on those issues make me worried that they are about another problem (and I could be wrong, I'm just trying to be careful with closing an issue without before certain to have fixed the real problem involved; if it ends up fixing the problem for those users, all the better).

Anyway, back on the PR, it does fix the issue, but I think it's also a tad too "coarse", in that it will result in type-explosion in cases where we could genuinely still avoid it without creating problem. And experience having shown that avoiding type-explosion when possible can sometimes make a measurable difference on some schema, I'm keen on avoiding a regression on that front where it can be avoided.

I believe there is 2 main reasons why the fix is not as precise as it could be.

The first one is that for a given type, both edges are always skipped in the patch, but the underlying problem is asymmetric. Take the following example:
```graphql
# Subgraph `S1`
type Query {
  i1: I1
  i2: I2
}

interface I1 {
  v: Int
}

interface I2 {
  v: Int
}

type A implements I1 {
  v: Int @shareable
}

type B implements I1 & I2 {
  v: Int
}
```
```graphql
# Subgraph `S2`
interface I2 {
  v: Int
}
type A implements I2 {
  v: Int @shareable
}
```
For query like:
```graphql
query {
  i1 {
   ... on I2 {
     v
   }
  }
}
```
we should absolutely type-explode `I2` (we can't just forward the query to `S` "as is"), because `i1` may return a `A` object and `A` is also a `I2` _in the supergraph_, but `S1` does not know that. This is what the PR fix, by ensuring we have no edge `I1` -> `I2`.

However, consider a query like:
```graphql
query {
  i2 {
   ... on I1 {
     v
   }
  }
}
```
In that case, `i2` will just never return a `A` object (that would be invalid for `S1` to do so), nor any other runtime types that the definition of `I1` in `S1` wouldn't catch. So in that case, it's fine to not type explode (the query can be forwarded to `S1` as is), so we do want to add the edge `I2` -> `I1`.

The second reason for imprecision is that the patch use a comparison on the length of the runtime types. But that means that every time an abstract type has more runtime types in the supergraph than in a local subgraph, then we'll always give up on the optimization. But in practice, any runtime type that don't intersects the 2 abstract types we're considering (globally) should have no impact whatsoever on what we do.

Aside from that, and for what it's worth, we're trying to avoid adding new tests that expand on the "fixtures" from `federation-integration-testsuite-js/src/fixtures`. we keep the existing tests using them, because no point in removing working tests, but are steering away from it. The reason is that new tests tend to need schema changes, but when tons of tests rely on said schema, any change tend to impact unrelated tests, which is at least annoying to review, but we've even have case when such changes inadvertently make other test stop testing what they were designed for and that's super hard to detect. Fwiw, I also personally prefer tests that are self-contained, and having to navigate between the fixtures back-and-forth continuously to understand a test was driving me nuts. To be clear, you couldn't have known, but I'm just trying to justify why I'd really prefer making more self-contained tests.

With all that said, this is an important fix to get in, and I was feeling pretty bad about my reaction time on this, so I took the liberty to take a stab at improving the 2 points I mention above, and moving the tests out of using fixtures, and I pushed that at https://github.com/pcmanus/federation/commit/4ec171c2def609aa2060bd9018ed9e734fee49d6. I did started from your patch initially and amended, so the commit has your name on it, and so I'd like to make sure this work for you/looks good.

pcmanus:
To close the loop here: we're about to release version 2.2, and I wanted to get this in, so I went ahead an merge the modified patch I mentioned above.

Again, thanks a lot for the contribution.

PR #2212, opened 2022-10-21
Ensure supergraph @defer/@stream definitions of supergraph are not in…

pcmanus:
…cluded in the API schema

Whether or not @defer and @stream ends up in the API schema used by the router/gateway depends on the runtime `config` provided to the call to `apiScheam.toGraphQLJSSchema(config)`, but whatever definition are in the supergraph should be ignored, because those correspond to what the subgraph were declaring (which is irrelevant to the defer support in the supergraph API schema).

### DIFF
--- a/internals-js/src/definitions.ts
+++ b/internals-js/src/definitions.ts
@@ -1270,6 +1270,19 @@ export class Schema {
       this.validate();
 
       const apiSchema = this.clone();
+
+      // As we compute the API schema of a supergraph, we want to ignore explicit definitions of `@defer` and `@stream` because
+      // those corresponds to the merging of potential definition from the subgraphs, but whether the supergraph API schema
+      // support defer or not is unrelated to the subgraph capacity. As far as gateway/router support goes, whether the defer/stream
+      // definitions ends up being provided or not will depends on the runtime `config` argument of the `toGraphQLJSSchema` that
+      // is the called on the API schema (the schema resulting of that method).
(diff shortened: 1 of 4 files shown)


internals-js/src/definitions.ts
```diff
@@ -1270,6 +1270,19 @@ export class Schema {
       this.validate();
 
       const apiSchema = this.clone();
+
+      // As we compute the API schema of a supergraph, we want to ignore explicit definitions of `@defer` and `@stream` because
+      // those corresponds to the merging of potential definition from the subgraphs, but whether the supergraph API schema
+      // support defer or not is unrelated to the subgraph capacity. As far as gateway/router support goes, whether the defer/stream
+      // definitions ends up being provided or not will depends on the runtime `config` argument of the `toGraphQLJSSchema` that
+      // is the called on the API schema (the schema resulting of that method).
```
clenfest, line 1278:
```suggestion
      // As we compute the API schema of a supergraph, we want to ignore explicit definitions of `@defer` and `@stream` because
      // those correspond to the merging of potential definitions from the subgraphs, but whether the supergraph API schema
      // supports defer or not is unrelated to the subgraph capacity. As far as gateway/router support goes, whether the defer/stream
      // definitions end up being provided or not will depend on the runtime `config` argument of the `toGraphQLJSSchema` that
      // is the called on the API schema (the schema resulting from that method).
```

clenfest (approved):

PR #2213, opened 2022-10-21
Align @external with @shareable for how it behaves on a type.

### LINKED ISSUE #2156: Remove or handle `OBJECT` location for `@external` directive
For some reason, we declare `@external` as:
```graphql
directive @external(reason: String) on OBJECT | FIELD_DEFINITION
```
but the `OBJECT` location is currently somewhat incorrect here as the code completely ignores it.

I'm actually not sure why this is the definition we have, but it seems somewhat historical as fed1 has it (https://github.com/apollographql/federation/blob/version-0.x/subgraph-js/src/directives.ts#L39-L42) and this from as far as the git history if this current repository goes. But:
- I couldn't really find any code handling in fed1, or any test  using it.
- I'm almost 100% certain the fed2 code has no code for this.

So we shouldn't leave it this way as it's currently confusing. And I see 2 options:
1. we just remove the `OBJECT` location in an upcoming release. The one slight tricky concern is that we have code that ensure the definition of federation directives is correct when reading subgraphs, and if we just remove the location without other changes, we'd start rejecting any subgraph having the `OBJECT` location, and that's essentially out of question (all subgraph libraries would need immediate upgrade to continue producing subgraph accepted). So we'll also need to make a special case in the validation for `@external` so that we still accept the `OBJECT` location, even if we stop producing it from our own code, remove it from our doc, and tell subgraph libraries they can remove it safely whenever they feel like it.
2. we keep the location, but actually make do something. `@external`, at least from a fed1 POV, is fundamentally a field property, as it means something along the lines of "the subgraph actually doesn't resolve that field, but it needs to be declared because of a `@requires` or other". That said, it's very similar to `@shareable` and we _do_ accept `@shareable` on types as a shortcut for applying it to all the fields of the type declaration. So it wouldn't be crazy to do exactly the same for `@external`, at least on principle. Is that useful? Well, possibly. For instance, you can sometime `@require` full objects and sometimes that's the only reason for declaring the types of those object, and in that case, all of the fields would be `@external`, and being able to put it once on a type could be convenient. Think:
```graphql
type T @key(fields: "id") {
  id: ID!
  x: X @external
  y: Int @requires(fields: "x { a b c d }")
}

type X @external {
  a: Int
  b: Int
  c: Int
  d: Int
}
```

I think both options are reasonable, though I find it somewhat appealing to create some symmetry around how `@shareable` and `@external` can be used, and it's probably not too much code to do anyway.

clenfest:
i.e. marking the type external makes all fields external.

Fixes #2156

### DIFF
--- a/composition-js/CHANGELOG.md
+++ b/composition-js/CHANGELOG.md
@@ -5,7 +5,7 @@ This CHANGELOG pertains only to Apollo Federation packages in the 2.x range. The
 ## vNext
 
 - Improves error message to help with misspelled source of an `@override` [PR #2181](https://github.com/apollographql/federation/pull/2181).
-- Provide support for marking @external on type [PR #2213](https://github.com/apollographql/federation/pull/2213)
+- Provide support for marking @external on object type [PR #2213](https://github.com/apollographql/federation/pull/2213)
 
 ## 2.1.2
 
--- a/composition-js/CHANGELOG.md
+++ b/composition-js/CHANGELOG.md
@@ -5,6 +5,7 @@ This CHANGELOG pertains only to Apollo Federation packages in the 2.x range. The
 ## vNext
 
 - Improves error message to help with misspelled source of an `@override` [PR #2181](https://github.com/apollographql/federation/pull/2181).
+- Provide support for marking @external on type [PR #2213](https://github.com/apollographql/federation/pull/2213)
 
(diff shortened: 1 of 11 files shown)


clenfest:
The negative case seems to be well covered in tests already, but let me know if you think there is a test that's missing.

PR #2214, opened 2022-10-21
Align @external with @shareable for how it behaves on a type

clenfest:
i.e. marking the type external makes all fields external.

Fixes https://github.com/apollographql/federation/issues/2156

### DIFF
--- a/composition-js/CHANGELOG.md
+++ b/composition-js/CHANGELOG.md
@@ -5,7 +5,7 @@ This CHANGELOG pertains only to Apollo Federation packages in the 2.x range. The
 ## vNext
 
 - Improves error message to help with misspelled source of an `@override` [PR #2181](https://github.com/apollographql/federation/pull/2181).
-- Provide support for marking @external on type [PR #2213](https://github.com/apollographql/federation/pull/2213)
+- Provide support for marking @external on object type [PR #2213](https://github.com/apollographql/federation/pull/2213)
 
 ## 2.1.2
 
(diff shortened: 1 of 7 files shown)


pcmanus (changes requested):
This looks mostly good as most testing of external appears to properly rely on the `ExternalTester`, but unfortunately there seems to be a few exceptions within `querygrapht.ts` (one [here](https://github.com/apollographql/federation/blob/main/query-graphs-js/src/querygraph.ts#L1031) and a few similar later).

I'd be suggest adding a new `isExternal` method similar to the `hasDirective` of this field, but that calls `metadata.isFieldExternal` instead of just checking the directive directly (this will leave the `hasDirective` only used once for `@requires` but that's probably fine).

query-graphs-js/src/querygraph.ts
```diff
@@ -965,6 +965,11 @@ class GraphBuilderFromSchema extends GraphBuilder {
     return !!metadata && field.hasAppliedDirective(directiveFct(metadata));
   }
 
+  private isExternal(field: FieldDefinition<any>): boolean {
+    const metadata = federationMetadata(this.schema);
+    return !!metadata && metadata.isFieldExternal(field);
```
clenfest, line 970: @pcmanus 
I followed the pattern you had above, but is there a reason we shouldn't just assert `metadata` here?

pcmanus: Yes. We build `QueryGraph` using this code both for subgraph schema _and_ for the supergraph API schema. The former will always have a `metadata` but the later won't, but we don't want to hard fail on it.

If we were to guard the calling of those methods with `this.isFederatedSubgraph`, then yes, we could assert `metadata` at that point, but not sure it would be much cleaner.

pcmanus (approved):

PR #2218, opened 2022-10-26
Preserve default values for input object fields when extracting subgraph

### LINKED ISSUE #2174: Gateway v2 validation rejects nullable query variable used as input to non-null field of input object with a default value
## Summary

Hi, we've got a schema with a field that takes an input object where a field of that input object is non-nullable and has a default value. We make a query for that field with a _nullable_ query variable, so that we don't need to repeat the default value at the client.
This was accepted by Apollo Gateway v0 and is still accepted by Apollo Server v3 and v4, but is rejected by a validation implemented in Apollo Gateway v2. TBH I don't know if this is supposed to be valid according to the graphql spec or if it has just slipped past validations thus far.

## Reduced testcase

Runnable example in gist: https://gist.github.com/cysp/7a9d8b5cfe808315b24ece291693afd8

### Schema:
```graphql
input Input {
  a: Int! = 1234
}

type Query {
  field(input: Input!): String
}
```

### Operation:
```graphql
query Foo($a: Int) {
  field(input: { a: $a })
}
```

## Expected result

The query is accepted and `input.a` in the field resolver takes on the default value of the input field, 1234

## Actual result

```json
{
  "errors": [
    {
      "message": "Invalid value {a: $a} for argument \"Query.field(input:)\" of type Input!",
      "extensions": […]
    }
  ]
}
```

cysp:
### Preserve default values for input object fields when extracting subgraph from super graph

Fixes #2174.

The previous attempt at this problem in https://github.com/apollographql/federation/pull/2176 took the default value of an input object field into account when validating a query document against the schema, but it turns out that input object fields' default values are being dropped before we get to that point, when we extract a subgraph from the supergraph.
This code change solves my problem in practice. 🙂


pcmanus (approved):
Thanks a lot for the patch, and sorry again for missing that. Would you mind adding a quick entry in `gateway-js/changelog` before we merge it (I would do it, but I don't think I can push to your own repo)?

cysp:
Sure no worries, I used about the same wording as I used in the PR title here but if you have other ideas about wording I believe you can modify this branch (I've ticked a box that sounds like it allows that, "allow edits by maintainers") 🙂

PR #2221, opened 2022-10-26
Update opentelemetry.mdx comment for default Collector URL

kevin-lin:
According to https://www.npmjs.com/package/@opentelemetry/exporter-trace-otlp-http, the default URL used by this particular constructor is http://localhost:4318/v1/traces. This updated comment should be more correct and explicit, in that we also expect a default path of `/v1/traces`


jeffjakub (approved):
Thanks for the correction!

PR #2223, opened 2022-10-31
update renovate to use next branch for major version changes

clenfest:


clenfest:
@trevor-scheer - Would you mind taking a look and double check that this will do what I want?

renovate.json5
```diff
@@ -81,6 +82,19 @@
...
       matchManagers: [ "npm" ],
     },
+    {
+      // major version changes should go to the next branch rather than main
+      groupName: "all major dependencies",
+      matchUpdateTypes: ["major"],
+      // Exclude all v0 dependencies from the major group. Changes to v0 packages
+      // should be treated as major updates, but renovate treats all patches and minors
+      // the same.
+      matchCurrentVersion: "!/^[\^~]?0/",
```
trevor-scheer, line 92: Hmm...I think it would be preferable to get 0.x updates on to the next branch (since they are semver major, as the comment notes), but it might be tricky to group them into this rule.

You could create an additional rule for capturing minor/patch v0.x bumps if you drop the `!`, then use the same group name `all major dependencies` to include them.

In this rule, with the `matchUpdateTypes` as `major` only, I expect renovate to explicitly ignore v0 packages, so I don't think this is necessary at all for this rule.

trevor-scheer (commented):
Is the goal of majors only on `next` cautionary? It should be ok to take new major versions of packages if API and behavior are preserved.

I also find the idea of grouping majors all in a batch to be a bit fraught. Taking on a new major often means parsing a changelog and updating usage if necessary. It's kinda nice to have those updates and related changes happen in individual commits.

PR #2230, opened 2022-11-02
Disable `exposeDocumentNodeInFetchNode` by default

### LINKED ISSUE #2227: Increase default in-memory cache size or disable AST parsing
With the addition of https://github.com/apollographql/federation/pull/1878, [by default](https://github.com/apollographql/federation/pull/1878/files#diff-92b604a7eda765787d04455bcbd2fe815b372f03a6b30a49fcfdaed9596a1fb1R26) in the gateway, we are now storing the full AST object in memory AND the query plan document for every operation. We turned on this feature without updating the default store size: https://github.com/apollographql/federation/blob/6bfec248ebd4735f2567704410a1506d7217c88e/gateway-js/src/index.ts#L196-L201

Some enterprise customers are starting to see increased usage and scaling needs to now handle this change of usage which means a big jump in resources or time process cache evictions and the latter is causing some timeouts

If we are not going to increase the default cache size we might want to consider storing the AST in a separate store or turning it on with a flag so users of the parsed AST can opt in and can also be aware of the increased memory requirements

pcmanus:
Since #1878 (added in 2.1.0), the full AST of subgraph fetch operations has been store in query plans. However, this is not used by the gateway by default, while it noticeably increases the size of query plans, leading to either a less efficient query plan cache (if its size is limited) or increased memory consumption/pressue (if the cache is given plenty of size).

This patch disable this option by default. Users that explicitely need to access the subgraph fetch ASTs can still make the trade-off of additional memory pressure by manually using the `GatewayConfig.queryPlannerConfig.exposeDocumentNodeInFetchNode` config option.

Fixes #2227.

### DIFF
--- a/gateway-js/CHANGELOG.md
+++ b/gateway-js/CHANGELOG.md
@@ -4,6 +4,10 @@ This CHANGELOG pertains only to Apollo Federation packages in the 2.x range. The
 
 ## vNext
 
+- __BREAKING__: Disable exposing full document to sub-query by default (introduce in in 2.1.0):
+  - This change decrease memory consumption in general (which is the reason for disabling this by
+    default), but users that have custom code making use of `GraphQLDataSourceProcessOptions.document`
+    will now need to explicitly set `GatewayConfig.queryPlannerConfig.exposeDocumentNodeInFetchNode`.
 - Allows `@shareable` to be repeatable so it can be allowed on both a type definition and its extensions [PR #2175](https://github.com/apollographql/federation/pull/2175).
   - Note that this require the use of the new 2.2 version of the federation spec introduced in this change.
 - Preserve default values of input object fields [PR #2218](https://github.com/apollographql/federation/pull/2218).
--- a/gateway-js/src/index.ts
+++ b/gateway-js/src/index.ts
@@ -192,13 +192,15 @@ export class ApolloGateway implements GatewayInterface {
   }
 
   private initQueryPlanStore(approximateQueryPlanStoreMiB?: number) {
+    // Create ~about~ a 30MiB InMemoryLRUCache (or 50MiB if the full operation ASTs are
+    // enabled in query plans as this plan then use quite a bit more memory). This is
+    // less than precise since the technique to calculate the size of a DocumentNode is
+    // only using JSON.stringify on the DocumentNode (and thus doesn't account
+    // for unicode characters, etc.), but it should do a reasonable job at
+    // providing a caching document store for most operations.
+    const defaultSize = (this.config.queryPlannerConfig?.exposeDocumentNodeInFetchNode ?? false) ? 50 : 30;
     return new LRUCache<string, QueryPlan>({
-      // Create ~about~ a 30MiB InMemoryLRUCache.  This is less than precise
-      // since the technique to calculate the size of a DocumentNode is
-      // only using JSON.stringify on the DocumentNode (and thus doesn't account
-      // for unicode characters, etc.), but it should do a reasonable job at
-      // providing a caching document store for most operations.
-      maxSize: Math.pow(2, 20) * (approximateQueryPlanStoreMiB || 30),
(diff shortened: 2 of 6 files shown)


query-planner-js/src/config.ts
```diff
@@ -1,6 +1,20 @@
...
+   * If enabled, the `FetchNode.operationDocumentNode` field in query plan will be populated with the AST
+   * of the underlying operation (_on top_ of the "serialized" string `FetchNode.operation` which is always
+   * present). This can used by specific gateway user code that needs read-only access to such AST in
+   * order to save having to parse `FetchNode.operation`. Without this option, `FetchNode.operationDocumentNode`
+   * will always be `undefined`.
+   *
+   * Enabling this option will make query plans use more memory and you should consider increasing the
+   * query plan cache size (though `GatewayConfig.experimental_approximateQueryPlanStoreMiB`) if you enable it.
+   *
+   * Defaults to false (at least since 2.2; it temporarily defaulted to true before 2.2, but this resulted
```
clenfest, line 14: I would get rid of everything after "but". You covered that in the previous sentence.

gateway-js/src/index.ts
```diff
@@ -192,13 +192,15 @@ export class ApolloGateway implements GatewayInterface {
   }
 
   private initQueryPlanStore(approximateQueryPlanStoreMiB?: number) {
+    // Create ~about~ a 30MiB InMemoryLRUCache (or 50MiB if the full operation ASTs are
+    // enabled in query plans as this plan then use quite a bit more memory). This is
```
clenfest, line 196: Wording here is a little off. Maybe "as this requires plans to use"

gateway-js/src/index.ts
```diff
@@ -192,13 +192,15 @@ export class ApolloGateway implements GatewayInterface {
...
+    // less than precise since the technique to calculate the size of a DocumentNode is
+    // only using JSON.stringify on the DocumentNode (and thus doesn't account
+    // for unicode characters, etc.), but it should do a reasonable job at
+    // providing a caching document store for most operations.
+    const defaultSize = (this.config.queryPlannerConfig?.exposeDocumentNodeInFetchNode ?? false) ? 50 : 30;
```
clenfest, line 201: `?? false` isn't necessary here.

gateway-js/CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ This CHANGELOG pertains only to Apollo Federation packages in the 2.x range. The
 
 ## vNext
 
+- __BREAKING__: Disable exposing full document to sub-query by default (introduce in in 2.1.0):
```
clenfest, line 7:
```suggestion
- __BREAKING__: Disable exposing full document to sub-query by default (introduced in 2.1.0):
```

gateway-js/CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ This CHANGELOG pertains only to Apollo Federation packages in the 2.x range. The
...
+  - This change decrease memory consumption in general (which is the reason for disabling this by
```
clenfest, line 8:
```suggestion
  - This change decreases memory consumption in general (which is the reason for disabling this by
```

query-planner-js/CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ This CHANGELOG pertains only to Apollo Federation packages in the 2.x range. The
 
 ## vNext
 
+- __BREAKING__: Disable exposing full document to sub-query by default (introduce in in 2.1.0):
+  - This change decrease memory consumption in general (which is the reason for disabling this by
```
clenfest, line 8:
```suggestion
- __BREAKING__: Disable exposing full document to sub-query by default (introduced in 2.1.0):
  - This change decreases memory consumption in general (which is the reason for disabling this by
```

clenfest (approved):
A couple of minor comments, but approved otherwise.

PR #2231, opened 2022-11-02
Link Import should be nullable list with non-nullable elements

clenfest:


StephenBarlow:
@clenfest Lookin' at some older PRs with docs involved. Are these `@link` changes true _now_, or is this part of potential future change?

PR #2239, opened 2022-11-08
Fix potential QP issue with shareable root fields

pcmanus:
Slightly complex to explain, but on some example, during query planning, the query planner had to consider plans whose dependencies where essentially:
```
`Root Fetch 1`
                         \
                          `Dependent fetch`
                         /
`Root Fetch 2`
```
but the "processing" code wasn't expecting that and that triggered an assert (the patch has a unit test that, if run against pre-patch `main` triggers that assertion; note that the assertion is thrown during query planning, but the final query plan of that test does _not_ look like the one above). The patch fixes this issue.

### DIFF
--- a/query-planner-js/src/__tests__/buildPlan.test.ts
+++ b/query-planner-js/src/__tests__/buildPlan.test.ts
@@ -24,69 +24,154 @@ export function composeAndCreatePlannerWithOptions(services: ServiceDefinition[]
   ];
 }
 
-test('can use same root operation from multiple subgraphs in parallel', () => {
-  const subgraph1 = {
-    name: 'Subgraph1',
-    typeDefs: gql`
-      type Query {
-        me: User! @shareable
-      }
+describe('shareable root fields', () => {
+  test('can use same root operation from multiple subgraphs in parallel', () => {
+    const subgraph1 = {
+      name: 'Subgraph1',
+      typeDefs: gql`
+        type Query {
+          me: User! @shareable
+        }
 
-      type User @key(fields: "id") {
-        id: ID!
-        prop1: String
-      }
-    `
-  }
-
-  const subgraph2 = {
-    name: 'Subgraph2',
-    typeDefs: gql`
-      type Query {
-        me: User! @shareable
-      }
-
-      type User @key(fields: "id") {
-        id: ID!
-        prop2: String
-      }
-    `
-  }
-
-  const [api, queryPlanner] = composeAndCreatePlanner(subgraph1, subgraph2);
-  const operation = operationFromDocument(api, gql`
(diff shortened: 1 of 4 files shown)


query-planner-js/src/__tests__/buildPlan.test.ts
```diff
@@ -4285,3 +4370,55 @@ describe('__typename handling', () => {
     `);
   });
 });
+
+describe('mutations', () => {
```
pcmanus, line 4374: Note to reviewer: that test is not really directly related to this patch, but I notice that `buildPlan.test.ts` had no simple Mutation test. Mutation were still tested elsewhere (in the plan execution tests mostly, but also in plenty of older test files) but as I often use that one test file as a shortcut for "have I broken query planning", it's convenient to have at least a simple mutation test there.

query-planner-js/src/buildPlan.ts
```diff
@@ -195,23 +196,14 @@ const defaultCostFunction: CostFunction = {
...
-  )
 };
 
 function parallelCost(values: number[]): number {
   return sum(values);
 }
 
 function sequenceCost(stages: number[]): number {
-  return stages.reduceRight((acc, stage, idx) => (acc + ((idx + 1) * pipeliningCost * stage)), 0);
+  return stages.reduceRight((acc, stage, idx) => (acc + (Math.max(1, idx * pipeliningCost) * stage)), 0);
```
pcmanus, line 206: Note to reviewer: this change is due to the fact that the code now more or less rely on the fact that calling `processor.reduceSequence` on a single element is essentially a no-op, and that was true for the `reduceSequence` that generates the actual plan, but for not the costing of the plan here. Doesn't really change the overall logic otherwise.

clenfest (approved):

query-planner-js/src/buildPlan.ts
```diff
@@ -1770,37 +1760,42 @@ class FetchDependencyGraph {
...
-      inParallel: processor.reduceParallel(parallelNodes),
+      processed: processInParallel ? processor.reduceParallel(processedNodes) : processor.reduceSequence(processedNodes),
       next: toHandleNext,
       unhandled: remainingNext,
       deferredGroups: allDeferredGroups,
     };
   }
 
   private mergeRemainings(r1: UnhandledGroups, r2: UnhandledGroups): [FetchGroup[], UnhandledGroups] {
+    if (r1.length === 0 && r2.length === 0) {
```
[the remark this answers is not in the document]
pcmanus: Yeah, that was me prematurely optimising for no reason. I've removed it, it's unnecessary, sorry about the noise.

PR #2241, opened 2022-11-09
doc: remove constructor syntax from datasource example

dnalborczyk:


pcmanus (approved):
Sorry this was missed when you created this, but thanks a lot for noticing this/taking the time to provide a patch.

PR #2247, opened 2022-11-14
Loosen validation of @shareable on interface fields

pcmanus:
The `@shareable` directive was always intended to be only on object types. However, its location being `FIELD_DEFINITION`, it is syntactically allowed on interface fields, and due to an oversight of the original code, doing so had not been propertly rejected (but it had no effects whatsoever).

This oversight was noticed during #2175 and a validation was added to reject a `@shareable` on interface fields. Unfortunately, it appears a number of users had started putting such `@shareable` instances, so the added validation breaks composition for those users on upgrades (the "fix" for those user is trivial, just removing `@shareable` on any interface field, it did nothing and so will have no other impact).

However, sticking with the status quo is highly undesirable: continuing to allow `@shareable` on interface fields but having it do nothing whatsoever is guaranteed to create confusion, and can even be harmful if a user is falsly led to believe that it does something it does not.

So this patch explores a middle-ground alternative: it gives a meaning to `@shareable` on interface fields. That meaning is to force implementations to also mark the field as `@shareable`, as this feel like the most intuitive behaviour that can be done simply. Note that this is a middle ground in the sense that a subset of the users that had started using `@shareable` on interface fields may still see composition break on upgrades, if their implementation does not "manually inherit" those `@shareable` already. But experimentally, it appears that this patch avoids breaking for almost all the cases we have access to.

### DIFF
--- a/docs/source/federated-types/federated-directives.mdx
+++ b/docs/source/federated-types/federated-directives.mdx
@@ -210,7 +210,9 @@ type Position @shareable {
 }
 ```
 
-If a field is marked `@shareable` in _any_ subgraph, it **must** be marked as either `@shareable` or [`@external`](#external) in _every_ Federation 2 subgraph that defines it.
+If an object type field is marked `@shareable` in _any_ subgraph, it **must** be marked as either `@shareable` or [`@external`](#external) in _every_ Federation 2 subgraph that defines it.
+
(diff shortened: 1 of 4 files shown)


clenfest:
@pcmanus  This is the one that you said you were going to close, right?

pcmanus:
Yes, sorry. Decided to not go ahead with this.

PR #2257, opened 2022-11-18
Document that `@shareable` on interface field is not allowed

pcmanus:
`@shareable` was never meant to be applied to interface fields, but before, and since #2175, it is correctly rejected. Before that, it was syntactically accepted but was doing nothing at all, as the code was entirely ignoring. This patch document the behaviour and change.


clenfest:
LGTM

clenfest (approved):

PR #2259, opened 2022-11-18
Fix building additional abstract type edges

### LINKED ISSUE #361: Union with types from different service implementing value type interface
* Package name: **apollo-gateway** occurs on `master` and latest versions
* Here a commit with failing test:
  - https://github.com/mpospelov/apollo-server/commit/0a996c609afdca4f314db89ff89802892511d5ce
  - Given 2 schemas:
  ```gql
  type Query {
    content: [Content!]!
  }
  union Content = Video | Article
  type Article implements WebResource @key(fields: "id") {
    id: ID
    url: String
  }
  extend type Video @key(fields: "id") {
    id: ID @external
  }
  interface WebResource {
    url: String
  }
  ```
  ```gql
  interface WebResource {
    url: String
  }
  type Video implements WebResource @key(fields: "id") {
    id: ID
    url: String
  }
  ```
  - query via gateway with
  ```gql
  query {
    content {
      ... on WebResource {
        url
      }
    }
  }
  ```
  - the expected result is 
  ```js
  { 
    data:{
      content: [
        { url: 'https://foobar.com/articles/1' },
        { url: 'https://foobar.com/videos/1' },
      ]
    }
  }
  ```
  - the actual result is 
  ```js
  { 
    data:{
      content: [
        { url: 'https://foobar.com/articles/1' },
        { url: null },
      ]
    }
  }
  ```
  - Query plan:
  ```gql
   QueryPlan {
      Fetch(service: "contentService") {
        {
          content {
            __typename
            ... on WebResource {
              url
            }
          }
        }
      },
   }
  ```
  - it ignores querying the Video service for some reason, I believe that such cases should detect services which implements the interface to prevent this issue 🤔

* Runnable reproduction: 
  1. clone my branch https://github.com/mpospelov/apollo-server/tree/union-with-interfaces-issue
  2. run `jest -- packages/apollo-gateway/src/__tests__/integration/unions-with-value-interfaces.test.ts`

Thank you for all the hard work! 🔥🚀

pcmanus:
As we build the query graph, some edges are added between abstract types to avoid type explosion when we can. Unfortunately, in some cases, adding those edge was incorrect and we need to type-explode due to some specific different between subgraphs on the definition of the types involved. This fixes this issue.

Fixes #361

### DIFF
--- a/gateway-js/CHANGELOG.md
+++ b/gateway-js/CHANGELOG.md
@@ -4,6 +4,7 @@ This CHANGELOG pertains only to Apollo Federation packages in the 2.x range. The
 
 ## vNext
 
+- Fix issue where QP was generating invalid plan missing some data [#361](https://github.com/apollographql/federation/issues/361).
 - Fix QP not always type-exploding interface when necessary [PR #2246](https://github.com/apollographql/federation/pull/2246).
 - Fix potential QP issue with shareable root fields [PR #2239](https://github.com/apollographql/federation/pull/2239).
 - Correctly reject field names starting with `__` [PR #2237](https://github.com/apollographql/federation/pull/2237).
--- a/query-graphs-js/src/querygraph.ts
+++ b/query-graphs-js/src/querygraph.ts
@@ -1129,7 +1129,7 @@ class GraphBuilderFromSchema extends GraphBuilder {
    *   }
    * }
    * ```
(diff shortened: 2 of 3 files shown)


pcmanus:
This is the modified version of #2194 that I mentioned in [my comment](https://github.com/apollographql/federation/pull/2194#issuecomment-1318771850), but creating a separate PR as I can't easily push there.

PR #2274, opened 2022-11-22
Fix always expanding the federation spec to the last version

pcmanus:
The code adding directives for the federation spec was incorrectly match versions, so that subgraphs `@link`-ed against 2.0 and 2.1 were incorrectly expanded as if they were 2.2. This fix this issue and add tests to catch any regressions.


internals-js/src/coreSpec.ts
```diff
@@ -507,18 +507,16 @@ export class FeatureDefinitions<T extends FeatureDefinition = FeatureDefinition>
...
     return this;
   }
 
   /**
-   * Returns the known definition with the greatest version that satisfies the requested version, or undefined if no
-   * known version can satisfy this version.
+   * Returns the definition corresponding to the requested version if known.
    */
   find(requested: FeatureVersion): T | undefined {
-    return this._definitions.find(def => def.version.satisfies(requested));
```
pcmanus, line 521: Note that this is the main fix here: because 2.2 "satisfies" 2.1 and 2.0, it was picked up every time. But in hindsight, if the user manually specify a given version in the subgraph, no reason to ever give it anything else, so juste made it an equal.

Note that it's possible that some core feature processor would want a method like the one that existed, which gives you whatever most recent version is able to "read" the requested version, but we can add a separate method for that if we ever need it. What we want to now is "expanding" the definitions for a version, so an equal match makes more sense.

cpeacock (approved):

internals-js/src/federation.ts
```diff
@@ -641,52 +641,66 @@ export class FederationMetadata {
...
   }
 
   tagDirective(): DirectiveDefinition<{name: string}> {
-    return this.getFederationDirective(FederationDirectiveName.TAG);
+    return this.getLegacyFederationDirective(FederationDirectiveName.TAG);
   }
 
-  composeDirective(): DirectiveDefinition<{name: string}> {
-    return this.getFederationDirective(FederationDirectiveName.COMPOSE_DIRECTIVE);
+  composeDirective(): DirectiveDefinition<{name: string}> | FederationDirectiveNotDefinedInSchema<{name: string}> {
```
pcmanus, line 695: That's the bit of "magic" needed so that new releases of `internals-js` continue to work with older `@apollo/composition`, because `composeDirectiveManager` calls
```typescript
const composeDirectives = sg.metadata()
    .composeDirective()
    .applications();
```
and that would start failing if we made `FederationMetadata.composeDirective()` sometimes return `undefined`.

Note that I tried to make this as safe as possible through typing, and I think that's an ok solution, but returning `undefined` would probably be simpler overall, and we can change to that later if we fix-up our locking of versions between packages.

internals-js/src/coreSpec.ts
cpeacock: There's no further traversal that would need to happen and would just be an exact equality in all cases, then?

cpeacock (approved):

internals-js/src/coreSpec.ts
pcmanus: There is still a traversal since we keep the definition as a sorted list and not a map keyed by version (javascript is a bit wanky when it goes to make keyed by something else than an base type), but the match is on exactly equality in all cases so we'll never expand to a version that wasn't the exact one asked (worst case, the version is unknown and we error saying that, but that's kind of what we want anyway).

cpeacock: Awesome, thank you for the explanation.  Looks good to me.

PR #2284, opened 2022-11-28
Use IResolvers instead of GraphQLResolverMap

### LINKED ISSUE #2283: Drop GraphQLResolverMap in favor of IResolvers for buildSubgraphSchema
Lots of customers use the `graphql-tools` definition `IResolvers` to specify resolvers rather than the apollo `GraphQLResolverMap` and would like to be able to pass that into `buildSubgraphSchema`. Since one definition is a subset of the other, it should be fairly easy to switch and it won't cause any incompatibilities.

clenfest:
Switch buildSubgraphSchema to use IResolvers rather than GraphQLResolverMap. Since IResolvers is a superset, this is a backward compatible change. 

Fixes #2283

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,7 +14,8 @@
         "@apollo/query-graphs": "file:query-graphs-js",
         "@apollo/query-planner": "file:query-planner-js",
         "@apollo/subgraph": "file:subgraph-js",
-        "apollo-federation-integration-testsuite": "file:federation-integration-testsuite-js"
+        "apollo-federation-integration-testsuite": "file:federation-integration-testsuite-js",
+        "graphql-tools": "^8.3.12"
       },
       "devDependencies": {
         "@apollo/cache-control-types": "1.0.2",
@@ -307,7 +308,7 @@
     },
     "node_modules/@apollo/client": {
       "version": "3.6.9",
-      "dev": true,
+      "devOptional": true,
       "license": "MIT",
       "dependencies": {
         "@graphql-typed-document-node/core": "^3.1.1",
@@ -2575,7 +2576,7 @@
     },
     "node_modules/@graphql-typed-document-node/core": {
       "version": "3.1.1",
-      "dev": true,
+      "devOptional": true,
       "license": "MIT",
       "peerDependencies": {
         "graphql": "^0.8.0 || ^0.9.0 || ^0.10.0 || ^0.11.0 || ^0.12.0 || ^0.13.0 || ^14.0.0 || ^15.0.0 || ^16.0.0"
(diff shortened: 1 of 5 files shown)


clenfest:
Note to reviewers. We could in theory keep both definitions (i.e. `resolvers?: IResolvers | GraphQLResolverMap<any>`, but I don't necessarily see the benefit.

pcmanus (changes requested):
I'm actually quite unsure about this.

`IResolvers` is indeed a superset, and while it does mean existing code shouldn't break, it also means that there a number of new things that a `IResolvers` may be, so by making `buildSubgraphSchema` accept `IResolvers`, we're pretending that we accept those new things. Except that that's not what the patch does: the is no change to `addResolversToSchema` to handle any of the new things that `IResolvers` brings.

I'll note that I haven't looked too closely to `IResolvers` (which is not a super simple definition) so I'm not sure how big of a deal it is in practice, but the difference appears substantial enough from a quick look that I'm not confident ignoring it without substantial justification.

And in general, adopting a type we don't control as argument should be done a bit carefully, because that also more or less bind us to handle any future additions made to that type.

Don't get me wrong, I do think we should consider moving toward `IResolvers` and what `graphqp-tools` provides. But if we accept `IResolvers`, then I think that imply we also need to start using `graph-tools` own [`addResolversToSchema`](https://the-guild.dev/graphql/tools/docs/resolvers#addresolverstoschema-schema-resolvers-resolvervalidationoptions-inheritresolversfrominterfaces-). I'm not sure what doing so entails exactly: our own `addResolversToSchema` relies on setting a few custom "extensions" for `__resolveReference`/`__resolveType`, but I suspect this could be done easily enough as a separate tasks from the rest of the work of setting resolvers. But this require careful looking to check if what the `graphql-tools` version does is indeed compatible with what our own version currently does.

darrellwarde:
Just want to chip in on the business case side of things here as the person (or one of) who flagged up this issue. I work on the GraphQL Library at Neo4j, a product which accepts simple type definitions representing entities as input, and outputs a schema definition for a fully-fledged GraphQL API which you could build an application against.

We are of the view that we shouldn't be opinionated in what we provide to our users, so we return a `DocumentNode` for type definitions and `IResolvers` for resolvers, as these are the industry adopted standards for these entities. Our users can then plug these into Apollo Server, but also have the freedom to use a different GraphQL server such as Yoga if they so wish. Due to the type signature of `buildSubgraphServer`, if they are using TypeScript then they will get compilation errors.

In my opinion, the current API for `buildSubgraphSchema` is very geared towards users who are directly defining their type definitions and resolvers, and presents a challenge to library authors who are looking to produce subgraph compatible schema definitions, which will slow adoption amongst these products.

Totally appreciate (and empathise with the challenges of) everything in your comment, just wanted to provide a bit of context! One issue I imagine you will run into is that `IResolvers` accepts `subscribe` functions for Subscriptions, which I believe has no impact in the world of Federation. Could it be an acceptable starting point to parse incoming resolvers and `console.warn` of any entries which will be ignored by `buildSubgraphSchema`?

Issue #2256, opened 2022-11-16
Union types are not rebuilt properly from the supergraph, creating potential runtime issues


pcmanus:
Union types are merged by composition, but we don't preserve information about which subgraph declares which types of the union in the supergraph (very much an oversight). As the result, the query planner does not know which subgraph declares which type as part of the union, and that can lead to send invalid queries to a subgraph (which are thus rejected at runtime).

Consider the following example:
```graphql
# subgraph `S1`
type Query {
  u: U
}

union U = A | B

type A @key(fields: "id") {
  id: ID!
  a1: Int
}

type B {
  id: ID!
  b: Int
}

type C  @key(fields: "id") {
  id: ID!
  c1: Int
}
```
```graphql
# subgraph `S2` 
type Query {
  otherQuery: U
}

union U = A | C

type A @key(fields: "id") {
  id: ID!
  a2: Int
}

type C @key(fields: "id") {
  id: ID!
  c2: Int
}
```

The definition of interface `U` in the supergraph ends up as:
```graphql
union U = A | B | C
```
and so the following query is valid (on the supergraph API):
```graphql
query {
  u {
    ... on C {
      c1
    }
  }
}
```
However, the query planner currently generate the following fetch:
```
Fetch(service: "S1") {
  {
    u {
      __typename
     ... on C {
       c1
     }
    }
  }
},
```
which is *not* valid GraphQL because for `S1`, union `U` does not contains `C` and so the subgraph would reject the query with an error looking like:
```
Fragment cannot be spread here as objects of type "U" can never be of type "C".
```

But the underlying reason this happens is that the composed supergraph does not preserve the information regarding which subgraphs defined which types for union `U`. As a result, the "view" of the subgraph that the query planner has is that both `S1` and `S2` declares `U` as `A | B | C`, which is incorrect.

pcmanus:
Note that fixing this require a bump of the underlying "join spec" used by supergraphs (which is ok, that's why we have versioning in place, but does mean that the supergraph generated by whichever release this goes in won't be readable by older versions).

Issue #2250, opened 2022-11-15
Supergraph Schema is not currently lossless


gwardwell:
# Description

Currently, when schema is composed into a supergraph schema, some of the schema can be omitted when the schema is merged via intersection, such as with Input Value Definitions or input enum values. This makes the supergraph schema lossy rather than lossless, and does not make it a sufficient source of truth for the current state of the supergraph.

# Example

Say I have two supergraphs with the following schema, which I will compose into a supergraph schema (currently using rover and federation v2.1):

```gql
# Subgraph A
type Query {
  booksByGenre(genre: BookGenreEnum!, address: Address): [String!]! @shareable
}

input Address {
  street: String!
  city: String
  state: String
  postalCode: Int!
}

enum BookGenreEnum {
  FICTION
  NON_FICTION
}

type Book @shareable {
  title: String!
  genre: String!
}
```

```gql
# Subgraph B
type Query {
  booksByGenre(genre: BookGenreEnum!): [String!]! @shareable
}

input Address {
  street: String!
  postalCode: Int!
}

enum BookGenreEnum {
  FICTION
}

type Book @shareable {
  title: String!
}
```

# Encountered behavior

The composed supergraph schema looks like this:

```gql
# Supergraph Schema
type Query
  @join__type(graph: SUBGRAPH_A)
  @join__type(graph: SUBGRAPH_B)
{
  booksByGenre(genre: BookGenreEnum!): [String!]! @join__field(graph: SUBGRAPH_A) @join__field(graph: SUBGRAPH_B)
}

input Address
  @join__type(graph: SUBGRAPH_A)
  @join__type(graph: SUBGRAPH_B)
{
  street: String!
  postalCode: Int!
}

enum BookGenreEnum
  @join__type(graph: SUBGRAPH_A)
  @join__type(graph: SUBGRAPH_B)
{
  FICTION
}

type Book
  @join__type(graph: SUBGRAPH_A)
  @join__type(graph: SUBGRAPH_B)
{
  title: String!
  genre: String! @join__field(graph: SUBGRAPH_A)
}
```

Note that `Query.booksByGenre.address`, `Address.city`, `Address.state`, and `BookGenreEnum.NON_FICTION` are all omitted from the supergraph schema.

# Expected behavior

[...]

PR #2288, opened 2022-11-30
Preserves source of union members and enum values in supergraph

pcmanus:
While composition allows unions and enums to differ between subgraphs (at least under certain conditions), the supergraph currently was not preserving which subgraph defined which union member/enum value, and in particular the query planner made the (sometimes incorrect) assumption that unions and enums were defined the same in all the subgraphs in which they are defined.

As shown in #2256, this was leading to genuine issues in the case of unions, where the query planner was generating invalid subgraph queries. I am not sure this created similar issues for enum values due to a combination of how the merging rule for enum work and the fact that values of enum are not types, but it is nonetheless a bad idea to run the query planner with incorrect information on the subgraph it generates queries for, and this can get in the way of other toolings that wants to use the supergraph (for instance, this gets in the way of #2250).

To fix this, this patch ensures the supergraph preserve the information regarding which subgraph defines which union member and enum values in the supergraph by adding 2 new dedicated "join spec" directives.

### DIFF
--- a/composition-js/CHANGELOG.md
+++ b/composition-js/CHANGELOG.md
@@ -4,6 +4,8 @@ This CHANGELOG pertains only to Apollo Federation packages in the 2.x range. The
 
 ## vNext
 
+- Preserves source of union members and enum values in supergraph [PR #2288](https://github.com/apollographql/federation/pull/2288).
+
 ## 2.2.0
 
 - __BREAKING__: composition now rejects `@shareable` on interface fields. The `@shareable` directive is about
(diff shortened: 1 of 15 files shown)


pcmanus:
I'll note that this PR is currently based on top of the #2279 PR. This is because that later PR was already adding the code to introduce a new `0.3` join spec, which this patch needs too. I can try to separated the aforementioned code so this can be merge first if we really really want, but as I do think we should make an effort in getting those 2 PRs in the same release, exactly because they both need new things from the join spec (and, all other things being equal, it's nice to avoid burning new versions too often), so maybe we just review #2279 first and then get to this.

clenfest (approved):

PR #2293, opened 2022-12-07
refactor: remove WHATWG URL import, use global

dnalborczyk:


trevor-scheer (approved):

clenfest:
Unfortunately, we had to revert this PR because it didn't play nicely with `deno` (used by the router). I can reopen it, but I'm not sure what will need to change on the deno side to allow us to merge it.

Issue #2277, opened 2022-11-24
Interfaces as abstractions across Subgraphs (`@interfaceObject` and `@key` on interfaces)


pcmanus:
## Tl;dr

This issues proposes 2 new related additions:
1. a new `@interfaceObject` directive that can be applied to an object type in a subgraph as way to "abstract" what is otherwise an interface in other subgraphs.
2. support for `@key` on interface types, in order to resolve values generated by subgraph using the `@interfaceObject` feature of the previous point where needed.

The combination of those 2 features will allow a subgraph to "use" an interface transparently, without needing to know anything about the implementations of that interface (which will be provided by another subgraph).

But for more details, read on.

## Motivation

Interfaces exists to abstract, they abstract various implementations. In federation, where one the goal is to allow subgraphs to develop in relative isolation, it feels natural to want to use interface as abstractions across subgraphs. By which we mean that a subgraph should be able to use an interface as a black-box abstraction, without needing any knowledge of its implementation types.

But this is not really possible at the moment, and this issue proposes a way to achieve such abstraction.

We'll note at the offset that the proposal of this issue is meant to be a useful starting point, but that this initial proposal does not solve every interface limitations one may have with federation. We will discuss (at the end of this description) a number of possible follow-ups and show that can naturally be built on the concept introduced here.

## Running example

Let introduce a simple example to illustrate the problem first, and introduce the proposed solution next.

The example is that of an extremely simplified only shop, and consider 2 subgraphs:
1. a `products` subgraph, that defines a generic `Product` interface, and defines a number of concrete `Product` implementations (`Book`, `Movie`, etc...)
2. a `reviews` subgraph, that allows users to leave reviews on any product.

And in that example, we'll assume that reviews are generic on the actual product being reviews, meaning that we ideally want the `Review` subgraph to know the `Product` interface, but not its implementation types. Typically, if a new `Product` implementation type is added, we'd like to not have to modify the `reviews` subgraph.

Unfortunately, this is not possible today. The best one can do at the moment looks something like this:
```graphql
# Subgraph `products`
type Query {
  products: [Product!]!
}

interface Product {
  id: ID!
  description: String
  price: Float
}

type Book implements Product @key(fields: "id") {
  id: ID!
  description: String
  price: Float
  pages: Int
}

type Movie implements Product @key(fields: "id") {
  id: ID!
  description: String
  price: Float
  duration: Int
}

# ... keeping the example short, but potentially more implementations of Product ...
```
```graphql
[...]

PR #2294, opened 2022-12-08
Use alias in QP when querying conflicting fields

### LINKED ISSUE #1257: Interface fragment expansion causes conflicting field types

pcmanus:
In a few situations, the query planner was generating queries where the same response name was queried at the same "level" with incompatible types, resulting in invalid queries (the queries were failing the [`FieldsInSetCanMerge`](https://spec.graphql.org/draft/#FieldsInSetCanMerge()))validation for the GraphQL spec).

This commit detects this case, and when it would happen, aliases oneof the occurence in the fetch to make the query valid. Once receivingthe fetch result, the aliased value is rewritten to it's originalresponse name.

Fixes #1257


pcmanus:
Note that the patch is currently on top of the patch for #2277. That's because the "post-fetch" rewriting done here piggy-back on the query plan changes made by the `@interfaceObject` patch, which also does some fetch rewriting (pre-fetch in that case, but still fairly similar). But given that this is a slightly involved fix, I think it's worth waiting on 2.3 for this anyway, not in a 2.2 patch release.

Note that problem/solution still very much follow what I describe in [this comment](https://github.com/apollographql/federation/issues/1257#issuecomment-1161875201). The one question I asked there was if we wanted to only add aliases when strictly necessary, and I answer this positively here, not only for the reason I mentioned in the comment (this avoid modifying most existing plan) but also because it is likely more efficient since adding aliases more often would 1) result in slightly larger fetch queries in general and 2) would imply post-fetch rewritings much more often, and while it's super expensive, it has a cost nonetheless.

gateway-js/src/__tests__/executeQueryPlan.test.ts
```diff
@@ -3516,4 +3516,1113 @@ describe('executeQueryPlan', () => {
         `);
     });
   });
```
clenfest, line 3518: You talk in the comments about potentially having a naming collision when it comes to aliasing and potentially needing to regenerate the name? Is there a test for that case?

pcmanus: Good call. I had not added a test for that, so pushed one (which obviously uncovered a small issue in some cases).

clenfest (approved):

PR #2297, opened 2022-12-12
Adds test that validates that older supergraphs can be read by new ve…

pcmanus:
…rsions

Because newer versions can't generate old subgraphs, this commit creates a new scripts that allow to easily generate a supergraph for the "current" version. The intent is that with each new version, we'll generate the corresponding supergraph and commit it to source control. As such, any following version will have older versions of the supergraph which it can test the reading of.

This is imperfect as this only test a relatively specific supergraph, and making change to it would be painful since we've have to backport the change to older versions and re-generate new versions of the supergraph (possible, but painful). But it is better than nothing as it will quickly catch any obvious breaking in the reading of older supergraphs.


pcmanus:
Going to mark this draft to convey the fact that I'm not entirely convinced this is a good approach.

But essentially, the problem we have is that newer versions of the query planner should be able to read and use older supergraphs, but we don't really have any test for that. Particularly as we're about to bump the join spec version to 0.3 in the next release, having tests that ensure we don't completely break when reading 0.2 versions would be nice.

But setting up that kind of tests is a bit painful as we need to keep around a bunch of older supergraphs that newer versions can test against and:
- as we move forward and add more and more versions, it would be nice to avoid having the process of generating those supergraphs to keep around and test against not too manual, or it'll fall by the wayside.
- it makes it very hard to change the tests themselves, and doing so would easily mean re-generating the supergraph for old the versions we want to test against, which isn't trivial.

In any case, this PR is some attempt at having "something". It tries to tackle the problem of the 1st bullet point above: it makes it easy to generate a test supergraph from the current version and to keep it around. In theory, if we don't forget to generate a new test supergraph on new versions (this part is still manual), we'd accumulate older supergraphs that never versions can test against (which is why this patch is against `main` btw; the idea is that we'd generate a supergraph against the 0.2 join spec and commit it before merging `main` into `next`, at which point we'll have a 0.2 supergraph to test against the newer code).

Overall, I'm on the fence on whether it's worth having something like this in-tree. Maybe this kind of backward compat is better left entirely to our offline harness.

query-planner-js/src/__tests__/supergraphBackwardCompatibility.test.ts
```diff
@@ -0,0 +1,336 @@
...
+    }
+
+    enum Currency {
+      USD
+      EUR
+    }
+  `
+}
+
+const reviews = {
```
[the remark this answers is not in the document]
clenfest: Oh I see you address it in the comment below. Again, it's fine if you don't think it's worth the hassle.

query-planner-js/src/__tests__/genTestSupergraph.ts
```diff
@@ -0,0 +1,22 @@
...
+// This is called through ts-node so argv[0] is ts-node itself, argv[1] is this script file name, and argv[2] is thus our argument.
+const version = process.argv[2];
+if (!version) {
+  console.error('Missing mandatory version argument.');
+  process.exit(1);
+}
+
+// Ensure we're not given a completely broken argument. This also reject any `-alpha` or `-rc` versions on purpose: no
+// point in keeping test supergraphs for those versions around.
+if (!(/[0-9]+\.[0-9]+(\.[0-9]+)?/.test(version))) {
```
clenfest, line 17: We don't actually check that the file we are trying to generate is the branch that we're currently on. Probably nbd, but I thought I'd call it out as a potential footgun.

pcmanus: Yeah; I wasn't actually sure how to do something useful here without being complicated. In theory, I wanted to use the current branch version instead of asking for a version manually, but as far as i can tell, we only bump versions just before releases, so the version of the "current branch" is usually the version of "whatever was the last release", and so it's not exactly the version we'd want for this.
I figured passing the version manually might end up less frustrating that something that tries to be automatic but does the wrong thing, but I'm open to suggestions (maybe someday we get this integrated to the release process...). Anyway, I'll note that we still need to manually commit the result of this script, which give more change to double check for mistakes.

PR #2299, opened 2022-12-13
Fix possible assertion error during query planning

pcmanus:
The code for handling @requires was, in rare cases, trying to add some selections to a subgraph that couldn't handle them. This fixes that issue by ensuring we only try the merging in question if the subgraph can handle the selection.

### DIFF
--- a/query-planner-js/src/buildPlan.ts
+++ b/query-planner-js/src/buildPlan.ts
@@ -669,7 +669,7 @@ class FetchGroup {
     readonly rootKind: SchemaRootKind,
     readonly parentType: CompositeType,
     readonly isEntityFetch: boolean,
-    private readonly _selection: LazySelectionSet,
+    private _selection: LazySelectionSet,
     private _inputs?: LazySelectionSet,
     readonly mergeAt?: ResponsePath,
     readonly deferRef?: string,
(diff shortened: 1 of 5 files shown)


query-planner-js/src/buildPlan.ts
```diff
@@ -3351,23 +3377,19 @@ function handleRequires(
...
+    // directly fetched from the parent. If it can, then we can just merge `newGroup` into that parent.
+    // Otherwise, we will have to "keep it".
     // Note: it is to be sure this test is not poluted by other things in `group` that we created `newGroup`.
-    // Note2: `__typename` selections adds a bit of complexity. That is, if `newGroup` selection is not
-    // strictly contained in its inputs but only due to the selection of some `__typename`, then we
-    // still want to ignore that group, because `__typename` are always trivially queriable from any
-    // type in any subgraph and so that `__typename` can always be fetched from the parent. Which is
-    // what the `newGroupIsUnneeded` check ignores `__typename` in the selection.
-    const newGroupIsUnneeded = newGroup.inputs!.contains(withoutTypename(newGroup.selection)) && parent.path;
+    newGroup.removeInputsFromSelection();
```
pcmanus, line 3391: Note that this is the core change here. The rest of the diff is mostly about supporting this change.

And the problem was that we could have cases where `newGroupIsUnneeded` was true, but some of selections of `newGroup` are not valid selections for its parent (despite the selections of `newGroup` being essentially its inputs, there is cases where the direct parent is not the group that has fetched those inputs). So when we tried to call `parent.group.mergeChildIn(newGroup)` below, this was actually throwing an assertion, and for good reason.

The change here make it so that we only try to merge `newGroup` to its parent if we can guarantee that the merge will not throw.

PR #2301, opened 2022-12-13
Add support for tag v0.3

matthew-gordon:
This PR introduces tag v0.3, which adds support for @tag in an additional location (schema).

### DIFF
--- a/internals-js/src/federationSpec.ts
+++ b/internals-js/src/federationSpec.ts
@@ -195,6 +195,7 @@ export class FederationSpecDefinition extends FeatureDefinition {
 export const FEDERATION_VERSIONS = new FeatureDefinitions<FederationSpecDefinition>(federationIdentity)
   .add(new FederationSpecDefinition(new FeatureVersion(2, 0)))
   .add(new FederationSpecDefinition(new FeatureVersion(2, 1)))
-  .add(new FederationSpecDefinition(new FeatureVersion(2, 2)));
+  .add(new FederationSpecDefinition(new FeatureVersion(2, 2)))
+  .add(new FederationSpecDefinition(new FeatureVersion(2, 3)));
 
 registerKnownFeature(FEDERATION_VERSIONS);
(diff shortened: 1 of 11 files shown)


internals-js/src/tagSpec.ts
```diff
@@ -24,14 +24,15 @@ export class TagSpecDefinition extends FeatureDefinition {
     this.printedTagDefinition = 'directive @tag(name: String!) repeatable on FIELD_DEFINITION | INTERFACE | OBJECT | UNION';
     if (!this.isV01()) {
       this.tagLocations.push(
+        DirectiveLocation.SCHEMA,
```
sachindshinde, line 27: This will cause the definition to change for `0.2` in addition to `0.3`.

Instead, you'll want to:
- Define a function called `isV02()` on this class, similar to `isV01()`.
- Inside this `if` block, at the end of it, add another `if` block checking `!this.isV02()`.
- Inside that `if` block, call `this.tagLocations.push()` with `DirectiveLocation.SCHEMA`.
- After that `push()` call, set `this.printedTagDefinition` to the one you have below, with the additional `SCHEMA` location.

internals-js/src/federationSpec.ts
```diff
@@ -195,6 +195,7 @@ export class FederationSpecDefinition extends FeatureDefinition {
 export const FEDERATION_VERSIONS = new FeatureDefinitions<FederationSpecDefinition>(federationIdentity)
   .add(new FederationSpecDefinition(new FeatureVersion(2, 0)))
   .add(new FederationSpecDefinition(new FeatureVersion(2, 1)))
-  .add(new FederationSpecDefinition(new FeatureVersion(2, 2)));
+  .add(new FederationSpecDefinition(new FeatureVersion(2, 2)))
+  .add(new FederationSpecDefinition(new FeatureVersion(2, 3)));
```
sachindshinde, line 199: You may need to rebase your branch on top of whatever branch adds in `2.3` support (presumably one of them does it, not sure which though).

sachindshinde (changes requested):

internals-js/src/federationSpec.ts
sachindshinde: Found it, it's the branch in this PR https://github.com/apollographql/federation/pull/2294

Note that it's a branch off of Sylvain's forked repo, so you'l need to add his repo as an upstream so you can pull his branch in.

pcmanus: The 2.3 spec was actually added in #2279, but that has been merged now. That said, this is all on the `next` branch, not the `main` one, so that's where this should be rebased.

PR #2303, opened 2022-12-16
Various fixes around `@interfaceObject`

pcmanus:
There were a few things that didn't worked as expected or weren't validated correctly in the handling of `@interfaceObject` and this commit fixes those issues. The issues concretely fixed are:
- Proper validation of shareability for fields provided by @interfaceObject, including cases where multiple @interfaceObject "abstract" the same field.
- If `@interfaceObject` is used for some interface `I`, reject if the subgraph also defines some of the implementations of `I` (this may be something that we lift later, but since the `@interfaceObject` abstract the type in question, we should take some time to think about the consequence if we do lift that limitation).
- Rejects `@override` on both `@interfaceObject` fields and object type fields which are otherwise provided by an @interfaceObject in another subgraph. Here again, we can probably lift such limitation later, but we need to think that through.
- Fix handling of external fields on implementation type when those field are provided by an `@interfaceObject`. This was preventing using such fields in `@requires` for instance.
- Rejects `@override` when used on interface fields. This is actually not strictly `@interfaceObject` related, but simply noticed that we were not properly rejecting it (much like we didn't reject `@shareable` on interface fields until a recent fix). The code definitively does not handle `@override` on an interface field (and like for `@shareable`, it's a tad unclear what it would mean really), and accepting it while ignoring it is just confusing.

### DIFF
--- a/docs/source/errors.md
+++ b/docs/source/errors.md
@@ -65,7 +65,7 @@ The following errors might be raised during composition:
 | `ONLY_INACCESSIBLE_CHILDREN` | A type visible in the API schema has only @inaccessible children. | 2.0.0 |  |
 | `OVERRIDE_COLLISION_WITH_ANOTHER_DIRECTIVE` | The @override directive cannot be used on external fields, nor to override fields with either @external, @provides, or @requires. | 2.0.0 |  |
 | `OVERRIDE_FROM_SELF_ERROR` | Field with `@override` directive has "from" location that references its own subgraph. | 2.0.0 |  |
-| `OVERRIDE_ON_INTERFACE` | The @override directive cannot be used on the fields of an interface type. | 2.0.0 |  |
+| `OVERRIDE_ON_INTERFACE` | The @override directive cannot be used on the fields of an interface type. | 2.3.0 |  |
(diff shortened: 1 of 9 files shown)


docs/source/errors.md
```diff
@@ -65,6 +65,7 @@ The following errors might be raised during composition:
 | `ONLY_INACCESSIBLE_CHILDREN` | A type visible in the API schema has only @inaccessible children. | 2.0.0 |  |
 | `OVERRIDE_COLLISION_WITH_ANOTHER_DIRECTIVE` | The @override directive cannot be used on external fields, nor to override fields with either @external, @provides, or @requires. | 2.0.0 |  |
 | `OVERRIDE_FROM_SELF_ERROR` | Field with `@override` directive has "from" location that references its own subgraph. | 2.0.0 |  |
+| `OVERRIDE_ON_INTERFACE` | The @override directive cannot be used on the fields of an interface type. | 2.0.0 |  |
```
clenfest, line 68: Is this being added back? If not, probably should be 2.3.0

clenfest (approved):

docs/source/errors.md
pcmanus: No, good catch, it should be `2.3.0`. Fixed.

Issue #2309, opened 2023-01-03
Unwarranted large QPcache sizes because of not hashing the operation name for queryPlanStoreKey


AneethAnand:
## Description

We are using the Federation Gateway in production and recently added the operation name to the request along with the query. After the change, we noticed an increase in the time taken to generate the query plan and all of the cache misses. Our analysis showed that the issue was that the operation name is not hashed for the query plan store key.

```javascript
const queryPlanStoreKey = queryHash + (request.operationName || '');
```

https://github.com/apollographql/federation/blob/main/gateway-js/src/index.ts#L758

## Impact
The increased cache size taken to store the query plan and cache misses are impacting the performance of our application.

## Possible Solution
We could hash the operation name along with the query hash to properly incorporate the operation name into the key. 
By hashing both the query and operation name saves more space/ cache size.

For example:
```javascript
          const queryPlanKey = print(document) + (request.operationName || '');
          const queryPlanStoreKey = computeQueryPlanKeyHash(queryPlanKey);
```

This would generate a unique hash for the query and operation name, which could be used as the query plan store key.

AneethAnand:
Created a PR:
 https://github.com/apollographql/federation/pull/2310

korinne:
Thanks for raising this issue and submitting a pull request! Our team will review and provide feedback shortly.

trevor-scheer:
Hey @AneethAnand, can you help me understand the problem a bit better or provide a reproduction of the issue you're seeing? Based on your explanation I feel like I'm missing something. Here's my understanding:

TL;DR, the only effect I see in this PR is changing the cache key from one unique ID to a slightly different one.

For simplicity I'll assume your documents contain only one operation. This is how I imagine the scenario you've described.
1. You send an operation with no `operationName`. A query plan is cached with some key: `abc123` (hash only).
2. You send the same operation with `operationName`. A duplicate query plan is cached with a new key: `abc123MyOperation` (hash + `operationName`).

For a long-running gateway, this results in 1 cache miss and 2 cache entries for the same query plan. But the next time (2) happens you shouldn't see another new cache entry or a cache miss. And once your gateway is restarted or your cache starts evicting old plans I wouldn't expect duplicate cache entries.

All of what I've said above would be expected whether or not the key included the `operationName` in the hash, but those are the consequences I would anticipate going from no `operationName` to including it.

> we noticed an increase in the time taken to generate the query plan
[...]

PR #2310, opened 2023-01-03
Make queryPlanStoreKey a hash of both the query and Operation name

AneethAnand:
…he query and Operation name

Fixes https://github.com/apollographql/federation/issues/2309


korinne:
Thanks for submitting a pull request! [Linking this to the original issue](https://github.com/apollographql/federation/issues/2309), so that we can review more thoroughly.

gateway-js/src/index.ts
```diff
@@ -754,8 +759,9 @@ export class ApolloGateway implements GatewayInterface {
       { attributes: spanAttributes },
       async (span) => {
         try {
-          const { request, document, queryHash } = requestContext;
-          const queryPlanStoreKey = queryHash + (request.operationName || '');
+          const { request, document } = requestContext;
+          const queryPlanKey = print(document) + (request.operationName || '');
```
pcmanus, line 763: Printing a document is not super cheap, especially for large queries, so it's worth avoiding if we can. In this case, I believe the query string is available in `requestContext.request.query` so that can be used.

But as computing sha256 is not extremely cheap either and we already have the hash of the query, why not re-hash that with the operation name. Something like (happy to extract in an helper method in practice, just keeping it short here):
```typescript
const queryPlanStoreKey = request.operationName
  ? createHash('sha256').update(queryHash).update(request.operationName).digest('hex')
  : queryHash;
```

trevor-scheer: Agreed, though maybe just hash the concatenation `queryHash + request.operationName`?

pcmanus: Admittedly, I don't think this make much practical difference, but concatenating necessarily means creating a new underlying string and some memory copies, while calling `update` twice most likely doesn't. And calling `update` repeatedly is pretty idiomatic too for a hash implementation as those are usually created with streaming in mind (and [node JS seems no exception](https://nodejs.org/api/crypto.html#hashupdatedata-inputencoding)).

trevor-scheer:
@AneethAnand can you please implement the change that Sylvain requested and add a patch changeset please?

AneethAnand:
> @AneethAnand can you please implement the change that Sylvain requested and add a patch changeset?

Incorporated your review comments!  Thanks again for your inputs

PR #2311, opened 2023-01-03
Add notes about N+1 and data loaders around entities

dbanty:
A first attempt at exposing users to data loaders earlier and more often.


docs/source/entities-advanced.mdx
```diff
@@ -575,3 +575,42 @@ Here we're using two directives in combination: `@provides` and `@external`.
...
+  Product: {
+    __resolveReference(productRepresentation) {
+      return fetchProductByID(productRepresentation.id);
+    }
+  },
+  // ...other resolvers...
+}
+```
+
+Imagine a naive implementation of `fetchProductByID` that makes a database call for each product ID. If we have `N` reviews, we'll make `N` database calls. These calls are _in addition_ to the 1 call made by the Reviews subgraph to fetch the initial list of reviews (and the `id` of each product). This is where the "N+1" problem gets its name. If not prevented, this problem will cause performance problems and can even enable "Denial of Service" attacks.
```
StephenBarlow, line 614:
```suggestion
A basic implementation of the `fetchProductByID` function might make a database call each time it's called. If we need to resolve `Product.price` for `N` different products, this results in `N` database calls. These calls are made _in addition to_ the call made by the Reviews subgraph to fetch the initial list of reviews (and the `id` of each product). This is where the "N+1" problem gets its name. If not prevented, this problem can cause performance problems and can even enable denial-of-service attacks.
```

docs/source/entities-advanced.mdx
```diff
@@ -575,3 +575,42 @@ Here we're using two directives in combination: `@provides` and `@external`.
...
+
+This is not only a problem with entity resolvers, it's a common problem with any GraphQL resolver—and it has a common solution: [the dataloader pattern](https://github.com/graphql/dataloader). Nearly every GraphQL server library has a dataloader implementation, and you should use it in **every resolver**. Even resolvers which are not for entities and do not return lists can _still_ cause N+1 issues via [batched requests](https://www.apollographql.com/docs/enterprise-guide/graph-security/#batched-requests).
```
StephenBarlow, line 616:
```suggestion
This problem is not limited to reference resolvers! In fact, it can occur with any resolver that fetches from a data store. To handle this problem, we strongly recommend using [the dataloader pattern](https://github.com/graphql/dataloader). Nearly every GraphQL server library provides a dataloader implementation, and you should use it in **every resolver**. This is true even for resolvers that _aren't_ for entities and that _don't_ return a list. These resolvers can _still_ cause N+1 issues via [batched requests](/enterprise-guide/graph-security/#batched-requests).
```

docs/source/entities.mdx
```diff
@@ -93,6 +93,8 @@ Let's break this example down:
 * A reference resolver is responsible for returning all of the entity fields that this subgraph defines.
     * In this example, the hypothetical `fetchProductByID` function fetches a particular `Product`'s field data based on its `id`.
 
+> ⚠️ **Reference resolvers _will_ be called more than once per query. It is crucial that they account for N+1 issues (e.g., via [data loaders](https://github.com/graphql/dataloader)). See [The N+1 problem](../entities-advanced#the-n1-problem) for more details.**
```
StephenBarlow, line 96:
```suggestion
> ⚠️ **A particular reference resolver might be called _many_ times to resolve a single query!** It is crucial that reference resolvers account for "N+1" issues (e.g., via [data loaders](https://github.com/graphql/dataloader)). For details, see [Handling the N+1 problem](../entities-advanced#handling-the-n1-problem).
```

StephenBarlow (approved):
Approved w/ suggestions!

dbanty:
Feedback incorporated, thanks @StephenBarlow!

docs/source/entities.mdx
```diff
@@ -93,6 +93,8 @@ Let's break this example down:
 * A reference resolver is responsible for returning all of the entity fields that this subgraph defines.
     * In this example, the hypothetical `fetchProductByID` function fetches a particular `Product`'s field data based on its `id`.
 
+> ⚠️ **A particular reference resolver might be called _many_ times to resolve a single query!** It is crucial that reference resolvers account for "N+1" issues (e.g., via [data loaders](https://github.com/graphql/dataloader)). For details, see [Handling the N+1 problem](../entities-advanced#handling-the-n1-problem).
```
StephenBarlow, line 96:
```suggestion
> ⚠️ **A particular reference resolver might be called _many_ times to resolve a single query!** It is crucial that reference resolvers account for "N+1" issues (e.g., via [data loaders](https://github.com/graphql/dataloader)). For details, see [Handling the N+1 problem](./entities-advanced#handling-the-n1-problem).
```

StephenBarlow (changes requested):

StephenBarlow (approved):

PR #2314, opened 2023-01-04
Add support for tag spec v0.3

sachindshinde:
This PR:
- Adds a tag spec definition for tag spec v0.3.
  - The only change is to add support for the `SCHEMA` location in the `@tag` directive definition.
  - This also changes composition to use tag spec v0.3 in output.
- Updates `SUPPORTED_FEATURES` to include tag v0.3 (which controls supported features during supergraph schema processing).
- Updates the federation spec to use the `@tag` directive definition of tag spec v0.3.
  - Note this change only uses that directive definition for federation spec v2.3.
  - For federation spec v2.2 and below, we maintain current behavior (which is to use the `@tag` directive definition of tag spec v0.2).
    - This is arguably a bug, as we should be using the `@tag` directive definition of tag spec v0.1 for Fed 1 instead of v0.2, but I'm guessing it's benign given no one's complained thus far (and we'd have to run checks if we wanted to change the behavior).
- Updates `TagDirective` in `@apollo/subgraph` to use the `@tag` directive definition of tag spec v0.3. 
- Updates tests.

### DIFF
--- a/composition-js/CHANGELOG.md
+++ b/composition-js/CHANGELOG.md
@@ -14,6 +14,7 @@ This CHANGELOG pertains only to Apollo Federation packages in the 2.x range. The
 - Error on composition when a `@shareable` field runtime types don't intersect between subgraphs: a `@shareable` field
   must resolve the same way in all the subgraphs, but this is impossible if the concrete runtime types have no
   intersection at all [PR #1556](https://github.com/apollographql/federation/pull/1556). 
+- Uses the 0.3 version of the tag spec in the supergraph, which adds `@tag` directive support for the `SCHEMA` location [PR #2314](https://github.com/apollographql/federation/pull/2314).
 
 ## 2.2.0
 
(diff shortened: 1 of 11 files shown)


sachindshinde:
@pcmanus 
Sounds good, updated the `CHANGELOG.md` for `subgraph-js` and `gateway-js` (also updated the `CHANGELOG.md` for `composition-js`). Will merge into `main`, let me know if there's anything in the `CHANGELOG.md` entries I should tweak and I can file follow-up PRs.

sachindshinde:
@geekmini Thanks for writing in! Your error looks similar to that reported in https://github.com/apollographql/federation/issues/2375 . We've written up some guidance in https://github.com/apollographql/federation/issues/2375#issuecomment-1423115401 , can you let us know whether that resolves the issue?

GeekEast:
Hi @sachindshinde yup, upgrade the federation version solves the issue. BTW, I am using Nest.Js as framwork

PR #2322, opened 2023-01-10
Changes for using npm workspaces

clenfest:
Updates to use npm workspaces. Thanks @trevor-scheer

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,6 +39,8 @@ jobs:
     docker:
       - image: cimg/node:14.20
     steps:
+      -run: 
+        command: npm i -g npm@^8.5.0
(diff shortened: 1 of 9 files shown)


.circleci/config.yml
```diff
@@ -39,6 +39,8 @@ jobs:
     docker:
       - image: cimg/node:14.20
     steps:
+      - run:
```
clenfest, line 42: npm workspaces don't work on the version of npm that is installed on the build image. We need to upgrade it. Thankfully this only adds ~10s to the build.

trevor-scheer (approved):
Lgtm. Probably some simplification we can do in Circle config in the future but nothing problematic for now.

PR #2330, opened 2023-01-11
github workflows for publish and release

clenfest:


clenfest:
Note to reviewers. Don't bother with the diff. Both of these files are new. I was dumb about how to test originally, but I figured it out.

.github/workflows/publish.yml
```diff
@@ -1,28 +1,44 @@
-name: Publish packages to npm
+name: Publish
 
 on:
```
dariuszkuc, line 3: I think this should be just (no push)

```
on:
  release:
    types: [published]
```

.github/workflows/release.yml
```diff
@@ -22,8 +23,17 @@ jobs:
 
       - name: Install Dependencies
         run: npm i
+        
+      - name: Extract branch name
+        shell: bash
+        run: echo "##[set-output name=branch;]$(echo ${GITHUB_REF#refs/heads/})"
```
dariuszkuc, line 29: I believe `set-output` is deprecated -> https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

You should just use `GITHUB_OUTPUT`

```
- name: Set output
  run: echo "{name}={value}" >> $GITHUB_OUTPUT
```

clenfest: I checked the docs, it looks like it's available in `${{ github.ref_name }}` now without me needing to parse it.

.github/workflows/publish.yml
```diff
@@ -1,28 +1,43 @@
...
+    name: Publish slack webhook notifying about approval
+    if: "contains(github.event.head_commit.message, 'release:')"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Send to Slack
+        env: 
+          SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
+        run: |
+          echo webhook url $SLACK_WEBHOOK
+#         TODO: slack integration
```
trevor-scheer, line 18: Is this still a TODO?

trevor-scheer (approved):

.github/workflows/publish.yml
clenfest: I'll leave it for now. Either we keep the approval step and have to add slack integration or we get rid of it and have to make other changes.

PR #2335, opened 2023-01-13
Rewrites gateway post-processing

### LINKED ISSUE #2320: Issue with gateway post-processing for @interfaceObject results
There is currently an issue when using `@interfaceObject` and data is fetched without resolving the concrete type. That is, to illustrate, if one has:
```graphql
type Query {
  t1: T
}

interface T @key(fields: "id") {
  id: ID!
}

# ... some implementations of `T`, details don't matter
```

```graphql
type Query {
  t2: T
}

type T @key(fields: "id") @interfaceObject {
  id: ID!
  v: Int
}
```
and the query is:
```graphql
query {
  t2 {
    v
  }
}
```
Then the intent is that we only need to query the 2nd subgraph and never need to resolve the underlying implementation type, since only fields of the interface is requested and `__typename` is not explicitly requested. And indeed, the query plan will simply send the query essentially as-is to the 2nd subgraph.

But while this correctly fetch the proper results internally, the gateway run the graphQL execution on those internal results as way to filter/reorder the subset of the internal results than needs to be returned, and this currently does not work in such example because in the supergraph, type `T` is an interface, and so graphQL-js execute _requires_ that that we have a `__typename` with a real implementation name, which we don't have. The result is that the code currently gets nullified instead of returning the proper results.

Overall, this is a restriction of our post-processing use of graphQL-js execute: we don't really want a full execution and we have the data for the field queried and the implementation type is not requested in any way, so we ought to be able to return the results.

The "right" fix here is imo to change our post-processing, and to stop using the full graphQL-js `execute` and instead add code that just do the bits of filtering and reordering on the internal data that we need, at which point that code can easily handle this case.

I'll note that this is something that has come up before as calling graphQL-js `execute` is known to have a somewhat large performance cost, despite us not needing a lot of what is done. There is even [a longstanding FIXME](https://github.com/apollographql/federation/blob/main/gateway-js/src/executeQueryPlan.ts#L96) in the code taht says that this should be improved/simplified.

Tl;dr, I suggest that it's time that we fix that FIXME once and for all.

pcmanus:
With `@interfaceObject`, some queries can sometimes be entirely fullfilled by a sugraph using the `@interfaceObject` (when only fields provided by that subgraph are queried and the query does not necessitate knowledge of the underlying implementation type as neither `__typename` is queried, nor is there any type condition). But the post-processing that the gateway does on the internal results to compute the final response was currently using GraphQL-js execution, and this was breaking in those cases because we get results for what is an interface in the supergraph but with no __typename.

Additionally, using GraphQL-js execution for this post-processing has long been a "FIXME" of the code, as this is a heavy-handed way to do this processing.

This patch rewrite this post-processing to stop using GraphQL-js execution and instead implements the subset of what is truly necessary for that post-processing. Doing so allows to ensure that the code work as needed with the `@interfaceObject` cases mentioned above.

Fixes #2320


gateway-js/src/__tests__/executeQueryPlan.test.ts
```diff
@@ -1114,13 +1068,11 @@ describe('executeQueryPlan', () => {
...
 
       expect(response.data?.vehicle).toEqual(null);
-      expect(response.errors).toBeUndefined();
-      // SEE COMMENT ABOVE THIS TEST.  SHOULD BE RE-ENABLED AFTER #981 IS FIXED!
-      // expect(response.errors).toMatchInlineSnapshot(`
-      //   Array [
-      //     [GraphQLError: Abstract type "Vehicle" was resolve to a type [inaccessible type] that does not exist inside schema.],
-      //   ]
-      // `);
+      expect(response.errors).toMatchInlineSnapshot(`
```
pcmanus, line 1071: Note to reviewers: there is a bunch of comments on the background around this in `executeQueryPlan`, which is part about `postProcessingErrors`. The tl;dr is that the existing code was unconditionally ignoring post-processing errors, which some users have been complaining about in the past because it can make it very hard to debug issue with broken subgraphs, and an earlier patch tried to change this by unconditionally returning post-processing errors, but this was quickly reverted because post-processing errors often duplicate subgraphs errors and people were annoyed.

This patch implements a middle ground where we do not include post-processing errors at all if we have some query plan execution errors (thus make sure we can't create duplicates), but include those post-processing errors if there is no query plan execution errors whatsoever, to at least help in that case.

This typically make this test kind of work again: the test have a subgraph that returns an object for an `@inacessible` type, and the old was nullifying that object (correctly, we shouldn't return it) but was not indicating any error (which isn't great). The updated code bring back the error.

Do not that the new error message send differs from what graphQL-js would have produced, but I think the new message is a bit more helpful at pinpointing the problem.

Lastly, this code made me noticed that we still had some `cleanErrorOfInaccessible` code that was completely unused, so this patch removes it. Now that we control the code generating the errors, we have no reason to ever need to "clean" those error message: we can just make sure we don't leak anything in the first place (which is what is done here).

gateway-js/src/executeQueryPlan.ts
```diff
@@ -49,12 +52,42 @@ interface ExecutionContext {
...
+  return data[introspectionSelection.name.value];
+}
+
 export async function executeQueryPlan(
   queryPlan: QueryPlan,
   serviceMap: ServiceMap,
   requestContext: GatewayGraphQLRequestContext,
   operationContext: OperationContext,
   supergraphSchema: GraphQLSchema,
+  apiSchema: Schema,
```
pcmanus, line 90: Note to reviewers: we do add an argument to an exported method, and I'll mention it in the changelog. We already added the `supergraphSchema` argument in a previous release the same way so I don't think it's an issue.

In general, the arguments of this method are a bit messy now because we pass the API schema as a `Schema` but we also have that same schema in the `OperationContext` as a `GraphQLSchema`, and while we do want to pass both versions to avoid costly conversions, we could maybe clean things up. I'll leave that for a followup though.

gateway-js/src/index.ts
```diff
@@ -559,9 +556,7 @@ export class ApolloGateway implements GatewayInterface {
   ): void {
     this.queryPlanStore.clear();
     this.apiSchema = coreSchema.toAPISchema();
-    this.schema = addExtensions(
-      wrapSchemaWithAliasResolver(this.apiSchema.toGraphQLJSSchema()),
```
pcmanus, line 563: Note to reviewers: `wrapSchemaWithAliasResolver` was needed due to the use of GraphQL-js execution. Namely, when GraphQL-js execute is passed an object, it expects that object to match the schema and so to use field names from the schema. But the internal data we gather during query plan execution will have data under the "response name", and so under aliases when aliases was used. That method made sure things were still working.

The new code is written to use response name directly, so no need for that.

clenfest (approved):

PR #2338, opened 2023-01-14
Fix Supergraph code example in composition docs

yoavbls:
At first glance, I thought that `find` adds to query names when they merge, 
but it's probably a tiny leftover. I changed `findBook` to `book` since it's more similar to the rest of the queries names in the docs.

CleanShot 2023-01-14 at 18 48 29@2x


trevor-scheer (approved):
Thanks!

Issue #2337, opened 2023-01-13
Setting the header Apollo-Query-Plan-Experimental has no effect


vtipparam:
Setting the request header `Apollo-Query-Plan-Experimental` is expected to populate the query plan in the response but it does not.

**Sample request**

```javascript
var request = require('request');
var options = {
  'method': 'POST',
  'url': 'http://localhost:8080/graphql',
  'headers': {
    'Apollo-Query-Plan-Experimental': 'true',
    'Content-Type': 'application/json'
  },
  body: JSON.stringify({
    query: `query Echo($message: String!) {
    echo(message: $message) {
      message
    }`
   },
    variables: {"message":"Hello world"}
  })
};
request(options, function (error, response) {
  if (error) throw new Error(error);
  console.log(response.body);
});
```
**Response**
```json
{
    "data": {
        "echo": {
            "message": "Hello world"
        }
    }
}
```
**Expecting**
```json
{
    "data": {
        "echo": {
            "message": "Hello world"
        }
    },
    "extensions": {
        "__queryPlanExperimental": "QueryPlan {\n  Fetch(service: \"echogql\") {\n    {\n      echo(message: $message) {\n        message\n      }\n    }\n  },\n}"
    }
}
```
A close look at the code appears to show it is looking up for the header `Apollo-Query-Plan-Experimental` in a case-sensitive manner. Forcing a case-insensitive lookup appears to fix the problem.

https://github.com/apollographql/federation/blob/main/gateway-js/src/index.ts#L837

vtipparam:
I thought of preparing a PR and started writing a test case. I expected the test case to fail but to my surprise the test case passed. After some more troubleshooting and digging into the code, it appears our usage of `expressMiddleware` handler is triggering the issue.

```javascript
  const graphqlRouter = express.Router();

  graphqlRouter.use(
      '/graphql',
      cors(corsOptions),
      json(),
      expressMiddleware(server, {
        context: async ({req}) => ({req}),
      })
  );
```

`expressMiddleware` is expected create `HTTPGraphQLRequest` with `HeaderMap` for `headers` but it is creating a plain `Map` instead. The lookups on plain Map are case sensitive while the lookups on HeaderMap are cease insensitive.

As far as I can see there are two possible fixes for this:
1) Fix just `Apollo-Query-Plan-Experimental` header lookup by passing lower case header name
https://github.com/apollographql/federation/blob/main/gateway-js/src/index.ts#L837
[...]

PR #2339, opened 2023-01-18
Change to case insensitive header lookup

vtipparam:
Setting the header Apollo-Query-Plan-Experimental has no effect. Setting the request header Apollo-Query-Plan-Experimental is expected to populate the query plan in the response but it does not. We are performing case sensitive lookup on the headers. See #2337 for details.

Adding `expressMiddleware` handler particularly exposes the issue. Will add a separate PR to address the expressMiddleware issue.


trevor-scheer:
This shouldn't be necessary, the `headers` object in question here is:
https://github.com/apollographql/apollo-server/blob/c835637be07929e3bebe8f3b262588c6d918e694/packages/server/src/utils/makeGatewayGraphQLRequestContext.ts#L163-L164 (which uses the case-insensitive `HeaderMap` internally).

This should no longer be an issue via https://github.com/apollographql/apollo-server/pull/7313

PR #2343, opened 2023-01-20
Fix fed1 schema upgraded when `@external` is on a type.

pcmanus:
For historical reasons, `@external` was legal to put on an object type since about forever (even in federation 1) but it used to be completely ignored. We "fixed" this support in #2214, but to limit surprised on upgrades from fed1, the schema upgrader had been modified to drop any `@external` on an object type (since again, those were legal but ignored in fed1). However, the code for adding @shareable in that same schema upgrader was not updated correctly and so it _was_ taking `@external` on types into account, and thus it ended up _not_ adding `@shareable` in some cases where it should have.

### DIFF
--- a/.changeset/flat-pandas-cross.md
+++ b/.changeset/flat-pandas-cross.md
@@ -1,0 +1,6 @@
+---
+"@apollo/federation-internals": patch
+---
+
+Fix fed1 schema upgraded when `@external` is on a type.
+  
--- a/.changeset/flat-pandas-cross.md
+++ b/.changeset/flat-pandas-cross.md
@@ -2,5 +2,6 @@
 "@apollo/federation-internals": patch
 ---
 
-Fix fed1 schema upgraded when `@external` is on a type.
-  
+Fix unexpected composition error about @shareable field when `@external` is on a type in a fed1 schema (one without
+`@link`)
+
--- a/internals-js/src/federation.ts
+++ b/internals-js/src/federation.ts
@@ -577,7 +577,7 @@ export class FederationMetadata {
 
   private externalTester(): ExternalTester {
     if (!this._externalTester) {
-      this._externalTester = new ExternalTester(this.schema);
+      this._externalTester = new ExternalTester(this.schema, this.isFed2Schema());
     }
     return this._externalTester;
   }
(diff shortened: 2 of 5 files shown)


internals-js/src/federation.ts
```diff
@@ -1827,6 +1827,13 @@ class ExternalTester {
   }
 
   private collectExternalsOnType() {
+    // We do not collect @external on types for fed1 schema as those are ignored and will be discarded by the schema upgrader
+    // Do note however that the schema upgrader relying on the methods of this tester to figure out if a subgraph resolves
+    // a field in order to know when @shareable should be automatically added, hence the need for the short-cut here.
+    if (!this.isFed2Schema) {
+      return;
+    }
```
trevor-scheer, line 1835: For my own understanding I've rewritten this comment a bit. Is this correct, and if so would you consider using it here or updating yours a bit? Your PR description explains this but the comment misses this part which was important for me to understand it:
> In the fed1 case, if the map is populated then `@shareable` won't be added in places where it should have.

My interpretation / suggestion:
We do not collect `@external` on types for fed1 schema since those will be discarded by the schema upgrader.
The schema upgrader relies on the populated `externalFieldsOnType` object to inform when `@shareable` should be automatically added. In the fed1 case, if the map is populated then `@shareable` won't be added in places where it should have.

.changeset/flat-pandas-cross.md
```diff
@@ -0,0 +1,6 @@
+---
+"@apollo/federation-internals": patch
+---
+
+Fix fed1 schema upgraded when `@external` is on a type.
```
trevor-scheer, line 5: I'm not sure a user reading this will understand how this affects their usage. I think the context in your PR description would be useful here (and in the manual changelog entries).

trevor-scheer (approved):
Just a couple minor comments on the documentation.

internals-js/src/federation.ts
pcmanus: The updates are correct, and are probably clearer; made the change.

.changeset/flat-pandas-cross.md
pcmanus: You're right, this wasn't that helpful as it could be. I've reworded it so it's more obvious for a user if they may be affected or not by this and care about.

PR #2349, opened 2023-01-24
Update lockfile and fix workflow

trevor-scheer:
There's a known issue in changesets where the lockfile isn't updated as part of the `version` command.
https://github.com/changesets/action/issues/203

This updates the lockfile from last release and implements the same workaround we've taken in the server repo.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,8 @@
     "error-code-doc:check": "npm run error-code-doc && git diff --exit-code",
     "hints-doc": "ts-node ./composition-js/src/genHintDoc.ts > ./docs/source/hints.md",
     "hints-doc:check": "npm run hints-doc && git diff --exit-code",
-    "gen-test-supergraph": "ts-node ./query-planner-js/src/__tests__/genTestSupergraph.ts"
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -24,7 +24,8 @@
     "error-code-doc:check": "npm run error-code-doc && git diff --exit-code",
     "hints-doc": "ts-node ./composition-js/src/genHintDoc.ts > ./docs/source/hints.md",
     "hints-doc:check": "npm run hints-doc && git diff --exit-code",
-    "gen-test-supergraph": "ts-node ./query-planner-js/src/__tests__/genTestSupergraph.ts"
+    "gen-test-supergraph": "ts-node ./query-planner-js/src/__tests__/genTestSupergraph.ts",
+    "changeset-version": "changeset version && npm i"
```
clenfest, line 28: It's ok there isn't a `npx` here? You're not relying on changesets being globally installed?

clenfest (approved):

package.json
trevor-scheer: Yeah. I'm either not searching for the right thing or documentation isn't clear but scripts can run locally installed binaries like this when they're installed as dependencies in the project.

PR #2354, opened 2023-01-25
exit pre release. Also modify release.yml to automatically publish and get rid of custom publish

clenfest:


package.json
```diff
@@ -25,7 +25,8 @@
     "hints-doc": "ts-node ./composition-js/src/genHintDoc.ts > ./docs/source/hints.md",
     "hints-doc:check": "npm run hints-doc && git diff --exit-code",
     "gen-test-supergraph": "ts-node ./query-planner-js/src/__tests__/genTestSupergraph.ts",
-    "changeset-version": "changeset version && npm i"
+    "changeset-version": "changeset version && npm i",
+    "publish": "changeset publish && git push --follow-tags"
```
trevor-scheer, line 29: I think the `git push --follow-tags` is only needed if you're not using the action. We don't have it in the AS repo but we do have all of our tags in git as we'd expect.

I think this is handling it?
https://github.com/changesets/action/blob/main/src/run.ts#L90

trevor-scheer (approved):

PR #2362, opened 2023-02-01
Fix QP bug querying some fields with `@interfaceObject`

pcmanus:
When an interface field is requested only for some implementation, and the query "transit" through a subgraph using `@interfaceObject`, then the query planning code was not recognizing early enough that the field was queried on an implementation type and not the interface itself, and this lead to a later assertion error.

This commit adds a reproduction test, and fix the issue by simply making sure, when we build the query plan "paths", to not use an interface field when it's an implementation field that is queried.

### DIFF
--- a/.changeset/tidy-cherries-fly.md
+++ b/.changeset/tidy-cherries-fly.md
@@ -3,5 +3,5 @@
 "@apollo/query-planner": patch
 ---
 
-Fix assertion errors thrown by the query planner when querying fields for a specific inteface implementation in some cases where `@interfaceObject` is involved
+Fix assertion errors thrown by the query planner when querying fields for a specific interface implementation in some cases where `@interfaceObject` is involved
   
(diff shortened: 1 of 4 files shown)


.changeset/tidy-cherries-fly.md
```diff
@@ -0,0 +1,7 @@
+---
+"@apollo/query-graphs": patch
+"@apollo/query-planner": patch
+---
+
+Fix assertion errors thrown by the query planner when querying fields for a specific inteface implementation in some cases where `@interfaceObject` is involved
```
clenfest, line 6:
```suggestion
Fix assertion errors thrown by the query planner when querying fields for a specific interface implementation in some cases where `@interfaceObject` is involved
```

clenfest:
Just one misspelling in the changelog. Other than that, it looks like we need to update the graphql type check thingy (is there an actual reason for having this be a part of our pipeline?) but I'll approve assuming you'll do that.

clenfest (approved):

PR #2363, opened 2023-02-01
adding finders to composition and query planning

clenfest:
Notes to reviewers:
- There are a couple of tests that I haven't yet written yet. Those are annotated with an `it.todo` and should be done shortly.
- There are still some TODOs in the code, mostly involving stuff that isn't going to make it into the GA version, but will be done at some point in the future. I'm guessing we may want another tag for this, but not sure what I should change it to.


query-planner-js/src/buildPlan.ts
```diff
@@ -1118,6 +1123,37 @@ class FetchGroup {
     return rewrites;
   }
 
+  getFindersForInput(inputs?: SelectionSet): Finder[] {
+    // TODO: Cache / optimize
+    assert(this.isEntityFetch, 'Must be a FetchGroup for an entity');
+    const entityName = inputs?.parentType.name;
```
pcmanus, line 1129: That seems to be assuming that the inputs parent type is the entity type, but that's not necessarily the case. `FetchNode` sometimes groups fetch to multiple entity types of the same interface, and in that case the input parent type will be that interface.

query-planner-js/src/buildPlan.ts
```diff
@@ -1138,13 +1174,49 @@ class FetchGroup {
...
+      const operationDocument = operationToDocument(operation);
+      const subgraphFetchNode: SubgraphFetchNode = {
+        kind: 'SubgraphFetch',
+        serviceName: this.subgraphName,
+        id: this.id,
+        operationName: operation.name ?? 'finder call',
+        operation: stripIgnoredCharacters(print(operationDocument)),
+        variableUsages: [finder.fieldName()],
+        inputs: [{
+          selectedType: finder.argument(0).inputType.toString(),
```
pcmanus, line 1200: Fwiw, neither `selectedType` nor `variableName` seems to be used during execution at all, so unclear what their purpose is.

query-planner-js/src/buildPlan.ts
```diff
@@ -4027,6 +4104,81 @@ function operationForEntitiesFetch(
...
+
+  arguments(): Readonly<ArgumentDefinition<FieldDefinition<any>>[]> {
+    return this.target.arguments();
+  }
+
+  argument(idx: number): { inputType: InputType, name: string } {
+    const arg = this.target.arguments()[idx];
+    assert(arg, `Argument at index ${idx} exists`);
+    assert(arg.type, 'type exists on argument');
+    return {
```
pcmanus, line 4128: Why create some new random structure instead of just returning `arg`? Also, this method is call twice, separately as `finder.argument(0).inputType` and `finder.argument(0).name`, which seem even more unnecessarily wasteful.

query-planner-js/src/buildPlan.ts
```diff
@@ -1138,13 +1174,49 @@ class FetchGroup {
...
+        allVariableDefinitions: variableDefinitions,
+        operationName,
+        finder,
+      });
+      const operationDocument = operationToDocument(operation);
+      const subgraphFetchNode: SubgraphFetchNode = {
+        kind: 'SubgraphFetch',
+        serviceName: this.subgraphName,
+        id: this.id,
+        operationName: operation.name ?? 'finder call',
```
pcmanus, line 1196: This defaulting of the operation name is what currently breaks any query involving finders if the user didn't explicitly named the query. The reason being that, if that's the case, we indicate here that the operation is called "finder call", but we don't actually name it that way later in `operationForFindersFetch`, so the execution fails with 'Unknown operation named "finder call".'. Note that even if we were to name the operation similarly, "finder call" is not a valid operation name in graphQL.

And taking a step back, what's the reasoning for setting a default operation name? We didn't do it before for `_entities` query. I'd personally not add one here too.

query-planner-js/src/buildPlan.ts
```diff
@@ -4027,6 +4104,81 @@ function operationForEntitiesFetch(
...
+  allVariableDefinitions: VariableDefinitions,
+  operationName?: string,
+  finder: Finder,
+}): Operation {
+  const inputField = finder.argument(0);
+  const variableDefinitions = new VariableDefinitions();
+  const variable = new Variable(inputField.name);
+  const metadata = federationMetadata(subgraphSchema);
+  assert(metadata, 'Expected schema to be a federation subgraph');
+  variableDefinitions.add(new VariableDefinition(subgraphSchema, variable, inputField.inputType));
```
pcmanus, line 4153: We should be checking for, and deal with clashes between the name of the variable we use for the key, and the other variables from the user query as those can absolutely clash. Fwiw, I think that's where the `variableName` of `FetchInput` could be useful: we cannot assume that use the name of the key as name of the variable will always work and we need to deal with that.

I'll note that for `_entities`, there is code in the query plan execution that validates that no user variables are named "representations" and that's how we avoid clash (it's not ideal as it's a bit of an arbitrary limitation on users, but it's at least handled gracefully and no-one has complained about it that I know of, so fine so far in practice). I don't think we can use that approach here: rejecting a variable name at execution because some key used in the query somewhere has the name is imo too limiting, and could even leak information it shouldn't if that key is not public.

pcmanus (commented):
I'll start by saying that I think we ultimately need to cover this extensively with execution tests as this is a new execution branch and many (most?) existing execution tests should probably be adapted to use `@finder`. 

As it stands, I'm a tad surprised this is marked "ready for review" when the PR doesn't have a single execution test (not even a `it.todo` one). In other words, not a single test actually validate that the added feature works end-to-test. Fwiw, it took me less than 10 minutes to write a simple execution test and to see that in fact, many things do not work end-to-end currently (I point what I found below and in inline comments, but this may not be exhaustive).

That aside, at a high level, I think the main problem is that a `FetchNode`, as written today, can fundamentally query multiple entities (due to lists), and can even batch fetches for different entity types in the same `FetchNode`. But the current `SubgraphFetchNode` handles neither correctly and assumes a single entity. For the issue of lists, we could theoretically deal with that during execution, by sending one fetch per-entity, and that was the original idea behind `MappingNode`, but this is not how it is implemented. But even then, that would still leave the issue of `FetchNode` than handles multiple entity types.

Besides, using the `MappingNode` to do one fetch per-entity would mean no batching at all whatsoever, and I though we had agreed that this probably wasn't acceptable. My understanding what that we agreed on try to do batching withing a single query using aliases, and I do think it should be part of the solution.

And just to make sure, what I mean by "batching using aliases", is that if we have a `FetchNode` that looks something like:
```
Fetch(service: "S2") {
  {
    ... on A {
      __typename
      id
    }
    ... on B {
      __typename
      id
    }
  } => {
    <some selection>
  }
```
and it is applied to the following list of entities:
```json
[ { "__typename": "A", "id": 0 }, { "__typename": "B", "id": 1 }, { "__typename": "A", "id": 2 }]
```
then the underlying query we'd send to the subgraph could look something like:
```graphql
# In practice, we'd likely use a fragment to avoid repeating <some selection>
query($id1: ID!, $id2: ID!, $id3: ID!) {
  alias1: finderOnA(id: $id1) {
    <some selection>
  }
  alias2: finderOnB(id: $id2) {
    <some selection>
  }
  alias3: finderOnA(id: $id3) {
    <some selection>
  }
}
```
Of course, we can't statically generate that query in the query plan, so that would imply a query node that has enough information so that such query can be easily generated at execution. 

Looking at this, I admittedly wonder a bit how this scale with bigger lists of entities: I'm sure there is a risk of
running into limitations of some subgraph servers around the number of arguments, or maybe the overall size of the
query. But at the same time, I don't feel comfortable calling `@finder` ready for production with no batching support
whatsoever: how could we recommend it for production to anyone if that's the case?

So I think a proper full solution here may have to use this trick of batching with aliases, but also enforce a maximum
batch size and split into multiple batches when necessary.

Other than that, I left comments inline when that made sense, but a few things not mentioned or indirectly mentioned:
- The is no handling of `@requires` at all. I understand that we don't want to have a finder-solution for `@requires` in this PR, but we should still have some sort of validation around it. Currently, if you use a `@finder` on an entity with `@requires`, it silently ignores the requirement. You might even be forced by the code to use a `@finder` on an entity that has `@requires` even though it is not working.
- I don't think keys on interfaces and `@interfaceObject` works properly currently.
- The code ignores the `resolvable` argument of keys. A much minor point obviously, but it would be trivial to handle it (by not requiring a `@finder` if the key is declared as non-resolvable (and even probably erroring if there is one in that case)).

query-planner-js/src/buildPlan.ts
```diff
@@ -1118,6 +1123,37 @@ class FetchGroup {
     return rewrites;
   }
 
+  getFindersForInput(inputs?: SelectionSet): Finder[] {
+    // TODO: Cache / optimize
```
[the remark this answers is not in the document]
clenfest: agreed.

PR #2369, opened 2023-02-02
When we publish a release, kick off workflows in federation-rs

clenfest:
If we publish a release, kick off workflows in federation-rs to create PRs for router-bridge and harmonizer.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,6 +5,7 @@ on:
     branches:
       - main
       - next
+      - clenfest/federation-rs-build-step
 
 concurrency: ${{ github.workflow }}-${{ github.ref }}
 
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -36,4 +39,13 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
-          
\ No newline at end of file
+      
+      - name: Kick off release in federation-rs
+        if: steps.changesets.outputs.hasChangesets == 'false'
+        uses: benc-uk/workflow-dispatch@v1
```
dariuszkuc, line 45: instead of some random action should we use octockit?

```yaml
- uses: actions/github-script@v6
  with:
    github-token: ${{ secrets.PAT_TOKEN }}
      script: |
        await github.rest.actions.createWorkflowDispatch({
          owner: 'apollographql',
          repo: 'federation-rs,
          workflow_id: '<TODO>',
          ref: 'main',
          inputs: { 
            version: "${{ env.FEDERATION_VERSION }}"
          })
```

clenfest: That's probably fair. I had this initially, but felt a little more comfortable with something that was done for me.

jeffjakub (approved):

Issue #2353, opened 2023-01-25
[Feature Request] @apollo/gateway - Expose the current request query path to the "willSendRequest" hook


talazenkot:
As far as I've researched, the path of the field currently resolved by the federation gateway is not exposed to any of the arguments passed to `willSendRequest` and `didReceiveResponse` hooks.
I've faced a scenario where I need to apply certain HTTP headers to the requests only sent under a specific node. 

![Screenshot 2023-01-25 at 18 23 41](https://user-images.githubusercontent.com/54143565/214618634-d26b0652-1341-4a6c-84a4-9eb97b5e9b93.png)
Given the image above, I would like to:

1. Resolve `node2` in some external apollo server
2. Respond with some header to the gateway
3. Make sure that the gateway passes this header to all children nodes (3, 4, 5, and their children) without its leaking into the sibling nodes (6 and its children).

I've managed to do that, only by caching this header for a certain path (the path from which it was received, in `didReceiveResponse`), and then in each request (`willSendRequest`), check if the current path is a children path of one of the cached paths, and if so, apply some logic, see the following code:
![Screenshot 2023-01-25 at 18 38 21](https://user-images.githubusercontent.com/54143565/214622564-5e5f5118-29f4-4f0e-ba6f-47d0e3fa77ea.png)

Bottom line, unless there's some more robust solution, I was wondering if it will be possible to expose the path of the current field.

korinne:
Thanks for raising this request! Our team has added this to our backlog, and we will keep this issue updated once the request has been prioritized. We are also always open to community contributions.

talazenkot:
@korinne Great! I will be glad to contribute and make this happen. Is there any other guidelines I need to follow except for the ones mentioned in https://github.com/apollographql/federation/blob/main/CONTRIBUTING.md#reporting-bugs ? Is there some sort of sync required?

PR #2384, opened 2023-02-08
@apollo/gateway - Expose the current request query path to the "willSendRequest" and "didReceiveResponse" hooks

talazenkot:
Description:
Exposed the path to the node currently resolved by the gateway to the willSendRequest and didReceiveResponse hooks.

Solves https://github.com/apollographql/federation/issues/2353


korinne:
Thanks again for opening a PR. We have filed this PR as under review, and will provide feedback shortly.

talazenkot:
> Thanks again for opening a PR. We have filed this PR as under review, and will provide feedback shortly.

@korinne hey! any news regarding this? maybe there's something I can do from my side to push it forwards?

gateway-js/src/datasources/types.ts
```diff
@@ -21,6 +22,11 @@ export type GraphQLDataSourceProcessOptions<
    * The request to send to the subgraph.
    */
   request: GatewayGraphQLRequestContext<TContext>['request'];
+
+  /**
+   * The path to the current processed node within the graph query
+   */
+  nodeGraphPath?: ResponsePath;
```
pcmanus, line 29: I'd move this inside the `INCOMING_OPERATION` part (that path makes no sense for a `HEALTH_CHECK` or `LOADING_SCHEMA` kind of request) and maybe rename it to something like `pathInIncomingRequest` (we call `incomingRequestContext` the total operation received by the gateway, and the "path" is really the path within that gateway operation at which this particular subgraph "sub-query" fits in, and that's probably how I would describe that field btw; I also wouldn't use `node` in the name as that's query plan specific naming and most users aren't meant to be familiar with that).

Could be nice to also document that the path may be `undefined` if it is unavailable, but will be empty (rather than `undefined`) for "top-level" fetches.

pcmanus (changes requested):

.changeset/flat-books-sin.md
```diff
@@ -0,0 +1,5 @@
+---
+"@apollo/gateway": patch
+---
+
+Current request graph path exposure
```
pcmanus, line 5: It would be awesome to expand this entry a bit. I'd suggest something along the lines of:
>  Exposes, for each subgraph request, the path in the overall gateway operation at which that subgraph request gets inserted. This path is now available as the `pathInIncomingRequest` field in the arguments of `RemoteGraphQLDataSource.willSendRequest` and `RemoteGraphQLDataSource.didReceiveResponse`.

talazenkot: Amazing. thanks!

gateway-js/src/datasources/types.ts
talazenkot: Thanks for the clarifications! I've adopted your guidelines.

gateway-js/src/datasources/types.ts
```diff
@@ -47,6 +48,12 @@ export type GraphQLDataSourceProcessOptions<
        * this will increase the memory used by the gateway query plan cache.
        */
       document?: GatewayGraphQLRequestContext<TContext>['document'];
+
+      /**
+      * The path in the overall gateway operation at which that subgraph request gets inserted.
+      * Please note that this could be set to `undefined` when the path is not available, or set to `null` for top-level fetch operations.
```
pcmanus, line 54: I don't believe the "or set to `null` for top-level fetch operations" part is quite right: it should say "or be an empty array for top-level fetch operations".

What I mean/understand by "top-level" fetch operations is just subgraph fetches that are not entity fetches but rather query some `Query`/`Mutation` fields, and those will still be `GraphQLDataSourceRequestKind.INCOMING_OPERATION` but where the `path` in `executeQueryPlan` will be empty.

Your current patch sets `null` for queries that are not `GraphQLDataSourceRequestKind.INCOMING_OPERATION`, but those are queries that don't belong to the execution of a gateway query: they are instead "maintenance" queries the gateway does for either refreshing it's understanding of the schema or "health checks" of the subgraph. As said in my other comment, having the path to `undefined` to mean "not available" is fine for those imo.

pcmanus (changes requested):
A few last remarks. Sorry, I may not have been super clear in my previous ones.

gateway-js/src/datasources/types.ts
talazenkot: Sorry. Misunderstood. Got you now.

talazenkot:
> A few last remarks. Sorry, I may not have been super clear in my previous ones.

No worries. Thanks a lot of your responsiveness and explanations. I've made a few adjustments according to your comments, and also explained why as far as I can tell I cannot use `options.pathInIncomingRequest` directly.

gateway-js/src/datasources/RemoteGraphQLDataSource.ts
```diff
@@ -67,6 +68,11 @@ export class RemoteGraphQLDataSource<
     options: GraphQLDataSourceProcessOptions<TContext>,
   ): Promise<GatewayGraphQLResponse> {
     const { request, context: originalContext } = options;
+    const pathInIncomingRequest =
+      options.kind === GraphQLDataSourceRequestKind.INCOMING_OPERATION
```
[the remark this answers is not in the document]
pcmanus: Oh, correct, this is my bad.

pcmanus (approved):
Lgtm. Thanks a lot for the contribution.

PR #2385, opened 2023-02-08
Add cache option to config to expose ApolloGateway's queryPlanStore 

### LINKED ISSUE #1034: feat(gateway): Flexible query plan caching interface
We should provide an interface for query plan caching to our users. This would allow them to cache in whatever manner they see fit. A great use case, for example, might be a distributed cache shared by `n` gateways running in a cluster.

This could be pretty nicely documented by example.

### LINKED ISSUE #2308: Expose ApolloGateway's queryPlanStore to client
## Description
Based on the understanding of the Apollo federation library source code, it seems that the queryPlanStore (LRUCache<string, QueryPlan>) size is configurable by the client, but the Cache object itself is not configurable. 

```javascript
    this.queryPlanStore = this.initQueryPlanStore(
      config?.experimental_approximateQueryPlanStoreMiB,
    );
```

## Issue
Since the cache is not configurable, we cannot initiate our own cache or even support Redis cache. 

## Possible Impact
By allowing the clients to add their own cache,  helps them add metrics for tracking hit rates of these caches.

## Solution
Add a cache option to "QueryPlannerConfig", clients could use them to add their implementation of LRUCache.

```javascript
export type QueryPlannerConfig = {
   ......
  incrementalDelivery?: {
    enableDefer?: boolean,
  }
  cache?: LRUCache<string, QueryPlan>,
}
```

AneethAnand:
Fixes #2308
Fixes #1034


korinne:
Thanks again for opening this PR. We've put in on our review backlog, and will provide feedback shortly.

trevor-scheer:
Hey @AneethAnand, I think this is a reasonable change in concept. I have one major concern with the PR, and that's the use of `LRUCache`.

Since we're going to be providing this as a configuration point for users, the interface should be more generic than "an `LRUCache` from the `lru-cache` library". We actually have a very similar concern in Apollo Server, where we use the `KeyValueCache` interface from the `@apollo/utils.keyvaluecache` package.

My recommendation is this:
* provide the `queryPlanConfig.cache` interface as a `type QueryPlanCache = KeyValueCache<QueryPlan> & { clear: () => void }`
* use the `InMemoryLRUCache` concrete class (also exported by the `@apollo/utils.keyvaluecache` package) for the default implementation - this cache wraps `lru-cache` and implements the `QueryPlanCache` type.

I'm happy to implement these changes if you'd prefer (just let me know you don't mind me pushing to your branch if so).

AneethAnand:
@trevor-scheer  Thanks for the comment. I had interface in the first place but it dint have the clear method and I dint think of this option 
```
type QueryPlanCache = KeyValueCache<QueryPlan> & { clear: () => void }
```
.Also I believe the set method also was not compatible. I will give it a try and push my change again.

AneethAnand:
HI @trevor-scheer  I have incorporated your review comments and pushed a commit to my branch.
https://github.com/AneethAnand/federation/commit/5bf44d78dad14c83729fa203958bf4e9a8903c42 for your perusal. Thanks!

gateway-js/src/__tests__/gateway/endToEnd.test.ts
```diff
@@ -8,9 +8,10 @@ import { GraphQLSchemaModule } from '@apollo/subgraph/src/schema-helper';
 import { buildSchema, ObjectType, ServiceDefinition } from '@apollo/federation-internals';
 import gql from 'graphql-tag';
 import { printSchema } from 'graphql';
-import LRUCache from 'lru-cache';
+import { InMemoryLRUCache } from '@apollo/utils.keyvaluecache';
 import { QueryPlan } from '@apollo/query-planner';
 import { createHash } from '@apollo/utils.createhash';
+import { QueryPlanCache } from '@apollo/query-planner/src/config'
```
trevor-scheer, line 14: We can't have imports from `src` directories (I know this is just an autocomplete / auto import). You can either import `QueryPlannerConfig` and access the type on it `QueryPlannerConfig['cache']` or export `QueryPlanCache` from the top-level index file (where `QueryPlannerConfig` is exported). I'm comfortable with either approach (I think the 2nd option is probably most helpful for users).

gateway-js/src/index.ts
```diff
@@ -1,7 +1,8 @@
 import { deprecate } from 'util';
 import { createHash } from '@apollo/utils.createhash';
 import type { Logger } from '@apollo/utils.logger';
-import LRUCache from 'lru-cache';
+import { QueryPlanCache } from '@apollo/query-planner/src/config'
```
trevor-scheer, line 4: Same import issue here

gateway-js/src/__tests__/gateway/endToEnd.test.ts
```diff
@@ -8,9 +8,10 @@ import { GraphQLSchemaModule } from '@apollo/subgraph/src/schema-helper';
 import { buildSchema, ObjectType, ServiceDefinition } from '@apollo/federation-internals';
 import gql from 'graphql-tag';
 import { printSchema } from 'graphql';
-import LRUCache from 'lru-cache';
```
trevor-scheer: There are no more usages of `lru-cache` in the gateway package. Can you remove it from the dependencies list in `gateway-js/package.json`? Then update the top-level `package-lock.json` by running `npm install`.

trevor-scheer (changes requested):
This is exactly what I had in mind, thanks for implementing those changes. Just a few minor comments and this will be good to go!

Can you change the target branch to `next` instead of `main`?

gateway-js/src/__tests__/gateway/endToEnd.test.ts
AneethAnand: Thanks @trevor-scheer resolved

AneethAnand:
> This is exactly what I had in mind, thanks for implementing those changes. Just a few minor comments and this will be good to go!
> 
> * Can you change the target branch to `next` instead of `main`?
> * Can you run `npx changeset` and add a `minor` entry which explains the new configuration option? This will end up in the gateway's changelog so please be descriptive for people who might be interested in using the new feature.
 Minor version bump for the following ?
  ◯ @apollo/query-planner@2.3.2
  ◯ @apollo/gateway@2.3.2

trevor-scheer (approved):

clenfest (approved):
lgtm!

trevor-scheer:
Thanks @AneethAnand!

trevor-scheer:
This change will land with v2.4.0 relatively soon, on the order of a couple weeks or so. In the meantime, you're welcome to use the build from your PR (look at the comment posted by the codesandbox bot).

Issue #2375, opened 2023-02-06
Subgraph doesn't work


Brilhante29:
When I run my project with subgraphs in the latest version of node and nestjs, I got this error

 super(
    ^
GraphQLError: The schema is not a valid GraphQL schema.. Caused by:
Invalid definition for directive "@tag": "@tag" should have locations FIELD_DEFINITION, OBJECT, INTERFACE, UNION, ARGUMENT_DEFINITION, SCALAR, ENUM, ENUM_VALUE, INPUT_OBJECT, INPUT_FIELD_DEFINITION, but found (non-subset) FIELD_DEFINITION, OBJECT, INTERFACE, UNION, ARGUMENT_DEFINITION, SCALAR, ENUM, ENUM_VALUE, INPUT_OBJECT, INPUT_FIELD_DEFINITION, SCHEMA

TLadd:
I experience the same error going from 2.2.3 -> 2.3.0 using [pothos](https://github.com/hayes/pothos) as my schema builder.

Brilhante29:
with 2.2.3 works perfectly?

alexbchr:
Same problem here with 2.3.0 and 2.3.1. No issue with 2.2.3

sachindshinde:
From what we can tell, this error occurs for subgraphs when they're using the `@tag` directive definition from federation spec version 2.3, i.e.
```graphql
directive @tag(name: String!) repeatable on
  | FIELD_DEFINITION
  | INTERFACE
  | OBJECT
  | UNION
  | ARGUMENT_DEFINITION
  | SCALAR
  | ENUM
  | ENUM_VALUE
  | INPUT_OBJECT
  | INPUT_FIELD_DEFINITION 
  | SCHEMA
```
but if their subgraph schema is declaring a lower federation spec, e.g.
```graphql
extend schema @link(url: "https://specs.apollo.dev/federation/v2.2", ...)
```

You'll need to either upgrade your subgraph to use federation spec v2.3 e.g. with
```graphql
extend schema @link(url: "https://specs.apollo.dev/federation/v2.3", ...)
```
or you'll need to use the `@tag` directive definition of your declared federation spec version (or alternatively omit the `@tag` directive definition if possible, as `buildSubgraphSchema()` should automatically add the appropriate definition).

zhumeisongsong:
With 2.2.3 works perfectly.

zhumeisongsong:
With 2.2.3 works perfectly.

[...]

PR #2386, opened 2023-02-08
Update @apollo/subgraph CHANGELOG.md entry for `2.3.0-beta.2`

sachindshinde:
We've seen folks accidentally forgetting to bump their federation spec to `v2.3` when using the latest `@tag` directive definition that includes `SCHEMA`, which will result in an error from `buildSubgraphSchema()` (see https://github.com/apollographql/federation/issues/2375 ). This PR updates the `CHANGELOG.md` to explicitly state this.

### DIFF
--- a/subgraph-js/CHANGELOG.md
+++ b/subgraph-js/CHANGELOG.md
@@ -9,7 +9,7 @@
 This CHANGELOG pertains only to Apollo Federation packages in the 2.x range. The Federation v0.x equivalent for this package can be found [here](https://github.com/apollographql/federation/blob/version-0.x/subgraph-js/CHANGELOG.md) on the `version-0.x` branch of this repo.
 
 ## 2.3.0-beta.2
-- `@tag` directive support for the `SCHEMA` location. This has been added to the 2.3 version of the federation spec, and you must declare usage of this spec version in your subgraph to use this functionality. [PR #2314](https://github.com/apollographql/federation/pull/2314).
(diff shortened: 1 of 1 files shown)


subgraph-js/CHANGELOG.md
```diff
@@ -9,7 +9,7 @@
 This CHANGELOG pertains only to Apollo Federation packages in the 2.x range. The Federation v0.x equivalent for this package can be found [here](https://github.com/apollographql/federation/blob/version-0.x/subgraph-js/CHANGELOG.md) on the `version-0.x` branch of this repo.
 
 ## 2.3.0-beta.2
-- `@tag` directive support for the `SCHEMA` location [PR #2314](https://github.com/apollographql/federation/pull/2314).
+- `@tag` directive support for the `SCHEMA` location. This has been added to the 2.3 version of the federation spec, and you must declare usage of this spec version in your subgraph to use this functionality. [PR #2314](https://github.com/apollographql/federation/pull/2314).
```
trevor-scheer, line 12: Can we be extra clear on the action item that's required for users? (you need to bump your `@link` federation spec version to v2.3 in your type definitions)

Providing a snippet certainly would remove any uncertainty.
```
@link(url: "https://specs.apollo.dev/federation/v2.3", import: [...])
```

trevor-scheer (approved):
Probably also helpful to update the PR description to reference the issue

subgraph-js/CHANGELOG.md
sachindshinde: @trevor-scheer Makes sense, made it more explicit in the entry. (I also removed the `import` since it's not strictly needed to run into this issue.)

PR #2388, opened 2023-02-09
Add subscription support to `@apollo/subgraph`

martijnwalraven:
The changes needed to support subscriptions in `@apollo/subgraph` are fairly minimal: we need to augment the resolver typings, and propagate the `subscribe` property to `GraphQLFieldConfig`.

One thing to call out is that this PR follows the definitions in [`graphql-js`](https://github.com/graphql/graphql-js/blob/f201681bf806a2c46c4ee8b2533287421327a302/src/type/definition.ts#L924-L925). That means both `resolve` and `subscribe` are optional, even though it could make sense to require at most once. The property types are also rather loosely defined: there is no attempt to validate that `subscribe` returns an `AsyncIterator` for example.


trevor-scheer (changes requested):
The code changes seem reasonable to me. Would you mind adding a test case and a changeset (via `npx changeset`)?

hwillson:
👋 @martijnwalraven @trevor-scheer - is there anything left to do with this, or do you think we can go ahead and merge it?

trevor-scheer:
I pushed up a test and changeset so this is ready to go. Validated that the test fails before @martijnwalraven's changes (and it clearly passes with them).

PR #2389, opened 2023-02-10
subscriptions support in the query planner

clenfest:
Created a couple of new node types:

- SubscriptionPlanNode which is a top level node indicating a subscription
- SubscriptionNode which is a subscription version of a FetchNode


bnjjj:
@clenfest ping me when you think it's ready to merge if you want me to try it in the router directly before merging.

query-planner-js/src/QueryPlan.ts
```diff
@@ -45,6 +51,10 @@ export interface FetchNode {
   outputRewrites?: FetchDataOutputRewrite[];
 }
 
+export interface SubscriptionNode extends Omit<FetchNode, 'kind' | 'hasDefers'> {
```
bnjjj, line 54: I think you should also omit the field `id` because it's related to defer. I'm wondering if we should handle the `inputRewrites` and `outputRewrites` for a subscription ?

query-planner-js/src/QueryPlan.ts
```diff
@@ -26,6 +26,12 @@ export interface ParallelNode {
   nodes: PlanNode[];
 }
 
+export interface SubscriptionPlanNode {
+  kind: 'SubscriptionPlan';
```
bnjjj, line 30: To be more consistent I suggest to just use `subscription`. What do you think ?

composition-js/src/merging/merge.ts
```diff
@@ -2798,4 +2823,16 @@ class Merger {
...
     });
   }
+
+  private validateSubscriptionField(sources: FieldOrUndefinedArray) {
+    // no subgraph marks field as @shareable
+    const fieldsWithShareable = sources.filter((src, idx) => src && src.appliedDirectivesOf(this.metadata(idx).shareableDirective()).length > 0);
+    if (fieldsWithShareable.length > 0) {
+      const nodes = sourceASTs(...fieldsWithShareable);
+      this.errors.push(ERRORS.INVALID_FIELD_SHARING.err(
+        `Fields on root level subscription object cannot be marked as shareable`,
```
dariuszkuc, line 2833: is this temporary thing or we will support it in the future? (I thought the intent was to support this?)

query-planner-js/src/buildPlan.ts
```diff
@@ -2751,6 +2745,42 @@ function computePlanInternal({
     ({ main, deferred } = dependencyGraph.process(processor, operation.rootKind));
     primarySelection = dependencyGraph.deferTracking.primarySelection;
 
+    // If this is a subscription, we want to make sure that we return a SubscriptionPlanNode rather than a PlanNode
```
pcmanus, line 2748: I wouldn't have put this code here. Instead, I would have continued to return a `PlanNode` here, and instead deal with transforming it for subscriptions at the end of `buildQueryPlan`. Conceptually, the subscription handling here is a top-level concern: we generate plans as for queries, but then extract the primary fetch separately. So having that change done in `computePlanInternal` feels like the wrong place, and in practice this shows up in the added checks in `computePlanForDeferConditionals` which shouldn't be necessary.

query-planner-js/src/snapshotSerializers/queryPlanSerializer.ts
```diff
@@ -119,6 +120,21 @@ function printNode(
           indentation + '}'
       }
       break;
+    case 'SubscriptionPlan': {
+      const primary = node.primary;
+      const rest = node.rest;
+      const indentationInner = indentationNext + config.indent;
+      // result +=
```
pcmanus, line 127: Nit: remove this left-over?

pcmanus (commented):
A couple remarks:
- have we given though about what `@provides` and `@requires` within the "primary" fetch of a subscription? I mean, those "work" in the sense that the query plan will handle them in the "normal" way, but are we sure this is what we want for subscription? That is, my understanding of our design is that what is in the primary fetch will not be "refreshed" automatically on events unless users manually include those fields in the events themselves, but it feels like `@requires` and `@provides` makes it quite a bit harder for a user to understand what is "primary" (and need to be manually included in events if a refresh is needed) and what isn't (which fwiw, strikes me as the one thing maybe a bit complex UX-wise in the design even when `@requires` and `@provides` are not involved).
-  the patch expects that subscriptions have no `@defer`, but there doesn't seem to be any validation that this is not the case. The [current validation](https://github.com/apollographql/federation/blob/main/internals-js/src/operations.ts#L1888-L1897) does reject `@defer` on subscription root fields, but only on root fields while I think we want to exclude them from "anywhere in a subscription operation" instead.

query-planner-js/src/QueryPlan.ts
bnjjj: When lookin at what we have here https://github.com/apollographql/router/blob/dev/apollo-router/src/query_planner/plan.rs#L65 nothing has a suffix but not a big deal, I will write an alias on our side to have consistency there

query-planner-js/src/QueryPlan.ts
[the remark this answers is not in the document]
bnjjj: My 2cents, as a consumer of your api it's cleaner to have another specific node for subscription, it's easier and cleaner to integrate in our codebase for example. It's just my opinion from a consumer point of view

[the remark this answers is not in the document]
pcmanus: > it's cleaner to have another specific node for subscription

Just to be extra clear, I'm not suggesting removing `SubscriptionPlanNode`, so there will be specific node for subscription. But I disagree otherwise that, within that already specific node, we need a new type of node to encode the initial fetch. The initial fetch is a normal fetch. It will be send to subgraph completely normally. In particular, why a new type for this primary fetch, but `SubscriptionPlanNode.rest` uses completely normal `FetchNode` within it? What's the rational for that inconsistency?

> it's easier and cleaner to integrate in our codebase for example

I don't mean to sound obtuse, but I have worked on those part of the router and I'd argue the complete opposite.

In fact, my advise on the router side would be to deserialise this as a [`FetchNode`](https://github.com/apollographql/router/blob/dev/apollo-router/src/query_planner/fetch.rs#L64-L93) directly, because you _will_ have to execute it as a normal fetch ultimately (yet again, this _is_ a subgraph fetch, not something else), so if you don't deserialise it as a fetch, you will have to convert to one anyway (or duplicate *tons* of code unnecessarily) and I'm missing how that's cleaner and easier. That is, this doesn't look too bad here only because it's typescript, with the `Omit` and duck-typing; it's going to be more duplication in rust.

But anyway, as far as this typescript code goes, as said above, I would at least insist that this `SubscriptionNode` is not added to `PlanNode` unless there is a justification for that that I don't see. I do think not just reusing `FetchNode` for a fetch is a mistake, but I've obviously make my point enough, so I won't insist on that point anymore.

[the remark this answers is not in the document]
bnjjj: Ah ok, I was only talking about `SubscriptionPlanNode`, my bad. And about implementation in the router, it's not exactly the same than for a fetch node (it's not so far but not really the same). For the different modes we want (websocket and callback) you need to spawn a specific task and do some specific stuffs to communicate between tasks to sync and send to right data to the right client. Happy to show you how it's implemented if you want... But TLDR that's not just a call to a subgraph like for a fetch node, it's a little bit more than that so having a new node is cleaner IMHO

composition-js/src/merging/merge.ts
clenfest: I don't think we intend on subgraphs being able to merge their subscriptions in the supergraph. Let me know if I'm missing something.

clenfest:
> A couple remarks:
> 
> * have we given though about what `@provides` and `@requires` within the "primary" fetch of a subscription? I mean, those "work" in the sense that the query plan will handle them in the "normal" way, but are we sure this is what we want for subscription? That is, my understanding of our design is that what is in the primary fetch will not be "refreshed" automatically on events unless users manually include those fields in the events themselves, but it feels like `@requires` and `@provides` makes it quite a bit harder for a user to understand what is "primary" (and need to be manually included in events if a refresh is needed) and what isn't (which fwiw, strikes me as the one thing maybe a bit complex UX-wise in the design even when `@requires` and `@provides` are not involved).
> * the patch expects that subscriptions have no `@defer`, but there doesn't seem to be any validation that this is not the case. The [current validation](https://github.com/apollographql/federation/blob/main/internals-js/src/operations.ts#L1888-L1897) does reject `@defer` on subscription root fields, but only on root fields while I think we want to exclude them from "anywhere in a subscription operation" instead.

Correct me if I'm wrong, but I believe that the call to `operation.withoutDefer();` on line 2515 should strip out all the defers.

pcmanus (approved):
Asides from a few nitpicks that would be nice but are unimportant, I'd appreciate:
- renaming `SubscriptionPlanNode` into `SubscriptionNode` now. It's genuinely feel cleaner to me.
- the point about @defer below.
But none are blockers per-se, so pre-emptively approving.

> Correct me if I'm wrong, but I believe that the call to `operation.withoutDefer();` on line 2515 should strip out all the defers.

Yes, it will. But my remark was more a UX suggestion. What I mean is that afaict, the behaviour with the current patch (in the router) would be:
1. for non-subscription operation, a @defer always does something (even when subgraph fetches can't be quite deferred, we still at least defer the response proc❯
2. for subscription operation, if the `@defer` is on a root field, then we reject it (as pointed in my previous comment).
3. for subscription operation, if the `@defer` is not on a root field, then this patch ignores it.

Now, point 3 is technically ok from a spec perspective, but from a consistency POV, merging 2 and 3 by always rejecting @defer within subscription operations wo❯

Note that rejecting @defer instead of ignoring it is, I believe, pretty trivial code wise: instead of skipping [this check for subscription](https://github.com/apollographql/federation/pull/2389/files#diff-4cca031169a6c6e3c5bbbb42ad4c29ddbb09e682e328ce90cf05f7782faf0169R2533), we could let it run but then error if `hasDefers && isSubscription`.

Anyway, I'd personally prefer rejecting @defer instead of ignoring it, but certainly not a huge point.

PR #2390, opened 2023-02-10
Improve deprecation script

trevor-scheer:
The script was struggling to run deprecate on so many things - OTP would expire before it finished, or the token would hit a rate limit.

I found out that `npm deprecate` accepts a version _range_ which means we can run just 4 commands instead of a command for every 0.x version that exists. Much better 😄


dariuszkuc (approved):
nice!

PR #2391, opened 2023-02-12
Docs for interfaceObject

StephenBarlow:
See primarily new file `interfaces.mdx`—other edits are primarily tidying

### DIFF
--- a/docs/source/federated-types/federated-directives.mdx
+++ b/docs/source/federated-types/federated-directives.mdx
@@ -176,7 +176,9 @@ The default value is `true`.
 directive @interfaceObject on OBJECT
 ```
 
-Indicates that an object definition serves as an extension of another subgraph's **entity interface**. This enables a subgraph to polymorphically contribute fields to _all_ entities that implement a particular entity interface.
+Indicates that an object definition serves as an abstraction of another subgraph's **entity interface**. This abstraction enables a subgraph to automatically contribute fields to _all_ entities that implement a particular entity interface.
+
+During composition, the fields of every `@interfaceObject` are added both to their corresponding `interface` definition _and_ to all entity types that _implement_ that interface.
 
 > [Learn more about entity interfaces.](./interfaces/)
 
--- a/docs/source/federated-types/federated-directives.mdx
+++ b/docs/source/federated-types/federated-directives.mdx
@@ -168,6 +168,18 @@ The default value is `true`.
 </tbody>
 </table>
 
+### `@interfaceObject`
+
+> **⚠️ This directive is available in Apollo Federation v2.3 and later.**
+
+```graphql
+directive @interfaceObject on OBJECT
+```
+
+Indicates that an object definition serves as an extension of another subgraph's **entity interface**. This enables a subgraph to polymorphically contribute fields to _all_ entities that implement a particular entity interface.
+
+> [Learn more about entity interfaces.](./interfaces/)
+
 ### `@extends`
 
 ```graphql
--- a/docs/source/federated-types/federated-directives.mdx
+++ b/docs/source/federated-types/federated-directives.mdx
@@ -110,7 +110,7 @@ type Product @key(fields: "upc") @key(fields: "sku") {
 
 To check whether your subgraph library supports repeatable directives, see the `repeatable @key` item in [Federation-compatible subgraph implementations](../building-supergraphs/supported-subgraphs/).
 
-> ⚠️ **Do not apply `@key` to interface definitions.** The `INTERFACE` location is included in this definition for backward compatibility with Federation 1 subgraphs. Currently, applying `@key` to an interface has no effect in _any_ version of federation, and Federation 2 subgraphs produce a composition error to emphasize this.
+**In Apollo Federation 2.3 and later,** you can also apply `@key` to `interface` definitions to create [entity interfaces](./interfaces/). If you apply `@key` to an `interface` in _earlier_ versions of Federation 2, a composition error occurs.
 
 #### Arguments
 
@@ -170,7 +170,7 @@ The default value is `true`.
 
 ### `@interfaceObject`
 
-> **⚠️ This directive is available in Apollo Federation v2.3 and later.**
+> **⚠️ This directive is available in Apollo Federation 2.3 and later.**
 
 ```graphql
 directive @interfaceObject on OBJECT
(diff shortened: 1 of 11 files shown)


docs/source/federated-types/federated-directives.mdx
```diff
@@ -168,6 +168,18 @@ The default value is `true`.
...
 
+### `@interfaceObject`
+
+> **⚠️ This directive is available in Apollo Federation v2.3 and later.**
+
+```graphql
+directive @interfaceObject on OBJECT
+```
+
+Indicates that an object definition serves as an extension of another subgraph's **entity interface**. This enables a subgraph to polymorphically contribute fields to _all_ entities that implement a particular entity interface.
```
korinne, line 179: I will say that Sylvain really does not like the word _extends_ for `@interfaceObject`, as he thinks of it more as an abstraction of the actual interface, rather than extending it. It's tricky, because I think "extends" is a quicker teaching moment to users, but it's not accurate in this case. 

For the definition in this case, I would say: "Indicates that an object definition serves as a local abstraction of another subgraph's **entity interface**. During composition, object types with the `@interfaceObject` annotation and the interface with the corresponding `@key` will be merged. This enables a subgraph to polymorphically contribute fields to _all_ entities that implement a particular entity interface. Additionally, all fields contributed by types marked by `@interfaceObject` will be added to the definition of each type implementing the entity interface in the supergraph API." 

Feel free to ignore or change as needed!

korinne (changes requested):
Added one comment, but otherwise looks good to me. We might need to revisit at some point to add in more detail, but as a first pass -- this should be good! The only thing that worries me is that "local abstraction of interfaces" is a really big part of understanding how `@interfaceObject` works in a subgraph, but it isn't mentioned in the docs here. Perhaps after GA we can revisit depending on any user feedback. Thanks again for putting this together!

docs/source/federated-types/interfaces.mdx
```diff
@@ -0,0 +1,315 @@
...
+    }
+    ```
+
+- _Every_ subgraph that defines `MyInterface` as an object type **must**:
+    - Apply `@interfaceObject` to its definition
+    - Include the exact same `@key`(s) as the _interface_ type's definition
+- Subgraph B **must not** _also_ define `MyInterface` as an _interface_ type.
+- Subgraph B **must not** define _any_ entity that _implements_ `MyInterface`.
+    - If a subgraph contributes entity fields via `@interfaceObject`, it "gives up" the ability to contribute fields to any _individual_ entity that implements that interface.
+    - This is a current limitation of entity interfaces.
```
patrick91, line 169: will we fix this in future?

docs/source/federated-types/interfaces.mdx
```diff
@@ -0,0 +1,315 @@
...
+
+
+- `@shareable` checks/rules on fields of   `Product` are performed “as if” the fields provided by `Product` in `reviews` were provided by all the types implementing `Product` in the supergraph.
```
patrick91, line 172:
```suggestion
- `@shareable` checks/rules on fields of `Product` are performed “as if” the fields provided by `Product` in `reviews` were provided by all the types implementing `Product` in the supergraph.
```

patrick91: I'm not sure we are talking about the product type in this file other than this line

docs/source/federated-types/interfaces.mdx
```diff
@@ -0,0 +1,315 @@
...
+
+```graphql title="Subgraph A"
+interface Media @key(fields: "id") {
+  id: ID!
+  title: String!
+}
+```
+
+As it does with any [standard entity](../entities/#1-define-a-key), `@key` indicates "this subgraph can resolve any instance of this type if provided its `@key` fields." This means Subgraph A needs to define a **reference resolver** for `Media`, just as it would for any other entity.
+
```
patrick91, line 189: ah, I didn't understand this!

docs/source/federated-types/interfaces.mdx
```diff
@@ -0,0 +1,315 @@
...
+// ....other resolvers ...
+```
+
+In this example, the hypothetical variable `allMedia` contains all `Media` data, including each object's `id`.
+
+
+#### `@interfaceObject` resolvers
+
+In [the example schemas above](#example-schemas), Subgraph B defines `Media` as an object type and applies `@interfaceObject` to it. It also defines a `Query.bestRatedMedia` field:
```
patrick91, line 209: don't we need a reference resolver in subgraph b too?

docs/source/federated-types/interfaces.mdx
```diff
@@ -0,0 +1,315 @@
...
+```
+
+To prevent these composition errors, the teams maintaining Subgraph A and Subgraph B need to coordinate their schema changes _every time_ a new implementation of `Media` is created. Imagine how complex that coordination becomes if the definition of `Media` is instead distributed across _ten_ subgraphs!
+
+In summary, Subgraph B shouldn't need to know every possible _kind_ of `Media` that exists in your supergraph. Instead, it should generically know how to fetch reviews for _any_ kind of `Media`. _This_ is the relationship that entity interfaces and `@interfaceObject` provide, as demonstrated in the [example above](#example-schemas).
+
+
+### Are there alternatives to using `@interfaceObject`?
+
+The primary alternative to using `@interfaceObject` is to use the discouraged strategy described in the [previous section](#why-is-interfaceobject-necessary). This requires duplicating _all_ implementations of a given interface in each subgraph that contributes fields to that interface.
```
lennyburdette, line 315: I'm not sure where to put it, but the caveat to this strategy that that subgraph must be able to resolve the interface's type. If the upstream data store only contains the key fields, but doesn't know if something if a Book or Movie, this strategy isn't even possible.

lennyburdette (approved):

docs/source/federated-types/interfaces.mdx
StephenBarlow: ahh good catch this is a line from the previous draft that I forgot to edit 😬

StephenBarlow: Going to cut this line for now, and we'll confirm the implications here w/ @pcmanus for a followup edit

docs/source/federated-types/interfaces.mdx
StephenBarlow: Good callout, I'm going to cut this line due to the fundamentally uncertain nature of _the future_

korinne (approved):
Looks great to me!

PR #2396, opened 2023-02-14
Add missing changelog for `@apollo/gateway@2.2.3`

michael-watson:
The `@apollo/gateway@2.2.3` package doesn't have an associated changelog for it. You can see the comparison of the two versions [here](https://github.com/apollographql/federation/compare/%40apollo/gateway%402.2.2...%40apollo/gateway%402.2.3).

This PR adds these details into the changelog


gateway-js/CHANGELOG.md
```diff
@@ -62,6 +62,11 @@ This CHANGELOG pertains only to Apollo Federation packages in the 2.x range. The
 - Adds support for `@interfaceObject` and keys on interfaces [PR #2279](https://github.com/apollographql/federation/pull/2279).
 - Preserves source of union members and enum values in supergraph [PR #2288](https://github.com/apollographql/federation/pull/2288).
 
+## 2.2.3
+
+- Fix possible assertion error during query planning [PR #2299](https://github.com/apollographql/federation/pull/2299)
```
trevor-scheer, line 67: Will you delete line 54? It looks like this was released from a separate branch which never made it to `main`, but the PR landed to main and was released as part of 2.3.0 as well. So the same fixes were landed in parallel to 2.2.3 and 2.3.0.

trevor-scheer (approved):

PR #2401, opened 2023-02-16
docs: update latest subgraph compatibility results

dariuszkuc:


dariuszkuc:
@StephenBarlow in the [testing repo](https://github.com/apollographql/apollo-federation-subgraph-compatibility) I've been wrapping all the test names (which generally just specify directive name) with `<code>` tags to avoid constantly mentioning `@`<someone> (yeah there are GH users that have same names as our directives). GH escapes those when rendering index markdown but not on comments/issues etc.

Let me know if this change is fine for this repo as well.

StephenBarlow (approved):

PR #2406, opened 2023-02-17
Node 12 compatibility testing (version-0.x)

trevor-scheer:
Part of #2413

Renovate has a handful of passing major version PRs which shouldn't be. Most of the major bumps drop support for node 12, and yet the node 12 tests still pass. It seems like as long as the lockfile is updated successfully, subsequent invocations of `npm i` will succeed (and renovate manages to update the lockfile successfully - either by ignoring engine warnings or some other method).

Landing this PR should turn a bunch of green v0.x PRs red, and we should then block those in the renovate config.


gateway-js/package.json
```diff
@@ -28,9 +28,10 @@
     "@apollo/core-schema": "^0.2.0",
     "@apollo/federation": "file:../federation-js",
     "@apollo/query-planner": "file:../query-planner-js",
-    "@apollo/server-gateway-interface": "^1.0.2",
+    "@apollo/usage-reporting-protobuf": "^4.0.2",
```
trevor-scheer, line 31: The `@apollo/server-gateway-interface` package actually requires node 14 so I removed it and copied the code into the repo instead. I don't think we'll be making any changes to the interface that affect v0 so this should be ok.

clenfest (approved):

PR #2420, opened 2023-02-23
Query planner skips fetches when possible (based on @skip and @include usages)

NoobMaster-96:

### DIFF
--- a/query-planner-js/src/buildPlan.ts
+++ b/query-planner-js/src/buildPlan.ts
@@ -92,8 +92,23 @@ import {
   createInitialOptions,
   buildFederatedQueryGraph,
 } from "@apollo/query-graphs";
-import { stripIgnoredCharacters, print, parse, OperationTypeNode } from "graphql";
-import { DeferredNode, FetchDataInputRewrite, FetchDataOutputRewrite } from ".";
+import {
+  stripIgnoredCharacters,
+  print,
+  parse,
+  OperationTypeNode,
+  DirectiveNode,
+  SelectionNode,
+  //ArgumentNode,
+  BooleanValueNode,
+  VariableNode,
+  //ValueNode,
+} from "graphql";
+import {
+  DeferredNode,
+  FetchDataInputRewrite,
+  FetchDataOutputRewrite
+} from ".";
 import { enforceQueryPlannerConfigDefaults, QueryPlannerConfig } from "./config";
 import { generateAllPlansAndFindBest } from "./generateAllPlans";
 import { QueryPlan, ResponsePath, SequenceNode, PlanNode, ParallelNode, FetchNode, trimSelectionNodes } from "./QueryPlan";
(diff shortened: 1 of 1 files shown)


pcmanus:
Hi @NoobMaster-96. Thanks for your interest on this issue. I've had a quick look at your PR, but I'm afraid there is unfortunately a bit more complexity involved here for a few reasons:
- The assumption that there is [always only one top-level selection](https://github.com/apollographql/federation/pull/2420/files#diff-4cca031169a6c6e3c5bbbb42ad4c29ddbb09e682e328ce90cf05f7782faf0169R1270) is unfortunately not correct: there can be multiple top-level selections in some cases (see [this test](https://github.com/apollographql/federation/blob/main/gateway-js/src/__tests__/executeQueryPlan.test.ts#L1949-L1957) for instance  but there is more examples in this file). So that needs handling.
- Sometimes the conditions are not all on the top-level selection (due to `@include` and `@skip` not being repeatable). For instance, a selection can look like `... on T @skip(if: $cond1) { ... on T @skip(if $cond2) { ... } }`. I think it's worth "extracting" `$cond2` too, not just `$cond1` (but I don't think the PR does this).
- The execution of `ConditionNode` is unfortunately not currently handled by the Gateway. This is because so far `ConditionNode` is only used for handling `@defer` which the gateway does not support. But if we generate `ConditionNode` for `@include`/`@skip`, we need to handle them and so this needs adding.
- It's a more minor details, but in principle generating a `ConditionNode` only around the immediate `FetchNode` having the conditions is not optimal. Ideally the `ConditionNode` would also include any other fetch that are direct descendant of that `FetchNode`. This is not a huge deal because descendant nodes would also have the conditions anyway so they would be wrapped into their own `ConditionNode`, but it's ever so slightly inefficient to re-check the condition multiple time when we could do it just once. This will also lead to slightly less clean and bigger plans.

Now, I'll mention that I had started implementing an approach that deal with those issue some times ago but I got distracted and didn't finish it. I've resumed work on it and I'm mostly done so I should have a PR for this shortly (hopefully next week) if that's ok with you.

NoobMaster-96:
@pcmanus thanks for taking a proper look into the PR. It will be really helpful if you can pick this up quickly. This is a really important resilience feature for us and would help our systems.

NoobMaster-96:
@pcmanus can you share that PR here as well once you are done? Also wanted to know about the ETA for this.

pcmanus:
@NoobMaster-96 Sorry for missing your earlier message. The PR is at https://github.com/apollographql/federation/pull/2424. As for ETA, my hope is that we get this in relatively quickly and that it's part of the 2.4 release, which I don't have an exact date for, but we're target mid-to-late March.

pcmanus:
We just merged #2424, so closing this. Fwiw, we should have an alpha of 2.4 out either today or early next week that will allow to test this.

PR #2424, opened 2023-02-24
Generates condition nodes for @include/@skip when possible

### LINKED ISSUE #2251: Pull fetches top-level `@include`/`@skip` into condition nodes
This ticket is essentially about porting the feature added in #1113 for federation 1 in federation 2, but adapting it to federation 2 specific.

Essentially, when `@include`/`@skip` are involved, it can happen that some `FetchNode` in the plan ends up with some top-level(s) `@include`/`@skip`, meaning that the whole fetch is conditional on some condition(s). In such case, we can pull up those conditions out of the `FetchNode` selection set and into a specific query plan construct so that the gateway/router can possibly entirely avoid the fetch (saving a bit of load on the underlying service in that case).

I'll note that federation 1 implementation of this uses a new somewhat adhoc `inclusionConditions` parameter to `FetchNode`, but in federation 2, we added generic `ConditionNode` with the implementation of `@defer`, so we should reuse this (we might need to generalize slightly was we have to handle this cleanly).

pcmanus:
When a fetch to a subgraph is entirely included inside some `@include`/`@skip`, we can use the `ConditionNode` introduced for `@defer` to completely save the fetch when it's not included/is skipped.

This commit detects when a fetch can be "wrapped" in one or more `ConditionNode` due to `@include`/`@skip` and generates the plan accordingly.

Additionally, this commit adds support for `ConditionNode` in the gateway execution code, since it was not handled before.

Fixes #2251


pcmanus:
I'll note that commit here is essentially ready for review, but I'm leaving this as a draft a bit for 2 reasons:
1. this adds support for executing `ConditionNode` in the gateway, but we can't properly report about those in traces until we modify https://github.com/apollographql/apollo-server/blob/main/packages/usage-reporting-protobuf/src/reports.proto to handle that new node. Right now, the patch simply don't report anything about the condition, and maybe that's ok initially, maybe we can merge as is and improve in a followup, but I'd like to have a quick chat about this with people that know traces before deciding.
2. while this patch does do what #2251 advertises, there is some cases where we still do fetches that we could avoid. This is because sometimes we have fetches that looks like this:
    ```graphql
    ... on T1 @include(if: $cond1) {
      <some stuff>
    }
    ... on T2 @include(if: $cond2) {
      <some stuff>
    }
    ```
    but we can't easily use a `ConditionNode` for this case because only parts of the fetch is covered by each condition. More precisely, we "could" use some `ConditionNode` with multiple branches that each include the sub-part they need, but this gets messy quickly, especially considering that there can be following fetches that depend on this one fetch (it could make some query plans a log bigger in particular). A better way is really to include the conditions (the original `@include` and `@skip` really) in the inputs selection of the fetch and handle that during execution. The result will be that when the gateway computes the inputs of the fetch, it will filter out any entity that isn't included, so that the fetch is only send for those entities. And if that filtering results in no entities, we just don't do the fetch. Anyway, I "think" doing this is relatively simple so I'd like to try it and include it here, as a way to somewhat "close the loop" on handling `@include` and `@skip` efficiently. That said, the first commit here can still be reviewed in the meantime.

gateway-js/src/executeQueryPlan.ts
```diff
@@ -326,12 +327,26 @@ async function executeNode(
       }
       return new Trace.QueryPlanNode({ fetch: traceNode });
     }
+    case 'Condition': {
+      // TODO: we probably should a add a proper protobuf message for this in https://github.com/apollographql/apollo-server/blob/main/packages/usage-reporting-protobuf/src/reports.proto
```
trevor-scheer, line 331: This is fixed in the latest version of the protobuf package (just published today)

query-graphs-js/src/pathContext.ts
```diff
@@ -8,12 +8,24 @@ export function isPathContext(v: any): v is PathContext {
   return v instanceof PathContext;
 }
 
-function addExtractedConditional(operation: OperationElement, directiveName: string, addTo: [string, any][]) {
-  const applied = operation.appliedDirectivesOf(directiveName);
+export type OperationConditional = {
+  kind: 'include' | 'skip',
+  value: any,
```
trevor-scheer, line 13: `'true'`, `'false'`, `'$foo'` - are you expecting something other than a string here? (should t/f be booleans?)
```suggestion
  value: string,
```

```suggestion
  value: string | boolean,
```

query-planner-js/src/conditions.ts
```diff
@@ -0,0 +1,229 @@
...
+
+export type VariableCondition = {
+  variable: Variable,
+  negated: boolean,
+}
+
+export type Condition = VariableCondition | boolean;
+
+// The invariant maintained on this type are that if is an array of variable conditions, then:
+// 1. that array is not empty (it has at least one condition).
```
trevor-scheer, line 20: Possibly interesting, you can enforce a non-empty array at the type level:
```ts
type NonEmptyArray<T> = [T, ...T[]]
```

trevor-scheer (changes requested):
Looking good! I appreciate all the new tests 👍

A few additional testing suggestions:
* Cases where both `@skip` and `@include` are applied on the same selection as constants and variables
* Skip/include directive usages on fragment spreads and inline fragments

gateway-js/src/executeQueryPlan.ts
pcmanus: Nice. Added.

query-graphs-js/src/pathContext.ts
pcmanus: It's actually properly "parsed" values at this point, so the proper type here is `Variable | boolean`, but you're right, we can do better than `any`. Updated.

query-planner-js/src/conditions.ts
pcmanus: Interesting. I wonder how well that actually work on more complex example, but I guess it works ok here so included a commit that adds that.

pcmanus:
Getting this out of draft mode since the trace issue has been resolved, and upon further consideration, the followup idea I mentioned in my previous comment is probably better done in ... a followup PR.

I had previously forgotten to handle variable default values, so I added code to handle it and a test, but at the time of this writing, that test fails due to #2443. So as soon as the later is merged, I'll rebase and CI should be happy. This is "ready" otherwise.

> A few additional testing suggestions:
> 
> * Cases where both `@skip` and `@include` are applied on the same selection as constants and variables

Added a bunch of tests for those.

> * Skip/include directive usages on fragment spreads and inline fragments

Added some for spreads, good call. Inline fragments are heavily used in the existing tests however, so I didn't had anything new for it.

gateway-js/src/__tests__/executeQueryPlan.conditions.test.ts
```diff
@@ -581,7 +580,196 @@ describe('Execution tests for @include/@skip', () => {
...
+                    }
+                  }
+                },
+              }
+            },
+          },
+        }
+      `);
+
+      s2Queries.length = 0;
```
trevor-scheer, line 637: This and the same below seems fine, it's just unconventional. My eyes were looking for a mistake (not sure I've seen `length` assigned in practice). If you do change it, `s2Queries = [];` would be more typical. You're welcome to ignore this comment.

trevor-scheer (approved):
LGTM (once tests are passing)

gateway-js/src/__tests__/executeQueryPlan.conditions.test.ts
```diff
@@ -557,31 +557,219 @@ describe('Execution tests for @include/@skip', () => {
...
+    it('handles condition on named fragments', async () => {
+      const operation = parseOperation(schema, `
+        query ($if: Boolean!){
+          t {
+            id
+            ... GetB1
+          }
+        }
+
+        fragment GetB1 on T1 @include(if: $if) {
```
trevor-scheer, line 595: I don't think this is a valid location for `@include`

From the spec:
```graphql
directive @include(if: Boolean!) on FIELD | FRAGMENT_SPREAD | INLINE_FRAGMENT
```

pcmanus: Oops, indeed. I meant to test the include on the spread itself, and changed the test to do that.

Side-note: if one wonder why the test was passing, this is because those execution tests don't use the full gateway but instead call the `buildQueryPlan` and `executeQueryPlan` "manually". As such, they don't run the full gamut of query validations. In practice, _most_ validations are done by `parseOperation` but there is a handful of "minor" ones that were left out as a shortcut, and validating directive locations is one of the left out (it might make sense to add them someday, but it's not really urgent since queries _are_ fully validated in the gateway and so this would be double-validation anyway). On the plus side, if the graph spec decides that allowing `@include` on named fragment definition makes sense, we know it works :).

Issue #2315, opened 2023-01-05
ability to specify core features whose directives will always compose into supergraph


clenfest:
As a part of upcoming Polaris work, they will want to ensure that new directives they design will compose. Rather than forcing each subgraph to specify `@composeDirective`, we should have the mechanism to add the core feature URL to a list (in code) and have all directives from that feature always compose into the supergraph.

PR #2431, opened 2023-03-02
support for composing federation directives into supergraph

clenfest:
Added some code to allow us to compose certain future federation directives automatically.

- There aren't any composition tests, those will come when we actually add a directive that uses this infrastructure, but should be fairly straightforward.
- You could maybe make the argument that the syntax is a little verbose. For example matching exact could potentially be implicit, COLLAPSE could potentially be inferred. But this is an internal function and I don't mind ensuring specificity so that whoever adds a directive make a choice at every step.

Refs #2315

### DIFF
--- a/internals-js/src/directiveCompositionRules.ts
+++ b/internals-js/src/directiveCompositionRules.ts
@@ -1,0 +1,255 @@
+import { DirectiveLocation } from 'graphql';
+import { Directive, DirectiveDefinition, FieldDefinition, NamedSchemaElement, ObjectType, Schema } from './definitions'
+import { assert, isDefined, isNotNull } from './utils';
+
+export enum DirectiveCompositionStrategy {
+  COLLAPSE = 'collapse',
+  COLLAPSE_FROM_ALL = 'collapseFromAll',
+  REMOVE_DUPLICATES = 'removeDuplicates',
+}
+
+export enum DirectivePropagationStrategy {
+  INHERIT_FROM_OBJECT = 'inheritFromObject',
+  CONSISTENT_LOCATION = 'consistentLocation',
+}
+
+export enum FieldPropagationStrategy {
+  MAX = 'max',
+  MIN = 'min',
+  SUM = 'sum',
+  AVERAGE = 'average',
+  AND = 'and',
+  OR = 'or',
+  INTERSECTION = 'intersection',
+  UNION = 'union',
+  EXACT = 'exact',
+}
+
+const SUPPORTED_LOCATIONS = [
+  DirectiveLocation.FIELD_DEFINITION,
+  DirectiveLocation.OBJECT,
+];
+
+export class FederationDirectiveCompositionManager {
+  private readonly directiveNameLookup: Map<string, string>[];
+  constructor(
+    readonly schemas: readonly Schema[],
+    readonly entries: DirectiveCompositionEntry[],
+  ) {
+    // we need to create a map so that for each directive definition, we need to what the name is in each schema
+    this.directiveNameLookup = new Array(schemas.length);
+    for (let i = 0; i < schemas.length; i += 1) {
+      this.directiveNameLookup[i] = new Map();
+      schemas[i].directives().forEach(directive => {
+        const nameInSchema = schemas[i].coreFeatures?.getByIdentity('federation')?.directiveNameInSchema(directive.name) ?? directive.name;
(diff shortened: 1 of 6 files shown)


internals-js/src/directiveCompositionRules.ts
```diff
@@ -0,0 +1,255 @@
...
+  SUM = 'sum',
+  AVERAGE = 'average',
+  AND = 'and',
+  OR = 'or',
+  INTERSECTION = 'intersection',
+  UNION = 'union',
+  EXACT = 'exact',
+}
+
+const SUPPORTED_LOCATIONS = [
```
dariuszkuc, line 28: should this include some other locations? e.g. interface, enums etc?

clenfest: My thought was that we'd expand support on an as-needed basis. For the stuff that's upcoming, I think OBJECT and FIELD_DEFINITION is sufficient.

lrlna: The cost directive will be on  ARGUMENT_DEFINITION, ENUM, FIELD_DEFINITION, INPUT_FIELD_DEFINITION, OBJECT and SCALAR. Does that make a difference here?

clenfest: Yeah, saw that once you sent me the definition. I'll make the changes to support those types as well.

internals-js/src/directiveCompositionRules.ts
```diff
@@ -0,0 +1,176 @@
+import { DirectiveLocation } from 'graphql';
+import { Directive, DirectiveDefinition, NamedSchemaElement, Schema } from './definitions'
+import { assert, assertUnreachable, isDefined } from './utils';
+
+export enum FieldPropagationStrategy {
```
pcmanus, line 5: I'd rename to `FieldArgumentMergeStrategy` or the like. `Argument` because it doesn't apply to the entire field, but rather  to argument values (a field with multiple arguments will declare more than one of this). And `s/Propagation/Merge/` because this really is just a sub-part of merging/composition and we never use "propagation" for that process in other places.

pcmanus (changes requested):
I think we can and really should integrate this with the existing code much more. Afaict, we're trying to solve 2 problems here:
1. makes it a tiny bit easier to add new (hard-coded) directives that gets composed automatically. I'll note that we already have such directives, `@tag` and `@inaccessible`, and they get merged automatically because they are in [this list](https://github.com/apollographql/federation/blob/main/composition-js/src/merging/merge.ts#L228). So all it would take for any other future directive would be to add it to this list, which is simple enough, so it's even debatable that there is a problem to solve here. That said, adding such future directive would require modifying 2 files, since the directive itself would have to be added in, say, `federationSpec.ts` (or another spec) and added to this list in `merge.ts`, so let's say that we want to refactor things a bit so only 1 file/place needs to be modified.
2. we want to add a new code path/behaviour to the merging of non-repeatable directives with arguments, so that when 2 subgraphs have applications on the same element with different argument values, then instead semi-randomly picking one of the application, we instead "merge" the argument values using one of a set of pre-defined strategies. We want to start by supporting this only for hardcoded directives at first, but mostly just because it simplifies the declaration of which directive use which pre-defined strategy for which argument.

The current PR ignores what we already have for `@tag` and `@inacessible` (I'll note on that front that the new code does not allow the future merged directives to be proper "core" directives in the supergraph, which this is handled for `@tag` and `@inaccessible` and I do think we want to handle it for future directives), and ignores the existing merging code for directive applications, creating a completely new path (which is bad for code maintenance/reuse in general and it'll make it harder to extend to directives merged by `@composeDirective` later in particular). I think the code should instead integrate [here](https://github.com/apollographql/federation/blob/main/composition-js/src/merging/merge.ts#L2447) (and it's btw probably worth keeping an hint even when we use the new strategies; whenever we get to allowing turning specific hints into errors, it will let users refuse merging on inconsisent arguments, which some users may prefer enforcing).

internals-js/src/directiveCompositionRules.ts
clenfest: Mainly so that if someone tries to add a directive with an unsupported location, they'll know there's a problem right away rather than assuming it works.

pcmanus: But what "problem" are we talking about here? Why would any location be a problem for any of this?

PR #2434, opened 2023-03-03
docs: update link imports to reference fed v2.3

dariuszkuc:
Our docs currently reference Federation spec v2.0. This is confusing to our users as they may not be aware that there are newer spec releases that introduced some additional functionality.

### DIFF
--- a/docs/source/entities-advanced.mdx
+++ b/docs/source/entities-advanced.mdx
@@ -206,7 +206,7 @@ We're done! `Bill` is now resolved in a new subgraph, and it was resolvable duri
 
 ```graphql {3} title="Billing subgraph"
 extend schema
-  @link(url: "https://specs.apollo.dev/federation/v2.0",
+  @link(url: "https://specs.apollo.dev/federation/v2.3",
         import: ["@key", "@shareable", "@override"])
 ```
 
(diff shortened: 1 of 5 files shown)


dariuszkuc:
We also explicitly link to Fed v2.0 version in [moving to federation 2](https://github.com/apollographql/federation/blob/main/docs/source/federation-2/moving-to-federation-2.mdx) and [quickstart setup](https://github.com/apollographql/federation/blob/main/docs/source/quickstart/setup.mdx) but I think those probably should remain as is.

clenfest:
Great job, but I think you missed a couple of files. `moving-to-federation-2.mdx` and `setup.mdx` had a couple of instances as well.

dariuszkuc:
> Why don't you want to change them?
I didn't update `moving-to-federation-2` as it talks about updating from fed v1 to v2 (we probably should have separate entries for updating from v2.0 to v2.1 and so on).
As for the the quickstart setup - it talks about apps deployed to heroku. Guess we probably should update those apps first?

clenfest:
> 

There's no reason that someone moving from v1 to v2 shouldn't immediately go to 2.3. It's just more functionality, the upgrade steps wouldn't be any different.

StephenBarlow (approved):

StephenBarlow:
Went ahead and merged this despite discussion around the couple remaining `link`s because the changes here are independently helpful and we can iterate further

PR #2437, opened 2023-03-04
Update Apollo Server to v4 (test code only)

trevor-scheer:
Update all of our tests to use AS4. This doesn't affect users (v3 will continue to work with gateway and v4 has been working).

### DIFF
--- a/gateway-js/src/__tests__/executeQueryPlan.test.ts
+++ b/gateway-js/src/__tests__/executeQueryPlan.test.ts
@@ -24,7 +24,7 @@ import {
   GraphQLResolverMap,
 } from '@apollo/subgraph/src/schema-helper';
 import {GatewayExecutionResult, GatewayGraphQLRequestContext} from '@apollo/server-gateway-interface';
-import assert from 'assert';
+import { unwrapSingleResultKind } from './testUtils';
 
 expect.addSnapshotSerializer(astSerializer);
 expect.addSnapshotSerializer(queryPlanSerializer);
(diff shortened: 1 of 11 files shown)


gateway-js/src/__tests__/executeQueryPlan.test.ts
```diff
@@ -1349,12 +1349,11 @@ describe('executeQueryPlan', () => {
         query,
       });
 
-      expect(response.data).toBeUndefined();
-      expect(response.errors).toMatchInlineSnapshot(`
-        Array [
-          [ValidationError: Cannot query field "ssn" on type "User".],
-        ]
-      `);
+      assert(response.body.kind === 'single');
```
pcmanus, line 1352: Suggestion: unpacking response is done in a few (test) places and it a tad more annoying now. Maybe a simple test helper that extract `data` and `errors` from the response would make things more readable. Fwiw, I just added a [`testUtils.ts`](https://github.com/apollographql/federation/blob/next/gateway-js/src/__tests__/gateway/testUtils.ts) recently for such helpers which could be a place for this.

pcmanus (commented):
A few very minor "stylistic" suggestions and with the caveat that I'm not super familiar with AS4 changes, but lgtm to me otherwise.

I do think it should be rebased on `next` however (I don't think this should go to a patch release) and this will require a bit of merge conflict resolution because I just did some minor refactoring to a few gateway tests that conflicts with this in #2441.

Also, AS4 has a number of breaking changes and I wonder if there is something specific that is particularly relevant to gateway usage that we want to call out in our changelog? (of course, I assume we want at least point to https://www.apollographql.com/docs/apollo-server/migration).

pcmanus:
> I don't think this should go to a patch release

> Also, AS4 has a number of breaking changes and I wonder if there is something specific that is particularly relevant to gateway usage that we want to call out in our changelog?

Now that I've had my 2nd coffee, I'm realising that all we're doing here is changing to AS4 _in tests_, and that this has no impact on actual users (which can continue using AS3 after this, and/or could have upgraded to AS4 without waiting for this). So disregard my remarks about changelog. And I guess having this is in a patch release would be ok too, though maybe still rebase on `next` to avoid having to deal with the conflicts I mentioned in my previous comment later?

pcmanus (approved):

trevor-scheer:
No change to runtime code or deps so skipping the changeset for this one.

PR #2441, opened 2023-03-07
Adds debug query planner option to skip query planning for a single s…

pcmanus:
…ubgraph

This commit adds a new option to the query planner config that, when used and when the supergraph is built from only a single subgraph, skip the query planning algorithm and instead generates a single fetch (to the one subgraph) that directly uses the gateway "input" query.

This option is disabled by default and marked as a "debug" option as it is a bit of a hack and so we're not suggesting committing to supporting it forever. But it is easy enough to have now and can be convenient at least for testing/debugging.

Note in particular that, when used, some of the features enabled by the query planner (for instance `@defer`) cannot be enabled. The option will also stop having any effect as soon as another subgraph is added, not matter what that 2nd subgraph contains.

### DIFF
--- a/.changeset/shaggy-knives-exercise.md
+++ b/.changeset/shaggy-knives-exercise.md
@@ -1,0 +1,9 @@
+---
+"@apollo/query-planner": minor
+"@apollo/gateway": minor
+---
+
+Adds debug/testing query planner options (`debug.bypassPlannerForSingleSubgraph`) to bypass the query planning
+process for federated supergraph having only a single subgraph. The option is disabled by default, is not recommended
+for production, and is not supported (it may be removed later). It is meant for debugging/testing purposes.
+  
--- a/query-planner-js/src/buildPlan.ts
+++ b/query-planner-js/src/buildPlan.ts
@@ -91,6 +91,7 @@ import {
   getLocallySatisfiableKey,
   createInitialOptions,
   buildFederatedQueryGraph,
+  FEDERATED_GRAPH_ROOT_SOURCE,
 } from "@apollo/query-graphs";
 import { stripIgnoredCharacters, print, parse, OperationTypeNode } from "graphql";
 import { DeferredNode, FetchDataInputRewrite, FetchDataOutputRewrite } from ".";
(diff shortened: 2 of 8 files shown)


query-planner-js/src/buildPlan.ts
```diff
@@ -2492,6 +2497,25 @@ export class QueryPlanner {
     };
     this._lastGeneratedPlanStatistics = statistics;
 
+    if (this.config.debug.bypassPlannerForSingleSubgraph) {
+      // A federated query graph always have 1 more sources than there is subgraph, because the root vertices
+      // belong to no subgraphs and use a special source named '_'. So we skip that "fake" source.
+      const subgraphs = mapKeys(this.federatedQueryGraph.sources).filter((name) => name !== FEDERATED_GRAPH_ROOT_SOURCE);
+      if (subgraphs.length === 1) {
```
clenfest, line 2504: Do you want the else case to error?

.changeset/shaggy-knives-exercise.md
```diff
@@ -0,0 +1,9 @@
+---
+"@apollo/query-planner": patch
+"@apollo/gateway": patch
+---
+
+Adds debug/testing query planner options (`debug.bypassPlannerQueryForSingleSubgraph`) to bypass the query planning
```
clenfest, line 6: I think this should be `debug.bypassPlannerForSingleSubgraph`
```suggestion
Adds debug/testing query planner options (`debug.bypassPlannerForSingleSubgraph`) to bypass the query planning
```

.changeset/shaggy-knives-exercise.md
```diff
@@ -0,0 +1,9 @@
+---
+"@apollo/query-planner": patch
```
clenfest, line 2: Not sure if it matters too much, but these should probably be changed to minor now that we're putting it in 2.4.0.

clenfest (approved):
A couple of asks, but otherwise LGTM.

.changeset/shaggy-knives-exercise.md
pcmanus: Good catch, thanks.

query-planner-js/src/buildPlan.ts
pcmanus: No. I think the config name and documentation makes it clear that it only applies to single subgraphs, and it's clearly flagged as for debug/testing so I'm not really worried about someone not understanding why they don't get what they want out of this. On the flip side, erroring could make it more annoying to use as you'd have to remember to disable the option when you go from 1 to 2 subgraph, or get a runtime crash. Admittedly this is all a bit subjective since again, it is clearly flagged for debug/testing, but I personally markedly prefer it not having a `else`.

PR #2443, opened 2023-03-08
fix: Handle defaulted variables correctly during post-processing

### LINKED ISSUE #300: Apollo Federation gateway breaks upon receiving a query with empty variables section
Apollo federation gateway breaks throwing ``` Cannot convert undefined or null to object ``` upon receiving a graphql query which does not have any variables section. (From our perspective this is a very common use case).

Passing variables as empty object {}, works fine, but it should not be mandated on the caller to explicitly pass empty object.

trevor-scheer:
Fixes #300
(more like https://github.com/apollographql/federation/issues/300#issuecomment-1458660420)

Post-processing doesn't currently account for defaulted variable values in an operation at all. This surfaces as a problem when using conditional directives (`@skip`/`@include`) that use variables with defaulted values (and no corresponding variable provided).

We should collect defaulted variable values from the operation and pass them along with the `variables` we provide to `computeResponse` so these are accounted for.

### DIFF
--- a/gateway-js/src/resultShaping.ts
+++ b/gateway-js/src/resultShaping.ts
@@ -72,7 +72,7 @@ export function computeResponse({
     schema: operation.schema,
     variables: {
       ...operation.collectDefaultedVariableValues(),
-      // overwrite and defaulted variables if they are provided
+      // overwrite any defaulted variables if they are provided
       ...variables,
     },
     errors: [],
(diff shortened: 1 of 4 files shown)


dariuszkuc (approved):

gateway-js/src/resultShaping.ts
```diff
@@ -70,7 +70,11 @@ export function computeResponse({
 
   const parameters = {
     schema: operation.schema,
-    variables: variables ?? {},
+    variables: {
+      ...operation.collectDefaultedVariableValues(),
+      // overwrite and defaulted variables if they are provided
```
pcmanus, line 75: Nit: I think the "and" should be removed in that sentence?

pcmanus (approved):

PR #2448, opened 2023-03-09
Update docs to use 2.4 instead of 2.3

clenfest:
Also, please look at https://github.com/apollographql/federation/commit/b0765a753a5c92562d0c8a8ee665a21c705893d1 which I accidentally nakedly committed to `next`. Will update branch protections.

### DIFF
--- a/.changeset/pre.json
+++ b/.changeset/pre.json
@@ -1,0 +1,14 @@
+{
+  "mode": "pre",
+  "tag": "alpha",
+  "initialVersions": {
+    "@apollo/composition": "2.3.3",
+    "apollo-federation-integration-testsuite": "2.3.3",
+    "@apollo/gateway": "2.3.3",
+    "@apollo/federation-internals": "2.3.3",
+    "@apollo/query-graphs": "2.3.3",
+    "@apollo/query-planner": "2.3.3",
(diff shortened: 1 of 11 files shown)


.changeset/pre.json
```diff
@@ -0,0 +1,14 @@
+{
+  "mode": "pre",
+  "tag": "alpha",
+  "initialVersions": {
+    "@apollo/composition": "2.3.3",
```
trevor-scheer, line 5: No change needed, but I just realized this gets kinda wonky if we land a patch to `main`?

trevor-scheer (approved):
LGTM. There are some test cases in `executeQueryPlan` (which I understand are not docs) that we should probably update eventually just to keep them current.

.changeset/pre.json
clenfest: I don't think so, because `next` will just continue to be on `2.4.0-alpha.*` until we exit `pre` mode, right? I don't think that releasing patches out of main should matter to it.

trevor-scheer: Right, for some reason I was thinking it'd be `2.3.4` on next but that's definitely not the case.

PR #2451, opened 2023-03-10
revert removal of url import

clenfest:
Revert change made in #2293, the router uses deno for the router-bridge, and url is not built in


trevor-scheer (approved):
Darn. Probably something we can polyfill with rollup but also probably not worth the trouble right now.

Issue #2457, opened 2023-03-12
reuseFragments in QueryPlannerConfig not working as expected


AneethAnand:
### Problem statement

- The issue is exactly what has been explained here :  https://github.com/apollographql/apollo-server/pull/3791

- For large queries (esp. ones that leverage many nested fragments), queries to subgraph services blows up astronomically because the gateway's query plan expands the query request. 

- Following is the Query Plan, which uses some of the fragments (not all) from the query but still number of lines 20523 for 2000 lines of query

<img width="232" alt="image" src="https://user-images.githubusercontent.com/43860019/224567186-e8fb9e83-b31c-4fee-be28-7df022d068d2.png">

### Existing Solution
QueryPlannerConfig.reuseQueryFragments or experimental_autoFragmentization did not help.

### Code
```
  if (fragments && reuseQueryFragments) {
    // For all subgraph fetches we query `__typename` on every abstract types (see `FetchGroup.toPlanNode`) so if we want
    // to have a chance to reuse fragments, we should make sure those fragments also query `__typename` for every abstract type.
    fragments = addTypenameFieldForAbstractTypesInNamedFragments(fragments)
  } else {
    fragments = undefined;
  }
```
"addTypenameFieldForAbstractTypesInNamedFragments" -- adds typename not to the fragment but on the selection set.

###  Possible Solution
After I read through the code, I added  "__typename" field to all fragments and then I saw reduction in the downstream query
From 20K to 1K
<img width="232" alt="image" src="https://user-images.githubusercontent.com/43860019/224567843-c6ab9d8d-7976-4c9d-a78e-02cb933a44f2.png">

I would like to know your thoughts on how adding __typename reduces the subgraph request query? Can it be automated in the queryPlan and then later prune those fields out from the result?

AneethAnand:
@trevor-scheer / @pcmanus 
I have created a PR (https://github.com/apollographql/federation/pull/2463) 
regarding the "Gateway was expanding all query fragments"

The issue I found is that query fragment as seen in the code below is used only if the selection set match exactly. 
```
        if (optimizedSelection.equals(candidate.selectionSet)) {
          const fragmentSelection = new FragmentSpreadSelection(fieldBaseType, fragments, candidate.name);
          return new FieldSelection(this.field, selectionSetOf(fieldBaseType, fragmentSelection));
        }
```
There are many uses cases where the Query can fragment a part of the type and use it in many places.  So I have modified the above code to
```
      const schema = options?.schema;
[...]

PR #2463, opened 2023-03-15
Adding an option to Auto-refragment the QueryPlan

AneethAnand:
PR related to: https://github.com/apollographql/federation/issues/2457

### DIFF
--- a/internals-js/src/operations.ts
+++ b/internals-js/src/operations.ts
@@ -48,9 +48,16 @@ import {
   Type,
 } from "./definitions";
 import { ERRORS } from "./error";
-import { isDirectSubtype, sameType } from "./types";
+import { isDirectSubtype, isSubtype, sameType } from "./types";
 import { assert, mapEntries, MapWithCachedArrays, MultiMap, SetMultiMap } from "./utils";
 import { argumentsEquals, argumentsFromAST, isValidValue, valueToAST, valueToString } from "./values";
+import { createHash } from '@apollo/utils.createhash';
+
+export interface OptimizeOptions {
+  minUsagesToOptimize?: number
+  autoFragmetize?: boolean
+  schema?: Schema
+}
 
 function validate(condition: any, message: () => string, sourceAST?: ASTNode): asserts condition {
   if (!condition) {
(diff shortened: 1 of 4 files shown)


internals-js/src/operations.ts
```diff
@@ -1071,8 +1082,8 @@ export class SelectionSet extends Freezable<SelectionSet> {
     }
   }
 
-  optimize(fragments?: NamedFragments): SelectionSet {
-    if (!fragments || fragments.isEmpty()) {
+  optimize(fragments?: NamedFragments, options: OptimizeOptions = {}): SelectionSet {
+    if (!fragments || (!options.autoFragmetize && fragments.isEmpty())) {
```
pcmanus, line 1086: Same remark that on `Operation` regarding `!fragments`. In that case however, it's probably stranger to create a new `NamedFragments` objects. But we actually never pass `undefined` to this particular method (we do for `Operation.optimize` however), so I might suggest just making `fragments` mandatory here.

internals-js/src/operations.ts
```diff
@@ -1629,13 +1641,44 @@ export class FieldSelection extends Freezable<FieldSelection> {
         //   }
         // To do that, we can change that `equals` to `contains`, but then we should also "extract" the remainder
         // of `optimizedSelection` that isn't covered by the fragment, and that is the part slighly more involved.
-        if (optimizedSelection.equals(candidate.selectionSet)) {
+        if (optimizedSelection && optimizedSelection.equals(candidate.selectionSet)) {
           const fragmentSelection = new FragmentSpreadSelection(fieldBaseType, fragments, candidate.name);
           return new FieldSelection(this.field, selectionSetOf(fieldBaseType, fragmentSelection));
+        } else if(options?.autoFragmetize && optimizedSelection && optimizedSelection.contains(candidate.selectionSet)) {
```
pcmanus, line 1647: Some similar `contains` check has been added on `main`, so the patch here will only need the later part where a fragment is optimistically generated if the remaining selection is not "trivial". But I'm afraid this will require a bit of rebase, sorry.

internals-js/src/operations.ts
```diff
@@ -1629,13 +1641,44 @@ export class FieldSelection extends Freezable<FieldSelection> {
...
+          // But defeinitely helps in many cases.
         }
       }
+      // Create a new named fragment with exact selection set, only if we did not find any matching fragment in the input query.
+      // Incase we see this same pattern in another part of the query,
+      // this will help reduce query expansion when sending requests to subgraphs.
+      // If we don't see the same pattern the final pass of optimizer will re-expand it.
+      const schema = options?.schema;
+      if(options?.autoFragmetize && schema && optimizedSelection.selections().length > 1) {
+        const hash = createHash('sha256').update(optimizedSelection.toString()).digest('hex');
```
pcmanus, line 1675: I believe the only goal of that `hash` is to generate a unique name for the created fragment. If so, I would suggest using a simple uuid here (using the `uuidv1()` method already imported in this file). This is going to be substantially cheaper to generate (`optimizedSelection.toString()` is not exactly cheap) and will lead to overall smaller identifiers (more readable, and I guess makes the overall query smaller).

pcmanus (commented):
Left some comments, but I think the first and main thing is that this will need to be rebased substantially due to #2497 having been merge. That later PR refactor the `optimise` method somewhat and actually already handle some of the changes in this PR (which will thus not be necessary).

I do feel a bit bad because that need for rebasing is on me, so please do let me know if you want me to help with that rebase and I can do it.

duckki:
Wait, is this superseded by generateQueryFragments (PR https://github.com/apollographql/federation/pull/2958)?

dariuszkuc:
Yes it was. Superseded by #2958

PR #2472, opened 2023-03-16
Add spell checking (without fixing existing errors)

trevor-scheer:
We've used `cspell` in the Apollo Server repo with success. The configuration / implementation is mostly copied from there. This should help prevent spelling errors in PRs / nitpicky spelling comments and meaningless iteration.

Rather than fixing all existing errors, I blindly added all existing violations to the dictionary just to keep the noise down (barring one `programing` which cspell recognized as a common misspelling and wouldn't allow it). Unfortunately this does mean those errors can be repeated, but maybe if we catch any repeats in review then we fix all instances of that error at that point.


dariuszkuc (approved):

cspell.yml
```diff
@@ -0,0 +1,41 @@
...
+  - workspace
+
+allowCompoundWords: true
+
+# Ignore files that aren't checked in to git as well as files that aren't written
+# by hand. Note that we do want to check, say, JSON files (as package.json
+# contains English text like package descriptions).
+useGitignore: true
+ignorePaths:
+  - cspell.yaml
```
clenfest, line 23: Did you want the ignore path to match this file? Because if so => cspell.yml.

cspell-dict.txt
```diff
@@ -0,0 +1,261 @@
+abolve
```
clenfest, line 1: Can we put these files somewhere other than at the root level?

clenfest (commented):
good idea! I have the vscode plugin installed locally, but it's definitely much cleaner to get this automatically.

cspell.yml
trevor-scheer: Good catch, maybe it never mattered

cspell-dict.txt
trevor-scheer: Yeah, moved config and dictionary to a `.cspell` dir. Also got the vscode extension pointed at the config 👍

package.json
```diff
@@ -25,7 +25,8 @@
     "hints-doc:check": "npm run hints-doc && git diff --exit-code",
     "gen-test-supergraph": "ts-node ./query-planner-js/src/__tests__/genTestSupergraph.ts",
     "changeset-version": "changeset version && npm i",
-    "publish": "changeset publish"
+    "publish": "changeset publish",
+    "spell:check": "cspell lint '**' '.changeset/**' --no-progress --config .cspell/cspell.yml  || (echo 'Add any real words to cspell-dict.txt.'; exit 1)"
```
trevor-scheer, line 29:
```suggestion
    "spell:check": "cspell lint '**' '.changeset/**' --no-progress --config .cspell/cspell.yml  || (echo 'Add any real words to ./cspell/cspell-dict.txt.'; exit 1)"
```

PR #2482, opened 2023-03-20
Alternative implementation for #2315

pcmanus:
This is a proposed alternative approach of #2431 to implementing:
- a slightly more generic way to declare directives that gets automatically composed.
- a new merging strategy for some non-repeatable directive with arguments that merges said argument values using a small set of predefined strategies (max, min, ...).

This approach follows my comments on #2431 as well as a few more points. In particular:
- it expands `createDirectiveSpecification` to allow both declaring directives that should be composed, as well as defining the new strategies for arguments when that's desired.
- it's added as an extension of the current directive merging code.
- it handles existing directives `@tag` and `@inaccessible`. In fact, to avoid needing to have a manual "list of directives that should merge automatically", I generalised the code a bit so that it iterates on every directives in the subgraphs and recognises known feature and directives, giving access to additional merging information on those directives.
- it handles the question of linking merged directives to a particular feature in the supergraph. That's what we do for `@tag` and `@inaccessible`, and I think that's what we want for future other similar directives. In particular, the patch includes a little bit of cleanup for feature definitions to make it easier to define new ones.


pcmanus:
I'll note that the patch has a test for the "happy path" of the new code, but it's missing tests around invalid declarations, to show that proper errors are raised. I'll add better coverage, but waiting on seeing if there is disagreement on the overall approach.

internals-js/src/directiveAndTypeSpecification.ts
```diff
@@ -34,43 +47,114 @@ export type TypeSpecification = {
...
+          // By the time we call this, the directive should have been added to the schema and so getting the type should not raise errors.
+          assert(!Array.isArray(argType), () => `Should have gotten error getting type for @${name}(${argName}:), but got ${argType}`)
+          const strategyTypes = strategy.types(schema);
+          if (!strategyTypes.some((t) => sameType(t, argType))) {
+            return new GraphQLError(
+              `Invalid composition strategy ${strategy.name} for argument @${name}(${argName}:) of type ${argType}; `
+                + `${strategy.name} only supports type(s) ${strategyTypes.join(', ')}`
+            );
+          }
+        };
```
clenfest, line 111:
```suggestion
        }
```

internals-js/src/fieldArgumentCompositionStrategies.ts
```diff
@@ -0,0 +1,39 @@
...
+export type ArgumentCompositionStrategy = {
+  name: string,
+  types: (schema: Schema) => InputType[],
+  mergeValues: (values: any[]) => any,
+}
+
+export const ARGUMENT_COMPOSITION_STRATEGIES = {
+  MAX: {
+    name: 'MAX',
+    types: (schema: Schema) => [new NonNullType(schema.intType())],
```
clenfest, line 12: Nit: Changing `types` to `supportedTypes` might be a bit more clear.

internals-js/src/directiveAndTypeSpecification.ts
```diff
@@ -34,43 +47,114 @@ export type TypeSpecification = {
...
+        return {
+          merge: (argName, values) => {
+            const strategy = argStrategies.get(argName); 
+            assert(strategy, () => `Should have a strategy for ${argName}`);
+            return strategy.mergeValues(values);
+          },
+          toString: () => {
+            return [...argStrategies.entries()].map(([arg, strat]) => `${arg}: ${strat.name}`).join(', ');
```
clenfest, line 119: Nit: `strategy` is only 3 chars more and is a bit more clear.

composition-js/src/merging/merge.ts
```diff
@@ -358,10 +353,19 @@ class Merger {
     // If we get here with `nameInSupergraph` unset, it means there is no usage for the directive at all and we
     // don't bother adding the spec to the supergraph.
     if (nameInSupergraph) {
+      const specInSupergraph = compositionSpec.supergraphSpecification();
       const errors = linkSpec.applyFeatureToSchema(this.merged, specInSupergraph, nameInSupergraph === specInSupergraph.url.name ? undefined : nameInSupergraph, specInSupergraph.defaultCorePurpose);
       assert(errors.length === 0, "We shouldn't have errors adding the join spec to the (still empty) supergraph schema");
+      const argumentsMerger = compositionSpec.argumentsMerger?.call(this, this.merged);
```
clenfest, line 359: I believe that the first argument should be null here since `argumentsMerger` is not part of an object.

```suggestion
      const argumentsMerger = compositionSpec.argumentsMerger?.call(null, this.merged);
```

composition-js/src/merging/merge.ts
```diff
@@ -2465,24 +2470,39 @@ class Merger {
...
+        const info = this.mergedFederationDirectiveInSupergraph.get(name);
+        if (info && info.argumentsMerger) {
+          const mergedArguments = Object.create(null);
+          const applicationsArguments = differentApplications.map((a) => a.arguments(true));
+          for (const argDef of info.definition.arguments()) {
+            const values = applicationsArguments.map((args) => args[argDef.name]);
+            mergedArguments[argDef.name] = info.argumentsMerger.merge(argDef.name, values);
+          }
+          dest.applyDirective(name, mergedArguments);
+          this.mismatchReporter.pushHint(new CompositionHint(
```
clenfest, line 2482: Do we need this hint? Seems like customers may want to turn this on/off on a per directive basis.

internals-js/src/tagSpec.ts
```diff
@@ -41,11 +41,11 @@ export class TagSpecDefinition extends FeatureDefinition {
       name:'tag',
       locations: this.tagLocations,
       repeatable: true,
-      argumentFct: (schema) => ({
-        args: [{ name: 'name', type: new NonNullType(schema.stringType()) }],
-        errors: [],
-      }),
+      args: [{ name: 'name', type: (schema) => new NonNullType(schema.stringType()) }],
+      composes: true,
+      supergraphSpecification: () => TAG_VERSIONS.latest(),
```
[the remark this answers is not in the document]
pcmanus: No, it's a good point, though I'm just perpetuating [the status quo](https://github.com/apollographql/federation/blob/main/composition-js/src/merging/merge.ts#L228-L231) here. But I agree we should probably change that behaviour and use the smallest version that is needed instead of always defaulting to the last.

I can do that change here if we really want to, but I feel it's related to using the latest `join` spec in the supergraph which I've been meaning to open a ticket for. So I've created #2495 and I would implement this change of strategy there.

composition-js/src/merging/merge.ts
pcmanus: I believe it was @lennyburdette which mentioned somewhere that some users strongly feel that discrepancies between subgraphs on non-repeatable directives should error out, and I do see the argument, so this is meant as a hook to allow this.

I would agree that it's high time we allow configuring a set of hints to turn into hard errors, and that until we have that native a user would have to catch hints manually, but filtering hints manually is not that hard either and this at least allow that.

And I do get the concern for noisiness, but hints are already super noisy if unfiltered and I don't think this one is worst than others we have (but I do think this hint has more concrete usefulness than many other we have).

composition-js/src/merging/merge.ts
pcmanus: I believe you're right, thanks.

clenfest (approved):
I think you said you were going to write more tests, but I'm approving in case you were waiting on me.

PR #2486, opened 2023-03-21
Fix assertion error due to not ignoring some unsatisfiable branches

pcmanus:
Code was added in #2449 which "optimize" selections by removing unecessary inline fragments and branches that can't be satisfied (when we cross type conditions so that the intersection of types leading here is empty).

Unfortunately that code misses some cases of when a branch is impossible, which leads to trying to build a selection set that is invalid and this in turn triggers an assertion error.

### DIFF
--- a/.changeset/lovely-walls-sneeze.md
+++ b/.changeset/lovely-walls-sneeze.md
@@ -1,0 +1,7 @@
+---
+"@apollo/federation-internals": patch
+---
+
+Fix assertion error during query planning in some cases where queries has some unsatisfiable branches (a part of the
+query goes through type conditions that no runtime types satisfies).
+  
--- a/internals-js/src/__tests__/operations.test.ts
+++ b/internals-js/src/__tests__/operations.test.ts
@@ -477,7 +477,6 @@ describe('basic operations', () => {
   })
 });
 
-
 describe('MutableSelectionSet', () => {
   test('memoizer', () => {
     const schema = parseSchema(`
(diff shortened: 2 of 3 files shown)


internals-js/src/__tests__/operations.test.ts
```diff
@@ -546,3 +545,62 @@ describe('MutableSelectionSet', () => {
...
+    '{ i { ... on T1 { a b c } } }',
+  ])('is identity if there is no unsatisfiable/unoptimized branches', (op) => {
+    expect(withoutUnsatisfiableBranches(op)).toBe(op);
+  });
+
+  it.each([
+    { input: '{ i { ... on I { a } } }', output: '{ i { a } }' },
+    { input: '{ i { ... on T1 { ... on I { a b } } } }', output: '{ i { ... on T1 { a b } } }' },
+    { input: '{ i { ... on I { a ... on T2 { d } } } }', output: '{ i { a ... on T2 { d } } }' },
+    { input: '{ i { ... on I { a ... on T2 { d } } } }', output: '{ i { a ... on T2 { d } } }' },
```
trevor-scheer, line 600: These are identical

.changeset/lovely-walls-sneeze.md
```diff
@@ -0,0 +1,7 @@
+---
+"@apollo/federation-internals": patch
+---
+
+Fix assertion error during query planning in some cases where queries has some unsatisfiable branches (a part of the
```
trevor-scheer, line 5:
```suggestion
Fix assertion error during query planning in some cases where queries have some unsatisfiable branches (a part of the
```

trevor-scheer (approved):

PR #2491, opened 2023-03-22
runs 'eslint --fix'

clenfest:
Just autofixing some linter problems.


trevor-scheer (approved):
Fine as-is, would be nice to clear up the remaining errors and enable in CI if we can pass with warnings only?

PR #2494, opened 2023-03-23
Fix incorrectly rejected composition and subgraph issues with `@inter…

### LINKED ISSUE #2485: The following supergraph API query: ... cannot be satisfied by the subgraphs because ... there is a more direct path in ... that avoids ... altogether
### Issue Description

I am trying to upgrade our stack to Federation 2.3. Everything is working except for one issue

```
Error: A valid schema couldn't be composed. The following composition errors were found:
        The following supergraph API query:
{
  search {
    goalConversionRate
  }
}
cannot be satisfied by the subgraphs because:
- from subgraph "search": cannot find field "Article.goalConversionRate".
- from subgraph "assets":
  - cannot find field "Article.goalConversionRate".
  - ignoring moving to subgraph "events" using @key(fields: "id") of "Article" because there is a more direct path in events that avoids assets altogether.
```

In the above query, `goalConversionRate` can _only_ be resolved by the `events` subgraph, but it seems that the query plan never attempts to fetch from that service.

If
1. this is an Apollo issue, we would appreciate a fix :)
2. If it's an issue on our side, would love some guidance on what is invalid (and perhaps a clearer error message).

### Link to Reproduction

https://codesandbox.io/p/sandbox/frosty-jones-njgsyt

### Reproduction Steps

Just run the `start` npm script in the repro link and you should see this error.

pcmanus:
…faceObject`

There is 2 main issues with the handling of `@interfaceObject` that this patch fixes:
1. given some `@interfaceObject` type `I` in some subgraph `S1`, if another subgraph `S2` was declaring an _implementation_ type `T` of `I` but _without_ declaring `I`, then the code missed creating an "edge" between `T` in `S2` and `I` in `S1`, which resulted in composition rejecting such cases, even though it should work.
2. there were an additional issue in the handling of `@requires` so that the "input rewrites" generated were not generic enough, leading to sometimes sending incorrect queries to subgraph (more precisely, we were sometimes not rewriting the typename when querying an `@interfaceObject` subgraph, leading that subgraph to complain that it is asked for type it doesn't know).

Note that fixing those 2 issues surfaced the fact that the handling of "rewrites" in the gateway was not working exactly as it should have. More precisely, hen a path had a fragment `... on I`, if `I` was an interface, then we would no select objects that implements `I` correctly. As it happens, the router implementation of those rewrites was both cleaner and didn't had that issue, so this patch also updates the handling of "rewrites" to mimick what the router implementation does (fixing the issue, and overall cleaning the code).

Fixes #2485

### DIFF
--- a/.changeset/hungry-berries-destroy.md
+++ b/.changeset/hungry-berries-destroy.md
@@ -1,0 +1,11 @@
+---
+"@apollo/query-planner": patch
+"@apollo/query-graphs": patch
+"@apollo/federation-internals": patch
+"@apollo/gateway": patch
+---
+
+Fix issues (incorrectly rejected composition and/or subgraph errors) with `@interfaceObject`. Those issues may occur
+either due to some use of `@requires` in an `@interfaceObject` type, or when some subgraph `S` defines a type that is an
+implementation of an interface `I` in the supergraph, and there is an `@interfaceObject` for `I` in another subgraph,
+but `S` does not itself defines `I`.
--- a/gateway-js/src/dataRewrites.ts
+++ b/gateway-js/src/dataRewrites.ts
@@ -1,0 +1,130 @@
+import { FetchDataRewrite } from "@apollo/query-planner";
+import { assert } from "console";
+import { GraphQLSchema, isAbstractType, isInterfaceType, isObjectType } from "graphql";
+
+const FRAGMENT_PREFIX = '... on ';
+
+export function applyRewrites(schema: GraphQLSchema, rewrites: FetchDataRewrite[] | undefined,  value: Record<string, any>) {
+  if (!rewrites) {
+    return;
+  }
+
+  for (const rewrite of rewrites) {
+    applyRewrite(schema, rewrite, value);
+  }
+}
+
+function applyRewrite(schema: GraphQLSchema, rewrite: FetchDataRewrite,  value: Record<string, any>) {
+  const splitted = splitPathLastElement(rewrite.path);
+  if (!splitted) {
+    return;
+  }
+
+  const [parent, last] = splitted;
+  const { kind, value: fieldName } = parsePathElement(last);
+  // So far, all rewrites finish by a field name. If this ever changes, this assertion will catch it early and we can update.
+  assert(kind === 'fieldName', () => `Unexpected fragment as last element of ${rewrite.path}`);
+  applyAtPath(schema, parent, value, rewriteAtPathFunction(rewrite, fieldName));
+}
+
+function rewriteAtPathFunction(rewrite: FetchDataRewrite, fieldAtPath: string): (obj: Record<string, any>) => void {
+  switch (rewrite.kind) {
+    case 'ValueSetter':
+      return (obj) => {
+        obj[fieldAtPath] = rewrite.setValueTo;
+      };
+    case 'KeyRenamer':
+      return (obj) => {
+        obj[rewrite.renameKeyTo] = obj[fieldAtPath];
+        obj[fieldAtPath] = undefined;
+      };
+  }
+}
+
+
+/**
+ * Given a path, separates the last element of path and the rest of it and return them as a pair.
+ * This will return `undefined` if the path is empty.
+ */
+function splitPathLastElement(path: string[]): [string[], string] | undefined {
+  if (path.length === 0) {
+    return undefined;
+  }
+
+  const lastIdx = path.length - 1;
+  return [path.slice(0, lastIdx), path[lastIdx]];
+}
+
+function applyAtPath(schema: GraphQLSchema, path: string[], value: any, fct: (objAtPath: Record<string, any>) => void) {
+  if (Array.isArray(value)) {
+    for (const arrayValue of value) {
+      applyAtPath(schema, path, arrayValue, fct);
+    }
+    return;
+  }
+
+  if (typeof value !== 'object') {
+    return;
+  }
+
+  if (path.length === 0) {
+    fct(value);
+    return;
+  }
+
+  const [first, ...rest] = path;
+  const { kind, value: eltValue } = parsePathElement(first);
+  switch (kind) {
+    case 'fieldName':
+      applyAtPath(schema, rest, value[eltValue], fct);
+      break;
+    case 'typeName':
+      // When we apply rewrites, we don't always have the __typename of all object we would need to, but the code expects that
+      // this does not stop the rewrite to applying, hence the modified to `true` when the object typename is not found.
+      if (isObjectOfType(schema, value, eltValue, true)) {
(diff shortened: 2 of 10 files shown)


query-planner-js/src/QueryPlan.ts
```diff
@@ -46,29 +46,19 @@ export interface FetchNode {
   operationKind: OperationTypeNode;
   operationDocumentNode?: DocumentNode;
   // Optionally describe a number of "rewrites" that query plan executors should apply to the data that is sent as input of this fetch.
-  inputRewrites?: FetchDataInputRewrite[];
```
pcmanus, line 49: Note to reviewer(s): the context behind the changes of this file is that I always though to ultimately have just `FetchDataRewrite`, but I didn't do a super good job of the first implementation of those rewrites on the execution side, so that before this patch, input and output rewrites use separate code at execution, and I used this separate typing to avoid dead code/ugly assertions in the execution code.

And fwiw, when I implemented this in the router side, I realise there was no point in having that distinction, so the router don't even bother distinguishing between those 2 types. Anyway, this patch cleans up the execution on the gateway side, using the same approach than in the router, and in doing so the distinction between `FetchDataOutputRewrite` and `FetchDataInputRewrite` becomes meaningless, so I figure it's worth getting rid of.

But do note this has no impact on the router since this is just some typing that is changed here: the underlying plans we generate are unchanged.

internals-js/src/operations.ts
```diff
@@ -304,16 +305,18 @@ export class Field<TArgs extends {[key: string]: any} = {[key: string]: any}> ex
   }
 
   private canRebaseOn(parentType: CompositeType) {
-    // There is 2 valid cases we want to allow:
+    const fieldParentType = this.definition.parent
+    // There is 3 valid cases we want to allow:
```
clenfest, line 309: Unless I'm missing it, you don't specify the third case.

.changeset/hungry-berries-destroy.md
```diff
@@ -0,0 +1,12 @@
+---
+"@apollo/query-planner": patch
+"@apollo/query-graphs": patch
+"@apollo/federation-internals": patch
+"@apollo/gateway": patch
+---
+
+Fix issues (incorrectly rejected composition and/or subgraph errors) with `@interfaceObject`. Those issues may occurs
```
clenfest, line 8:
```suggestion
Fix issues (incorrectly rejected composition and/or subgraph errors) with `@interfaceObject`. Those issues may occur
```

.changeset/hungry-berries-destroy.md
```diff
@@ -0,0 +1,12 @@
...
+either due to some use of `@requires` in an `@interfaceObject` type, or when some subgraph `S` defines a type that is an
+implementation of an interface `I` in the supergraph, and there is an `@interfaceObject` for `I` in another subgraph,
+but `S` does not itself defines `I`.
```
clenfest, line 11: Do you mean that it doesn't declare `@interfaceObject` for `I`?

query-graphs-js/src/querygraph.ts
```diff
@@ -664,42 +664,54 @@ function federateSubgraphs(supergraph: Schema, subgraphs: QueryGraph[]): QueryGr
...
+            if (otherVertices.length > 0) {
+              // Note that later, when we've handled @provides, this might not be true anymore as @provides may create copy of a
+              // certain type. But for now, it's true.
+              assert(
+                otherVertices.length == 1,
+                () => `Subgraph ${j} should have a single vertex for type ${type.name} but got ${otherVertices.length}: ${inspect(otherVertices)}`);
+
+              const otherVertice = otherVertices[0];
+              // The edge goes from the otherSubgraph to the current one.
+              const head = copyPointers[j].copiedVertex(otherVertice);
```
clenfest, line 685:
```suggestion
              const otherVertex = otherVertices[0];
              // The edge goes from the otherSubgraph to the current one.
              const head = copyPointers[j].copiedVertex(otherVertex);
```

query-graphs-js/src/querygraph.ts
```diff
@@ -664,42 +664,54 @@ function federateSubgraphs(supergraph: Schema, subgraphs: QueryGraph[]): QueryGr
...
             // requested fields are only provided by that @interfaceObject.
-            const otherType = otherVertice.type;
-            if (isInterfaceObject && isInterfaceType(otherType)) {
-              for (const implemType of otherType.possibleRuntimeTypes()) {
-                // Note that we're only taking the implementation types from "otherSubgraph", so we're guaranteed
-                // to have a corresponding vertice (and only one for the same reason than mentioned in the assert above).
-                const implemVertice = otherSubgraph.verticesForType(implemType.name)[0];
+            if (isInterfaceObject) {
+              const typeInSupergraph = supergraph.type(type.name);
+              assert(typeInSupergraph && isInterfaceType(typeInSupergraph), () => `Type ${type} is an interfaceObject in subgraph ${i}; should be an interface in the supergraph`);
```
clenfest, line 696: Assuming that these get wrapped in an Error further up?

gateway-js/src/dataRewrites.ts
```diff
@@ -0,0 +1,130 @@
...
+
+function applyAtPath(schema: GraphQLSchema, path: string[], value: any, fct: (objAtPath: Record<string, any>) => void) {
+  if (Array.isArray(value)) {
+    for (const arrayValue of value) {
+      applyAtPath(schema, path, arrayValue, fct);
+    }
+    return;
+  }
+
+  if (typeof value !== 'object') {
```
clenfest, line 66: I'm a little confused. Further down, we seem to be treating `value` as a `Record<string,any>`. Wouldn't this cause that case to exit here? I'm probably missing something.

clenfest (approved):

.changeset/hungry-berries-destroy.md
pcmanus: No, I mean that `S` does not have type `I` at all (but it declares a type that is an implementation of `I` in another subgraph).

query-graphs-js/src/querygraph.ts
pcmanus: I'm not sure to understand what "these" refers to. But `assert` throws an `Error` if it ever does anything, so unsure why there would need to be additional wrapping further up?

gateway-js/src/dataRewrites.ts
pcmanus: If value is a `Record<string, any>`, then it is an `object`, so it shouldn't exit since the condition exits if it is _not_ an object.

pcmanus: If value is a `Record<string, any>`, then it is an `object`, so it shouldn't exit since the condition exits if it is _not_ an object.

PR #2519, opened 2023-04-07
Document using requires with fields that take arguments

StephenBarlow:


docs/source/entities-advanced.mdx
```diff
@@ -468,6 +468,27 @@ type Product @key(fields: "id") {
...
+  #highlight-start
+  shippingEstimate: String @requires(fields: "weight(units:\"KILOGRAMS\")")
+  #highlight-end
+}
+```
+
+- The router provides the specified values in its query to whichever subgraph defines the required field.
+- Each specified argument value is _static_ (i.e., the router always provides the same value).
+- You can omit values for nullable arguments. You _must_ provide values for non-nullable arguments.
+- If you define your subgraph schema in an SDL file (instead of programmatically), you escape quotes for string and enum values with backslashes (as shown above).
```
lennyburdette, line 490:
```suggestion
- If you define your subgraph schema in an SDL file (instead of programmatically), you must escape quotes for string and enum values with backslashes (as shown above).
```
missing word?

lennyburdette (approved):

dariuszkuc (approved):

Issue #2495, opened 2023-03-23
Composition should try to use the smallest usable version for the specs it adds to the supergraph


pcmanus:
Currently, when composition creates a new supergraph, it defaults both the `join` spec and any other specs it adds (`@tag` typically when used) to the latest possible version (known by said composition). That's ok as long as on follows the recommended  upgrade process of updating the router/gateway first before using a new version of composition, and it *is* recommended to follow that process, but it still make sense to be as forgiving as possible when that process is not followed.

In many cases[1], the version the supergraph specs are bumped to accommodate brand new features but have no changes when those new features aren't yet used. The proposal of this ticket is to have composition use the smallest "usable" version of specs in the generated subgraph, instead of always defaulting to the latest one. Doing so will mean that in most case, upgrading composition will not impact the generated supergraph up until new features are actually used, which will make upgrades a bit more flexible in practice.

[1]: in some cases, the `join` spec may have to be bumped not for a new feature, but because fixing a bug requires information in the supergraph that didn't previously exists. In such case, upgrading composition would still automatically have to bump the `join` spec even with no changes to the subgaphs.

sachindshinde:
Regarding versioning, the tricky part here is:
1. People want to know easily and definitively what versions of gateway/router they can run that "work" with whatever composition/build track they're running.
2. Right now that's relatively well-known, since composition and gateway/router code are developed in lock-step; the Studio build configuration modal describes specific version requirements for a build track.
   a. We also have [docs that describe the update process](https://www.apollographql.com/docs/graphos/graphs/updating/), which let users know that they should update their gateway/router before they update their composition/build track.

If we change composition in the way you've described, while that might prevent accidentally breaking gateways/routers due to new spec versions:
1. As you've mentioned, this is only really an issue for folks who don't follow the update docs above.
2. This makes determining the required gateway/router version for a given build pipeline/composition version variable. I.e., you'd have to look at your subgraph schemas (or the supergraph schema) to determine which gateway/router versions you could use, and how you make that determination seems non-trivial (we'd probably need to write and maintain code for it).
3. This makes determining the required gateway/router version for a given build pipeline/composition version dynamic. I.e., because subgraphs can change over time, you have to worry about changes in them requiring changes in gateway/router versions. (We could use the code mentioned above to write a Studio check for this.)

[...]

PR #2528, opened 2023-04-14
First crack at an implementation of #2495

clenfest:
Important points
- We get the the version of any core feature implied by that version of federation, unless there has been a major version bump
- If there has been a major version bump of the core spec, we get the minimum required version that matches the major version of latest

### DIFF
--- a/composition-js/src/merging/merge.ts
+++ b/composition-js/src/merging/merge.ts
@@ -66,6 +66,11 @@ import {
   addSubgraphToError,
   printHumanReadableList,
   ArgumentMerger,
+  JoinSpecDefinition,
+  CoreSpecDefinition,
+  FeatureVersion,
+  FEDERATION_VERSIONS,
+  InaccessibleSpecDefinition,
 } from "@apollo/federation-internals";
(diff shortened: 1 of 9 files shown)


clenfest:
Note to reviewers: I'm slightly unsure on two points. 
1. federation.ts still uses `latest()` for the tag/link/federation specs. I think this is ok, in almost all cases, since as far as I can tell Merger doesn't use a FederationBlueprint for the schema. However,
2. Will this cause a problem with `extractSubgraphFromSupergraph`? I think it may, but I may have to give this some more thought.

trevor-scheer (changes requested):
Just a few minor suggestions.

Can we add some integration-ish tests for the `composeServices` fn which show various versions being used and the resulting supergraph?

Needs a changeset as well.

clenfest:
> 

So the reason I didn't put any tests into composeServices is that there isn't really anything to test just yet. Since we are saying that this isn't relevant for older versions, We'll always get the latest versions of joinSpec, etc. for now, and it won't become relevant until there is a new version of one of them.

trevor-scheer (approved):

PR #2538, opened 2023-04-20
dist-tag path was not found when updating packages

clenfest:
See 

https://github.com/apollographql/federation/actions/runs/4755972382

for an example of the failure


trevor-scheer (approved):
Oops, lgtm. Looks like I left a debugger and some stray logs in there too.

trevor-scheer (approved):

PR #2551, opened 2023-04-30
allow overriding sendRequest in RemoteGraphQLDataSource

srolel:
Hello! We had to copy `RemoteGraphQLDataSource` over to our codebase and make the changes in this PR in order to extend it with some caching functionality. The changes are:

- Rename existing `sendRequest` to `_sendRequest` (more of a placeholder name, open to suggestions)
- Add a replacement, public `sendRequest` method with a different type signature (takes `options` and returns both request actually made and the response)
- Move `apq` behavior  to `sendRequest` from `process`
- Change `sendRequest` call in `process`

Having this included upstream would definitely benefit us but perhaps others as well.


gateway-js/src/datasources/RemoteGraphQLDataSource.ts
```diff
@@ -225,6 +181,58 @@ export class RemoteGraphQLDataSource<
...
+      // the necessary "extensions" for APQ handshaking.
+      requestWithoutQuery.extensions = {
+        ...request.extensions,
+        persistedQuery: {
+          version: 1,
+          sha256Hash: apqHash,
+        },
+      };
+
+      const apqOptimisticResponse = await this._sendRequest(requestWithoutQuery, context as TContext);
```
srolel, line 211: had to cast `context` here as well as is done in `process`, could probably be refactored to only do it once somewhere else

PR #2556, opened 2023-05-02
Resolve `Promise` references for interface types before handing off to `resolveType` fn

### LINKED ISSUE #2552: __resolveType receiving Promise after __resolveReference for interfaces
### Issue Description

When using __resolveReference to resolve an interface, it ends up passing a promise to __resolveType. It is my understanding that while __resolveType can return a promise, it should not receive a promise.

In my example, if I take out the __resolveType, I get an expected error about needing to implement __resolveType or isTypeOf.  If I make the __resolveReference synchronous, everything works fine. But if it's async (like it would be if it hit an actual database), then the type of the object going into __resolveType is a promise which means that it won't match any of my conditionals. 

I was able to make a codesandbox with this issue. I was originally trying to do this with typegraph, but when I ran into the issue, I decided to take that out of the equation and came up with this. I have a PR over there as well: https://github.com/MichalLytek/type-graphql/pull/1448

### Link to Reproduction

https://codesandbox.io/p/sandbox/naughty-keldysh-phhss6?file=%2Fsrc%2Fproducts.ts

### Reproduction Steps

I've set it up so "Product" is an interface for `Book`s and `Movie`s.

Run this of :4000 (the supergraph):

```graphql
query ExampleQuery {
  reviews {
    id
      product {
        id
        name
        __typename
        ... on Book {
          author
        }
        ... on Movie {
          director
        }
      }
  
  }
}
```

You can also run the reference resolution query directly on the products subgraph (:4001):
```graphql
query ($representations: [_Any!]!) {
    _entities(representations: $representations) {
      ... on Product {
        name
      }
    }
  }

variable:
{"representations": [{"__typename": "Product", "id": "1"}]}
```

trevor-scheer:
Fixes #2552

Commits in order first reproduce the issue then resolve it.

Interface entities are a relatively new thing, but I'm surprised this hasn't surfaced yet since it seems like it'd be a showstopper for anyone trying to use it. If my understanding of the issue is correct, this seems like a reasonable solution... though it's possible the _issue_ is wrong.

FWIW, none of our `__resolveReference` functions in tests return `Promise`s, it might be more realistic if they did.

### DIFF
--- a/.changeset/weak-dryers-sort.md
+++ b/.changeset/weak-dryers-sort.md
@@ -1,0 +1,21 @@
+---
+"@apollo/subgraph": patch
+---
+
+Resolve `Promise` references before calling `__resolveType` on interface
+
+Since the introduction of entity interfaces, users could not return
+a `Promise` from `__resolveReference` while implementing a synchronous,
+custom `__resolveType` function. This change fixes/permits this use case.
+
+Additional background / implementation details:
+
+Returning a `Promise` from `__resolveReference` has historically never
+been an issue. However, with the introduction of entity interfaces, the
+calling of an interface's `__resolveType` function became a new concern.
+
+`__resolveType` functions expect a reference (and shouldn't be concerned
+with whether those references are wrapped in a `Promise`). In order to
+address this, we can `await` the reference before calling the
+`__resolveType` (this handles both the non-`Promise` and `Promise` case).
+  
--- a/gateway-js/src/__tests__/executeQueryPlan.test.ts
+++ b/gateway-js/src/__tests__/executeQueryPlan.test.ts
@@ -3802,10 +3802,15 @@ describe('executeQueryPlan', () => {
       s1,
     }: {
       s1?: {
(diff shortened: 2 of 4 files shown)


pcmanus (commented):
I do think this is the issue and the fix looks good to me.

Testing wise, I'll note that there is few variants on how you can make this work: you can skip the `__resolveType` but return the `__typename` in the `__resolveReference` instead, or you can use `isTypeOf` for instance, and those variants are tested [here](https://github.com/apollographql/federation/blob/main/gateway-js/src/__tests__/executeQueryPlan.test.ts#L4079-L4216), but indeed only for there synchronous versions. It might be nice to try to modify this (parameterised) test to include asynchronous variants of the various options. In particular, the added test in this patch tests an async `__resolveReference` but a sync `__resolveType`, but you can also have async `__resolveType` (with or without async `__resolveReference`), and while I believe the code is fine for those now, would be nice to cover everything.

pcmanus (approved):

PR #2559, opened 2023-05-04
Start testing against Node 20, simplify CI config

trevor-scheer:
This PR standardizes our CCI config around the commonly used `circleci/node` orb which handles caching and installing specific node versions for us. Additionally, it removes a bunch of cruft from our previous publishing workflow which is no longer useful to us.

This PR adds Node v20 to the testing matrix (and updates one test accordingly, where JSON parse error messages have changed slightly).

This appears to speed up CI by about a minute (this should be a fair comparison, both are "cached" runs):
Before:
https://app.circleci.com/pipelines/github/apollographql/federation/12359/workflows/ff2b0df2-e5c7-4b5f-81b2-f9dda8602fc4

After:
https://app.circleci.com/pipelines/github/apollographql/federation/12378/workflows/7e7dc21b-8efc-44c8-add7-0aaf90edcb54


trevor-scheer:
Note to reviewer: scroll the checks box for CI - the names have updated (so the expected checks will remain in the expected state)

trevor-scheer:
Merging with an offline approval from @pcmanus

PR #2575, opened 2023-05-15
Fix assertion error on some multi-field `@requires` dependencies

pcmanus:
When handling a `@requires`, the code gathers what it expected to be the set of "fetch groups" needed to fulfill the requirement, but in the rare cases, the same group can be added multiple times and in that case it was not de-duplicated. This duplication was unexpected by the rest of the code, ultimately leading to triggering an assertion.

This would happen when a `@requires` was on multiple fields, and one of those field was also a condition (say a key) for another of the required field.

### DIFF
--- a/query-planner-js/src/buildPlan.ts
+++ b/query-planner-js/src/buildPlan.ts
@@ -3628,6 +3628,14 @@ function maybeSubstratPathPrefix(basePath: OperationPath, maybePrefix: Operation
   return undefined;
 }
 
+function updateCreatedGroups(createdGroups: FetchGroup[], ...newCreatedGroups: FetchGroup[]) {
+  for (const newGroup of newCreatedGroups) {
+    if (!createdGroups.includes(newGroup)) {
+      createdGroups.push(newGroup);
+    }
+  }
+}
+
 function computeGroupsForTree(
   dependencyGraph: FetchDependencyGraph,
   pathTree: OpPathTree<any>,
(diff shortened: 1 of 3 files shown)


query-planner-js/src/buildPlan.ts
```diff
@@ -3649,7 +3657,7 @@ function computeGroupsForTree(
     context: initialContext,
     deferContext: initialDeferContext,
   }];
-  const createdGroups = [ ];
+  const createdGroups: FetchGroup[] = [ ];
```
pcmanus, line 3660: Note to reviewers: I hesitated changing this to a `Set<FetchGroup>`, but this turned out to a bit more changes that felt worth it. But if there is strong feeling about it, it can be changed.

query-planner-js/src/buildPlan.ts
```diff
@@ -3628,6 +3628,14 @@ function maybeSubstratPathPrefix(basePath: OperationPath, maybePrefix: Operation
   return undefined;
 }
 
+function updateCreatedGroups(createdGroups: FetchGroup[], ...newCreatedGroups: FetchGroup[]) {
```
clenfest, line 3631: So if I understand this correctly, the issue was that if a FetchGroup was present multiple times, the hander would not process the duplicates, leading to an assertion?

query-planner-js/src/buildPlan.ts
clenfest: A set would certainly be cleaner, but I'll trust your judgement about whether it's worth it to make the change or not.

PR #2593, opened 2023-05-22
Exposes `debug.maxEvaluatedPlans` planning config

pcmanus:
So far, the maximum number of query plan evaluated (above which the query planning eliminate choices to evaluated, thus potentially reducing the generated plan quality) has only be hard-coded. This exposes a config option to set that cap, mostly to help debugging query planning runtime issues.

### DIFF
--- a/query-planner-js/src/buildPlan.ts
+++ b/query-planner-js/src/buildPlan.ts
@@ -558,9 +558,7 @@ class QueryPlanningTraversal<RV extends Vertex> {
     debug.log(() => `Query has ${planCount} possible plans`);
 
     let firstBranch = this.closedBranches[0];
-    // Note: typing is not able to know that `maxEvaluatedPlans` is guaranteed to be set here, but it is
-    // due to early call to `enforceQueryPlannerConfigDefaults`.
-    const maxPlansToCompute = this.parameters.config.debug.maxEvaluatedPlans!;
+    const maxPlansToCompute = this.parameters.config.debug.maxEvaluatedPlans;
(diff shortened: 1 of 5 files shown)


query-planner-js/src/buildPlan.ts
```diff
@@ -565,7 +558,10 @@ class QueryPlanningTraversal<RV extends Vertex> {
     debug.log(() => `Query has ${planCount} possible plans`);
 
     let firstBranch = this.closedBranches[0];
-    while (planCount > MAX_COMPUTED_PLANS && firstBranch.length > 1) {
+    // Note: typing is not able to know that `maxEvaluatedPlans` is guaranteed to be set here, but it is
+    // due to early call to `enforceQueryPlannerConfigDefaults`.
+    const maxPlansToCompute = this.parameters.config.debug.maxEvaluatedPlans!;
```
clenfest, line 563: I really hate this. What if you created a new type and used it in place of `Concrete<QueryPlannerConfig>` in places it's used.

```
export type ConcreteQueryPlannerConfig = {
  [Property in keyof QueryPlannerConfig]-?: Concrete<QueryPlannerConfig[Property]>
}
```

That will let the properties of `debug` also be concrete.

pcmanus: Good point, didn't realised I could just call `Concrete` "recursively" on the property types. I did that in the definition of `Concrete` directly though, instead of creating a new specific type.

clenfest: Oh nice. I didn't realize you could do that either.

PR #2596, opened 2023-05-23
Fix possible fragment-related assertion error during query planning.

pcmanus:
The assertion throw had message of the form `Cannot add fragment of condition X (runtimes: ...) to parent type Y (runtimes: ...)` and was due to not always properly maintaining the "parent type" information when fragment spreads expanding into some other spread.

Additionally, this commit fixes a small issue in the code computing the "diff" of what remains in a selection set after a fragment has be "reused", which could lead to inefficient (and weird looking) selections in fetches.

### DIFF
--- a/query-planner-js/src/__tests__/buildPlan.test.ts
+++ b/query-planner-js/src/__tests__/buildPlan.test.ts
@@ -5748,3 +5748,76 @@ test('does not error out handling fragments when interface subtyping is involved
     }
   `);
 });
+
+
+test('handles mix of fragments indirection and unions', () => {
+  const subgraph1 = {
+    name: 'Subgraph1',
+    typeDefs: gql`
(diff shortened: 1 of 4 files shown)


query-planner-js/src/__tests__/buildPlan.test.ts
```diff
@@ -5672,3 +5672,76 @@ test('handles types with no common supertype at the same "mergeAt"', () => {
...
+        id: ID!
+      }
+
+      type Cat {
+        name: String
+      }
+    `
+  }
+
+  global.console = require('console')
```
trevor-scheer, line 5700: Stray / debugging change perhaps?
```suggestion
```

trevor-scheer (approved):
Thanks for the comments, lgtm!

PR #2598, opened 2023-05-23
Tweak managed fed docs to decommission duplicates in router

StephenBarlow:
We duplicate some managed federation content in the Router docs. These updates fold in changes made on the Router side so that we can decommission those duplicates

### DIFF
--- a/docs/source/managed-federation/setup.mdx
+++ b/docs/source/managed-federation/setup.mdx
@@ -3,30 +3,65 @@ title: Setting up managed federation
 ---
 
 import ObtainGraphApiKey from '../../shared/obtain-graph-api-key.mdx';
-import RegisterFederatedCli from '../../shared/register-federated-cli.mdx';
 
-This article describes how to set up Apollo Studio for a graph that uses Apollo Federation.
+This article describes how to connect your supergraph to [Apollo GraphOS](/graphos/) to enable managed federation features.
 
-> As with all changes, you should first set up managed federation in a non-production environment, such as staging. To support this, you can use [variants](https://www.apollographql.com/docs/studio/schema/registry/#managing-environments-with-variants), which are distinct versions of the same graph for different environments.
+> As with all changes, you should first set up managed federation in a non-production environment, such as staging. To support this, you can create multiple [variants](/graphos/graphs/#variants) of your supergraph in GraphOS Studio. Each variant represents a distinct version of the same graph for different environments.
 
 ## 1. Get started
 
-If you haven't yet, complete the first two steps from the Apollo Studio getting started guide:
+First, complete the **Set up Apollo tools** step from [this tutorial](/graphos/quickstart/self-hosted#1-set-up-apollo-tools), including:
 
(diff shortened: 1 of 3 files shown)


docs/source/managed-federation/setup.mdx
```diff
@@ -3,30 +3,65 @@ title: Setting up managed federation
...
-If you've already set up Apollo Federation _without_ Apollo Studio, the constructor of your `ApolloGateway` instance probably includes a `supergraphSdl` option, like this:
+As you register your subgraph schemas, the schema registry attempts to **compose** their latest versions into a single **supergraph schema**. Whenever composition succeeds, your managed router can fetch the latest supergraph schema from GraphOS.
+
+You can also manually fetch your latest supergraph schema with the `rover supergraph fetch` command, or retrieve it from your graph's **Schema > SDL** tab in GraphOS Studio.
+
+## 3. Modify your router's startup command
+
+<ExpansionPanel title="Using Apollo Gateway?">
+
+If you've already set up Apollo Federation _without_ conneting your gateway to GraphOS, the constructor of your `ApolloGateway` instance probably includes a `supergraphSdl` option, like this:
```
trevor-scheer, line 64:
```suggestion
If you've already set up Apollo Federation _without_ connecting your gateway to GraphOS, the constructor of your `ApolloGateway` instance probably includes a `supergraphSdl` option, like this:
```

docs/source/managed-federation/setup.mdx
```diff
@@ -36,35 +71,49 @@ const gateway = new ApolloGateway({
...
+    - This variable tells the router which variant of which graph to use (for example, `docs-example-graph@production`). You can find your variant's graph ref at the very top of its README page in GraphOS Studio.
 
-## 5. Deploy the modified gateway
+## 5. Deploy the modified router
 
-You can now deploy your modified gateway to begin fetching your federated schema from Studio instead of directly from your subgraphs.
+You can now deploy your modified router to begin fetching your supergraph schema from GraphOS instead of using a locally generated one.
 
-On startup, your gateway will use its API key to fetch its federation config from Apollo. It can then begin executing operations across your subgraphs.
+On startup, your router will use its API key to fetch its federation config from GraphOS. It can then begin executing operations across your subgraphs.
```
trevor-scheer, line 119: I think "federation config" is an ancient reference to the original gateway's method for constructing a supergraph, I think we can get rid of it 😄

```suggestion
On startup, your router will use its API key to fetch its supergraph from GraphOS. It can then begin executing operations across your subgraphs.
```

trevor-scheer (approved):

docs/source/managed-federation/setup.mdx
trevor-scheer: I'm disappointed our spell checker didn't catch this, I had to double check that it was working at all! It is working, and it thinks that `conneting` is a word.

StephenBarlow: Gasp! I am ashamed 😞

PR #2604, opened 2023-05-26
Faster approach to code for reusing fragments in fetches

pcmanus:
This commit somehwat "inverse" the way fragment reuse is tried. Before this commit, fragment reuse was tried depth-first, meaning that we tried reusing fragment on the leaf of the result set and then up on every selection. Unfortunately, doing so creates a number of subtlety that force the checks done on every selection to be somewhat costly, which adds up.

This commit change the process a bit to test for fragment reuse "at the top level" first (and so always fully expanded selections), and recurse down for the part that didn't matched anything. In practice, doing so simplify things a bit due to always dealing with expanded selections, and this allow to the checks simpler and more efficient.

Overall, this commit usually improves the time spend for trying fragment reuse during query planning (sometimes substantially), hence lowering query planning time.

### DIFF
--- a/.changeset/brown-bats-float.md
+++ b/.changeset/brown-bats-float.md
@@ -5,6 +5,7 @@
 "@apollo/gateway": patch
 ---
 
-Improves the performance of the code used to try to reuse query named fragments, thus improving query planning
-performance (at least for queries having named fragments).
-  
+Re-work the code use to try to reuse query named fragments to improve performance (thus sometimes improving query
+planning performance), to fix a possibly raised assertion error (with a message of form like `Cannot add selection of
+field X to selection set of parent type Y`), and to fix a rare issue where an interface or union field was not being
+queried for all the types it should be.
--- a/.changeset/brown-bats-float.md
+++ b/.changeset/brown-bats-float.md
@@ -1,0 +1,10 @@
+---
+"@apollo/query-planner": patch
+"@apollo/composition": patch
+"@apollo/federation-internals": patch
+"@apollo/gateway": patch
+---
+
+Improves the performance of the code used to try to reuse query named fragments, thus improving query planning
+performance (at least for queries having named fragments).
+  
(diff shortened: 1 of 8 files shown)


pcmanus:
I'll note that as mentioned on https://github.com/apollographql/router/issues/3172, while the motivation of this patch started as a performance improvement for the fragment reuse code, it also conceptually simplify said code, making easier to maintain. And it happens to fix the issue at https://github.com/apollographql/router/issues/3172, so this is technically a bug fix.

.changeset/brown-bats-float.md
```diff
@@ -0,0 +1,10 @@
+---
+"@apollo/query-planner": patch
+"@apollo/composition": patch
+"@apollo/federation-internals": patch
+"@apollo/gateway": patch
+---
+
+Improves the performance of the code used to try to reuse query named fragments, thus improving query planning
```
clenfest, line 8: Add a note about the other issues being fixed?

internals-js/src/operations.ts
```diff
@@ -696,13 +696,172 @@ export type RootOperationPath = {
...
+function computeFragmentsReverseDependencies(fragments: NamedFragments): SetMultiMap<string, string> {
+  const reverseDeps = new SetMultiMap<string, string>();
+  for (const fragment of fragments.definitions()) {
+    for (const dependency of fragment.fragmentUsages().keys()) {
+      reverseDeps.add(dependency, fragment.name);
+    }
+  }
+  return reverseDeps;
+}
+
```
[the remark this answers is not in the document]
pcmanus: Right, `computedFragmentDependents` is better. Changed to that.

.changeset/brown-bats-float.md
pcmanus: Good point. Re-worked the message to include them.

clenfest (approved):

PR #2610, opened 2023-05-31
Query planning performance improvements

pcmanus:
This PR contains 2 relatively small (in patch size) performance improvement for query planning (performance in the sense of the time it takes to compute query plans):
1. the first commit memoize when possible the result of the `FetchGroup.isUseless`, which was showing near the top of profiling some slow planning. This optimisation is arguably a minor one in that this probably only help specific cases, but it consistently reduced query planning time by ~15% on at least one example, so probably worth the small additional complexity.
2. the 2nd commit is arguably more impactful: in some situation with multiple keys, the planner was not discarding some options that are trivially inefficient (see commit message and included test for details). Which while not incorrect per-se, was sometime multiplying the number of plans evaluated by a lot, leading to very inefficient query planning time. This change sometimes decrease query planning time by an order of magnitude (some plan that took close to 2s now takes less than 100ms).

### DIFF
--- a/query-graphs-js/src/graphPath.ts
+++ b/query-graphs-js/src/graphPath.ts
@@ -1842,6 +1842,40 @@ export function advanceOptionsToString(options: (SimultaneousPaths<any> | Simult
   return '[\n  ' + options.map(opt => Array.isArray(opt) ? simultaneousPathsToString(opt, "  ") : opt.toString()).join('\n  ') + '\n]';
 }
 
+// Given a list of just computed indirect paths and a field that we're trying to advance after those paths, this
+// method fields any path that should note be considered.
+//
+// Currently, this handle the case where the key used at the end of the indirect path contains (at top level) the field
+// being queried. Or to make this more concrete, if we're trying to collect field `id`, and the path last edge was using
+// key `id`, then we can ignore that path because this imply that there is a way to `id` "some other way" (also see
+// the `does not evaluate plans relying on a key field to fetch that same field` test in `buildPlan` for more details).
+function filterNonCollectingPathsForField<V extends Vertex>(
+  paths: OpIndirectPaths<V>,
+  field: Field,
+): OpIndirectPaths<V> {
+  // We only handle leafs. Things are more complex non-leaf.
+  if (!field.isLeafField()) {
+    return paths;
+  }
+
+  const filtered = paths.paths.filter((p) => {
+    const lastEdge = p.lastEdge();
+    if (!lastEdge || lastEdge.transition.kind !== 'KeyResolution') {
+      return true;
+    }
+
+    const conditions = lastEdge.conditions;
+    return !(conditions && conditions.containsTopLevelField(field));
+  });
+  return filtered.length === paths.paths.length
+    ? paths
+    : {
+      ...paths,
+      paths: filtered
+    };
+
+}
+
 // Returns undefined if the operation cannot be dealt with/advanced. Otherwise, it returns a list of options we can be in after advancing the operation, each option
 // being a set of simultaneous paths in the subgraphs (a single path in the simple case, but type exploding may make us explore multiple paths simultaneously).
 // The lists of options can be empty, which has the special meaning that the operation is guaranteed to have no results (it corresponds to unsatisfiable conditions),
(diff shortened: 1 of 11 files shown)


query-planner-js/src/buildPlan.ts
```diff
@@ -1122,6 +1145,9 @@ class FetchGroup {
       return implementationInputSelections.length > 0
         && implementationInputSelections.every((input) => input.selectionSet.contains(subSelectionSet));
     });
+
+    this.isKnownUseful = !isUseless;
+    return isUseless;;
```
clenfest, line 1150:
```suggestion
    return isUseless;
```

query-planner-js/src/buildPlan.ts
```diff
@@ -1051,6 +1070,10 @@ class FetchGroup {
       return false;
     }
 
+    if (this.isKnownUseful) {
```
clenfest, line 1073: I'd just as this to the above clause on line 1069.

query-planner-js/src/buildPlan.ts
```diff
@@ -810,7 +813,21 @@ class FetchGroup {
...
+    // Cache used to save unecessary recomputation of the `isUseless` method.
+    private isKnownUseful: boolean = false,
   ) {
+    if (this._inputs) {
+      this._inputs.onUpdateCallback = () => {
+        // We're trying to avoid the full recomputation of `isUseless` when we're already
+        // shown that the group is known usefull (if it is shown useless, the group is removed,
+        // so we're not caching that result but it's ok). And `isUseless` basically checks if
+        // `inputs.contains(selection)`, so if a group is shown usefull, it means that there
+        // is some selections not in the inputs, but as long as we add to selections (and we
```
clenfest, line 825:
```suggestion
        // shown that the group is known useful (if it is shown useless, the group is removed,
        // so we're not caching that result but it's ok). And `isUseless` basically checks if
        // `inputs.contains(selection)`, so if a group is shown useful, it means that there
        // is some selections not in the inputs, but as long as we add to selections (and we
```

query-planner-js/src/buildPlan.ts
```diff
@@ -810,7 +813,21 @@ class FetchGroup {
     // key for that. Having it here saves us from re-computing it more than once.
     readonly subgraphAndMergeAtKey?: string,
     private cachedCost?: number,
+    // Cache used to save unecessary recomputation of the `isUseless` method.
+    private isKnownUseful: boolean = false,
```
clenfest, line 817: Wouldn't it be better to cache the status either way? i.e. the type should be `boolean | undefined` and you can just return the value if it's not undefined and run isUseful() otherwise? Obviously you should change the variable name if you do that.

query-graphs-js/src/graphPath.ts
```diff
@@ -1733,6 +1733,40 @@ export function advanceOptionsToString(options: (SimultaneousPaths<any> | Simult
...
+  });
+  return filtered.length === paths.paths.length
+    ? paths
+    : {
+      ...paths,
+      paths: filtered
+    };
+
+}
+
```
clenfest, line 1769: Can you add tests for this function?

clenfest (approved):

query-planner-js/src/buildPlan.ts
pcmanus: Done.

query-planner-js/src/buildPlan.ts
pcmanus: I don't think it'd be better. First, as I mentioned in one of the comment of the patch, when a group is shown to be useless, it is removed right away, so it doesn't really make matter whether that information is cached or not But adding to that, to properly cache the status either way, we'd have to add invalidation for that other way: that, changes to the group `selection` would have to invalid this cache, when currently only change to the `inputs` (which happens to be less frequent in practice) does so. So implementing the other would make things possibly less efficient overall.

query-graphs-js/src/graphPath.ts
pcmanus: Added some tests. Not directly for this function because I preferred keeping it non-exported/private, but for the function that directly call this (`advanceSimultaneousPathsWithOperation`) and for the behaviours that this method tackle.

PR #2612, opened 2023-06-02
Bug expanding the diff making reusing the fragment ineffective. 

AneethAnand:
Added an example in operation.test.ts file to illustrate the bug in ReUsing Fragments


pcmanus:
Thanks for the report, though this issue happens to be fixed on the last 2.4.7 which now correctly regenerate the original query (with all `OnSubT2` reused 4 times).

PR #2619, opened 2023-06-07
Fix issues with named fragment handling code

pcmanus:
This PR attempts to fix a number of remaining issues around the code to handle named fragments, namely at least:
1. the assertion failure reported in https://github.com/apollographql/router/issues/3227.
2. the assertion failure reported in https://github.com/apollographql/router/issues/3217 (which is the same assertion failing than in the previous point, but on a slightly different code path and due to a different issue).
3. an issue whereby named fragments are reused in a rare situation where this is invalid, resulting in an invalid (in the sense of graphQL validations) subgraph fetch.

The 2 first point have repro on the linked tickets, but for the last point, the problem is that the graphQL validation for conflicting fields specified by [FieldsInSetCanMerge](https://spec.graphql.org/draft/#FieldsInSetCanMerge()) traverses all fragments to look for conflicts, even "branches" that can be statically proven to be dead branches, but because the named fragment reuse code was taking those dead branch into account, it was sometimes using a named fragment that created a validation issue (in such a dead branch). See https://github.com/apollographql/federation/commit/e1d104484b466e5eafde306820543a08131e8419 for an example of the problem with some explanations.

### DIFF
--- a/internals-js/src/operations.ts
+++ b/internals-js/src/operations.ts
@@ -1082,13 +1082,32 @@ export class NamedFragmentDefinition extends DirectiveTargetElement<NamedFragmen
   }
 
   /**
-   * Whether this fragment may apply at the provided type, that is if either:
-   *  - its type condition is equal to the provided type.
-   *  - or the runtime types of the provided type include all of those of the fragment condition. 
+   * Whether this fragment may apply _directly_ at the provided type, meaning that the fragment sub-selection
+   * could be put directly inside a `... on type {}` inline fragment (_without_ re-adding the fragment condition
+   * that is), and both be valid and not "lose context".
+   *
+   * For that to be true, we need one of this to be true:
+   * 1. the runtime types of the fragment condition must be at least as general as those of the provided `type`.
+   *    Otherwise, putting it at `type` without its condition would "generalize" more than fragment meant to (and
+   *    so we'd "lose context"
+   * 2. either `type` and `this.typeCondition` are equal, or `type` is an object or `this.typeCondition` is a union
+   *    The idea is that, assuming our 1st point, then:
+   *    - if both are equal, things works trivially.
+   *    - if `type` is an object, `this.typeCondition` is either the same object, or a union/interface for which
+   *      type is a valid runtime. In all case, anything valid on `this.typeCondition` would apply to `type` too.
+   *    - if `this.typeCondition` is a union, then it's selection can only have fragments on object types at top-level
+   *      (save for `__typename`), and all those selection will work at `type` too.
+   *    But in any other case, both types must be abstract (if `this.typeCondition` is an object, the 1st condition
+   *    imply `type` can only be the same type) and we're in one of:
+   *    - `type` and `this.typeCondition` are both different interfaces (that intersect but are different).
+   *    - `type` is aunion and `this.typeCondition` an interface.
+   *       And in both cases, the selection of the fragment may selection an interface that is not valid at `type` (if `type`
+   *       is a union because a direct field is always wrong, and if `type` is another interface because that interface may
+   *       not have that particular field).
    *
    * @param type - the type at which we're looking at applying the fragment
    */
-  canApplyAtType(type: CompositeType): boolean {
+  canApplyDirectlyAtType(type: CompositeType): boolean {
     if (sameType(type, this.typeCondition)) {
       return true;
     }
@@ -1101,17 +1120,20 @@ export class NamedFragmentDefinition extends DirectiveTargetElement<NamedFragmen
 
     const conditionRuntimes = possibleRuntimeTypes(this.typeCondition);
     const typeRuntimes = possibleRuntimeTypes(type);
-
-    // The fragment condition must be at least as general as the provided type (so that if we use the fragment
-    // inside `type`, then it doesn't add restriction that weren't there without the fragment).
+    // The fragment condition must be at least as general as the provided type (in other words, all of the
+    // runtimes of `type` must be in `conditionRuntimes`).
     // Note: the `length` test is technically redundant, but just avoid the more costly sub-set check if we
     // can cheaply show it's unnecessary.
-    return conditionRuntimes.length >= typeRuntimes.length
-     && typeRuntimes.every((t1) => conditionRuntimes.some((t2) => sameType(t1, t2)));
+    if (conditionRuntimes.length < typeRuntimes.length
+      || !typeRuntimes.every((t1) => conditionRuntimes.some((t2) => sameType(t1, t2)))) {
+      return false;
+    }
+
+    return isObjectType(type) || isUnionType(this.typeCondition);
   }
 
   /**
-   * This methods *assumes* that `this.canApplyAtType(type)` is `true` (and may crash if this is not true), and returns
+   * This methods *assumes* that `this.canApplyDirectlyAtType(type)` is `true` (and may crash if this is not true), and returns
    * a version fo this named fragment selection set that corresponds to the "expansion" of this named fragment at `type`
    *
    * The overall idea here is that if we have an interface I with 2 implementations T1 and T2, and we have a fragment like:
@@ -1129,7 +1151,7 @@ export class NamedFragmentDefinition extends DirectiveTargetElement<NamedFragmen
    * us that part.
    */
   expandedSelectionSetAtType(type: CompositeType): SelectionSetAtType {
-    // First, if the candidate condition is an object or is the type passed, then there isn't anyrestriction to do.
+    // First, if the candidate condition is an object or is the type passed, then there isn't any restriction to do.
     if (sameType(type, this.typeCondition) || isObjectType(this.typeCondition)) {
       return { selectionSet: this.expandedSelectionSet() };
     }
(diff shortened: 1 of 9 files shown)


query-planner-js/src/buildPlan.ts
```diff
@@ -2806,7 +2788,6 @@ export class QueryPlanner {
     // going to expand everything during the algorithm anyway. We'll re-optimize subgraph fetches with fragments
     // later if possible (which is why we saved them above before expansion).
     operation = operation.expandAllFragments();
```
clenfest, line 2790: There is no reason to normalize at this point?

internals-js/src/operations.ts
```diff
@@ -865,6 +866,25 @@ export class Operation {
     readonly name?: string) {
   }
 
+  // Returns a copy of this operation with the provided updated selection set. Optionally, a new set of fragments
+  // can be also provided. If `newFragments` is undefined, then the existing fragments will be reused, but if it
+  // `null`, then the new operation will have no fragments.
+  private withUpdatedSelectionSet(newSelectionSet: SelectionSet, newFragments?: NamedFragments | null): Operation {
```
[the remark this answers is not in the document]
pcmanus: Alright, I've split it into 2 separately named methods.

internals-js/src/operations.ts
```diff
@@ -1523,8 +1554,82 @@ export class SelectionSet {
...
+   * For this operation to be valid (to not throw), `parentType` must be such this selection set would
+   * be valid as a subselection of an inline fragment `... on parentType { <this selection set> }` (and
+   * so `this.normalize(this.parentType)` is always valid and useful, but it is possible to pass a `parentType`
+   * that is more "restrictive" than the selection current parent type).
+   *
+   * Passing the option `recursive == false` makes the normalization only apply at the top-level, removing
+   * any unecessary top-level inline fragments, possibly multiple layers of them, but we never recurse
+   * inside the sub-selection of an selection that is not removed by the normalization.
+   */
+  normalize(parentType: CompositeType, options?: { recursive? : boolean }): SelectionSet {
```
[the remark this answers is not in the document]
pcmanus: I've changed it for the sake of not being too contrarian, but I'll admit not entirely seeing how this isn't almost entirely a stylistic preference.

query-planner-js/src/buildPlan.ts
pcmanus: Not now that `expandAllFragments` always ensure it's output is normalized.

internals-js/src/operations.ts
```diff
@@ -2889,13 +3104,13 @@ class InlineFragmentSelection extends FragmentSelection {
...
         newSet.add(this.withUpdatedSelectionSet(
-          trimmedSelectionSet.filter((s) => !liftableSelections.includes(s)),
+          normalizedSelectionSet.filter((s) => !liftableSelections.includes(s)),
         ));
         return newSet.toSelectionSet(this.parentType);
       }
     }
 
-    return this.selectionSet === trimmedSelectionSet ? this : this.withUpdatedSelectionSet(trimmedSelectionSet);
+    return this.selectionSet === normalizedSelectionSet ? this : this.withUpdatedSelectionSet(normalizedSelectionSet);
```
sachindshinde, line 3113: I'm noticing that unlike `FieldSelection.normalize()` and `InlineFragmentSelection.normalize()` which update `this.parentType` to `parentType`, `InlineFragmentSelection.normalize()` doesn't seem to, e.g.:
- The `return this.selectionSet === normalizedSelectionSet ? this : this.withUpdatedSelectionSet(normalizedSelectionSet);` doesn't update the element's `parentType`.
- The `return this.withUpdatedSelectionSet(selectionSetOfElement( ... )` above doesn't update the element's `parentType`.
- The `return newSet.toSelectionSet(this.parentType)` above uses `this.parentType` instead of `parentType`.
- There's a `(this.element.typeCondition ?? this.parentType).typenameField()!` call above instead of `(thisCondition ?? parentType).typenameField()!,`

Also I noticed that when you lift selections out of the fragment in this function, the `parentType` isn't accordingly updated for those selections.

Is this fine/expected? (It might be that the field is only used in certain cases, or maybe some other code requires it to not be changed, but I'm not familiar enough with the code here.)

sachindshinde (changes requested):

internals-js/src/operations.ts
```diff
@@ -2369,23 +2494,93 @@ abstract class AbstractSelection<TElement extends OperationElement, TIsLeaf exte
...
     const optimized = new SelectionSetUpdates();
-    // TODO: doing repeated calls to `minus` for every fragment is simple, but a `minusAll` method that
-    // takes the fragment selections at once would be more efficient in pratice.
-    for (const fragment of filteredApplyingFragments) {
-      // Note: we call `expandedSelectionSetAType` twice in this method for the applying fragments, but
-      // we know it's cached so rely on that fact.
-      notCoveredByFragments = notCoveredByFragments.minus(fragment.expandedSelectionSetAtType(parentType));
+    for (const { fragment, atType} of filteredApplyingFragments) {
+      const notCovered = subSelection.minus(atType.selectionSet);
+      if (atType.trimmed && !selectionSetsDoMerge(notCovered, atType.trimmed)) {
```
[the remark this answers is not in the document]
pcmanus: That's a very good point. Added some testing for this case and fixed.

internals-js/src/operations.ts
```diff
@@ -1117,34 +1128,51 @@ export class NamedFragmentDefinition extends DirectiveTargetElement<NamedFragmen
...
+      // of the fragment that are kept, which don't apply here. And calling `normalize` would not be valid because
+      // one of its effect is to "rebase" on `type` which may be invalid here. Namely, we could have `type === I1`
+      // and the fragment be: `fragment X on I2 { f }`, and even having `runtimes(I1)` included in `runtimes(I2)`
+      // does not mean that `f` if a valid field for `I1`.
+      const typeRuntimes = possibleRuntimeTypes(type);
+      const conditionRuntimes = possibleRuntimeTypes(this.typeCondition);
+      selectionSet = typeRuntimes.length === conditionRuntimes.length
+        ? expandedSelectionSet
+        : expandedSelectionSet.filter((s) =>
+          s.kind !== 'FragmentSelection' || !s.element.typeCondition || typeRuntimes.some((t) => t.name == s.element.typeCondition?.name)
```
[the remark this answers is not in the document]
pcmanus: Yes, you're right; I brain-farted. I added test for this, and in the process realised the  that I was using the `filter` method wrong as the one implemented was recursing inside sub-selection, which we don't want. Anyway, this was confusing and I actually renamed the existing filter into `filterRecursiveDepthFirst` to be extra clear and kept a simpler `filter` method which only apply to top-level (so this part of the code hasn't changed, but just explaining why there is some filter-related changes in the last commit).

internals-js/src/operations.ts
pcmanus: No,I was being sloppy.

I mean, it is probably mostly fine in the grand scheme of stuff because when we build a `SelectionSet` with `SelectionSetUpdates` (which `SelectionSet.normalize` does through `lazyMap`), then it ensures the elements have the proper parent type by [calling `rebasOn`](https://github.com/apollographql/federation/blob/main/internals-js/src/operations.ts#L1996C1-L2022), so I believe the selections still ended up with the proper parent type ultimately.

Still, that was sloppy and would be unexpected at least out of the `Selection.normlize` call and could cause issue someday, so fixed it.

> you lift selections out of the fragment in this function, the `parentType` isn't accordingly updated for those selections

As mentioned above, the fact that we passed the lifted selections to a `SelectionSetUpdates` took care of updating the parent type of the elements (of course, as you pointed out, the `toSelectionSet` call for that `SelectionSetUpdates` was incorrect, so did fixed that, but just calling out that there is nothing additional to do).

pcmanus:
Note: the node 20 circleCI check on this PR seems to have failed a bunch of times in a row, yet the error seems CI related. Check the link for detail, but the error is:
```
npm ERR! code 2
npm ERR! path /home/circleci/project/node_modules/@parcel/watcher
npm ERR! command failed
npm ERR! command sh -c node-gyp-build
npm ERR! sh: 1: node-gyp-build: Text file busy
```
during npm package installing, and for a PR that don't even change any dependency, so it screams "unrelated to the PR". Yet, I haven't (yet) seen this on other PRs and as said, I got this here a few time in a row now, so wondering if anyone has a clue.

internals-js/src/operations.ts
```diff
@@ -1117,34 +1128,63 @@ export class NamedFragmentDefinition extends DirectiveTargetElement<NamedFragmen
...
+      // one of its effect is to "rebase" on `type` which may be invalid here. Namely, we could have `type === I1`
+      // and the fragment be: `fragment X on I2 { f }`, and even having `runtimes(I1)` included in `runtimes(I2)`
+      // does not mean that `f` if a valid field for `I1`.
+      //
+      // Note: reminder here that this method assumes `this.canApplyAtType(type)` and that means that we already
+      // know that `typeRuntimes` is a (non-necessarily strict) subset of `conditionRuntimes`. And so if they
+      // have the same number of runtimes, then they have the same set of runtimes and no point doing any filtering.
+      const typeRuntimes = possibleRuntimeTypes(type);
+      const conditionRuntimes = possibleRuntimeTypes(this.typeCondition);
+      selectionSet = typeRuntimes.length === conditionRuntimes.length
```
sachindshinde, line 1171: `selectionSet` in the `if` block above will be rebased on `type` due to the `SelectionSet.normalize()` call, but it looks like we'd have to manually do that rebasing here. (Not sure if it really matters, since it looks like this function is only used by `tryOptimizeSubselectionWithFragments()`.)

sachindshinde (approved):
Thanks for making the fixes! There's a few small things below, but other than that LGTM

sachindshinde:
Regarding the Node 20 issues, I couldn't immediately pinpoint a cause.

I tried to SSH into the box and look at the debug logs. Toward the end I saw:
```
4635 timing build:link Completed in 11ms
4636 info run @parcel/watcher@2.1.0 install node_modules/@parcel/watcher node-gyp-build
4637 info run dtrace-provider@0.8.8 install node_modules/dtrace-provider node-gyp rebuild || node suppress-error.js
4638 info run @parcel/watcher@2.1.0 install { code: 2, signal: null }
4639 timing reify:rollback:createSparse Completed in 1708ms
4640 timing reify:rollback:retireShallow Completed in 0ms
4641 timing command:ci Completed in 19138ms
4642 verbose stack Error: command failed
4642 verbose stack     at ChildProcess.<anonymous> (/home/circleci/.nvm/versions/node/v20.3.0/lib/node_modules/npm/node_modules/@npmcli/promise-spawn/lib/index.js:53:27)
4642 verbose stack     at ChildProcess.emit (node:events:511:28)
4642 verbose stack     at maybeClose (node:internal/child_process:1098:16)
4642 verbose stack     at Socket.<anonymous> (node:internal/child_process:456:11)
4642 verbose stack     at Socket.emit (node:events:511:28)
4642 verbose stack     at Pipe.<anonymous> (node:net:334:12)
4643 verbose pkgid @parcel/watcher@2.1.0
4644 verbose cwd /home/circleci/project
4645 verbose Linux 5.15.0-1030-aws
4646 verbose node v20.3.0
4647 verbose npm  v9.7.1
4648 error code 2
4649 error path /home/circleci/project/node_modules/@parcel/watcher
4650 error command failed
4651 error command sh -c node-gyp-build
4652 error sh: 1: node-gyp-build: Text file busy
4653 verbose exit 2
4654 timing npm Completed in 19317ms
4655 verbose unfinished npm timer reify 1686630640693
4656 verbose unfinished npm timer reify:build 1686630657224
4657 verbose unfinished npm timer build 1686630657226
4658 verbose unfinished npm timer build:deps 1686630657227
4659 verbose unfinished npm timer build:run:install 1686630657258
4660 verbose unfinished npm timer build:run:install:node_modules/@parcel/watcher 1686630657259
4661 verbose unfinished npm timer build:run:install:node_modules/dtrace-provider 1686630657309
4662 verbose code 2
```

From some googling, `sh: 1: node-gyp-build: Text file busy` seems to mean that something was trying to overwrite `node-gyp-build`, but it was in the middle of being executed.

In any case, I changed my volta config in my `package.json` to the above (node `20.3.0` and npm `9.7.1`) and ran tests, and they passed. So it's probably fine to ignore the Node 20 CI error (hopefully it won't re-appear on `main`).

abernix:
@pcmanus:

> Note: the node 20 circleCI check on this PR seems to have failed a bunch of times in a row, yet the error seems CI related. Check the link for detail, but the error is:

Just triangulating a theme I'm seeing around Node.js v20.3 with the `Text file busy` error, a very similar occurrence of that error happened with something _totally_ different in this PR from @jerelmiller on another repository: https://github.com/apollographql/apollo-utils/pull/302.

I think that either Node.js v20.3 **OR** CircleCI's Node.js v20.3 image are potentially the culprit. I don't yet know if that's foreshadowing something bigger that will break in that version or just a bug.  Hoping for the latter.

internals-js/src/operations.ts
```diff
@@ -3106,11 +3156,13 @@ class InlineFragmentSelection extends FragmentSelection {
         newSet.add(this.withUpdatedSelectionSet(
           normalizedSelectionSet.filter((s) => !liftableSelections.includes(s)),
         ));
-        return newSet.toSelectionSet(this.parentType);
+        return newSet.toSelectionSet(parentType);
       }
     }
 
-    return this.selectionSet === normalizedSelectionSet ? this : this.withUpdatedSelectionSet(normalizedSelectionSet);
+    return this.selectionSet === normalizedSelectionSet
```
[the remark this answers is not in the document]
pcmanus: Yes, good catch, thanks.

internals-js/src/operations.ts
pcmanus: You're right, and this actually made me realise that I was making this more complicated than it needed to be.

Because calling `rebaseOn(type)` on the `if` block above is actually usually incorrect. Typically, that `if` block was taken when, say, `type` was some interface `I1` while `this.typeCondition` was another interface `I2` (assuming the runtimes match what expect). But in general, you can't just rebase fields of `I2` into `I1` directly, even it happens that it "would" be alright from the POV of the underlying possible runtimes.

So really, the issue was that the `canApplyAtType` method, which was guarding calls to this `expandedSelectionSetAtType` should just not allow those cases where calling `normalize(type)` does not work. So fixed that and simplified accordingly, and I've validated that fragments are still reused as well as before.

@sachindshinde since the change is probably a bit bigger than you intended, I'll give you the chance to have a look at the last commit/check my logic before merging.

jerelmiller:
@abernix interestingly I was able to get it working again using the Node 20.2 image: https://github.com/apollographql/apollo-utils/pull/306 so seems its specific to the 20.3 image. I hope this is just a temp fix and as you said, hope its not a foreshadowing of something worse.

internals-js/src/operations.ts
pcmanus: > I'll give you the chance to have a look at the last commit/check my logic before merging.

Slightly changed my mind given circumstance: I optimistically merged, but do raise any concern if you have and we'll address.

PR #2622, opened 2023-06-09
Fix fetch group processing handling of dependencies

pcmanus:
The code that process `FetchGroup` in dependency order has a bug that, for some types of dependencies, can finish with some group not handled, even though they could and should have been.

In practice, this shows itself by failing the post-processing assertion that double-check that no groups are left unhandled.

The error message when this triggered looks like `Root groups X should have no remaining groups unhandled (...)` (where `X` is a dump of the fetch groups involved and so this can be large and hard to parse, but basically the assertion error starts with "Root groups").

### DIFF
--- a/query-planner-js/src/buildPlan.ts
+++ b/query-planner-js/src/buildPlan.ts
@@ -793,15 +793,6 @@ class QueryPlanningTraversal<RV extends Vertex> {
   }
 }
 
-type UnhandledGroups = [FetchGroup, UnhandledParentRelations][];
-type UnhandledParentRelations = ParentRelation[];
-
-function printUnhandled(u: UnhandledGroups): string {
-  return '[' + u.map(([g, relations]) =>
-    `${g.index} (missing: [${relations.map((r) => r.group.index).join(', ')}])`
-  ).join(', ') + ']';
-}
-
 /**
  * Used in `FetchDependencyGraph` to store, for a given group, information about one of its parent.
  * Namely, this structure stores:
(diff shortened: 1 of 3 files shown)


query-planner-js/src/buildPlan.ts
```diff
@@ -2495,6 +2500,26 @@ class FetchDependencyGraph {
     };
   }
 
+  private updateUnhandledOnProcessedGroups(
+    processed: readonly FetchGroup[],
+    currentUnhandled: UnhandledGroups,
+    readyToHandle: FetchGroup[],
+  ): UnhandledGroups {
+    const unhandled: UnhandledGroups = [];
+    for (const current of currentUnhandled) {
```
sachindshinde, line 2509: minor nit: To be consistent with `mergeRemainings()`/make code more clear, could you change `current` to e.g. `[g, edges]`? (`edges` is more clear than `current[1]`)

sachindshinde (changes requested):

query-planner-js/src/buildPlan.ts
```diff
@@ -2423,19 +2429,18 @@ class FetchDependencyGraph {
...
     for (const [g, edges] of r1) {
       const newEdges = this.mergeRemaingsAndRemoveIfFound(g, edges, r2);
       if (newEdges.length == 0) {
-        toHandle.push(g);
+        readyToHandle.push(g);
       } else {
         unhandled.push([g, newEdges])
       }
     }
     unhandled.push(...r2);
```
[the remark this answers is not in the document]
pcmanus: You're right this was fishy here. The intend though was that `UnhandledGroups` would never contain group whose "edges" are empty (both methods that modify it make sure to remove groups as soon as their edges are all empty). But the "bug" was indeed that `processGroup` could "generate" an `UnhandledGroups` with "zero-edges" groups and that's where the code was confused. I've addressed in the small refactor I did (it now does genuinely avoid created zero-edge entries in `UnhandledGroups`, well, at least unless I brain-farted again obviously).

query-planner-js/src/buildPlan.ts
```diff
@@ -2483,9 +2488,9 @@ class FetchDependencyGraph {
       // After the root groups, handled on the first iteration, we can process everything in parallel.
       processInParallel = true;
       mainSequence.push(processed);
-      const [canHandle, newRemaining] = this.mergeRemainings(remainingNext, unhandled);
-      remainingNext = newRemaining;
-      nextGroups = canHandle.concat(next);
+      remainingNext = this.mergeRemainings(remainingNext, unhandled, next);
```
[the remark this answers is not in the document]
pcmanus: You're right, this was unnecessary (and probably dangerous).

The overall code around this was admittedly a bit confusing, so I did a small refactor to hopefully clear things up a bit. It should be somewhat equivalent to what we have before (save the real bug you rightfully pointed in the previous comment) but I think encapsulate things a tad better/is more explicit. It's probably not perfect, I think this processing code could maybe be cleaned up som more, but I'll leave that for later (started a slightly bigger refactor and messed it up, so figured I should leave that separate from this bug-fixing).

query-planner-js/src/buildPlan.ts
```diff
@@ -2452,19 +2561,19 @@ class FetchDependencyGraph {
...
     const processed = processor.onFetchGroup(group, newHandledConditions);
     if (children.length == 0) {
-      return { main: processor.onConditions(conditions, processed), unhandled: [], deferredGroups };
+      return { main: processor.onConditions(conditions, processed), state: ProcessingState.empty(), deferredGroups };
     }
 
-    const groupIsOnlyParentOfAllChildren = children.every(g => g.parents().length === 1);
-    if (groupIsOnlyParentOfAllChildren) {
+    const state = ProcessingState.forChildrenOfProcessedGroup(group, children);
+    if (state.next.length > 0) {
```
sachindshinde, line 2576: Note that there's a logic change here.

Before this change, this `if` block executes only if all children have exactly one parent. After this change, this `if` block executes if at least one child has exactly one parent.

This might be fine/intentional, given that you've also swapped out `children` down below for `state.next`. Though in that case, you should update the comment below to say e.g. `ready children` instead of just `children`.

query-planner-js/src/buildPlan.ts
```diff
@@ -2473,86 +2582,55 @@ class FetchDependencyGraph {
         deferredGroups: allDeferredGroups,
       } = this.processRootMainGroups({
```
sachindshinde, line 2583: Note that because `processRootMainGroups()` doesn't take the old `state`/return a new `state`, presumably when it eventually calls `updateForProcessedGroups()` for the `state` it internally creates, it won't alter the separate `state` in the upstream `processGroups()` call. This was the case before this PR too though, so it's probably fine to punt that to another PR if you'd like. (It might also be the case this was intentional.)

sachindshinde (approved):
Thanks for making the changes! There's some comments below, but nothing really blocking this PR from merging.

PR #2623, opened 2023-06-09
Over-aggressive optimisation can lead to generating non-optimal query plans

pcmanus:
The query planner has a number of heuristic it applies when computing possible paths to avoid generating options that can be proved early  to never be better than other ones (as a basic example, if we have the same key between A -> B, A -> C and B -> C, and we're trying to go from A to C, then no point in considering doing A -> B -> C since we can do A -> C directly). Anyway, the point is that if we're not a bit smart in eliminating such "unnecessary" options, the number of final plans to evaluate can quickly explode (and make query planning long, up to the point where the planner stops degrading its output).

But one of the heuristic we're applying is unfortunately a bit too aggressive: it does remove options that are unnecessary, but was also removing options that should have been kept.  The result is that the planner will sometime return plans that are not optimal (see the added test for an example).

This commit fix that problem by making the heuristic more precise and only remove the option that are safe to remove.

Side-note: the first commit of this PR is actually slightly unrelated to the description above, it is a small optimisation when generating plans. But I discovered both at the same time, and that optimisation was fairly simple, so took the liberty to include.

### DIFF
--- a/query-graphs-js/src/graphPath.ts
+++ b/query-graphs-js/src/graphPath.ts
@@ -240,18 +240,18 @@ export class GraphPath<TTrigger, RV extends Vertex = Vertex, TNullEdge extends n
    *
    * Please note that this method assumes that the 2 paths have the same root, and will fail if that's not the case.
    */
-  findBiggestCommonPrefixAndCountMainSubgraphJumpsAfterIt(that: GraphPath<TTrigger, RV, TNullEdge>): {
+  countSubgraphJumpsAfterLastCommonVertex(that: GraphPath<TTrigger, RV, TNullEdge>): {
     thisJumps: number,
     thatJumps: number
   } {
-    const { vertex, index } = this.findBiggestCommonPrefixEndIndex(that);
+    const { vertex, index } = this.findLastCommonVertex(that);
     return {
       thisJumps: this.subgraphJumpsAtIdx(vertex, index),
       thatJumps: that.subgraphJumpsAtIdx(vertex, index),
     };
   }
 
-  private findBiggestCommonPrefixEndIndex(that: GraphPath<TTrigger, RV, TNullEdge>): { vertex: Vertex, index: number } {
+  private findLastCommonVertex(that: GraphPath<TTrigger, RV, TNullEdge>): { vertex: Vertex, index: number } {
     let vertex: Vertex = this.root;
     assert(that.root === vertex, () => `Expected both path to start on the same root, but 'this' has root ${vertex} while 'that' has ${that.root}`);
 
--- a/query-graphs-js/src/graphPath.ts
+++ b/query-graphs-js/src/graphPath.ts
@@ -232,20 +232,129 @@ export class GraphPath<TTrigger, RV extends Vertex = Vertex, TNullEdge extends n
     return this.props.edgeIndexes.length;
   }
 
-  subgraphJumps(): number {
+  /**
+   * That method first look for the biggest common prefix to `this` and `that` (assuming that both path are build as choices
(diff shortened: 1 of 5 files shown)


query-graphs-js/src/graphPath.ts
```diff
@@ -232,22 +232,131 @@ export class GraphPath<TTrigger, RV extends Vertex = Vertex, TNullEdge extends n
...
-  subgraphJumps(): number {
+  /**
+   * That method first look for the biggest common prefix to `this` and `that` (assuming that both path are build as choices
+   * of the same "query path"), and the count how many subgraph jumps each of the path has after said prefix.
+   *
+   * Note that this method always return someting but the biggest common prefix considered might well be empty. 
+   *
+   * Please note that this method assumes that the 2 paths have the same root, and will fail if that's not the case.
+   */
+  findBiggestCommonPrefixAndCountMainSubgraphJumpsAfterIt(that: GraphPath<TTrigger, RV, TNullEdge>): {
```
clenfest, line 243: That's a mouthful. How about something like `countSubgraphJumpsAfterCommon` or something?

query-graphs-js/src/graphPath.ts
```diff
@@ -232,22 +232,131 @@ export class GraphPath<TTrigger, RV extends Vertex = Vertex, TNullEdge extends n
...
+    thatJumps: number
+  } {
+    const { vertex, index } = this.findBiggestCommonPrefixEndIndex(that);
+    return {
+      thisJumps: this.subgraphJumpsAtIdx(vertex, index),
+      thatJumps: that.subgraphJumpsAtIdx(vertex, index),
+    };
+  }
+
+  private findBiggestCommonPrefixEndIndex(that: GraphPath<TTrigger, RV, TNullEdge>): { vertex: Vertex, index: number } {
```
clenfest, line 254: Maybe `findLastCommonVertexIndex` instead?

query-graphs-js/src/graphPath.ts
```diff
@@ -232,22 +232,131 @@ export class GraphPath<TTrigger, RV extends Vertex = Vertex, TNullEdge extends n
...
+          || !thisEdge.conditions!.equals(thatEdge.conditions!)
+        ) {
+          return false;
+        }
+
+        // So far, so good. `thisV` and `thatV` are positioned on the vertex after which the path
+        // must be equal again. So check that it's true, and if it is, we're good.
+        // Note that for `this`, the last edge we looked at was `i`, so the next is `i+1`. And
+        // for `that`, we've skipped over one more edge, so need to use `j+1`.
+        for (let j = i + 1; j < this.size; j++) {
```
clenfest, line 337: It's a little ugly to have a loop inside the previous loop that is really just a continuation of the first loop. Would be great if you could move this outside the enclosing for loop.

clenfest (approved):
Just as a general note, it would be great if we could get to the point where we have unit tests for each of these individual functions we're creating.

query-graphs-js/src/graphPath.ts
pcmanus: Yeah, make sense, renamed to `findLastCommonVertex` (we return both the vertex and index, so figure the `Index` part was unnecessary).

query-graphs-js/src/graphPath.ts
pcmanus: Renamed to `countSubgraphJumpsAfterLastCommonVertex` to match the point below.

query-graphs-js/src/graphPath.ts
pcmanus: I think that one might be a personal preference: as it is a continuation in only a specific case, moving that inner loop out means some sort of state tracking that I don't find more obvious.

PR #2640, opened 2023-06-23
Update description of override directive

StephenBarlow:


docs/source/federated-types/federated-directives.mdx
```diff
@@ -338,8 +337,10 @@ For more information, see [Migrating entities and fields](../entities-advanced/#
 
 **Required.** The name of the other subgraph that no longer resolves the field.
 
-* **If you're performing composition with managed federation,** this must match the name of the subgraph registered to GraphOS.
-* **If you're performing composition with the Rover CLI,** this must match the name of the subgraph in the YAML config file you provide to `rover supergraph compose`.
+* **If you perform composition with managed federation,** this corresponds to the subgraph name you've published (or will publish) to GraphOS.
+* **If you perform composition with the Rover CLI,** this corresponds to the subgraph name in the YAML config file you provide to `rover supergraph compose`.
+
+You _can_ provide the name of a subgraph that you haven't yet added to your supergraph. When you later add that subgraph, composition honors the directive.
```
StephenBarlow, line 343: The primary content _addition_ is this line

trevor-scheer: I think this would be more helpful with an explanation of the use case. It might be helpful if we contrive an example (maybe above) to demonstrate why someone might want to use this.

PR #2649, opened 2023-06-30
Introduce `@requiresScopes` directive in composition

trevor-scheer:
Effectively duplicated and adapted work from #2644

This PR introduces the new `@requiresScopes` directive for composition

With this change, users can now compose `@requiresScopes` applications from their subgraphs into a supergraph. This addition will support a future version of Apollo Router that enables authenticated access to specific types and fields via directive applications.

Since the implementation of `@requiresScopes` is strictly a composition and execution concern, there's no change to the query planner with this work. The execution work is well under way in Apollo Router (and won't be built at all for Gateway). So as far as this repo is concerned, only composition is concerned with `@requiresScopes`.

TODO:
- [x] changelog entry (pending some internal feedback)
- [x] possibly some more interesting test cases
- [x] pending discussion on use of custom `Scope` scalar for input type


composition-js/src/__tests__/compose.test.ts
```diff
@@ -4219,4 +4219,272 @@ describe('composition', () => {
...
+      const result1 = composeAsFed2Subgraphs([a1, a2]);
+      const result2 = composeAsFed2Subgraphs([a2, a1]);
+      assertCompositionSuccess(result1);
+      assertCompositionSuccess(result2);
+
+      expect(result1.schema.type('A')?.hasAppliedDirective('requiresScopes')).toBeTruthy();
+      expect(result2.schema.type('A')?.hasAppliedDirective('requiresScopes')).toBeTruthy();
+    });
+
+    it('merges @requiresScopes lists', () => {
```
pcmanus, line 4391: For the sake of completeness, might we worth having a test that merge scopes that intersects (say `["a", "b"]` on the one side and `["b", "c"]` on the other).

gateway-js/src/__tests__/gateway/testUtils.ts
```diff
@@ -66,8 +66,16 @@ export async function startSubgraphsAndGateway(
     serviceList.push({ name: serviceDef.name, url });
   }
 
+  const mockLogger = {
```
pcmanus, line 69: Out of curiosity, is that to avoid the warnings about `serviceList` being deprecated during tests?

Not strongly opposed to this, but I am a little uneasy to blanket silence all warn and error logs. Basically, I'd worry about us making a stupid mistake that results in spamming some warning/error and it making it to a release. I'd agree the status quo, logging stuff during tests, is not perfect, but it at least give us a better chance to catch such error.

pcmanus (commented):
Had a look, and currently lgtm. Holding off on approving until this get out of "draft" and we've dealt with the few remaining TODO, but so far so good.

gateway-js/src/__tests__/gateway/testUtils.ts
trevor-scheer: Removed. I actually don't remember why this ended up here, possibly stray mid-development stuff that I left. Definitely agree we don't want to silence completely.

We could eventually adopt this and create expectations on the logger which would allow us to both quiet the logs and be specific with our log expectations but I don't think it belongs in this PR.

composition-js/src/__tests__/compose.test.ts
trevor-scheer: Good call, added

internals-js/src/requiresScopesSpec.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  get defaultCorePurpose(): CorePurpose {
+    return 'SECURITY';
+  }
+}
+
+export const REQUIRES_SCOPES_VERSIONS =
+  new FeatureDefinitions<RequiresScopesSpecDefinition>(
+    RequiresScopesSpecDefinition.identity
+  ).add(new RequiresScopesSpecDefinition(new FeatureVersion(1, 0)));
```
clenfest, line 66: Same comment here as in the `@authenticated` PR. You should specify the federation version that you're tying this to.

clenfest:
I may have missed a conversation on this, but did we want to have two separate specs or a single spec for this?

clenfest:
Oh, one thing that I remembered. Could you add both directives into the `directivesComposedByDefault` array on line 172 of `composeDirectiveManager.ts`? It will prevent people from trying to manually promote those directives to the supergraph.

composition-js/src/__tests__/compose.test.ts
```diff
@@ -4228,4 +4228,313 @@ describe('composition', () => {
...
+      const result = composeAsFed2Subgraphs([a1, a2], { supportedFeatures });
+      assertCompositionSuccess(result);
+      expect(
+        result.schema.type('A')
+          ?.appliedDirectivesOf('requiresScopes')
+          ?.[0]?.arguments()?.scopes).toStrictEqual(['a', 'b']
+      );
+    });
+
+    it('merges @requiresScopes lists (intersection)', () => {
```
clenfest, line 4441: change the name of this test since we don't want it to be the intersection between the lists.

clenfest:
looks good

internals-js/src/directiveAndTypeSpecification.ts
```diff
@@ -86,26 +87,28 @@ export function createDirectiveSpecification({
...
         // we have a schema but well, not a huge deal either).
         for (const { name: argName, type } of args) {
           const strategy = argStrategies.get(argName);
           // Note that we've built `argStrategies` from the declared args and checked that all argument had a strategy, so it would be
           // a bug in the code if we didn't get a strategy (not an issue in the directive declaration).
           assert(strategy, () => `Should have a strategy for ${argName}`);
-          const argType = type(schema);
+          console.log(`Strategy=${strategy}`);
+          const argType = type(schema, feature);
+          console.log(`argType=${argType}`);
```
trevor-scheer, line 104:
```suggestion
          const argType = type(schema, feature);
```

internals-js/src/federationSpec.ts
```diff
@@ -140,17 +142,12 @@ export class FederationSpecDefinition extends FeatureDefinition {
         name: FederationDirectiveName.INTERFACE_OBJECT,
         locations: [DirectiveLocation.OBJECT],
       }));
-      this.registerDirective(
-        TAG_VERSIONS.find(new FeatureVersion(0, 3))!.tagDirectiveSpec
-      );
+      this.registerSubFeature(TAG_VERSIONS.find(new FeatureVersion(0, 3))!);
```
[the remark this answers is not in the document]
trevor-scheer: I don't disagree with the ergonomic benefit, but there's a bit to untangle underneath that I'd rather punt on.

trevor-scheer:
@clenfest approved this in Slack and I'd like to get this on `next` today so we can kick off a harness run.

PR #2651, opened 2023-07-01
executeFetch: fix entities loop condition

erankor:
the loop iterates on receivedEntities, not entities. normally the two arrays will have the same length, however if the condition `if (representation &&
representation[TypeNameMetaFieldDef.name])` is skipped, receivedEntities will contain fewer elements than entities.

Hi, something small that I noticed while looking at the code...


jeffjakub:
Thanks for raising this issue! Could you provide a test so we can reproduce this problem?

PR #2654, opened 2023-07-04
Fix not using `@join__type` for fields with no `@join__field`

pcmanus:
When a field is defined in all the subgraphs in which its type is defined (typically the case for value types) and the field has nothing specific (no `@requires`, `@provides`, `@override`, ...), then we don't mark it in the supergraph with a `@join__field`. It's not really needed, and this keeps supergraphs smaller.

But the code extracting subgraphs from supergraph was not handling that case gracefully, and it was instead trying to add such fields to all the subgraphs. This didn't break anything because when a subgraph didn't had the type, the code was still giving up on adding the field, but for supergraph with a fair number of subgraphs, this led to a lot of busy work, making the extraction code slower than it needed to be.

This commit fix that issue.

Note that a reason this probably occured is that the extraction is trying to handle all versions of supergraphs with the same code, but fed1 supergraphs and fed2 supergraphs have actually a fair number of differences, and trying to handle both with the same code is, in hindsight, a bit confusing and we should clean it up. Leaving this as a follow-up however.


dariuszkuc (approved):

internals-js/src/extractSubgraphsFromSupergraph.ts
```diff
@@ -328,15 +328,46 @@ export function extractSubgraphsFromSupergraph(supergraph: Schema): Subgraphs {
           }
           // Fall-through on purpose.
         case 'InputObjectType':
+          const typeApplications = type.appliedDirectivesOf(typeDirective);
+          // Note that we can have more that one `@join__type` for a given graph if there is multiple keys, so we collect
+          // the actual set of subgraphs defining the type.
+          const subgraphsDefiningType: Subgraph[] = [];
+          for (const app of typeApplications) {
+            const subgraph = subgraphs.get(graphEnumNameToSubgraphName.get(app.arguments().graph)!)!;
```
clenfest, line 336: I know that these type assertions should be theoretically fine, but I do worry slightly about hand edited supergraphs. Maybe convert these to use `assert`'s?

pcmanus: I'll change that in #2655 as it centralises a bit the places where this is done, and would rather get that one in quickly.

PR #2655, opened 2023-07-05
Refactor code to extract subgraphs from supergraphs

pcmanus:
The subgraph extraction code is currently written as a single method that handles all join spec versions. There is however a fair number of differences betwee the version 0.1 of the join spec (the one generated by federation 1) and its successor (generated by federation 2), due to 1) having no ownership in fed2 and 2) the join spec 0.2+ versions tracking types much more systematically.

So this commit actually separate the code for extracting each versions, or more precisely, has one method for join 0.1 and one for join 0.2+ (the difference between join 0.2 and 0.3 are a lot more minor). Doing may increase the total LOC technically, but it also makes each version a lot clearer. This also allows to refactor the 0.2+ version a bit to use the fact that those version are more systematic in the metadata they provide, and do things a bit more efficiently.

### DIFF
--- a/internals-js/src/extractSubgraphsFromSupergraph.ts
+++ b/internals-js/src/extractSubgraphsFromSupergraph.ts
@@ -20,6 +20,8 @@ import {
   ObjectType,
   Schema,
   Type,
+  EnumType,
+  UnionType,
 } from "./definitions";
 import {
   newEmptyFederation2Schema,
@@ -27,7 +29,7 @@ import {
   removeInactiveProvidesAndRequires,
 } from "./federation";
 import { CoreSpecDefinition, FeatureVersion } from "./coreSpec";
-import { JoinSpecDefinition } from "./joinSpec";
+import { JoinFieldDirectiveArguments, JoinSpecDefinition, JoinTypeDirectiveArguments } from "./joinSpec";
 import { FederationMetadata, Subgraph, Subgraphs } from "./federation";
 import { assert } from "./utils";
 import { validateSupergraph } from "./supergraphs";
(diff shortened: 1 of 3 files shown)


pcmanus:
Note that this PR is currently on top of #2654 as both change the same places and it makes sense to get #2654 first anyway. I'll rebase once #2654 is committed, but for now, only the 2nd commit is new to this PR.

internals-js/src/extractSubgraphsFromSupergraph.ts
```diff
@@ -625,51 +577,360 @@ function errorToString(e: any,): string {
...
+  // Note that we first add all types empty and populate the types next. This avoids having to care about the iteration
+  // order if we have fields than depends on other types.
+  for (const type of filteredTypes) {
+    const typeApplications = type.appliedDirectivesOf(typeDirective);
+    if (!typeApplications.length) {
+      // Imply we don't know in which subgraph the type is, so we had it in all subgraph in which the type is reachable.
+      subgraphs
+        .values()
+        .filter((sg) => includeTypeInSubgraph(type, sg.name))
+        .map(sg => sg.schema).forEach(schema => schema.addType(newNamedType(type.kind, type.name)));
```
trevor-scheer, line 712: I know this is previously existing code but this filter/map is pretty trivial and adds 100s of iterations (in the case we're hearing about lately). Fine to leave as is but this looks like an easy win.
```suggestion
        .filter((sg) => includeTypeInSubgraph(type, sg.name))
        .map(sg => sg.schema)
        
        .forEach(sg => {
          if (!includeTypeInSubgraph(type, sg.name)) return;
          sg.schema.addType(newNamedType(type.kind, type.name));
        });
```

internals-js/src/extractSubgraphsFromSupergraph.ts
```diff
@@ -625,51 +577,360 @@ function errorToString(e: any,): string {
...
+  }
+
+  // We can now populate all those types (with relevant @provides and @requires on fields).
+  for (const type of filteredTypes) {
+    switch (type.kind) {
+      case 'ObjectType':
+        // @ts-expect-error: we fall-through the inputObjectType for fields.
+      case 'InterfaceType':
+        for (const implementations of type.interfaceImplementations()) {
+          // There is no `@join__implements` in fed1 supergraphs, so we have to choice to but mark the
```
trevor-scheer, line 746:
```suggestion
          // There is no `@join__implements` in fed1 supergraphs, so we have no choice but to mark the
```

internals-js/src/extractSubgraphsFromSupergraph.ts
```diff
@@ -625,51 +577,360 @@ function errorToString(e: any,): string {
...
   }
+
+  const allExecutableDirectives = supergraph.directives().filter((def) => def.hasExecutableLocations());
+  for (const subgraph of subgraphs) {
+    // The join spec in fed1 was not including external fields. Let's make sure we had them or we'll get validation
+    // errors later.
+    addExternalFields(subgraph, supergraph, true);
+    removeInactiveProvidesAndRequires(subgraph.schema);
+
+    // We now do an additional path on all types because we sometimes added types to subgraphs without
```
trevor-scheer, line 841:
```suggestion
    // We now do an additional pass on all types because we sometimes added types to subgraphs without
```

internals-js/src/extractSubgraphsFromSupergraph.ts
```diff
@@ -625,51 +577,360 @@ function errorToString(e: any,): string {
...
+        if (!implementations.every(implem => implem.field(field.name))) {
+          field.remove();
+        }
+      }
+      // And it may be that the interface wasn't part of the subgraph at all!
+      if (!itf.hasFields()) {
+        itf.remove();
+      }
+    }
+  }
```
trevor-scheer, line 905: Maybe doesn't have to be this PR but should we remove this as suggested by the comment? Looks like tests pass without it but that's the extent of my understanding of its importance. Another potentially hundreds/thousands of iterations.

trevor-scheer (approved):
Just minor / non-blocking stuff. LGTM!

internals-js/src/extractSubgraphsFromSupergraph.ts
pcmanus: I'm ok changing that. Made it a good'ol for-loop while at it though as it's imo more readable at that point.

I'll note though that this is code that only run for fed1 supergraphs, so only for use that never ever composed with the fed2 composition code, and so not necessary where performance matters much.

internals-js/src/extractSubgraphsFromSupergraph.ts
pcmanus: Good call. When I noticed this, I was doing something unrelated and didn't wanted to risk breaking something, hence the TODO instead of removing it right away. But this patch is a refactor, so it's as good a place as any to make the removal. And double checked that this code indeed is not needed anymore (as the TODO says, the code that adds missing external fields that is called before that will ensure there is no implementation missing a field needed for an interface).

But I'll also point out that I rebased this/will merge this against `next` because this is a refactor, not a bug fix, so really belong in `next` (and the other PR that rely on this also really belong in `next` anyway). Which is all the more reason not to be shy about removing this.

PR #2657, opened 2023-07-05
Skip some supergraph validations in production by default

pcmanus:
When receiving a supergraph, we were always running all the graphQL validations on both the supergraph itslef, and on each of the subgraphs we extract from it.

This is useful for us when doing dev, as it allows to get decent error message when we break the supergraph generation process or the subgraph extraction process, but this is of very limited value otherwise: even if we don't run those full validations, the supergraph still has to be valid graphQL syntax to parse in the first place, and the subgraph extraction will still throw if the supergraph is not essentially a valid supergraph (and note that the code checking that the `join` spec version is valid, or that the supergraph does not use non-supported feature for `SECURITY` or `EXECUTION` is *not* disabled by this).

Besides, it's not like uplink will send basically-valid-but-slightly-corrupted supergraphs, and when using `IntrospectAndCompose` there isn't really room for the supergraph to be corrupted.

And those validations are actually not super cheap, especially on large supergraphs, so running them feels wasteful.

This commit thus disable those validation in production environment (when `NODE_ENV === "production"`) by default. It keeps those validation during dev mainly for 2 reasons:
1. as mentioned above, so that those validate are run when we run our own unit tests, to help catch/identify any subtle mistake we would introduce in the supergraph generation and extaction codes.
2. when debugging something, it can happens that one start a gateway from a statically provided supergraph SDL, SDL that one may have manually edited for some reason (testing something, playing with it, ...). In that case, the validation can provide nicer errors if the manual edits are invalid.

The commit also introduce a `validateReceivedSupergraphs` in the gateway config to allows overriding the default behaviour described above.


.changeset/dull-lamps-invent.md
```diff
@@ -0,0 +1,17 @@
+---
+"@apollo/query-planner": patch
+"@apollo/query-graphs": patch
+"@apollo/composition": patch
+"@apollo/federation-internals": patch
+"@apollo/gateway": patch
```
trevor-scheer, line 6: Thinking this should be `minor` since we're treating this as a new feature and landing to `next`

```suggestion
"@apollo/query-planner": minor
"@apollo/query-graphs": minor
"@apollo/composition": minor
"@apollo/federation-internals": minor
"@apollo/gateway": minor
```

gateway-js/src/config.ts
```diff
@@ -129,6 +129,24 @@ interface GatewayConfigBase {
...
+   * be a valid supergraph for the gateway to be able to use it, and so this option is not
+   * necessary to protected against corrupted/invalid supergraphs, but it may produce more legible
+   * errors when facing such invalid supergraph.
+   *
+   * By default, this depends on the value of `NODE_ENV`: for production, validation is disabled
+   * (as it is somewhat expensive and not that valuable as mentioned above), but it is enabled
+   * for development (mostly to provide better error messages when provided with an incorrect
+   * supergraph).
+   */
+  validateReceivedSupergraphs?: boolean;
```
trevor-scheer, line 149: Small naming nitpick, not a big deal though. `validateSupergraph(s)` maybe? "received" seems extraneous. I can see the argument to keep it plural but it doesn't seem necessary, fine either way.

trevor-scheer (approved):
Changes lgtm. Maybe we should add some tests around the `assumeValid` behavior? I think it would be fine to follow up with those later, just to preserve the behavior we expect going forward.

internals-js/src/supergraphs.ts
```diff
@@ -118,7 +123,7 @@ export class Supergraph {
       // Note that `extractSubgraphsFromSupergraph` redo a little bit of work we're already one, like validating
       // the supergraph. We could refactor things to avoid it, but it's completely negligible in practice so we
       // can leave that to "some day, maybe".
-      this._subgraphs = extractSubgraphsFromSupergraph(this.schema);
+      this._subgraphs = extractSubgraphsFromSupergraph(this.schema, !this.assumeValid);
```
clenfest, line 126: I don't love that we start with `validateReceivedSupergraphs` then go to `assumeValid` then back to `validateExtractedSubgraphs`. It would be nice to not keep negating the poor boolean.

gateway-js/src/index.ts
```diff
@@ -634,8 +634,8 @@ export class ApolloGateway implements GatewayInterface {
   }
 
   private createSchemaFromSupergraphSdl(supergraphSdl: string) {
-    const assumeValid = !(this.config.validateReceivedSupergraphs ?? process.env.NODE_ENV !== 'production');
-    const supergraph = Supergraph.build(supergraphSdl, { assumeValid });
+    const validateSupergraph = this.config.validateSupergraph ?? process.env.NODE_ENV !== 'production';
```
clenfest, line 637: I think this is wrong. If they explicitly set `validateSupergraph` to false, we should probably honor that, no?

trevor-scheer: `false ?? true` equates to false, unlike `false || true`, so I think this is doing what we want.

clenfest: Yeah, of course. Clearly been doing too much rust lately. ship it.

PR #2658, opened 2023-07-06
feat: Add the schema coordinate to all CompositionHints in composition-js

swcollard:
## Overview

As part of pulling Composition Hints and all build messages into Apollo Studio and GraphOS we want to align on specifying errors, warnings, and lint errors around the [schema coordinate](https://github.com/graphql/graphql-wg/blob/main/rfcs/SchemaCoordinates.md). The easiest way to start aligning all those would be to include the schema coordinate in the Composition Hint. This PR adds the optional String  to the CompositionHint class and implements setting the coordinate in all instances where CompositionHint objects are constructed.

## Changes

* Expand the CompositionHint class to include an optional String `coordinate`
* In every instance of creating a CompositionHint object, include the coordinate or the string that seems closest to the coordinate?

## Testing

Expanded existing unit tests to also verify the coordinate is correct


composition-js/src/composeDirectiveManager.ts
```diff
@@ -174,6 +175,7 @@ export class ComposeDirectiveManager {
       this.pushHint(new CompositionHint(
         HINTS.DIRECTIVE_COMPOSITION_INFO,
         `Directive "@${directive.name}" should not be explicitly manually composed since it is a federation directive composed by default`,
+        directive.name,
```
jsegaran, line 178: nit: Directive schema coordinates should start with `@`

swcollard: Good catch. Updated in latency revision

composition-js/src/composeDirectiveManager.ts
```diff
@@ -143,6 +143,7 @@ export class ComposeDirectiveManager {
           this.pushHint(new CompositionHint(
             HINTS.DIRECTIVE_COMPOSITION_INFO,
             `Non-composed core feature "${coreIdentity}" has major version mismatch across subgraphs`,
+            coreIdentity,
```
trevor-scheer, line 146: It looks like our testing story is a bit inconsistent, which is not your fault.

i.e. if I change this to, say, `undefined`, tests still pass. But if I remove the `this.pushHint(...)` there is a test that fails (which doesn't use the `toRaiseHint` matcher.

I'd like it if we addressed these inconsistencies so there's coverage for all of your additions. Let me know if you'd like me to help with that, since I think that's an "us" problem.

trevor-scheer: Note: this is the only one I checked, so for now I'm not checking the others or leaving comments on them.

trevor-scheer: Pretty sure if we get the testing story consistent it'll reveal the directive name errors that I pointed out above (or reveal that we're just lacking coverage on those particular hints altogether).

trevor-scheer:
Please add a `minor` changeset by running `npx changeset` and following the prompts

composition-js/src/composeDirectiveManager.ts
```diff
@@ -174,6 +175,7 @@ export class ComposeDirectiveManager {
       this.pushHint(new CompositionHint(
         HINTS.DIRECTIVE_COMPOSITION_INFO,
         `Directive "@${directive.name}" should not be explicitly manually composed since it is a federation directive composed by default`,
+        `@${directive.name}`,
```
pcmanus, line 178: I think we should avoid manually crafting any coordinate when the [`coordinate`](https://github.com/apollographql/federation/blob/main/internals-js/src/definitions.ts#L671) property exists.

In fact, see my suggestion on the ctor of `CompositionHint` which would pretty much guarantee this by design.

composition-js/src/hints.ts
```diff
@@ -235,6 +235,7 @@ export class CompositionHint {
   constructor(
     readonly definition: HintCodeDefinition,
     readonly message: string,
+    readonly coordinate: string | undefined,
```
pcmanus, line 238: I would suggest changing this argument to be the (main) element that is the target of the hint, so something like:
```typescript
export class CompositionHint {
  public readonly nodes?: readonly SubgraphASTNode[];
  public readonly coordinate?: string;
  
  constructor(
    readonly definition: HintCodeDefinition,
    readonly message: string,
    element: NamedSchemaElement | undefined,
  ) {
    this.coordinate = element?.coordinate;
    ....
  }
  ...
}
```
Imo this has a bunch of advantages:
- this is more precisely typed. `string` is not the most helpful type: it's really easy to pass something that isn't a valid coordinate and have this go undetected. Passing `NamedSchemaElement` makes it less error prone and guaranteed we will get a valid coordinate at the end of the day.
- practically, this avoid having to repeat `.coodinate` on almost all callsite, which feel kinda nice.

composition-js/src/merging/reporter.ts
```diff
@@ -112,6 +113,7 @@ export class MismatchReporter {
     message: string,
     supergraphElement: TMismatched,
     subgraphElements: (TMismatched | undefined)[],
+    coordinate: string | undefined,
```
pcmanus, line 116: Same remarks as in `CompositionHint`: we should imo pass `element: NamedSchemaElement | undefined`.

In fact, it's worth double checking, but I suspect we can avoid the need of that argument, at least in most case, as `supergraphElement` is usually what we want here (when it is a `NamedSchemaElement` at least; and when it isn't, we want `undefined`). From a quick look, it seems that there is 2 special cases where that's not quite true, for hints on enum values and union members where `supergraphElement` is the type itself, but we ideally would want to pass the value/member itself. But I'd personally suggest adding an optional argument (say, `targetedElement?: NamedSchemaElement` and only pass it when it's not equal to `supergraphElement` (and default to the later when `targetedElement` is not provided).

pcmanus (changes requested):

composition-js/src/hints.ts
swcollard: Thanks for the review! This makes sense to me, I will make this change and post a new revision

composition-js/src/merging/reporter.ts
swcollard: You were correct, I noticed that while going through and fixing unit tests. I have added the `targetedElement?: NamedSchemaElement` to this function and filled in it for the Enum value cases.

PR #2678, opened 2023-07-17
remove unused types from both fed1 and fed2 extractSubgraphs

clenfest:
This can potentially happen because use of the `@override` directive can result in a type being empty.

### DIFF
--- a/internals-js/src/__tests__/extractSubgraphsFromSupergraph.test.ts
+++ b/internals-js/src/__tests__/extractSubgraphsFromSupergraph.test.ts
@@ -738,3 +738,80 @@ test('throw meaningful error for type erased from supergraph due to extending an
     + 'Error: Cannot find type "T" in subgraph "serviceB"'
   );
 })
+
+test('types that are empty because of overridden fields are erased', () => {
+  const supergraph = `
+    schema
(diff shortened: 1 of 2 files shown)


trevor-scheer (changes requested):
Pretty sure the answer is no since this is fixing a new issue - but no changeset needed, right?

Can we add a test for this? Otherwise lgtm.

internals-js/src/__tests__/extractSubgraphsFromSupergraph.test.ts
```diff
@@ -738,3 +738,80 @@ test('throw meaningful error for type erased from supergraph due to extending an
...
+
+      baz: String @join__field(graph: A)
+    }
+  `;
+
+  const subgraphs = Supergraph.build(supergraph).subgraphs();
+
+  const subgraph = subgraphs.get('b');
+  const inputType = subgraph?.schema.type('User') as InputObjectType | undefined;
+  expect(inputType).toBeUndefined();
```
trevor-scheer, line 816: Inconsequential really but this should be an `ObjectType` and renamed. I'd suggest but it requires an import as well.

trevor-scheer (approved):

Issue #2673, opened 2023-07-14
@authenticated name collision


jeffjakub:
### Issue Description

Please review test results for name collision error.

### Link to Reproduction

n/a

dariuszkuc:
This is not a new issue but it looks like it surfaced because `@authenticated` is more common directive name than other fed v2 directive names.

Error is due to our Fed v1 -> Fed v2 subgraph schema upgrade logic. When we upgrade the schema we add default (latest) federation link definition that imports all fed v2 directives with default names (i.e. no namespacing). We then attempt to add all fed directive definitions verifying that if they are already defined in the schema, their definitions should match ([link](https://github.com/apollographql/federation/blob/main/internals-js/src/directiveAndTypeSpecification.ts#L155).

Simple test to repro (it fails with the same error for other fed v2 directives, e.g. `@tag`)

```js
it('existing @authenticated directive with fed 1', () => {
  const subgraphA = {
    typeDefs: gql`
      directive @authenticated(scope: [String!]) repeatable on FIELD_DEFINITION

      extend type Foo @key(fields: "id") {
        id: ID!
        protected: String @authenticated(scope: ["foo"])
      }
    `,
    name: 'subgraphA',
  };

  const subgraphB = {
    typeDefs: gql`
      type Query {
        foo: Foo
      }

      type Foo @key(fields: "id") {
        id: ID!
        name: String!
      }
      `,
    name: 'subgraphB',
  };

  debugger;
  const result = composeServices([subgraphA, subgraphB]);
  const schema = expectNoErrors(result); // blows up with error below
});
```

Error:
```
Array [
  Array [
    "DIRECTIVE_DEFINITION_INVALID",
      "[subgraphA] Invalid definition for directive \"@authenticated\": unknown/unsupported argument \"scope\"",
    ],
  Array [
    "DIRECTIVE_DEFINITION_INVALID",
      "[subgraphA] Invalid definition for directive \"@authenticated\": \"@authenticated\" should not be repeatable",
    ],
]
```

dariuszkuc:
One potential fix is to change the upgrade logic to only include fed v1 directive definitions imports (as it was require to omit fed v1 directive definitions) so all the fed v2 specific directive are namespaced, i.e.

instead of
```
[...]

PR #2679, opened 2023-07-17
fix: directive name collisions

dariuszkuc:
We are currently auto-expanding Fed v1 subgraphs with ALL Fed v2 definitions which causes an issue if users already defined directive with the same name. For convenience we also rely on this behavior in our tests... Since newly introduced `@authenticated` directive is already used by some customers in their v1 schemas, composition fails in those scenarios. In order to preserve backwards compatibility, we will continue auto-expanding directive definitions up to Fed v2.4.

Fixes: https://github.com/apollographql/federation/issues/2673

### DIFF
--- a/composition-js/src/__tests__/compose.composeDirective.test.ts
+++ b/composition-js/src/__tests__/compose.composeDirective.test.ts
@@ -998,4 +998,37 @@ describe('composing custom core directives', () => {
     const appliedDirectives = schema.elementByCoordinate('Query.shared')?.appliedDirectives;
     expect(appliedDirectives?.map(d => [d.name, d.arguments()])).toMatchObject([['auth', { scope: 'VIEWER'}], ['auth', {}]]);
   });
+
+  it('existing @authenticated directive with fed 1', () => {
+    const subgraphA = {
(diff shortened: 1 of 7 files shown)


trevor-scheer (approved):

composition-js/src/__tests__/compose.composeDirective.test.ts
```diff
@@ -998,4 +998,37 @@ describe('composing custom core directives', () => {
     const appliedDirectives = schema.elementByCoordinate('Query.shared')?.appliedDirectives;
     expect(appliedDirectives?.map(d => [d.name, d.arguments()])).toMatchObject([['auth', { scope: 'VIEWER'}], ['auth', {}]]);
   });
+
+  it('existing @authenticated directive with fed 1', () => {
```
clenfest, line 1002: Very minor, but this file is for tests that use `@composeDirective`. I'd put it into `compose.test.ts` instead.

dariuszkuc: my bad i was using this for testing and left it in a wrong spot 👍 will move it

PR #2700, opened 2023-07-28
Gateway OpenTelemetry improvements

unflxw:
This PR adds the GraphQL document as a span attribute on the gateway's request span, and makes it so that errors are recorded as exception events in the spans in which they occur.

### [Record errors as OTel exception events](https://github.com/apollographql/federation/commit/d8a36a8c0e4a45553435e62a68d4d139c53cfa72)

When an error is received or thrown, in addition to setting the
span status to error, report the error as an OpenTelemetry
exception event.

If more than one error is received in a GraphQL error array, report
only the first one.

### [Report the GraphQL document as a span attribute](https://github.com/apollographql/federation/commit/f92f200ddd48ced4b6250ced0a8a8faa86d6c569)

On the request OpenTelemetry span, add the GraphQL document as an
OpenTelemetry span attribute, using the name for this
property recommended by the OpenTelemetry Semantic Conventions.

## Notes

- I made it so that only the first error in a GraphQL error array is added, erring on the side of caution, as I'm unsure just how many errors could be reported. If you'd rather it reports all errors in the array, let me know.
- The exception events are not tested through the snapshots, but separately, checking only for their presence in the desired spans. This is convenient, as the details of the error (backtrace, timing) are not interesting to check against.
- I could not figure out how to add the sub-query that is sent on fetch operations as a span attribute on the fetch spans. I believe this would also be useful. I am quite unfamiliar with the codebase, any pointers would be appreciated.
- I did not rename the existing `operationName` attribute to the recommended `graphql.operation.name` in the OpenTelemetry Semantic Conventions. Let me know if you want me to do that.

### DIFF
--- a/gateway-js/src/executeQueryPlan.ts
+++ b/gateway-js/src/executeQueryPlan.ts
@@ -239,7 +239,7 @@ export async function executeQueryPlan(
           // Note that this behavior is also what the router does (and in fact, the exact name of the `extensions` we use,
           // "valueCompletion", comes from the router and we use it for alignment.
           if (errors.length === 0 && postProcessingErrors.length > 0) {
-            span.recordException(postProcessingErrors[0]);
+            postProcessingErrors.forEach(error => span.recordException(error));
             span.setStatus({ code:SpanStatusCode.ERROR });
             return { extensions: { "valueCompletion":  postProcessingErrors }, data };
           }
@@ -282,7 +282,7 @@ export async function executeQueryPlan(
       });
 
       if(result.errors) {
-        span.recordException(result.errors[0]);
+        result.errors.forEach(error => span.recordException(error));
         span.setStatus({ code:SpanStatusCode.ERROR });
       }
       return result;
--- a/gateway-js/src/executeQueryPlan.ts
+++ b/gateway-js/src/executeQueryPlan.ts
@@ -239,10 +239,12 @@ export async function executeQueryPlan(
           // Note that this behavior is also what the router does (and in fact, the exact name of the `extensions` we use,
           // "valueCompletion", comes from the router and we use it for alignment.
           if (errors.length === 0 && postProcessingErrors.length > 0) {
+            span.recordException(postProcessingErrors[0]);
             span.setStatus({ code:SpanStatusCode.ERROR });
             return { extensions: { "valueCompletion":  postProcessingErrors }, data };
           }
         } catch (error) {
+          span.recordException(error);
           span.setStatus({ code:SpanStatusCode.ERROR });
           if (error instanceof GraphQLError) {
             return { errors: [error] };
@@ -280,11 +282,13 @@ export async function executeQueryPlan(
       });
 
       if(result.errors) {
+        span.recordException(result.errors[0]);
         span.setStatus({ code:SpanStatusCode.ERROR });
       }
       return result;
     }
     catch (err) {
+      span.recordException(err)
       span.setStatus({ code:SpanStatusCode.ERROR });
       throw err;
     }
@@ -517,6 +521,7 @@ async function executeFetch(
       }
     }
     catch (err) {
+      span.recordException(err);
       span.setStatus({ code:SpanStatusCode.ERROR });
       throw err;
     }
--- a/gateway-js/src/executeQueryPlan.ts
+++ b/gateway-js/src/executeQueryPlan.ts
@@ -31,7 +31,7 @@ import {
 import { deepMerge } from './utilities/deepMerge';
 import { isNotNullOrUndefined } from './utilities/array';
 import { SpanStatusCode } from "@opentelemetry/api";
-import { OpenTelemetrySpanNames, tracer } from "./utilities/opentelemetry";
+import { OpenTelemetryConfig, OpenTelemetrySpanNames, recordExceptions, tracer } from "./utilities/opentelemetry";
 import { assert, defaultRootName, errorCodeDef, ERRORS, Operation, operationFromDocument, Schema } from '@apollo/federation-internals';
 import { GatewayGraphQLRequestContext, GatewayExecutionResult } from '@apollo/server-gateway-interface';
 import { computeResponse } from './resultShaping';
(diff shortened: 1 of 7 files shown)


gateway-js/src/executeQueryPlan.ts
```diff
@@ -239,10 +239,12 @@ export async function executeQueryPlan(
           // Note that this behavior is also what the router does (and in fact, the exact name of the `extensions` we use,
           // "valueCompletion", comes from the router and we use it for alignment.
           if (errors.length === 0 && postProcessingErrors.length > 0) {
+            span.recordException(postProcessingErrors[0]);
```
trevor-scheer, line 242: It's not obvious to me in their docs whether calling `recordException` multiple times works, but I do think it would be nice to capture all of them rather than just the first if that's an option.

gateway-js/src/utilities/opentelemetry.ts
```diff
@@ -11,3 +12,16 @@ export enum OpenTelemetrySpanNames {
...
+
+  if (requestContext.operationName) {
+    spanAttributes["operationName"] = requestContext.operationName;
+  }
+  if (requestContext.source) {
+    spanAttributes["graphql.document"] = requestContext.source;
+  }
+
+  return spanAttributes;
+}
```
trevor-scheer, line 27: How about something like this? This incorporates your suggestion to add the semantic naming for a graphql operation, which I think is a good suggestion. We can just keep `operationName` around as deprecated and duplicate it for now.

```suggestion
import type { Attributes } from '@opentelemetry/api';

interface SpanAttributes extends Attributes {
  /**
   * @deprecated in favor of `graphql.operation.name`
   */
  operationName: string | undefined;
  'graphql.operation.name': string | undefined;
  'graphql.document': string;
}

export function requestContextSpanAttributes(
  requestContext: GatewayGraphQLRequestContext,
): SpanAttributes {
  return {
    operationName: requestContext.operationName ?? undefined,
    'graphql.operation.name': requestContext.operationName ?? undefined,
    'graphql.document': requestContext.source,
  };
}
```

trevor-scheer (changes requested):
Can you add a changeset by running `npx changeset`? I'd consider this a minor. Please be descriptive in the changeset you write, it becomes the changelog entry. It should tell users what change they'll observe, why it might be useful, etc.

Looks pretty good, thanks for the addition! Just a couple comments.

trevor-scheer:
@unflxw FYI I've changed the base branch to `next` in case you decide to rebase or anything locally.

gateway-js/src/executeQueryPlan.ts
unflxw: It is possible to call `recordException` multiple times to report multiple exception events. I will change it to do that then. 👍

gateway-js/src/utilities/opentelemetry.ts
unflxw: I like this! I'll change it to use this intermediate span attributes type. 👍

The OpenTelemetry span attribute types allow `undefined` as a value, but [the `Attributes` and `AttributeValue` types are also documented](https://github.com/open-telemetry/opentelemetry-js/blob/2b20565d105d4267de5e19a730c2401eacad43a4/api/src/common/Attributes.ts) as follows:

> [...] the own enumerable keys are counted as valid attribute keys [...]
> [...] null or undefined attribute values are invalid and will result in undefined behavior.

So I'm thinking that this implementation should still avoid setting the key at all if its value will be `undefined`, like in the previous implementation, and not find out what this "undefined behaviour" is.

unflxw:
Thank you for reviewing this @trevor-scheer 🙌 will address the feedback (and rebase on `next`) shortly.

I also tried adding the operation type (`graphql.operation.type`) but I could not get it to work. It seemed to me that `requestContext.operation?.operation` should be it, but when running the tests it came up empty. 🤷

Any thoughts on how it could be obtained? This would be nice to have, as the combined operation type and name make for a good human-readable identifier of the query, with low cardinality, for the purpose of grouping traces.

unflxw:
Done as described @trevor-scheer, please re-review 👍 I also figured out how to add the `graphql.operation.type` attribute from a different source.

trevor-scheer (changes requested):
@unflxw changes look good!

@dariuszkuc raised a concern to me that capturing the documents could really blow up the size of these spans for some. I think it's best if we make the new behavior configurable and opt-in. I figure we might as well do the same for the new exceptions capturing while we're in there.

Thoughts? Would you mind implementing some configurability into this? I'm open to the naming here, wouldn't mind @dariuszkuc's input as well.
```ts
new ApolloGateway({
  // ...
  telemetry: {
    includeDocument: true, // default false
    recordExceptions: true, // default false
  },
});
```

unflxw:
@trevor-scheer @dariuszkuc I'm not opposed to adding this configuration option, I will look into adding it.

That said, I think it's worth pointing out that the `@opentelemetry/instrumentation-graphql` package, which Apollo-not-Gateway is compatible with, does not have such a configuration option. Presumably that means that this isn't a problem for those servers, which in many setups would be the ones implementing the sub-graphs that Gateway federates with.

trevor-scheer:
@unflxw In principle I don't disagree with what you're saying, and that might be the ideal future state of this configuration. My primary concern is making that change in a minor version. If this change in default behavior drove up my operating costs in a non-negligible way I would be pretty frustrated and expect that to be part of a major bump that's explicitly called out. I have no idea what to expect as far as span size and cost goes, so I'd rather make no assumptions there. With what I've proposed, it's visible in the changelog and a new feature that users can opt in to.

unflxw:
Yeah, that makes sense to me. I'll implement it in that way.

trevor-scheer (approved):
Thanks for incorporating those changes @unflxw, this looks good to me. I'm looking for @dariuszkuc to take a final pass on this next week just to make sure this aligns with what he and I spoke about.

unflxw:
Sounds good! Ping me if there's anything I can help with.

dariuszkuc (approved):
LGTM.

While opt-in flag is currently the best we can do, ideally open telemetry spec should define a mechanism that would allow you to dynamically opt-in to this additional information on a per trace basis (similar to [zipkin B3 debug flag](https://github.com/openzipkin/b3-propagation#debug-flag)). As Trevor mentioned, those documents could potentially be rather large (as in hundreds of KB big). Sadly [`TraceFlags` currently only define single `sampled` flag](https://www.w3.org/TR/trace-context/#trace-flags) so there is no built-in mechanism yet :(.

trevor-scheer:
Thanks for the contribution @unflxw!

PR #2704, opened 2023-07-30
Create README.md for federation-internals

smyrick:
Add a simple doc to comment that this is an internal package and redirecting users to the more public packages

### DIFF
--- a/internals-js/README.md
+++ b/internals-js/README.md
@@ -1,0 +1,10 @@
+# @apollo/federation-internals
+
+Apollo Federation internal utilities. This package may ship breaking changes and does not hold the same semver guarentees as our other Federation pacakges.
+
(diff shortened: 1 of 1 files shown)


internals-js/README.md
```diff
@@ -0,0 +1,10 @@
+# @apollo/federation-internals
+
+Apollo Federation internal utilities. This package may ship breaking changes and does not hold the same semver guarentees as our other Federation pacakges.
```
trevor-scheer, line 3: This repo doesn't necessarily adhere to semver (though we do our best to), I don't want to make claims anywhere that that's the case until it's true.

```suggestion
This is an **internal** package for core Federation components. This package may ship breaking changes at any time and does not make the same efforts to avoid breaking changes as our other Federation packages.
```

trevor-scheer (changes requested):
Thanks for the PR!

trevor-scheer (approved):

Issue #2695, opened 2023-07-27
Query Plan selects most efficient `@key` instead of most specific `@key` for entity resolution


rhino88:
### Issue Description

In a scenario where a query plan could be supported by multiple keys, the generated query plan seems to choose the most efficient `@key` (fewer fields) vs. the most specific (more fields) when multiple `@key`s are available.

*Subgraph A*
```
extend schema
  @link(url: "https://specs.apollo.dev/federation/v2.3", import: ["@key"])

extend type Query {
  getFoo: Foo
}

type Foo {
  bar: Bar
}

type Bar @key(fields: "key1 key2 key3") {
  key1: String!
  key2: String!
  key3: String
}

```

*Subgraph B*
```
extend schema
  @link(url: "https://specs.apollo.dev/federation/v2.3", import: ["@key"])

type Bar @key(fields: "key1 key2") @key(fields: "key1 key2 key3") {
  key1: String!
  key2: String!
  key3: String
  nonKeyField: String!
}
```

*Query*
```
query FooBar {
  getFoo {
    bar {
      nonKeyField
    }
  }
}
```

*Query Plan*
```
QueryPlan {
  Sequence {
    Fetch(service: "foo_subgraph") {
      {
        getFoo {
          bar {
            __typename
            key1
            key2
          }
        }
      }
    },
    Flatten(path: "getFoo.bar") {
      Fetch(service: "bar_subgraph") {
        {
          ... on Bar {
            __typename
            key1
            key2
          }
        } =>
        {
          ... on Bar {
            nonKeyField
          }
        }
      },
    },
  },
}
```

In the above scenario, the query plan is using `@key(fields: "key1 key2")` instead of what is likely the more specific `@key(fields: "key1 key2 key3")`.

This can pose challenges for incremental evolution of a graph where new fields are added to an entity to improve specificity or otherwise evolve an entity which may require a new set of `@key` fields while maintaining backwards compatibility with the old `@key`.

**Solution Ideas**

Could the query plan select the first `@key` in Subgraph B compatible with Subgraph A? This would better support the incremental evolution of keys and deprecation of old keys.

For example, to declare preference for the 3 field key, we would change the ordering of the `@key`s in Subgraph B like so:

`@key(fields: "key1 key2") @key(fields: "key1 key2 key3")` -> `@key(fields: "key1 key2 key3") @key(fields: "key1 key2") `

### Link to Reproduction

N/A

### Reproduction Steps

N/A

[...]

PR #2714, opened 2023-08-04
[docs] add info about mismatched keys

Meschreiber:
This PR adds information about "mismatched" `@keys` across subgraphs.

Refer to https://apollograph.slack.com/archives/C03RH11Q6S1/p1685024491858819 for more info.

### DIFF
--- a/docs/source/entities-advanced.mdx
+++ b/docs/source/entities-advanced.mdx
@@ -104,30 +104,6 @@ type Product @key(fields: "upc") {
 
 </CodeColumns>
 
-If subgraphs share any fields that act as unique identifiers, you can use these shared field(s) as `@keys` for optimal [query planner](./entities#the-query-plan) performance.
-
-<p style="margin-bottom: 0">✅</p>
-
-<CodeColumns>
-
-```graphql title="Products subgraph"
-type Product @key(fields: "upc") {
-  sku: ID!
-  upc: String!
-  name: String!
-  price: Int
-}
-```
-
-```graphql title="Inventory subgraph"
-type Product @key(fields: "upc") {
-  upc: String!
-  inStock: Boolean!
-}
-```
-
-</CodeColumns>
-
 ## Migrating entities and fields
 
 As your supergraph grows, you might want to move parts of an entity to a different subgraph. This section describes how to perform these migrations safely.
--- a/docs/source/entities-advanced.mdx
+++ b/docs/source/entities-advanced.mdx
@@ -81,7 +81,7 @@ type Product @key(fields: "upc") {
 
 </CodeColumns>
 
-To merge entities between subgraphs, the entity must have at least one shared field between subgraphs. For example, you can't merge the `Product` entity defined in the following subgraphs because they don't share any fields:
+To merge entities between subgraphs, the entity must have at least one shared field between subgraphs. For example, you can't merge the `Product` entity defined in the following subgraphs because they don't share any fields specified in `@key` selection set:
 
 <p style="margin-bottom: 0">❌</p>
 
--- a/docs/source/entities-advanced.mdx
+++ b/docs/source/entities-advanced.mdx
@@ -107,7 +107,7 @@ type Product @key(fields: "upc") {
 
 #### Operations with mismatched `@key`s
 
-Mismatched keys also affect which fields can be resolved. Requests can resolve an entity's fields _if there is a traversable path from the root query to the fields_.
+Mismatched keys affect which fields from an entity can be resolved. Requests can resolve an entity's fields _if there is a traversable path from the root query to the fields_.
 
 Take these subgraph schemas as an example:
 
@@ -138,7 +138,7 @@ type Product @key(fields: "upc") {
 
 The queries defined in the products subgraph can always resolve all product fields because the product entity can be joined via the `upc` field present in both schemas.
 
-On the other hand, queries added to the inventory subgraph can't resolve fields from the products subgraph.
+On the other hand, queries added to the inventory subgraph can't resolve fields from the products subgraph:
 
 <CodeColumns>
 
--- a/docs/source/entities-advanced.mdx
+++ b/docs/source/entities-advanced.mdx
@@ -59,29 +59,96 @@ type Organization {
 
 ### Differing `@key`s across subgraphs
 
-An entity often has the same @key fields across subgraphs, but this isn't required. For example, you can define a `Product` entity shared between products and inventory subgraphs, one with `id` as the `@key` field and one with `sku` as the `@key` field:
+An entity often has the exact same `@key` field(s) across subgraphs, but this isn't required. For example, you can define a `Product` entity shared between subgraphs, one with `sku` and `upc` as the `@key`s and one with only `upc` as the `@key` field:
 
 <CodeColumns>
 
 ```graphql title="Products subgraph"
-type Product @key(fields: "id") {
-  id: ID!
+type Product @key(fields: "sku") @key(fields: "upc") {
+  sku: ID!
+  upc: String!
   name: String!
   price: Int
 }
 ```
 
 ```graphql title="Inventory subgraph"
-type Product @key(fields: "sku") {
-  sku: ID!
+type Product @key(fields: "upc") {
+  upc: String!
   inStock: Boolean!
 }
 ```
 
 </CodeColumns>
 
-As long as `id` is a unique identifier within the products subgraph and `sku` is a unique identifier for the same product in the inventory subgraph, the subgraphs can each resolve the `Product` type.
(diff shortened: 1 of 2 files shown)


docs/source/entities-advanced.mdx
```diff
@@ -57,6 +57,32 @@ type Organization {
...
+type Product @key(fields: "sku") {
+  sku: ID!
+  inStock: Boolean!
+}
+```
+
+</CodeColumns>
+
+As long as `id` is a unique identifier within the products subgraph and `sku` is a unique identifier for the same product in the inventory subgraph, the subgraphs can each resolve the `Product` type.
+In other words, **each key declaration is specific to the subgraph it belongs to**, and you don't need to match `@key` fields between graphs.
```
lleadbet, line 84: To clarify this a bit, the `key` acts as a sort of sign to the query planner that says "hey you! i need this data to be able to identify this thing!". 

Just like talking to them, you'd also need to have the thing they're asking for in order for it to work. In the above example, the query planner wouldn't work. It'd choke because entering from either products or inventory can't ever lead to the other subgraph. 

This is the crux of _why_ we've historically made it so that the keys must match. 

A concrete example of this working would be: 

sub a:
```gql
type Query {
  pi: Product
}

type Product @key(fields: "sku") {
  sku: ID!
  upc: String!
  inStock: Boolean!
}
```

sub b: 
```gql
type Product @key(fields: "upc") {
  upc: String! @external
  name: String!
  price: Int
}
```

But as soon as you add a query to B, it needs to reference _something_ from sub a. It's a really weird edge-case and isn't going to come up tremendously often. 

@pcmanus may be able to explain this or at least confirm I'm not totally wrong here.

lleadbet (approved):

docs/source/entities.mdx
```diff
@@ -55,14 +55,14 @@ type Product @key(fields: "id") {
 
 The `@key` directive defines the entity's **primary key**, which consists of one or more of the type's `fields`. In the example above, the `Product` entity's primary key is its `id` field.
 
-**Every instance of an entity must be uniquely identifiable by its `@key` fields.** This is what enables your router to associate field data from _different_ subgraphs with the _same_ entity instance.
```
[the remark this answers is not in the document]
Meschreiber: I guess it must be true. (Otherwise, how would entities work?)

I'm a little confused though—if the keys can be mismatched, how do the keys help the router associate entity fields across subgraphs?

docs/source/entities.mdx
```diff
@@ -55,14 +55,14 @@ type Product @key(fields: "id") {
...
+**Every instance of an entity must be uniquely identifiable by its `@key` fields within its own subgraph.**
```
Meschreiber, line 58:
```suggestion
**Every instance of an entity must be uniquely identifiable by its `@key` fields within its own subgraph.** This is what enables your router to associate field data from _different_ subgraphs with the _same_ entity instance.
```

docs/source/entities.mdx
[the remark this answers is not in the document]
lleadbet: The other message describes it better (I think at least)- it's a weird thing.

docs/source/entities-advanced.mdx
dariuszkuc: I guess the most common use case for different `@key`s is that in one subgraph you could have multiple `@key`s and in another subgraph just one of them. 

While technically subgraphs don't have to share the same `@key`, practically 
- if they share the same fields (e.g. `upc` in the example above, which doesn't need the `@external`) - why wouldn't you define it as a `@key` in all subgraphs that have that field (which gives more flexibility to query planner)?
- if they don't share any of the fields - you won't be able to cross subgraph boundaries, i.e. you won't be able to merge entities as there is no way to merge entities using information from just one subgraph. **Queries against single subgraph will work though**

Meschreiber: This is very helpful, thank you @lleadbet and @dariuszkuc ! Just updated the text based on this info, would you take another look whenever you get a chance?

docs/source/entities-advanced.mdx
```diff
@@ -57,6 +57,99 @@ type Organization {
...
+```graphql title="Inventory subgraph"
+type Product @key(fields: "upc") {
+  upc: String!
+  inStock: Boolean!
+}
+```
+
+</CodeColumns>
+
+The following would allow you to merge the `Product` entity since both subgraph schemas include the `upc` field, even if it's only the `@key` for one of them:
```
dariuszkuc, line 107: well in this example it will be possible to resolve entity when starting from `products` subgraph but you won't be able to federate calls coming from `inventory` (as `upc` is not a `@key`)

docs/source/entities-advanced.mdx
```diff
@@ -57,6 +57,99 @@ type Organization {
...
+```graphql title="Inventory subgraph"
+type Product @key(fields: "upc") {
+  upc: String!
+  inStock: Boolean!
+}
+```
+
+</CodeColumns>
+
+If subgraphs share any fields—for example `upc` in the preceding example—it's best to use these shared field(s) as `@keys` for optimal [query planner](./entities#the-query-plan) performance.
```
dariuszkuc, line 129: subgraphs can share number of fields (for optimization purposes -> e.g. this is a common use case behind `@provides`) which cannot be used as keys as they do not uniquely identify the entity. Maybe rephrase it to indicate that if subgraphs share any fields that can be used to uniquely identify underlying entity, then they could be included as `@keys`. But again if there is a "better" key (e.g. single field vs composite key from multiple keys) they may never be used so...

docs/source/entities-advanced.mdx
```diff
@@ -57,6 +57,99 @@ type Organization {
...
+```graphql title="Inventory subgraph"
+type Product @key(fields: "upc") {
+  upc: String!
+  inStock: Boolean!
+}
+```
+
+</CodeColumns>
+
+To merge entities between subgraphs, the entity must have at least one shared field between subgraphs. For example, you can't merge the `Product` entity defined in the following subgraphs because they don't share any fields:
```
dariuszkuc, line 84: maybe something along those lines
```suggestion
To merge entities between subgraphs, the entity must have at least one shared field between subgraphs. For example, you can't merge the `Product` entity defined in the following subgraphs because they don't share any fields specified in `@key` selection set:
```

docs/source/entities-advanced.mdx
Meschreiber: Ah, got it. It seems this is maybe an unrealistic example and this should just be omitted :)

docs/source/entities-advanced.mdx
Meschreiber: > if there is a "better" key (e.g. single field vs composite key from multiple keys) they may never be used so...

What is best for query planner optimization? Just to use one single key field as long as its available on all subgraphs? Or is allowing for multiple keys better?

dariuszkuc: based on https://github.com/apollographql/federation/issues/2695 it appears that we prefer the simpler `@key`s

docs/source/entities-advanced.mdx
```diff
@@ -57,6 +57,77 @@ type Organization {
...
+```
+
+```graphql title="Inventory subgraph"
+type Product @key(fields: "upc") {
+  upc: String!
+  inStock: Boolean!
+}
+```
+
+</CodeColumns>
```
dariuszkuc, line 129: this section now seems redundant as thats what the `@key`s are for
```suggestion
```

Meschreiber: Agreed!

docs/source/entities-advanced.mdx
Meschreiber: Got it, thank you!

smyrick (approved):

dariuszkuc (approved):
Approved but just wondering whether we should call out explicitly the implication of having mismatched keys -> entities can be expanded as long as there exists a traversable path from the root query to access the fields

for example given the following schema
```graphql
// products subgraph
type Product @key(fields: "sku") {
  sku: ID!
  upc: String!
  name: String!
  price: Int
}

type Query {
  product(sku: ID!): Product
  products: [Product!]!
}

// inventory subgraph
type Product @key(fields: "upc") {
  upc: String!
  inStock: Boolean!
}
```

It will always be possible to resolve all entity fields through all queries as we can join entities through `upc` field. On the other hand if we add a query to the inventory subgraph

```graphql
type Product @key(fields: "upc") {
  upc: String!
  inStock: Boolean!
}

// new query
type Query {
  productsInStock: [Product!]!
}
```

not all queries will be able to federate the calls -> queries from product subgraph will be able to resolve `inStock` field from inventory subgraph BUT `productsInStock` query will not be able to federate the call as there is no `@key(fields: "upc")` in product subgraph and we don't have `sku` field.

lleadbet (approved):

Issue #2715, opened 2023-08-04
Unexpected `INTERNAL_SERVER_ERROR`


id-ilych:
Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `@apollo/gateway@2.5.1` for the project I'm working on.

I was doing a significant bump of `@apollo/gateway` package (`2.1.3.` -> `2.5.1`) and one of my tests failed.
I am not sure what is expected logic in `dataRewrites.js` but I guessed that the author assumed `typeof null` to be not an `object`, but it is.
Did I guess right? Why does it try to access fields of a `null` object (the parent field is nullable so it was perfectly correct for the downstream service to return `null`)?

Here is the diff that solved my problem:

```diff
diff --git a/node_modules/@apollo/gateway/dist/dataRewrites.js b/node_modules/@apollo/gateway/dist/dataRewrites.js
index 4c7e40d..c3d7a66 100644
--- a/node_modules/@apollo/gateway/dist/dataRewrites.js
+++ b/node_modules/@apollo/gateway/dist/dataRewrites.js
@@ -50,7 +50,7 @@ function applyAtPath(schema, path, value, fct) {
         }
         return;
     }
-    if (typeof value !== 'object') {
+    if (typeof value !== 'object' || value === null) {
         return;
     }
     if (path.length === 0) {
```

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

PR #2716, opened 2023-08-04
Fix "TypeError: Cannot read properties of null"

id-ilych:
Fixes https://github.com/apollographql/federation/issues/2715

### DIFF
--- a/gateway-js/src/dataRewrites.ts
+++ b/gateway-js/src/dataRewrites.ts
@@ -63,7 +63,7 @@ function applyAtPath(schema: GraphQLSchema, path: string[], value: any, fct: (ob
     return;
   }
 
-  if (typeof value !== 'object') {
+  if (typeof value !== 'object' || value === null) {
     return;
   }
 


trevor-scheer:
Thanks @id-ilych, the change looks correct to me (agreed that the `typeof null === object` is likely just an oversight). Would you mind sharing the condition to reproduce this error (or push up a test case)?

I'm hoping to release a patch today / soon, would love to include this so whatever is the fastest option is preferred. I don't mind pushing up a test if you can explain how to reproduce easily 😄

id-ilych:
Hey @trevor-scheer :wave:

https://codesandbox.io/p/sandbox/xenodochial-margulis-w66spv

You can change `test` method to return either value or a `null` for `node` field.
```js
tests() {
  // return [{node: {foo: 'Foo'}}]
  // return [{node: {bar: 'Bar'}}]
  return [{node: null}]
},
```

The value case works but the `null` case produces the error even though the `node` field is defined as nullable.
```json
{
  "data": {},
  "errors": [
    {
      "message": "Cannot read properties of null (reading '__typename')",
      "extensions": {
        "code": "INTERNAL_SERVER_ERROR",
        "stacktrace": [
          "TypeError: Cannot read properties of null (reading '__typename')",
          "    at isObjectOfType (/project/sandbox/node_modules/@apollo/gateway/src/dataRewrites.ts:106:26)",
          "    at applyAtPath (/project/sandbox/node_modules/@apollo/gateway/src/dataRewrites.ts:84:11)",
          "    at applyAtPath (/project/sandbox/node_modules/@apollo/gateway/src/dataRewrites.ts:79:7)",
          "    at applyAtPath (/project/sandbox/node_modules/@apollo/gateway/src/dataRewrites.ts:61:7)",
          "    at applyAtPath (/project/sandbox/node_modules/@apollo/gateway/src/dataRewrites.ts:79:7)",
          "    at applyRewrite (/project/sandbox/node_modules/@apollo/gateway/src/dataRewrites.ts:27:3)",
          "    at applyRewrites (/project/sandbox/node_modules/@apollo/gateway/src/dataRewrites.ts:13:5)",
          "    at /project/sandbox/node_modules/@apollo/gateway/src/executeQueryPlan.ts:452:24",
          "    at processTicksAndRejections (node:internal/process/task_queues:96:5)",
          "    at async executeNode (/project/sandbox/node_modules/@apollo/gateway/src/executeQueryPlan.ts:372:9)"
        ]
      }
    }
  ]
}
```

Not sure how to turn it into a unit test.

pcmanus:
The change itself does seem reasonable, but @id-ilych, I don't seem to be able to reproduce with the example you linked. I'll not that when I open the sandbox, it doesn't seem to show a valid query, but if try something like:
```graphql
query ExampleQuery {
  tests { node { id }}
}
````
I get a reasonable result and no error.

And actually, I was initially surprised when I saw your example because it uses a single subgraph, but the "data rewrites" code that the commit changes is only used in a few very specific cases (one involves `@interfaceObject`, another is a subtle case of interface and `@external` or `@requires`), and I believe those case require at least 2 subgraphs. I'll admit that I'm currently missing how the data rewrite would ever run on a single subgraph example, but then why would you have gotten the stack trace you pasted above? There is likely just something I'm missing/forgetting about here, but I'd really like to get to the bottom of what that is for my own sanity.

**Comment by id-ilych** [comment]
@pcmanus I don't know what happened to the sandbox :(

The problematic query is following (note that the two fields are aliased to the same name)
```graphql
query {
  tests {
    node {
      ... on Foo { test: foo }
      ... on Bar { test: bar }
    }
  }
}
```

pcmanus (approved):
Thanks @id-ilych, this makes more sense :). This case with 2 aliases happens to fall into the case I was thinking of when I said "another is a subtle case of interface and `@external` or `@requires`", I just had forgotten that the logic implied was a tad conservative (because it was harmless and easier to do this way) and so could applied in case like this (that don't involve `@external` or `@requires`).

Anyway, I've converted that repro into a unit test and added a changeset and I think we're good to go so gonna merge. Thanks again for the contrib.

PR #2722, opened 2023-08-07
docs: subgraph compatibility updates

dariuszkuc:
latest compatibility results


jeffjakub:
Doesn't pertain to the Ballerina update but is there a max width to the library description container? 
https://jam.dev/c/e40816c5-1408-4fe9-a6ed-a4009abc74b9

dariuszkuc:
> Doesn't pertain to the Ballerina update but is there a max width to the library description container? https://jam.dev/c/e40816c5-1408-4fe9-a6ed-a4009abc74b9

Good callout. It looks like Gatsby ignores the deprecated `width` attribute and the CSS style as well. @Meschreiber any suggestions on how to force consistent width on the tables?

PR #2755, opened 2023-08-30
docs: specify minVersion for composition hints

Meschreiber:

### DIFF
--- a/docs/source/hints.mdx
+++ b/docs/source/hints.mdx
@@ -2,9 +2,11 @@
 title: Composition hints
 ---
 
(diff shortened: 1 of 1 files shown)


docs/source/hints.mdx
```diff
@@ -2,9 +2,11 @@
 title: Composition hints
 ---
```
Meschreiber, line 3: I opted not to include a minVersion in the frontmatter since Composition hints were available before 2.4 via the `rover supergraph compose` command.

dariuszkuc (approved):

PR #2769, opened 2023-09-04
feat: add support for smallrye-graphql

ivanp:
smallrye-graphql is using weird `content-type` response: application/graphql+json


dariuszkuc:
Hello 👋 
Thanks for raising this but `application/graphql+json` is not a standard content type and as such we will not be adding generic support for it. [GraphQL over HTTP spec](https://github.com/graphql/graphql-over-http/blob/main/spec/GraphQLOverHTTP.md#serialization-format) defines `application/json` and `application/graphql-response+json` as officially recognized formats and those are the ones we support.

This should be fixed in the `smallrye-graphql` library.

PR #2771, opened 2023-09-05
docs: add note about query planner key usage

Meschreiber:
This PR adds a note about how the query planner handles a key selection set with redundancies.
It also reorders and copyedits some of the surrounding sections.


docs/source/entities-advanced.mdx
```diff
@@ -6,13 +6,39 @@ This article describes complex behaviors of federated entities beyond those cove
...
+
+### Compound `@key`s
+
+A single `@key` can consist of multiple fields, the combination of which is required to uniquely identify an entity. This is called a **compound** or composite key. In the following example, the combination of both `username` and `domain` fields is required to uniquely identify the `User` entity:
+
+```graphql {1} title="Users subgraph"
+type User @key(fields: "username domain") {
+  username: String!
+  domain: String!
+}
```
shorgi, line 19: Add, or link to, explanation: must a @key field be required (`!`)? What happens if a @key field isn't required?

docs/source/entities-advanced.mdx
```diff
@@ -22,10 +48,34 @@ type Product @key(fields: "id") @key(fields: "sku") {
   price: Int
 }
 ```
+<blockquote>
+
+**Note:** If you include multiple sets of `@key` fields, the query planner always uses the most efficient set for entity resolution. For example, suppose you allow a type to be identified by `@key(fields: "id")` _or_ `@key(fields: "id sku")`:
```
shorgi, line 53: Copyedit suggestion: is there a more precise definition for `most efficient`? Is it the least number of operands?

dariuszkuc (approved):
nice cleanup!

docs/source/entities-advanced.mdx
```diff
@@ -40,28 +90,9 @@ type Product @key(fields: "sku", resolvable: false) {
...
-  id: ID!
-}
-```
-
 ### Differing `@key`s across subgraphs
 
-An entity often has the exact same `@key` field(s) across subgraphs, but this isn't required. For example, you can define a `Product` entity shared between subgraphs, one with `sku` and `upc` as the `@key`s and one with only `upc` as the `@key` field:
-
-<CodeColumns>
+An entity often uses the exact same `@key` field(s) across subgraphs, but this isn't required. For example, you can define a `Product` entity shared between subgraphs, one with `sku` and `upc` as the `@key`s and one with only `upc` as the `@key` field:
```
shorgi, line 95:
```suggestion
Although an entity commonly uses the exact same `@key` field(s) across subgraphs, you can alternatively use different `@key`s with different fields. For example, you can define a `Product` entity shared between subgraphs, one with `sku` and `upc` as its `@key`s, and the other with only `upc` as the `@key` field:
```

docs/source/entities-advanced.mdx
Meschreiber: Good question. That's how I understood it from [this conversation](https://github.com/apollographql/federation/issues/2695), but I wonder what else could be at play here. For example, if two different keys both use the same number of fields, how does the query planner decide which is more efficient? Does it calculate the shortest "path" to resolve fields?

@dariuszkuc , would you be able to help with this?

docs/source/entities-advanced.mdx
Meschreiber: Good callout! I'd assume keys can only use non-nullable field but based on [this conversation](https://apollograph.slack.com/archives/C03RH11Q6S1/p1692821163122989), it seems that's not the case.

I think this info is better suited for the "basics" page and will add there. https://github.com/apollographql/federation/pull/2771/commits/df834d31809f996a8c27331545fe08e5c92cbfd1

dariuszkuc: yep ideally `@key`s are non-nullable but they can be nullable

docs/source/entities-advanced.mdx
dariuszkuc: > if two different keys both use the same number of fields, how does the query planner decide which is more efficient? Does it calculate the shortest "path" to resolve fields?



@pcmanus by any chance do you know whats the optimization logic here?

Meschreiber: Got a chance to check in with @pcmanus and learned that the query planner a cost function to score keys, and it usually selects the key with the fewest fields in case of ties, although there are no strict guarantees. Since this is likely to come up infrequently, I'm going to leave it relatively vague.

Issue #2743, opened 2023-08-22
query execution error when using @interfaceObject


maxrad82:
### Issue Description

I'm experiencing a very strange behaviour using the @interfaceObject directive

Considering 3 services

Service 1:
```graphql
        extend schema
        @link(url: "https://specs.apollo.dev/federation/v2.3",
            import: ["@key"])

        interface Product @key(fields:"id") {
            id: ID!
        }

        type Book implements Product @key(fields:"id") {
            id:ID!
            title: String
        }

        type Query {
            allProducts: [Product!]!
        }
```
Service 2:
```graphql
        extend schema
        @link(url: "https://specs.apollo.dev/federation/v2.3",
            import: ["@key", "@interfaceObject"])

        type Product @key(fields:"id") @interfaceObject {
            id: ID!
            reviews: [String!]!
        }
```
Service 3:
```graphql
        extend schema
        @link(url: "https://specs.apollo.dev/federation/v2.3",
            import: ["@key", "@interfaceObject"])

        type Product @key(fields:"id") @interfaceObject {
            id: ID!
        }

        type Query {
            searchProducts: [Product!]!
        }
```
depending on the query, it may failed, suceed or be inconsistant
```graphql
  searchProducts {
    id
    reviews    
  }
```
=> ok
```graphql
  searchProducts {
    __typename
    id
    reviews    
  } 
```
=> error "Cannot return null for non-nullable array element of type Product at index 0"
```graphql
  searchProducts {
    id
    ...on Book {
      reviews
    }    
  }
```
=> inconsistant result, missing reviews field

I have linked a github repo to reproduce this issue. I'm pretty sure the implementation is correct according the documentation

### Link to Reproduction

https://github.com/maxrad82/apollo-federation-bug-interface

### Reproduction Steps

Prerequisite:
- docker
-  docker-compose
-  rover

clone the repository and run make run  to start the subgraphs and the supergraph
go to http://localhost:4007

run the followng queries:
```graphql
query ExampleQuery {  
  searchProducts {
    id
    reviews
  }
}
```
=> ok
```graphql
query ExampleQuery {  
  searchProducts {
   __typename
    id
    reviews
  }
}
```
=> fails
```graphql
query ExampleQuery {  
  searchProducts {
    id
    ...on Book {
      reviews
    }    
  }
}
```
=> reviews is not return

PR #2775, opened 2023-09-08
Don't treat entity interface `__typename` refinements as useless

trevor-scheer:
Fix specific case for requesting `__typename` on interface entity type

In certain cases, when resolving a `__typename` on an interface entity (due to it actual being requested in the operation), that fetch group could previously be trimmed / treated as useless. At a glance, it appears to be a redundant step, i.e.:
```
{ ... on Product { __typename id }} => { ... on Product { __typename} }
```
It's actually necessary to preserve this in the case that we're coming from an interface object to an (entity) interface so that we can resolve the concrete `__typename` correctly.

Ref: #2743
(this fixes one of the queries from the issue but not both)


query-planner-js/src/buildPlan.ts
```diff
@@ -1226,9 +1226,39 @@ class FetchGroup {
...
+    // keep this since this is our only path to resolving the concrete `__typename`.
+    const isInterfaceTypeConditionOnInterfaceObject = (
+      selection: Selection
+    ): boolean => {
+      if (selection.kind === "FragmentSelection") {
+        const parentType = selection.element.typeCondition;
+        if (parentType && isInterfaceType(parentType)) {
+          // Lastly, we just need to check that we're coming from a subgraph
+          // that has the type as an interface object in its schema.
+          return this.parents().some((p) => {
```
clenfest, line 1242: Just for my own edification, probably just being dense, but why are there multiple parents here?

clenfest (approved):

query-planner-js/src/buildPlan.ts
trevor-scheer: My understanding (and would appreciate @pcmanus to confirm) is that we'd rarely have more than one parent in practice, but it's possible in the case that we also happen to be in a [diamond-shaped dependency](https://github.com/apollographql/federation/pull/1900) situation.

PR #2776, opened 2023-09-08
fix: propagate type information when renaming entity fields

dariuszkuc:


gateway-js/src/dataRewrites.ts
```diff
@@ -35,8 +35,11 @@ function rewriteAtPathFunction(rewrite: FetchDataRewrite, fieldAtPath: string):
       };
     case 'KeyRenamer':
       return (obj) => {
-        obj[rewrite.renameKeyTo] = obj[fieldAtPath];
-        obj[fieldAtPath] = undefined;
+        const objAtPath = obj[fieldAtPath];
+        if (objAtPath) {
```
pcmanus, line 39: I'll note that this change is actually enough to fix the issue at hand, even without the changes in `executeQueryPlan.ts`. Because what happened in this example was that the "rewrite" was saying "rename field `foo__alias0` into `foo`, but it was applied to an object looking like `{ foo: { field: "something" }}`, so one would except the rewrite to do nothing (the field we're asked to rename does not exists), but instead it was still doing the assignment and so was overriding `foo`.

That said, the change in `executeQueryPlan.ts` basically make sure we reuse type information when we can (so, if the __typename is not in the fetch result, but was known pre-fetch, we use that), and that certainly make sense and ensure the rewrite logic is more precise and isn't even tried in the case where the object is truly a `T1` (referring to the added test here).

Anyway, I'd say that the change in `executeQueryPlan.ts` is really the best fix here, but it still feel worth to me to fix the rewriting logic to be a bit more intuitive/less destructive, hence including both change in the commit.

trevor-scheer (approved):
LGTM. Have we confirmed if this bug exists in the router as well? I don't know if the router's execution was modeled directly after ours. Might make for a good test case over there anyway.

PR #2779, opened 2023-09-13
Adds code to output mermaid graph for query graphs

pcmanus:
When learning about the "query graph" abstraction used by the code, or debugging one, it can sometimes be convenient to display such query graph as an actual graph. There is already old code to output a query graph in the [graphviz](https://graphviz.org/) dot format, but it can a bit more convenient to use the [mermaid](https://mermaid.js.org/) format as [github display it automatically](https://github.blog/2022-02-14-include-diagrams-markdown-files-mermaid/). This PR adds some simple code to do so that I used recently. It's probably not perfect, but it works well enough on simple example (besides, complex example gets quickly unreadable anyway when display as a graph, so this is really only useful on relatively small example).


query-graphs-js/src/querygraph.ts
```diff
@@ -61,6 +61,13 @@ export function isFederatedGraphRootType(type: NamedType) {
  */
 export class Vertex {
   hasReachableCrossSubgraphEdges: boolean = false;
+  // @provides works by creating duplicates of the vertex/type involved in the provides and adding the provided
+  // edges only to those copy. This means that with @provides, you can have more than one vertex per-type-and-subgraph
+  // in a query graph. Which is fined, but this `provideId` allows to distinguish if a vertex was created as part of
+  // this @provides duplication or not. The value of this field has no other meaning than to be unique per-@provide,
+  // and so all the vertex copied for a given @provides application will have the same `provideId`. Overall, this
+  // mostly exists for debugging visualisation.
+  provideId: number | undefined;
```
pcmanus, line 70: Pretty much all the change in this file are just to populate this, because I noticed there wasn't a simple way to display things correctly in mermaid without that. I think it's kind of convenient to have for debugging in general, and it should be fairly obvious that this addition doesn't break anything.

query-graphs-js/src/querygraph.ts
```diff
@@ -1092,8 +1105,7 @@ class GraphBuilderFromSchema extends GraphBuilder {
     private readonly supergraph?: { apiSchema: Schema, isFed1: boolean },
   ) {
     super();
-    this.isFederatedSubgraph = isFederationSubgraphSchema(schema);
-    assert(!this.isFederatedSubgraph || supergraph, `Missing supergraph schema for building the federated subgraph graph`);
+    this.isFederatedSubgraph = !!supergraph && isFederationSubgraphSchema(schema);
```
pcmanus, line 1108: This one change is the only exception to "this just adds `provideId`. The reason for this is that the assertion was getting in the way of creating a standalone query graph for a simple subgraph (outside of a full federated query graph that is), which is never done by the code in practice, but can still be nice to do when debugging/learning about query graphs (basically to be able to visualise what is created for a single subgraph in isolation). Again, shouldn't change anything for existing code.

trevor-scheer (approved):

PR #2785, opened 2023-09-18
Enforce `prettier` in test files

trevor-scheer:
Jest's inline snapshotting feature is used throughout our tests and offers a pretty great DX for diffing and updating test expectations. However, any time a snapshot is written or updated, Jest prettifies the entire file. Without prettier enforcement on our tests, the result is either:
* commit a bunch of unrelated noise whenever a snapshot is added/updated, or
* obtain the snapshot, copy it, re-checkout the file/revert the changes, and paste the snapshot back in

Both less-than-ideal. For the sake of usability alone and the fact that we're not actively moving away from inline snapshotting, might as well buy in and work with the tooling instead of against it.

TODO:
- [ ] add commit hash to `.git-blame-ignore-revs`


trevor-scheer:
Note to self: merge don't squash to preserve commit hash for hiding blame

.circleci/config.yml
```diff
@@ -31,7 +31,6 @@ jobs:
           path: junit.xml
 
 workflows:
-  version: 2
```
trevor-scheer, line 34: `version` is deprecated, this is just cleanup

.circleci/config.yml
```diff
@@ -56,3 +55,6 @@ workflows:
       - node/run:
           name: Check Spelling
           npm-run: spell:check
+      - node/run:
```
clenfest, line 58: Do we want to consider adding a pre-commit hook? Might be annoying to not notice build failures until late.

trevor-scheer: Yeah, that's a nice to have. I'll be using it. Added it to our (now growing) set of opt-in repo git hooks [cdf70e1](https://github.com/apollographql/federation/pull/2785/commits/cdf70e1839e3cc1ff68a6ed4fa0217de5a65eae3).

clenfest (approved):

PR #2791, opened 2023-09-25
Fix `@interfaceObject` + `__typename` / key bug

trevor-scheer:
Just a repro for now. As it stands currently, the query plan in the snapshot is correct but the test fails. The entire `... on T1` is unexpectedly missing.

### DIFF
--- a/gateway-js/src/__tests__/executeQueryPlan.test.ts
+++ b/gateway-js/src/__tests__/executeQueryPlan.test.ts
@@ -3796,7 +3796,7 @@ describe('executeQueryPlan', () => {
             },
           },
         }
-      `);
+        `);
 
         const response = await executePlan(
           queryPlan,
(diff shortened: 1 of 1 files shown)


michael-watson:
Just wanted to add a confirmed workaround for this problem. Any field on the `interface` could be defined at the root of that interface and not duplicated in the fragments. 

From the test in this repo:
```
query {
    itfObj {
      __typename
      ... on T1 {
        __typename
        id
      }
      ... on T2 {
        # Uncomment id to break operation
        # id
        __typename
      }
    }
  }
```
Could be this to fix it and generate the appropriate plan:
```
query {
    itfObj {
        id
      __typename
      ... on T1 {
        __typename
      }
      ... on T2 {
        __typename
      }
    }
  }
```

Issue #2795, opened 2023-09-28
Remove "primary key" language


trevor-scheer:
We refer to `@key`s as "primary keys" in various places, but `@key` doesn't align well with the actual definition of a PK so it's a misnomer that might add confusion. We allow multiple keys and varying keys on the same entity from one subgraph to another, whereas there can only be one PK in a database context. In most cases `@key` will be used as a PK so it can be helpful to build the mental model but it should be clear that that isn't a _constraint_ that `@key` is bound to.

PR #2796, opened 2023-09-29
Clarify primary key terminology for `@key`

Meschreiber:
Addresses https://github.com/apollographql/federation/issues/2795

### DIFF
--- a/docs/source/entities.mdx
+++ b/docs/source/entities.mdx
@@ -57,7 +57,7 @@ The `@key` directive defines an entity's **unique key**, which consists of one o
 
 In most cases, the `@key` field(s) for the same entity will be the same across subgraphs. For example, if one subgraph uses `id` as the `@key` field for the `Product` entity, other subgraphs should do the same. However, this [isn't strictly required](./entities-advanced#differing-keys-across-subgraphs).
 
-> If coming from a database context, it can be helpful to think of a `@key` as an entity's [**primary key**](https://en.wikipedia.org/wiki/Primary_key). This term isn't completely accurate for entities since a single entity can have [multiple `@key`s](./entities-advanced#multiple-keys). The field(s) you select for an entity's `@key` must, however, uniquely identify the entity. In that way, `@key`s are similar to [candidate keys](https://en.wikipedia.org/wiki/Candidate_key).
+> If coming from a database context, it can be helpful to think of a `@key` as an entity's [primary key](https://en.wikipedia.org/wiki/Primary_key). This term isn't completely accurate for entities since a single entity can have [multiple `@key`s](./entities-advanced#multiple-keys). The field(s) you select for an entity's `@key` must, however, uniquely identify the entity. In that way, `@key`s are similar to [candidate keys](https://en.wikipedia.org/wiki/Candidate_key).
 
 <CodeColumns>
 
--- a/docs/source/entities.mdx
+++ b/docs/source/entities.mdx
@@ -57,7 +57,7 @@ The `@key` directive defines an entity's **unique key**, which consists of one o
 
 In most cases, the `@key` field(s) for the same entity will be the same across subgraphs. For example, if one subgraph uses `id` as the `@key` field for the `Product` entity, other subgraphs should do the same. However, this [isn't strictly required](./entities-advanced#differing-keys-across-subgraphs).
 
-> If coming from a database context, it can be helpful to think of a `@key` as an entity's [**primary key**](https://en.wikipedia.org/wiki/Primary_key). The term isn't completely accurate for entities since a single entity can have [multiple `@key`s](./entities-advanced#multiple-keys). The field(s) you select for an entity's `@key` must, however, uniquely identify the entity.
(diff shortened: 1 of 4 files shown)


docs/source/federated-types/federated-directives.mdx
```diff
@@ -27,10 +27,10 @@ If an imported directive's default name matches one of your own custom directive
 ```graphql
 extend schema
   @link(url: "https://specs.apollo.dev/federation/v2.3",
-        import: [{ name: "@key", as: "@primaryKey"}, "@shareable"])
+        import: [{ name: "@key", as: "@uniqueKey"}, "@shareable"])
```
Meschreiber, line 30: I originally wasn't going to change this renaming directives example since users can rename a directive however they wish. I decided to in the end for consistency purposes. Curious if folks feel strongly otherwise though.

docs/source/entities.mdx
```diff
@@ -53,10 +53,12 @@ type Product @key(fields: "id") {
 }
 ```
 
-The `@key` directive defines the entity's **primary key**, which consists of one or more of the type's `fields`. In the example above, the `Product` entity's primary key is its `id` field. **Every instance of an entity must be uniquely identifiable by its `@key` fields.** This is what enables your router to associate field data from _different_ subgraphs with the _same_ entity instance.
+The `@key` directive defines the entity's **unique key**, which consists of one or more of the type's `fields`. In the example above, the `Product` entity's unique key is its `id` field. **Every instance of an entity must be uniquely identifiable by its `@key` fields.** This is what enables your router to associate field data from _different_ subgraphs with the _same_ entity instance.
```
trevor-scheer, line 56:
```suggestion
The `@key` directive defines an entity's **unique key**, which consists of one or more of the type's `fields`. In the example above, the `Product` entity's unique key is its `id` field. **Every instance of an entity must be uniquely identifiable by its `@key` fields.** This is what enables your router to associate field data from _different_ subgraphs with the _same_ entity instance.
```

the -> an due to possibility of multiple?

docs/source/entities.mdx
```diff
@@ -53,10 +53,12 @@ type Product @key(fields: "id") {
...
 
 In most cases, the `@key` field(s) for the same entity will be the same across subgraphs. For example, if one subgraph uses `id` as the `@key` field for the `Product` entity, other subgraphs should do the same. However, this [isn't strictly required](./entities-advanced#differing-keys-across-subgraphs).
 
+> If coming from a database context, it can be helpful to think of a `@key` as an entity's [**primary key**](https://en.wikipedia.org/wiki/Primary_key). The term isn't completely accurate for entities since a single entity can have [multiple `@key`s](./entities-advanced#multiple-keys). The field(s) you select for an entity's `@key` must, however, uniquely identify the entity.
```
trevor-scheer, line 60: I am comfortable with making this association since most people have at least a good idea around PKs. However, I will also suggest an alternative "candidate key" which is a more accurate fit for the concept.
https://en.wikipedia.org/wiki/Candidate_key

trevor-scheer: I'm not suggesting that we change all the unique key terminology to candidate key though. Candidate key is probably a _bit_ too specific to make it the broad-use term.

docs/source/federated-types/federated-directives.mdx
trevor-scheer: I see no reason not to

trevor-scheer (approved):

docs/source/entities.mdx
shorgi: @trevor-scheer - is an entity having multiple `@keys` analogous to a database having composite keys? If so, is that worth describing here?

shorgi (approved):
Copy LGTM, with one [comment](https://github.com/apollographql/federation/pull/2796/files#r1342206590)

docs/source/entities.mdx
trevor-scheer: @shorgi not quite. `@key`s can _also_ be composite. As long as each `@key` (on its own) can uniquely identify an entity within its respective subgraph, it's valid. As an example, maybe a product can be uniquely identified by its sku OR the combination of model + variant + color + size. That would look like:

```graphql
type Product @key(fields: "id") @key(fields: "model variant color size") {
  sku: String!
  model: String!
  variant: String!
  color: String!
  size: String!
  # ...
}

The main differentiation here is that tables only have one primary key.

File: docs/source/entities.mdx (hunk 2 above)
Comment (Meschreiber) [reply docs/source/entities.mdx:56]: Good call

File: docs/source/entities.mdx (hunk 3 above)
Comment (Meschreiber) [reply docs/source/entities.mdx:60]: (@shorgi , just FYI you can create composite/compound `@keys` too: https://www.apollographql.com/docs/federation/entities-advanced/#compound-keys, but as Trevor mentioned, we're differentiate vs having multiple keys (composite or not) here.)

Comment (Meschreiber) [reply docs/source/entities.mdx:60]:
```suggestion
> If coming from a database context, it can be helpful to think of a `@key` as an entity's [**primary key**](https://en.wikipedia.org/wiki/Primary_key). This term isn't completely accurate for entities since a single entity can have [multiple `@key`s](./entities-advanced#multiple-keys). The field(s) you select for an entity's `@key` must, however, uniquely identify the entity. In that way, `@key`s are similar to [candidate keys](https://en.wikipedia.org/wiki/Candidate_key).
```

PR #2800, opened 2023-10-02
WIP: Fix nested `@requires` bug

trevor-scheer:
For now this is just a reproduction. There are two tests here which illustrate the two bullets below.

Either:
* Composition should error since our `@requires` field set passes _through_ a non-external top-level field OR
* The query planner should be able to handle this case and generate a valid query plan without throwing an error


trevor-scheer:
Fixed via https://github.com/apollographql/federation/pulls/2895

PR #2813, opened 2023-10-06
feat: Extend support for content types that include JSON

2012160085:
### Problem:
While federating a `smallrye-graphql` based API server using Apollo Gateway, I encountered an issue. The server responds with a `application/graphql+json` content type, which is currently not supported by the gateway.

### Proposed Solution:
This PR aims to broaden the supported content types in Apollo Gateway, specifically to include responses with `application/graphql+json` and other JSON subtypes. This change ensures that APIs like those built with `smallrye-graphql` can be seamlessly integrated without content type restrictions.

### Changes:
- Enhanced the content type detection logic to support types that include `json`.

### DIFF
--- a/gateway-js/src/datasources/RemoteGraphQLDataSource.ts
+++ b/gateway-js/src/datasources/RemoteGraphQLDataSource.ts
@@ -293,11 +293,10 @@ export class RemoteGraphQLDataSource<
     _fetchRequest?: NodeFetchRequest,
     _context?: TContext,
   ): Promise<object | string> {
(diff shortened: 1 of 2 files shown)


dariuszkuc:
Thanks for the PR. Closing this with the same reason as was provided for https://github.com/apollographql/federation/pull/2769#issuecomment-1706891804.

Looks like the underlying issue was already fixed on the [quarkus side](https://github.com/quarkusio/quarkus/pull/35769).

PR #2838, opened 2023-10-25
[docs] Add __typename example to advanced entity docs

rhino88:
In some cases it may be useful to define an entity as a [singleton](https://en.wikipedia.org/wiki/Singleton_pattern). The `@key` directive supports the `__typename` [Meta Field](https://graphql.org/learn/queries/#meta-fields) to declare and entity as a singleton.

In the following example, the `Meta` entity can be uniquely identified by its `__typename` (because there is only ever one instance):

```graphql title="Meta singleton"
type Meta @key(fields: "__typename") {
    version: String!
}
```

I was able to confirm this works as expected in a repro: https://codesandbox.io/p/sandbox/tender-engelbart-wympgk?file=%2Ftsconfig.json%3A1%2C1

### DIFF
--- a/docs/source/entities-advanced.mdx
+++ b/docs/source/entities-advanced.mdx
@@ -216,6 +216,10 @@ type Query {
 }
 ```
 
+### Using `__typename` as `@key`
+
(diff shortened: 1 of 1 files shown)


Meschreiber:
Thanks for this @rhino88 ! The team is discussing whether this is a pattern we want to promote, we'll keep you updated.

rhino88:
Thanks @Meschreiber.

This is something we've found useful for a number of scenarios in production (e.g `HomePage` entities). The only current alternative we found would be to introduce another dummy field to use as a `@key`, which isn't really desirable.

PR #2839, opened 2023-10-26
Remove invalid advice about omitting fields from `@shareable` types

dylan-apollo:
I'm pretty sure the example that was present was never valid, because I don't think the field can be omitted _and_ resolvable unless it's an entity.

This article probably needs a larger pass to separate it more clearly from entities, but the immediate goal here is to remove the example that caused some confusion for a Discord user.

### DIFF
--- a/docs/source/federated-types/sharing-types.mdx
+++ b/docs/source/federated-types/sharing-types.mdx
@@ -137,7 +137,8 @@ To make `Position.z` shareable, you can do one of the following:
 
 ## Differing shared fields
 
-The type of a shared field can vary in nullability (`String` / `String!`), but _can't_ vary in its core type (`Int` vs. `String`) or in whether it returns a list (`Int` vs. `[Int]`). Shareable value types must also have the same fields in _every_ subgraph that defines them.
+Shared fields can only differ in their [return types](#return-types) and [arguments](#arguments) in specific ways.
+If fields you want to share between subgraphs differ more than is permitted, use  [entities](../entities) instead of shareable value types.
 
 > To contribute different fields from different subgraphs, you can use [entity types](../entities) instead of shareable value types.
 
--- a/docs/source/federated-types/sharing-types.mdx
+++ b/docs/source/federated-types/sharing-types.mdx
@@ -137,35 +137,9 @@ To make `Position.z` shareable, you can do one of the following:
 
 ## Differing shared fields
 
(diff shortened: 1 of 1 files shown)


docs/source/federated-types/sharing-types.mdx
```diff
@@ -137,35 +137,9 @@ To make `Position.z` shareable, you can do one of the following:
 
 ## Differing shared fields
 
-Shared fields can differ between subgraphs in specific ways:
-* The return type of a shared field can vary in nullability (`String` / `String!`).
-  * A shared field's return type _can't_ vary in its core type (`Int` vs. `String`) or in whether it returns a list (`Int` vs. `[Int]`).
-* A field can be omitted from a subgraph, _if_ that field is always [resolvable](../federated-types/composition/#rules-of-composition).
+The type of a shared field can vary in nullability (`String` / `String!`), but _can't_ vary in its core type (`Int` vs. `String`) or in whether it returns a list (`Int` vs. `[Int]`). Shareable value types must also have the same fields in _every_ subgraph that defines them.
```
Meschreiber, line 140: It actually looks as though this is covered (with examples) in the following subsections, e.g.: https://www.apollographql.com/docs/federation/federated-types/sharing-types#return-types

(That's not to say these sections couldn't use a little copyedittng too...)

Instead of rehashing it in the intro, I might say something like this:

```suggestion
Shared fields can only differ in their [return types](#return-types) and [arguments](#arguments) in specific ways.
If fields you want to share between subgraphs differ more than is permitted, use  [entities](../entities) instead of shareable value types.
```

docs/source/federated-types/sharing-types.mdx
```diff
@@ -137,35 +137,9 @@ To make `Position.z` shareable, you can do one of the following:
...
-  inStock: Boolean! # Not in A
-}
-```
-
-<></>
-
-</CodeColumns>
-
-The above `Food` types differ in the nullability of their fields _and_ the fields included in each type.
+> To contribute different fields from different subgraphs, you can use [entity types](../entities) instead of shareable value types.
```
Meschreiber, line 142: And then you could delete this.

```suggestion
```

dylan-apollo:
Thanks @Meschreiber, suggestions accepted.

Meschreiber (approved):

PR #2848, opened 2023-11-07
Adding note on resolving an entity when it has multiple keys

andrewmcgivery:
Adding some clarity around resolving an entity when there are multiple keys present.

This is due to a custom request about how to handle this... they couldn't find anything in the docs... and either could we. :)

### DIFF
--- a/docs/source/entities-advanced.mdx
+++ b/docs/source/entities-advanced.mdx
@@ -87,7 +87,24 @@ type Product @key(fields: "id", resolvable: false) {
 type Product @key(fields: "sku", resolvable: false) {
   sku: String!
 }
+```
 
+When resolving a reference for an entity with multiple keys, you can determine how to resolve based on which key is present:
+
+```js title="resolvers.js"
+// Products subgraph
+const resolvers = {
+  Product: {
+    __resolveReference(productRepresentation) {
+      if(productRepresentation.sku){
+        return fetchProductBySku(productRepresentation.id);
+      } else {
+        return fetchProductByID(productRepresentation.id);
+      }
+    }
+  },
+  // ...other resolvers...
+}
 ```
 
 ### Differing `@key`s across subgraphs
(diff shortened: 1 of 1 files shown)


Meschreiber (approved):
Thank you for this addition!  🙏

docs/source/entities-advanced.mdx
```diff
@@ -87,7 +87,24 @@ type Product @key(fields: "id", resolvable: false) {
 type Product @key(fields: "sku", resolvable: false) {
   sku: String!
 }
+```
+
+When resolving a reference for an entity with multiple keys, you can determine how to resolve it based on which key is present:
 
+```js title="resolvers.js"
+// Products subgraph
+const resolvers = {
```
dariuszkuc, line 96: this is an implementation detail that is only going to work for `subgraph-js` based implementation

trevor-scheer: Do you just mean `__resolveReference` is a `subgraph-js` thing? The example is still good and other libs will all have to do something similar, right? Maybe nice to just mention "if you're using `@apollo/subgraph` this could look like..."?

Meschreiber: Thanks @trevor-scheer , I like that suggestion! Also agree that eventually subgraph library specific implementation examples might be better placed elsewhere. For now, this is probably the best place.

docs/source/entities-advanced.mdx
```diff
@@ -87,7 +87,24 @@ type Product @key(fields: "id", resolvable: false) {
 type Product @key(fields: "sku", resolvable: false) {
   sku: String!
 }
+```
+
+When resolving a reference for an entity with multiple keys, you can determine how to resolve it based on which key is present:
```
Meschreiber, line 92:
```suggestion
When resolving a reference for an entity with multiple keys, you can determine how to resolve it based on which key is present. For example, if you're using [`@apollo/subgraph`](/apollo-server/using-federation/api/apollo-subgraph/), it could look like this:
```

PR #2858, opened 2023-11-14
chore: sangria compatibility updates

dariuszkuc:


dariuszkuc:
Updated caliban results as failures were due to flaky tests.

PR #2859, opened 2023-11-14
Fix removeInputsFromSelection to only remove inputs

clenfest:
removeInputsFromSelection is tasked with removing inputs to a FetchGroup from the fetch group so that we don't retrieve the same values twice (and FetchGroups that don't functionally do anything can be removed). There was a bug in the code that meant that in certain cases, things that were required were actually being removed. This could manifest in subgraph jumps not being possible because the key it was using to make the jump was not available.


query-planner-js/src/__tests__/buildPlan.test.ts
```diff
@@ -3810,7 +3810,7 @@ describe('Fed1 supergraph handling', () => {
 });
 
 describe('Named fragments preservation', () => {
-  it('works with nested fragments', () => {
+  it('works with nested fragments 1', () => {
```
trevor-scheer, line 3813: stray?
```suggestion
  it('works with nested fragments', () => {
```

.changeset/pink-ducks-fail.md
```diff
@@ -0,0 +1,6 @@
+---
+"@apollo/query-planner": patch
+---
+
+removeInputsFromSelection is tasked with removing inputs to a FetchGroup from the fetch group so that we don't retrieve the same values twice (and FetchGroups that don't functionally do anything can be removed). There was a bug in the code that meant that in certain cases, things that were required were actually being removed. This could manifest in subgraph jumps not being possible because the key it was using to make the jump was not available.
```
trevor-scheer, line 5: This is nbd, but the details of the fix are probably less interesting to changelog readers than a description of the bug itself. I might just reorder this to be something like:
"Fix a bug where...xyz." then followed up with implementation details that might explain why it happened.

trevor-scheer (approved):
Nice sleuthing!

query-planner-js/src/__tests__/buildPlan.test.ts
clenfest: The reason I had this in here is that if you try to run a test by name, jest will actually run all tests for which that name is a subset. So if you run "test1", but a "test11" and "test12" exist, it will run all three. I was trying to debug this test, and got confused because another test would run immediately afterwards, and it took me a while to realize it was a separate execution.

Issue #2862, opened 2023-11-16
Build Plan failure when using fragments and @include


tinnou:
### Issue Description

We are getting the following error when using queries that have a combination of non intersecting fragments and `@include` directive.

Error:
```
Cannot add fragment of condition "Video" (runtimes: [Show]) to parent type "Show" (runtimes: Show)
GraphQLError: Cannot add fragment of condition "Video" (runtimes: [Show]) to parent type "Show" (runtimes: Show)
    at Object.err (/Users/antoineb/IdeaProjects/federation/internals-js/src/error.ts:36:60)
    at validate (/Users/antoineb/IdeaProjects/federation/internals-js/src/operations.ts:61:34)
    at FragmentElement.rebaseOn (/Users/antoineb/IdeaProjects/federation/internals-js/src/operations.ts:492:7)
    at FragmentElement.rebaseOnOrError (/Users/antoineb/IdeaProjects/federation/internals-js/src/operations.ts:91:17)
    at makeSelection (/Users/antoineb/IdeaProjects/federation/internals-js/src/operations.ts:2239:40)
(...)
```

### Link to Reproduction

https://github.com/apollographql/federation/compare/main...tinnou:federation:non_intersecting_fragment_with_include

### Reproduction Steps

I added a unit test that reproduces the issue with a minimal schema and query in `buildPlan.test.ts` see the commit details.

If you remove `@include` the plan succeeds.

PR #2864, opened 2023-11-22
Fix nested fragments on interfaces (when a directive is present)

trevor-scheer:
The following query:

```gql
query {
  i {
    _id
    ... on I2 {
      ... on I2 @test {
        id
      }
    }
  }
}
```
Blows up with the following error:
>Cannot add fragment of condition "I2" (runtimes: [T1,T2]) to parent type "T1" (runtimes: T1)

When the `@test` directive is removed, we collapse the inline fragments and this error doesn't occur, but we can't do that when a directive application is present.

This _seems_ to be due to the way that the query planner handles interface conditions, by expanding them into their runtime types. My understanding right now is roughly this:

Expanding the above query in theory looks something like:
```gql
query {
  i {
    _id
    ... on T1 {
      ... on I2 @test {
        id
      }
    }
    ... on T2 {
      ... on I2 @test {
        id
      }
    }
  }
}
```
...but this T2 as parent type of an I2 is triggering the error since `canRebaseOn` is returning false.

Minimal reproduction of https://github.com/apollographql/federation/issues/2862

### DIFF
--- a/internals-js/src/operations.ts
+++ b/internals-js/src/operations.ts
@@ -2239,7 +2239,7 @@ function makeSelection(parentType: CompositeType, updates: SelectionUpdate[], fr
   const element = updateElement(first).rebaseOnOrError(parentType);
   const subSelectionParentType = element.kind === 'Field' ? element.baseType() : element.castedType();
   if (!isCompositeType(subSelectionParentType)) {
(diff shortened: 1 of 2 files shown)


dariuszkuc:
Superseded by #2875

PR #2879, opened 2023-12-06
Progressive `@override` (composition only)

trevor-scheer:
Bump fed spec to v2.7
Bump join spec to v0.4

Add new optional `label` arg to `@override` which is a `String`.
Capture label in the supergraph via the new `@join__field` arg `overrideLabel` so these values can be used during query graph creation and query planning.

### DIFF
--- a/composition-js/src/merging/merge.ts
+++ b/composition-js/src/merging/merge.ts
@@ -121,7 +121,7 @@ class FieldMergeContext {
     return this._props[idx].overrideWithUnknownTarget;
   }
 
-  overrideLabelForIdx(idx: number) {
+  overrideLabel(idx: number) {
     return this._props[idx].overrideLabel;
   }
 
@@ -1614,7 +1614,7 @@ class Merger {
         type: allTypesEqual ? undefined : source.type?.toString(),
         external: external ? true : undefined,
         usedOverridden: usedOverridden ? true : undefined,
-        overrideLabel: mergeContext.overrideLabelForIdx(idx),
+        overrideLabel: mergeContext.overrideLabel(idx),
       });
     }
   }
--- a/composition-js/src/merging/merge.ts
+++ b/composition-js/src/merging/merge.ts
@@ -92,6 +92,7 @@ type FieldMergeContextProperties = {
   usedOverridden: boolean,
   unusedOverridden: boolean,
   overrideWithUnknownTarget: boolean,
+  overrideLabel: string | undefined,
 }
 
 // for each source, specify additional properties that validate functions can set
@@ -99,7 +100,13 @@ class FieldMergeContext {
   _props: FieldMergeContextProperties[];
 
   constructor(sources: unknown[]) {
-    this._props = (new Array(sources.length)).fill(true).map(_ => ({ usedOverridden: false, unusedOverridden: false, overrideWithUnknownTarget: false}));
+    this._props = (
+      new Array(sources.length)).fill(true).map(_ => ({
+        usedOverridden: false,
+        unusedOverridden: false,
+        overrideWithUnknownTarget: false,
+        overrideLabel: undefined,
+      }));
   }
 
   isUsedOverridden(idx: number) {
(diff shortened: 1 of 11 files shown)


composition-js/src/merging/merge.ts
```diff
@@ -1248,6 +1270,12 @@ class Merger {
               overriddenSubgraphASTNode,
             ));
           }
+
+          // capture an override label if it exists
```
sachindshinde, line 1274: We might want to do some kind of validation on the `label` here (it's easier to relax validations in the future than to add them). For example, we could require that either (1) the label has to start with a letter and consist of alphanumeric+dashes+underscore(+periods+slashes+colons if we feel like it), or (2) the label has to be of the form `percent(x)` where `x` is a number between 0 and 100 inclusive (maybe something like [js-big-decimal](https://www.npmjs.com/package/js-big-decimal) could help here). If you have any other syntax in mind for `label` we could use that too.

sachindshinde (changes requested):

composition-js/src/merging/merge.ts
```diff
@@ -114,6 +121,10 @@ class FieldMergeContext {
     return this._props[idx].overrideWithUnknownTarget;
   }
 
+  overrideLabelForIdx(idx: number) {
```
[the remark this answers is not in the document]
trevor-scheer: [9ace0a0](https://github.com/apollographql/federation/pull/2879/commits/9ace0a0a7bb65f04b33637fc15fb87c8c875564b)

composition-js/src/merging/merge.ts
trevor-scheer: Good call out. I don't _think_ we need a lib for this but let me know if you think I've missed anything in how we handle floats.
[cf27bcb](https://github.com/apollographql/federation/pull/2879/commits/cf27bcbea0eea036039563dc47721d66582742ab)

sachindshinde: It was mainly to handle cases where JS `Number`s lose precision and we don't faithfully copy the argument from subgraphs to supergraph, but if you're using 10 digits total (1-2 before the decimal point and 8 after), then JS `Number`s should be able to faithfully represent that.

internals-js/src/error.ts
```diff
@@ -513,6 +513,12 @@ const OVERRIDE_ON_INTERFACE = makeCodeDefinition(
   { addedIn: '2.3.0' },
 );
 
+const OVERRIDE_LABEL_INVALID = makeCodeDefinition(
+  'OVERRIDE_LABEL_INVALID',
+  'The @override directive `label` argument must match the pattern /^[a-zA-Z][a-zA-Z0-9_\-:./]*$/ or /^percent\(([0-9]|[1-9][0-9]|100)\)$/',
```
sachindshinde, line 518: Need to update the regex for the `percent` case here.

docs/source/errors.md
```diff
@@ -65,6 +65,7 @@ The following errors might be raised during composition:
 | `ONLY_INACCESSIBLE_CHILDREN` | A type visible in the API schema has only @inaccessible children. | 2.0.0 |  |
 | `OVERRIDE_COLLISION_WITH_ANOTHER_DIRECTIVE` | The @override directive cannot be used on external fields, nor to override fields with either @external, @provides, or @requires. | 2.0.0 |  |
 | `OVERRIDE_FROM_SELF_ERROR` | Field with `@override` directive has "from" location that references its own subgraph. | 2.0.0 |  |
+| `OVERRIDE_LABEL_INVALID` | The @override directive `label` argument must match the pattern /^[a-zA-Z][a-zA-Z0-9_-:./]*$/ or /^percent(([0-9]|[1-9][0-9]|100))$/ | 2.7.0 |  |
```
sachindshinde, line 68: Will need to regenerate this file after updating `error.ts`.

sachindshinde (changes requested):
Just some minor things below. (You'll also want to update the changeset regarding testing strategy, as noted [here](https://github.com/apollographql/federation/pull/2879#discussion_r1431512857).)

docs/source/errors.md
trevor-scheer: c0a3638452b8fbcf7407c847e60a8ec8d120451d

internals-js/src/error.ts
trevor-scheer: c0a3638452b8fbcf7407c847e60a8ec8d120451d

PR #2880, opened 2023-12-07
Add a limit to the number of options for a selection

Geal:
In some cases, the number of options generated for a selection can grow to a huge number, which results in a lot of time spent pruning options and creating plans. The new `pathsLimit` option will put a limit on that number of options


query-planner-js/src/config.ts
```diff
@@ -84,6 +84,16 @@ export type QueryPlannerConfig = {
...
+
+  /**
+   * Before creating query plans, we need to evaluate all the possible paths to access every field:
+   * through which subgraph providing it, through which type in union or type implementing an
+   * interface. The number of paths generated can grow if the schema or query are complex,
+   * and that will increase the time spent planning.
+   *
+   * The default value is set to 10000
+   */
+  pathsLimit: number
```
sachindshinde, line 96: You'll want to use `?` here so that users won't be required to provide the property (and also to not break the API).
```suggestion
  pathsLimit?: number
```
Also could you place this property under the `debug` section for now? (Could potentially move it out later, but for now I'd rather we don't make stability guarantees around it.)

query-planner-js/src/config.ts
```diff
@@ -83,7 +83,19 @@ export type QueryPlannerConfig = {
...
     maxEvaluatedPlans?: number,
+
+      /**
+   * Before creating query plans, we need to evaluate all the possible paths to access every field:
+   * through which subgraph providing it, through which type in union or type implementing an
+   * interface. The number of paths generated can grow if the schema or query are complex,
+   * and that will increase the time spent planning.
+   *
+   * The default value is set to 10000
+   */
```
sachindshinde, line 94: Adjusted the indentation and phrasing a bit.
```suggestion
    /**
     * Before creating query plans, for each path of fields in the query we compute all the
     * possible options to traverse that path via the subgraphs. Multiple options can arise because
     * fields in the path can be provided by multiple subgraphs, and abstract types (i.e. unions
     * and interfaces) returned by fields sometimes require the query planner to traverse through
     * each constituent object type. The number of options generated in this computation can grow
     * large if the schema or query are sufficiently complex, and that will increase the time spent
     * planning.
     * 
     * This config allows specifying a per-path limit to the number of options considered. If any
     * path's options exceeds this limit, query planning will abort and the operation will fail.
     *
     * The default value is null, which specifies no limit.
     */
```

sachindshinde (changes requested):
Mainly mirroring some feedback from standup and some tweaks to wording in the doc string. (Also for some reason I don't understand, we only run `prettier`/formatting on the test files and not the main source files, so there's some manual whitespace fixes below as well.)

sachindshinde:
Also you'll need to add a changeset file to this PR, there's instructions for what changesets are and how to add one in this comment https://github.com/apollographql/federation/pull/2880#issuecomment-1845127174 .

Specifically you'll want to click the [second link](https://github.com/apollographql/federation/new/geal/options-limit?filename=.changeset/fast-cherries-tan.md&value=---%0A%22%40apollo%2Fquery-planner%22%3A%20patch%0A---%0A%0AAdd%20a%20limit%20to%20the%20number%20of%20options%20for%20a%20selection%0A), which lets you add/commit the file via the GitHub UI.

Geal:
> Also for some reason I don't understand, we only run prettier/formatting on the test files and not the main source files, so there's some manual whitespace fixes below as well

There may be some automatic formatter somewhere in my vscode. I don't remember setting it up for typescript though

Geal:
alright, I think this can be merged now

sachindshinde (approved):
LGTM, merging it in

PR #2881, opened 2023-12-07
reduce the number of pruning operations

Geal:
Instead of checking a path against the entire list of paths every time, we only check against the list of already accepted paths.

### DIFF
--- a/query-planner-js/src/buildPlan.ts
+++ b/query-planner-js/src/buildPlan.ts
@@ -610,7 +610,7 @@ class QueryPlanningTraversal<RV extends Vertex> {
 
       const pruned: ClosedBranch<RV> = [];
       for (const toCheck of branch) {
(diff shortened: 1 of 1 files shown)


Geal:
there's another potnetial optimization here: making`ownPathIds` a Set instead of an array:
https://github.com/apollographql/federation/blob/main/query-graphs-js/src/graphPath.ts#L153

Geal:
I am not sure this PR is correct. In `markOverriding`:
https://github.com/apollographql/federation/blob/3f7392b84f8b626b248b59ce81f193d0f0272045/query-graphs-js/src/graphPath.ts#L753-L768

we add the `newId` to `thisPath.ownPathIds` but we add it to `overridingPathIds` in the paths under `otherOptions`. So I don't think there's a transitive dependency here where A overrides B and B overrides C would mean A overrides C

PR #2883, opened 2023-12-07
Avoid `>=` comparison for `FeatureVersion` objects

benjamn:
This won't be a problem in practice until we have `major` or `minor` version numbers greater than 9, because `toString`-based comparison of single digits works as expected.

I wish TypeScript could warn about implicit conversions during evaluation of expressions like `a >= b`, but I haven't found a way to enable that. Maybe a lint rule of some sort?

### DIFF
--- a/internals-js/src/specs/__tests__/coreSpec.test.ts
+++ b/internals-js/src/specs/__tests__/coreSpec.test.ts
@@ -217,6 +217,47 @@ class TestFeatureDefinition extends FeatureDefinition {
   }
 }
 
+describe('FeatureVersion', () => {
+  it('toString-based comparisons', () => {
+    const v2_3 = new FeatureVersion(2, 3);
+    const v10_0 = new FeatureVersion(10, 0);
+
+    expect(v2_3.toString()).toBe('v2.3');
+    expect(v10_0.toString()).toBe('v10.0');
+
+    // Operators like <, <=, >, and >= use lexicographic comparison on
+    // version.toString() strings, but do not perform numeric lexicographic
+    // comparison of the major and minor numbers, so 'v10...' < 'v2...' and the
+    // following comparisons fail to produce intuitive results.
+    expect(() => {
+      expect(v2_3 < v10_0).toBe(true);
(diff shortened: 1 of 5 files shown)


internals-js/src/specs/__tests__/coreSpec.test.ts
```diff
@@ -217,6 +217,47 @@ class TestFeatureDefinition extends FeatureDefinition {
...
+    // Operators like <, <=, >, and >= use lexicographic comparison on
+    // version.toString() strings, but do not perform numeric lexicographic
+    // comparison of the major and minor numbers, so 'v10...' < 'v2...' and the
+    // following comparisons fail to produce intuitive results.
+    expect(() => {
+      expect(v2_3 < v10_0).toBe(true);
+      expect(v2_3 <= v10_0).toBe(true);
+      expect(v2_3 > v10_0).toBe(false);
+      expect(v2_3 >= v10_0).toBe(false);
+    }).toThrow();
```
benjamn, line 237: These are some tests you'd intuitively like to see passing, but they fail.

I don't see a way to make this syntax work in JavaScript without returning something very different from `toString` (note: not sure if possible, in general), which could be a breaking change for consumers of `toString`.

trevor-scheer: I understand these assertions are more for illustration, and I suppose you're already aware of this, but only the first one actually runs since it throws.

Just so someone doesn't look at this confusedly later, I might suggest either:
* comment the others out, or
* reconfigure this so that each assertion runs (maybe each with it's own `.toThrow()` or just flip all the booleans so the assertion is positive and leave the explanation up to the comment).

trevor-scheer (approved):
Nice catch!

internals-js/src/specs/__tests__/coreSpec.test.ts
benjamn: Thanks for the nudge @trevor-scheer… I think/hope https://github.com/apollographql/federation/pull/2883/commits/79297702cad94b6823a9beb4faa3089bfa757ffb makes the test clearer.

PR #2894, opened 2023-12-19
Support `@join__directive(graphs, name, args)` directives

benjamn:
This new `@join__directive` directive, like other `@join__` directives, is a mechanism for linking supergraph schema elements back to their originating subgraph elements.

In the case of `@join__directive`, the specific kind of subgraph schema elements to be linked are _directive applications_. In addition to its `graphs: [join__Graph!]!` argument, `@join__directive` represents the `name` and `args` of directive applications found in subgraphs, without literally reapplying the directives in the supergraph, as `@composeDirective` would do.

Since directive applications can appear on a variety of schema elements in subgraphs, `@join__directive` is allowed on a similar variety of locations in the supergraph schema (anywhere `@join__type` or `@join__field` can appear, plus on `schema` definitions).

In principle, this new `@join__directive` behavior can be enabled at the granularity of any `@link` specification URL, and is initially enabled for directives imported from the hypothetical `https://specs.apollo.dev/source` URL, as well as for any `@link` directives involved in those imports. This limited initial usage of `@join__directive` allows preserving/recovering accurate per-subgraph information about the original directive applications.

We can extend this behavior to other specification URLs in the future if it works and proves useful, and we might consider opening the behavior up to subgraph authors by allowing, say, `@composeDirective` to generate `@join__directive` applications instead of reapplying subgraph directives verbatim (as it does by default), perhaps with an optional `@composeDirective(name: "someDirective", strategy: "@join__directive")` argument.

One way `@join__directive` deviates from the rest of the `@join__` directives is in taking a plural `graphs: [join__Graph!]!` argument, rather than a single `graph: join__Graph!` argument. This API allows directives that are repeated exactly in different subgraphs (same name, deep-equal arguments) to be represented with a single `@join__directive` directive in the supergraph (with multiple graph names in the `graphs` list), which tends to save a bunch of repetition of `@join__directive` in the supergraph.

### DIFF
--- a/composition-js/src/__tests__/compose.test.ts
+++ b/composition-js/src/__tests__/compose.test.ts
@@ -4645,38 +4645,38 @@ describe('composition', () => {
 
   it('@source{API,Type,Field} directives', () => {
     const subgraphA = {
-      typeDefs: gql`
-extend schema
-  @link(url: "https://specs.apollo.dev/federation/v2.1", import: [
-    "@key"
-  ])
-  @link(url: "https://specs.apollo.dev/source/v0.1", import: [
-    "@sourceAPI"
-    "@sourceType"
-    "@sourceField"
-  ])
-  @sourceAPI(
-    name: "A"
-    http: { baseURL: "https://api.a.com/v1" }
-  )
-
-type Query {
-  resources: [Resource!]! @sourceField(
-    api: "A"
-    http: { GET: "/resources" }
-  )
-}
-
-type Resource @key(fields: "id") @sourceType(
-  api: "A"
-  http: { GET: "/resources/{id}" }
-  selection: "id description"
-) {
-  id: ID!
-  description: String!
-}
-      `,
       name: 'subgraphA',
+      typeDefs: gql`
+        extend schema
+          @link(url: "https://specs.apollo.dev/federation/v2.1", import: [
+            "@key"
+          ])
+          @link(url: "https://specs.apollo.dev/source/v0.1", import: [
+            "@sourceAPI"
+            "@sourceType"
+            "@sourceField"
+          ])
+          @sourceAPI(
+            name: "A"
+            http: { baseURL: "https://api.a.com/v1" }
+          )
+
+        type Query {
+          resources: [Resource!]! @sourceField(
+            api: "A"
+            http: { GET: "/resources" }
+          )
+        }
+
+        type Resource @key(fields: "id") @sourceType(
+          api: "A"
+          http: { GET: "/resources/{id}" }
+          selection: "id description"
+        ) {
+          id: ID!
+          description: String!
+        }
+      `,
     };
     const result = composeServices([subgraphA]);
     expect(result.errors ?? []).toEqual([]);
--- a/composition-js/src/__tests__/compose.test.ts
+++ b/composition-js/src/__tests__/compose.test.ts
@@ -76,6 +76,8 @@ describe('composition', () => {
         query: Query
       }
 
+      directive @join__directive(graphs: [join__Graph!], name: String!, args: join__DirectiveArguments) repeatable on SCHEMA | OBJECT | INTERFACE | UNION | ENUM | INPUT_OBJECT | SCALAR | FIELD_DEFINITION | INPUT_FIELD_DEFINITION
+
       directive @join__enumValue(graph: join__Graph!) repeatable on ENUM_VALUE
 
       directive @join__field(graph: join__Graph, requires: join__FieldSet, provides: join__FieldSet, type: String, external: Boolean, override: String, usedOverridden: Boolean) repeatable on FIELD_DEFINITION | INPUT_FIELD_DEFINITION
@@ -96,6 +98,8 @@ describe('composition', () => {
         V1 @join__enumValue(graph: SUBGRAPH2)
         V2 @join__enumValue(graph: SUBGRAPH2)
       }
+
+      scalar join__DirectiveArguments
 
       scalar join__FieldSet
 
@@ -224,6 +228,8 @@ describe('composition', () => {
         query: Query
       }
 
+      directive @join__directive(graphs: [join__Graph!], name: String!, args: join__DirectiveArguments) repeatable on SCHEMA | OBJECT | INTERFACE | UNION | ENUM | INPUT_OBJECT | SCALAR | FIELD_DEFINITION | INPUT_FIELD_DEFINITION
+
       directive @join__enumValue(graph: join__Graph!) repeatable on ENUM_VALUE
 
       directive @join__field(graph: join__Graph, requires: join__FieldSet, provides: join__FieldSet, type: String, external: Boolean, override: String, usedOverridden: Boolean) repeatable on FIELD_DEFINITION | INPUT_FIELD_DEFINITION
@@ -244,6 +250,8 @@ describe('composition', () => {
         V1 @join__enumValue(graph: SUBGRAPH2)
         V2 @join__enumValue(graph: SUBGRAPH2)
       }
+
+      scalar join__DirectiveArguments
 
       scalar join__FieldSet
 
@@ -2489,6 +2497,8 @@ describe('composition', () => {
         query: Query
       }
 
+      directive @join__directive(graphs: [join__Graph!], name: String!, args: join__DirectiveArguments) repeatable on SCHEMA | OBJECT | INTERFACE | UNION | ENUM | INPUT_OBJECT | SCALAR | FIELD_DEFINITION | INPUT_FIELD_DEFINITION
+
       directive @join__enumValue(graph: join__Graph!) repeatable on ENUM_VALUE
 
       directive @join__field(graph: join__Graph, requires: join__FieldSet, provides: join__FieldSet, type: String, external: Boolean, override: String, usedOverridden: Boolean) repeatable on FIELD_DEFINITION | INPUT_FIELD_DEFINITION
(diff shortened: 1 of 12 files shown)


benjamn:
Marking this PR as ready for review despite known/unknown remaining gaps (reversibility, additional tests, 100% code coverage, to name a few), because I'd love some directional feedback from the team.

composition-js/src/merging/merge.ts
```diff
@@ -2572,6 +2584,116 @@ class Merger {
       // Because we rename all root type in subgraphs to their default names, we shouldn't ever have incompatibilities here.
       assert(!isIncompatible, () => `Should not have incompatible root type for ${rootKind}`);
     }
+    this.addJoinDirectiveDirectives(sources, dest);
+  }
+
+  private normalizeSpecIdentityURL(identityURL: string): string {
```
clenfest, line 2590: consider using FeatureURL.parse as I think it does what you need.

composition-js/src/merging/merge.ts
```diff
@@ -2572,6 +2584,116 @@ class Merger {
...
+    } = Object.create(null);
+
+    for (const [idx, source] of sources.entries()) {
+      if (!source) continue;
+      const graph = this.joinSpecName(idx);
+
+      // We compute this map only once per subgraph, as it takes time
+      // proportional to the size of the schema.
+      const linkImportIdentityURLMap =
+        this.computeMapFromImportedNamesToIdentityURLs(source.schema());
```
clenfest, line 2652: Are you really doing this once? Why not do this in the constructor?

composition-js/src/merging/merge.ts
```diff
@@ -2572,6 +2584,116 @@ class Merger {
...
+
+      for (const directive of source.appliedDirectives) {
+        const isRelevantLink =
```
clenfest, line 2655: Is there a reason we want to add @link? I'll probably get to it later on, but I would avoid this if we can.

composition-js/src/__tests__/compose.test.ts
```diff
@@ -4630,4 +4642,75 @@ describe('composition', () => {
     const authenticatedDirectiveExists = schema.directives().find(d => d.name === 'authenticated');
     expect(authenticatedDirectiveExists).toBeUndefined();
   });
+
```
clenfest, line 4645: formatting is off.

internals-js/src/specs/sourceSpec.ts
```diff
@@ -0,0 +1,146 @@
...
+    return this.directive(schema, 'sourceType')!;
+  }
+
+  sourceFieldDirective(schema: Schema) {
+    return this.directive(schema, 'sourceField')!;
+  }
+}
+
+export const SOURCE_VERSIONS = new FeatureDefinitions<SourceSpecDefinition>(sourceIdentity)
+  .add(new SourceSpecDefinition(new FeatureVersion(0, 1)));
```
clenfest, line 144: we want a second argument on the SourceSpecDefinition that will tie this to a specific federation version. When we release, this will be `2.7`, but we haven't bumped elsewhere yet.

internals-js/src/specs/joinSpec.ts
```diff
@@ -126,6 +133,25 @@ export class JoinSpecDefinition extends FeatureDefinition {
       joinEnumValue.addArgument('graph', new NonNullType(graphEnum));
     }
 
+    const joinDirective = this.addDirective(schema, 'directive').addLocations(
+      // Allow @join__directive at any location where @joinType or @joinField
+      // can appear, as well as on the schema element.
+      ...new Set([
+        DirectiveLocation.SCHEMA,
+        ...joinType.locations,
```
clenfest, line 141: Do we really want to support it on all these locations? The locations for source are much more narrow.

internals-js/src/specs/joinSpec.ts
```diff
@@ -192,6 +218,10 @@ export class JoinSpecDefinition extends FeatureDefinition {
     return this.directive(schema, 'graph')!;
   }
 
+  directiveDirective(schema: Schema): DirectiveDefinition<JoinDirectiveArguments> {
```
clenfest, line 221: good old directivedirective :)

internals-js/src/specs/sourceSpec.ts
```diff
@@ -0,0 +1,146 @@
+import { DirectiveLocation, GraphQLError } from 'graphql';
+import { FeatureDefinition, FeatureDefinitions, FeatureUrl, FeatureVersion } from "./coreSpec";
+import {
+  Schema,
+  NonNullType,
+  InputObjectType,
+  InputFieldDefinition,
+  ListType,
+} from "../definitions";
```
clenfest, line 9: Nit, but it would be nice to be consistent about single/double quotes within a file.

internals-js/src/specs/sourceSpec.ts
```diff
@@ -0,0 +1,146 @@
...
+      DirectiveLocation.INTERFACE,
+      // TODO Allow @sourceType on unions, similar to interfaces?
+      // DirectiveLocation.UNION,
+    );
+    sourceType.repeatable = true;
+    sourceType.addArgument('api', new NonNullType(schema.stringType()));
+
+    const URLPathTemplate = this.addScalarType(schema, 'URLPathTemplate');
+    const JSONSelection = this.addScalarType(schema, 'JSONSelection');
+    const JSON = this.addScalarType(schema, 'JSON');
```
clenfest, line 76: What would happen if a JSON definition already existed in the schema? Would it be safer to namespace the scalars to avoid conflicting definintions?

composition-js/src/__tests__/compose.test.ts
```diff
@@ -4630,4 +4642,75 @@ describe('composition', () => {
...
+  it('@source{API,Type,Field} directives', () => {
+    const subgraphA = {
+      typeDefs: gql`
```
clenfest, line 4648: Add some more tests for some of the special cases? i.e. multiple graphs with the same args, etc.

clenfest (commented):
Also, make sure we're merging to `next`, rather than `main`.

clenfest:
Missed that you should add the directives in `FederationMetadata` in `federation.ts`. You need to add getters for the directives as well as adding them in `allFederationDirectives`.

internals-js/src/specs/sourceSpec.ts
benjamn: I think I used to have bigger plans for this `JSON` than the one place it's used now: the `typeMap` field of the `KeyTypeMap` input object type. Given it's only used there, I'll adjust the scalar typename to match that use case.

composition-js/src/merging/merge.ts
benjamn: https://github.com/apollographql/federation/pull/2894/commits/2b6228fb83e8e15927fbac636ccbd98eb41c65e2

composition-js/src/__tests__/compose.test.ts
benjamn: https://github.com/apollographql/federation/pull/2894/commits/e4076f039b09637097b3e314740e06dbeb27935b

internals-js/src/specs/sourceSpec.ts
benjamn: https://github.com/apollographql/federation/pull/2894/commits/928b2942028fcd0709d6a537b4443c061690da27

internals-js/src/specs/sourceSpec.ts
benjamn: https://github.com/apollographql/federation/pull/2894/commits/d6f7e8a5815ca65fc7150aaafcc38c8cf5baa810

internals-js/src/specs/joinSpec.ts
benjamn: https://github.com/apollographql/federation/pull/2894/commits/8127b6983975e49dd090f3df2bb7243c2d85de23

internals-js/src/specs/sourceSpec.ts
benjamn: https://github.com/apollographql/federation/pull/2894/commits/d10a518a80146346d5213681adafe91ed2f56542

composition-js/src/merging/merge.ts
benjamn: https://github.com/apollographql/federation/pull/2894/commits/d803501c4085eddc0717881d6e90a798e034a0b2

composition-js/src/__tests__/compose.test.ts
benjamn: https://github.com/apollographql/federation/pull/2894/commits/59fd58f994a1cf106bf9bd478f0b9bb511e7f68a

composition-js/src/merging/merge.ts
benjamn: I think it's important to preserve the `@link` associated with importing any of the directives we're wrapping with `@join__directive`, because the `@link` tells us whether/how the imported directives have been renamed or prefixed. Without that information, there'd be no way to tell which of the original `@source*` directives is intended when the supergraph SDL contains `@join__directive(graphs: [...], name: "sourceRenamed", args: {...})`.

An alternative would be to disallow renaming the `@source*` directives in subgraph schemas, but that's a surprising gotcha to impose, and not consistent with how `@link` typically works.

Including `@link` is something @sachindshinde recommended, and I agree it's worthwhile.

clenfest (approved):

PR #2898, opened 2023-12-22
keep inputRewrites when cloning a FetchGroup

Geal:
if we do not keep them, they may disappear from the final fetch node

### DIFF
--- a/query-planner-js/src/__tests__/buildPlan.interfaceObject.test.ts
+++ b/query-planner-js/src/__tests__/buildPlan.interfaceObject.test.ts
@@ -697,7 +697,7 @@ it('handles @interfaceObject in nested entity', () => {
   `);
 });
 
-it('handles @interfaceObject with complex graphs', () => {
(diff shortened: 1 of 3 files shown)


query-planner-js/src/__tests__/buildPlan.interfaceObject.test.ts
```diff
@@ -696,3 +696,151 @@ it('handles @interfaceObject in nested entity', () => {
...
+      }
+    `,
+  };
+
+  const [api, queryPlanner] = composeAndCreatePlanner(
+    subgraph1,
+    subgraph2,
+    subgraph3,
+    subgraph4,
+  );
```
dariuszkuc, line 767: in order to trigger input rewrites we need to federate the call to `@interfaceObject` (S1 -> S2) and then in order to trigger dependency graph copy we need to have multiple query plan options (i.e. S2 -> S3 or S2 -> S4).

trevor-scheer (approved):

PR #2902, opened 2024-01-03
Progressive override (query planning / validation)

trevor-scheer:
TODO:
- [ ] Changeset comment https://github.com/apollographql/federation/pull/2879#discussion_r1431512857


sachindshinde (changes requested):
Sorry for the delays 🙇 , partial review here (still looking at composition validation logic).

composition-js/src/validate.ts
```diff
@@ -474,7 +479,16 @@ export class ValidationState {
       return { error: satisfiabilityError(newPath, this.subgraphPaths.map((p) => p.path), deadEnds) };
     }
 
-    const updatedState = new ValidationState(newPath, newSubgraphPaths);
+    // If the edge has an override condition, we should capture it in the state so
+    // that we can ignore later edges that don't satisfy the condition.
+    const newOverrideCondition = supergraphEdge.overrideCondition ? {
```
sachindshinde, line 484: Earlier in this function, there's a call to `advancePathWithTransition()`. That code has to take in this `newOverrideCondition`, so that it can take the corresponding virtual edge in the subgraph, so you'll need to move this `newOverrideCondition` definition to before that `advancePathWithTransition()` call.

Within `advancePathWithTransition()`, you'll pass `newOverrideCondition` to `advancePathWithDirectTransition()`. Then in `advancePathWithDirectTransition()`, specifically [here](https://github.com/apollographql/federation/blob/c83c25a430327da5caa0b04db79b575420e88e50/query-graphs-js/src/graphPath.ts#L1577), you'll need to:
1. Check if the subgraph edge has an override condition.
2. If it does, assert that the `newOverrideCondition` has a value for the corresponding label (this is more just sanity checking).
3. If the `XOR` of the edge's override condition boolean and the `newOverrideCondition`'s boolean is true, then `continue`/skip the rest of that `for` loop.

composition-js/src/validate.ts
```diff
@@ -401,32 +402,36 @@ export class ValidationContext {
...
     // Path in the supergraph corresponding to the current state.
     public readonly supergraphPath: RootPath<Transition>,
     // All the possible paths we could be in the subgraph.
-    public readonly subgraphPaths: TransitionPathWithLazyIndirectPaths<RootVertex>[]
+    public readonly subgraphPaths: TransitionPathWithLazyIndirectPaths<RootVertex>[],
+    // When we encounter an `@override`n field with a label condition, we record
+    // its value (T/F) as we traverse the graph. This allows us to ignore paths
+    // that can never be taken by the query planner (i.e. a path where the
+    // condition is T in one case and F in another).
+    public selectedOverrideConditions: Record<string, boolean> = {},
```
sachindshinde, line 417: If we're using JS objects here instead of `Map`s, we should use `Object.create(null)` instead of `{}`.

composition-js/src/validate.ts
```diff
@@ -651,6 +665,16 @@ class ValidationTraversal {
         continue;
       }
 
+      // `state.selectedOverrideConditions` indicates the labels (and their
```
sachindshinde, line 668: Earlier in this function, you'll notice a call to `isSupersetOrEqual(currentSources, previousSources)`; that is the part of the code that stops us from recursing infinitely into the API query graph.

That will need some changes since we're adding new state in the form of `selectedOverrideConditions`. In particular, the current type of `previousVisits` is:
```typescript
previousVisits: QueryGraphState<string[][]>;
```
it would need to change to:
```typescript
previousVisits: QueryGraphState<{ sources: string[], selectedOverrideConditions: Record<string, boolean> }[]>;
```
So you'd push both the `currentSources` and the current `selectedOverrideCondition` onto this instead of just `currentSources`.

The other change is that in addition to `isSupersetOrEqual(currentSources, previousSources)` needing be true to early `return`, it must also be the case that the current `selectedOverrideCondition` is a superset of the previous `selectedOverrideCondition` (with the booleans matching for the labels in common).

sachindshinde (changes requested):
Here's the rest of the review, it's mainly about composition validation.

internals-js/src/extractSubgraphsFromSupergraph.ts
```diff
@@ -418,15 +418,26 @@ function extractObjOrItfContent(args: ExtractArguments, info: TypeInfo<ObjectTyp
...
+          // If an override label is present, either subgraph might be
+          // responsible for resolving the field so we need to preserve it in
+          // both places.
+          if (joinFieldArgs.overrideLabel) {
+            const overriddenFrom = Array.from(subgraphsInfo.values()).find(
+              ({ subgraph }) => subgraph.name === joinFieldArgs.override
+            );
+            assert(overriddenFrom, `Could not find overridden subgraph "${joinFieldArgs.override}" for field "${type.name}.${field.name}"`);
+            addSubgraphField({ ...overriddenFrom, field, isShareable, joinFieldArgs });
+          }
```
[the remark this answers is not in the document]
trevor-scheer: I addressed this comment as well as a couple others via this commit. Let me know what you think.
[4797410](https://github.com/apollographql/federation/pull/2902/commits/4797410e08bf09544634af03ae9b19fbb17a9afb)

query-graphs-js/src/querygraph.ts
```diff
@@ -163,8 +168,24 @@ export class Edge {
...
+    if (!this.overrideCondition && this.transition instanceof FieldCollection) {
+      const [overrideApplication] = this.transition.definition.appliedDirectivesOf('override') ?? [];
+      const label = overrideApplication?.arguments().label;
+      if (label) {
+        this.overrideCondition = {
+          label,
+          condition: this.head.source !== overrideApplication?.arguments().from,
+        };
+      }
+    }
```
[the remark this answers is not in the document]
trevor-scheer: [4797410](https://github.com/apollographql/federation/pull/2902/commits/4797410e08bf09544634af03ae9b19fbb17a9afb)

query-graphs-js/src/querygraph.ts
```diff
@@ -516,23 +552,26 @@ export class QueryGraphState<VertexState, EdgeState = undefined> {
...
+  return buildGraphInternal(name, schema, false, undefined, virtualEdges);
 }
 
 function buildGraphInternal(
   name: string,
   schema: Schema,
   addAdditionalAbstractTypeEdges: boolean,
-  supergraphSchema?: Schema
+  supergraphSchema?: Schema,
+  virtualEdges?: VirtualEdges,
```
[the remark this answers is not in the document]
trevor-scheer: [bdad1a9](https://github.com/apollographql/federation/pull/2902/commits/bdad1a9efa791bd21128985d6ad16f41b161e9b7)

query-graphs-js/src/querygraph.ts
```diff
@@ -1195,9 +1256,30 @@ class GraphBuilderFromSchema extends GraphBuilder {
...
+        const otherEdge = this.edges(head).find(
+          e => {
+            return e.head.type.name === typeName
+              && e.transition.kind === 'FieldCollection'
+              && e.transition.definition.name === fieldName
+          });
+        assert(otherEdge, () => `Edge for field ${fieldName} not found in graph`);
+        otherEdge.overrideCondition = { label, condition: false };
+      }
+    }
```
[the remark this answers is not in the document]
trevor-scheer: Makes sense. Good catch! [bdad1a9](https://github.com/apollographql/federation/pull/2902/commits/bdad1a9efa791bd21128985d6ad16f41b161e9b7)

query-graphs-js/src/querygraph.ts
```diff
@@ -928,10 +987,11 @@ class GraphBuilder {
     return this.rootVertices.get(kind);
   }
 
-  addEdge(head: Vertex, tail: Vertex, transition: Transition, conditions?: SelectionSet) {
+  addEdge(head: Vertex, tail: Vertex, transition: Transition, conditions?: SelectionSet, label?: string) {
```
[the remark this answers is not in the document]
trevor-scheer: [bdad1a9](https://github.com/apollographql/federation/pull/2902/commits/bdad1a9efa791bd21128985d6ad16f41b161e9b7)

composition-js/src/validate.ts
trevor-scheer: I may have misunderstood, or maybe it's fine as is, you tell me. I passed all conditions rather than just the new one.
[dbb26c1](https://github.com/apollographql/federation/pull/2902/commits/dbb26c1b11cf6c3f10490c83121a0df9504078c4)

composition-js/src/validate.ts
trevor-scheer: Switched to map in [dbb26c1](https://github.com/apollographql/federation/pull/2902/commits/dbb26c1b11cf6c3f10490c83121a0df9504078c4)

composition-js/src/validate.ts
trevor-scheer: Introducing this started failing my fpf test with some errors. I experimented a bit with that `continue`, by doing nothing, returning an unadvanceable, and finally trying out a different "bypass" approach, all of which cause different errors that I'm unsure about. Maybe we can take a look tomorrow.

This commit is interesting.
https://github.com/apollographql/federation/pull/2902/commits/d1a5bf719e5e170cbe5e2532327f02620f7950c0

query-graphs-js/src/querygraph.ts
```diff
@@ -163,8 +168,24 @@ export class Edge {
      * Outside of keys, @requires also rely on conditions.
      */
     conditions?: SelectionSet,
+    public overrideCondition?: OverrideCondition,
```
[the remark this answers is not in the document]
trevor-scheer: 5eac6159

query-graphs-js/src/querygraph.ts
```diff
@@ -194,7 +215,14 @@ export class Edge {
...
     }
-    return this._conditions ? `${this._conditions} ⊢ ${this.transition}` : this.transition.toString();
+    const overrideConditionString = this.overrideCondition
+      ? ` ⊢ ${this.overrideCondition.label}(${this.overrideCondition.condition})`
+      : '';
+    return (
+      this._conditions
+        ? `${this._conditions} ⊢ ${this.transition}`
+        : this.transition.toString()
+    ) + overrideConditionString;
```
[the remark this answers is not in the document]
trevor-scheer: 5eac6159

query-planner-js/src/buildPlan.ts
```diff
@@ -3470,6 +3482,7 @@ function computeRootParallelBestPlan(
     hasDefers,
     parameters.root.rootKind,
     defaultCostFunction,
+    // TODO: trevor - is the reuse of this emptyContext problematic?
```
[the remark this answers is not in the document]
trevor-scheer: 5eac6159

query-planner-js/src/buildPlan.ts
```diff
@@ -2929,6 +2935,11 @@ type PlanningParameters<RV extends Vertex> = {
   root: RV,
   inconsistentAbstractTypesRuntimes: Set<string>,
   config: Concrete<QueryPlannerConfig>,
+  overriddenLabels: Set<string>,
+}
+
+interface BuildQueryPlanOptions {
+  overriddenLabels?: Set<string>,
```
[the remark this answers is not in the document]
trevor-scheer: 5eac6159

query-graphs-js/src/graphPath.ts
```diff
@@ -877,8 +877,13 @@ export function traversePath(
...
 
 // Note that ConditionResolver are guaranteed to be only called for edge with conditions.
-export type ConditionResolver =
-  (edge: Edge, context: PathContext, excludedDestinations: ExcludedDestinations, excludedConditions: ExcludedConditions) => ConditionResolution;
+export type ConditionResolver = (
+  edge: Edge,
+  context: PathContext,
+  excludedDestinations: ExcludedDestinations,
+  excludedConditions: ExcludedConditions,
+  overriddenLabels: Set<string>
```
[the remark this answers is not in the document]
trevor-scheer: 3005ee6a

internals-js/src/extractSubgraphsFromSupergraph.ts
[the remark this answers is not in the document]
trevor-scheer: Follow up in [8ba6cce](https://github.com/apollographql/federation/pull/2902/commits/8ba6ccea89766aa1f4cde9d4682d39429cf7d72a)

composition-js/src/validate.ts
trevor-scheer: e8c8a0bb

composition-js/src/validate.ts
trevor-scheer: Resolved by Sachin in [674bb17](https://github.com/apollographql/federation/pull/2902/commits/674bb17c4487ea78ae8230c7321e0f0f40f66bfd)

query-planner-js/src/buildPlan.ts
```diff
@@ -3059,6 +3080,7 @@ export class QueryPlanner {
       statistics,
       inconsistentAbstractTypesRuntimes: this.inconsistentAbstractTypesRuntimes,
       config: this.config,
+      overrideConditions: options?.overrideConditions ?? new Map(),
```
sachindshinde, line 3083: One caveat to note here is that if the user forgets to provide a label, then certain fields will be impossible to traverse (and query planning will be fail). It may be good to add code here to manually set such labels to be `false` if not present. (You'll want to precompute all the labels in a supergraph schema inside the constructor for `QueryPlanner`, by looking at the applications of the `@join__field directive.)

query-graphs-js/src/querygraph.ts
```diff
@@ -748,6 +798,49 @@ function federateSubgraphs(supergraph: Schema, subgraphs: QueryGraph[]): QueryGr
...
+      const edges = subgraph.outEdges(vertex);
+      for (const edge of edges) {
+        if (
+          edge.transition.kind === "FieldCollection"
+          && edge.transition.definition.name === field.name
+        ) {
+          const head = copyPointers[i].copiedVertex(vertex);
+          const copiedEdge = builder.edge(head, edge.index);
+
+          // TODO: from sachin - should this only be applied in 2 places?
```
sachindshinde, line 834: Still need to address this comment (this `overrideCondition` should be applied in two places, the overriding subgraph and the overridden subgraph).

query-graphs-js/src/querygraph.ts
```diff
@@ -748,6 +798,49 @@ function federateSubgraphs(supergraph: Schema, subgraphs: QueryGraph[]): QueryGr
...
+    const label = application.arguments().label;
+    if (!label) continue;
+
+    const field = application.parent;
+    assert(field instanceof NamedSchemaElement, () => `@override should have been on a field, got ${field}`)
+    const type = field.parent;
+
+    for (const [i, subgraph] of subgraphs.entries()) {
+      const [vertex, ...unexpectedAdditionalVertices] = subgraph.verticesForType(type.name);
+      assert(vertex && unexpectedAdditionalVertices.length === 0, () => `Subgraph ${i} should have exactly one vertex for type ${type.name}`);
```
sachindshinde, line 824: (Somewhat unrelated since this section needs updating, but not every subgraph will have the type.)

query-graphs-js/src/graphPath.ts
```diff
@@ -1575,6 +1600,19 @@ function advancePathWithDirectTransition<V extends Vertex>(
       continue;
     }
 
+    if (edge.overrideCondition) {
+      const isOverridden = overrideConditions.get(edge.overrideCondition.label) === true;
```
sachindshinde, line 1604: So this code technically deals with empty `Map`s by assuming that no label means `false`. Which sounds fine at first glance, but other code that checks for satisfiability doesn't behave like this (and will instead always return `false` for override condition satisfiability if the `Map` is empty).

It would be better to ensure all labels are in `Map` at the start of `buildQueryPlan()` as mentioned in the previous comment.

query-graphs-js/src/graphPath.ts
```diff
@@ -1633,6 +1671,7 @@ function advancePathWithDirectTransition<V extends Vertex>(
         // it uses @interfaceObject on an interface of that implementation.
         details = `cannot find implementation type "${fieldTypeName}" (supergraph interface "${path.tail.type.name}" is declared with @interfaceObject in "${subgraph}")`;
       } else {
+        // TODO: from sachin - this is the place to "pretend" the field doesn't exist
```
sachindshinde, line 1674: You can remove this comment since it's handled above now.

query-graphs-js/src/graphPath.ts
```diff
@@ -2314,17 +2369,24 @@ function advanceWithOperation<V extends Vertex>(
...
           debug.groupEnd(() => `No edge for field ${field} on object type ${currentType}`);
           return { options: undefined };
         }
+        if (edge.overrideCondition) {
+          const { label } = edge.overrideCondition;
+          if (!edge.satisfiesOverrideConditions(overrideConditions)) {
+            debug.groupEnd(() => `Ignoring edge due to override label "${label}" in subgraph ${edge.head.source} on field ${field} on object type ${currentType}`);
+            return { options: undefined };
+          }
+        }
```
sachindshinde, line 2383: This block shouldn't be needed, since satisfiability is checked inside `nextEdgeForField()`

composition-js/src/validate.ts
```diff
@@ -623,11 +665,15 @@ class ValidationTraversal {
...
     const vertex = state.supergraphPath.tail;
-    const currentSources = state.currentSubgraphNames();
-    const previousSeenSources = this.previousVisits.getVertexState(vertex);
-    if (previousSeenSources) {
-      for (const previousSources of previousSeenSources) {
-        if (isSupersetOrEqual(currentSources, previousSources)) {
+
+    const currentVertexVisit: VertexVisit = {
+      subgraphs: state.currentSubgraphNames(),
+      overriddenLabels: state.selectedOverrideConditions
```
sachindshinde, line 671: nit: `overrideLabels` is the old name back from when this was a `Set`, would be good to update it since it's a `Map` now.

composition-js/src/validate.ts
```diff
@@ -474,7 +491,20 @@ export class ValidationState {
...
+      newOverrideConditions.set(
+        supergraphEdge.overrideCondition.label,
+        supergraphEdge.overrideCondition.condition
+      );
+    }
+    const updatedState = new ValidationState(
+      newPath,
+      newSubgraphPaths,
+      newOverrideConditions,
+    );
```
sachindshinde, line 507: These `newOverrideConditions` need to be computed prior to the call to `advancePathWithTransition()`, and they should be passed into `advancePathWithTransition()` instead of `this.selectedOverrideConditions`.

sachindshinde (changes requested):

composition-js/src/validate.ts
```diff
@@ -651,6 +697,21 @@ class ValidationTraversal {
...
 
+      // `state.selectedOverrideConditions` indicates the labels (and their
+      // respective conditions) that we've selected so far in our traversal
+      // (i.e. "foo" -> true). There's no need to validate edges that share the
+      // same label with the opposite condition since they're unreachable during
+      // query planning.
+      if (
+        edge.overrideCondition
+        && state.selectedOverrideConditions.has(edge.overrideCondition.label)
+        && edge.satisfiesOverrideConditions(state.selectedOverrideConditions)
```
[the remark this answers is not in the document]
trevor-scheer: [2391f43](https://github.com/apollographql/federation/pull/2902/commits/2391f43bfcb9e698f38ee7da77dee9696076c82c)

query-graphs-js/src/graphPath.ts
trevor-scheer: Agreed. As it is this is un-updated remains from the `Set` approach.
[2391f43](https://github.com/apollographql/federation/pull/2902/commits/2391f43bfcb9e698f38ee7da77dee9696076c82c)

query-graphs-js/src/graphPath.ts
trevor-scheer: [2391f43](https://github.com/apollographql/federation/pull/2902/commits/2391f43bfcb9e698f38ee7da77dee9696076c82c)

query-graphs-js/src/graphPath.ts
trevor-scheer: [2391f43](https://github.com/apollographql/federation/pull/2902/commits/2391f43bfcb9e698f38ee7da77dee9696076c82c)

composition-js/src/validate.ts
trevor-scheer: [2391f43](https://github.com/apollographql/federation/pull/2902/commits/2391f43bfcb9e698f38ee7da77dee9696076c82c)

query-graphs-js/src/querygraph.ts
trevor-scheer: I'm sure this can be cleaned up but it works and I'll take it for now. Will revisit last.
e44e0846

query-graphs-js/src/querygraph.ts
trevor-scheer: e44e0846

query-planner-js/src/buildPlan.ts
trevor-scheer: f7646134

query-graphs-js/src/querygraph.ts
```diff
@@ -807,67 +807,53 @@ function federateSubgraphs(supergraph: Schema, subgraphs: QueryGraph[]): QueryGr
...
-          condition: true,
+    const overrideDirectiveApplicationsWithLabel =
+      subgraphMetadata
+        .overrideDirective()
+        .applications()
+        .filter(a => a.arguments().label);
+
+    for (const application of overrideDirectiveApplicationsWithLabel) {
+      const { from, label } = application.arguments();
+      assert(label, () => `Expected @override application to have a label`);
```
sachindshinde: nit: This `for` loop could be over `subgraphMetadata.overrideDirective().applications()`, and this `assert` could then be replaced with `if (!label) { continue; }`

sachindshinde (approved):
LGTM, thanks for addressing all the feedback! Just one minor nit below

query-graphs-js/src/querygraph.ts
trevor-scheer: [f08a347](https://github.com/apollographql/federation/pull/2902/commits/f08a347fd05b125d5f3ca54531a7c2f4ebe130c4)

trevor-scheer:
@dariuszkuc I'm moving the `backport-to-rust` label over to #2911 which is the full PR for this feature (this one doesn't have the composition component).

PR #2910, opened 2024-01-16
Allow known `FeatureDefinition` subclasses to define custom subgraph schema validation rules

benjamn:
There are a number of `// TODO` comments left in this implementation, but this PR demonstrates a basic framework for allowing known `FeatureDefinition` subclasses, like `SourceSpecDefinition`, to define their own subgraph validation rules, by overriding the `validateSubgraphSchema(schema: Schema): GraphQLError[]` method.

It's the responsibility of `validateSubgraphSchema` to determine if its validation rules are applicable to a given schema, ideally returning an empty array quickly if the feature in question is not used in the schema.

In order for `validateKnownFeatures` to work, a given `FeatureDefinition` subclass needs to have been registered with `registerKnownFeature`.

I'll keep working on the `TODO`s and the tests.

### DIFF
--- a/docs/source/errors.md
+++ b/docs/source/errors.md
@@ -89,23 +89,23 @@ The following errors might be raised during composition:
 | `ROOT_SUBSCRIPTION_USED` | A subgraph's schema defines a type with the name `subscription`, while also specifying a _different_ type name as the root query object. This is not allowed. | 0.x |  |
 | `SATISFIABILITY_ERROR` | Subgraphs can be merged, but the resulting supergraph API would have queries that cannot be satisfied by those subgraphs. | 2.0.0 |  |
 | `SHAREABLE_HAS_MISMATCHED_RUNTIME_TYPES` | A shareable field return type has mismatched possible runtime types in the subgraphs in which the field is declared. As shared fields must resolve the same way in all subgraphs, this is almost surely a mistake. | 2.0.0 |  |
-| `SOURCE_API_HTTP_BASE_URL_INVALID` | The `@sourceAPI` directive must specify a valid http.baseURL | 2.6.0 |  |
-| `SOURCE_API_NAME_INVALID` | Each `@sourceAPI` directive must take a unique and valid name as an argument | 2.6.0 |  |
-| `SOURCE_API_PROTOCOL_INVALID` | Each `@sourceAPI` directive must specify exactly one of the known protocols | 2.6.0 |  |
-| `SOURCE_FIELD_API_ERROR` | The `api` argument of the `@sourceField` directive must match a valid `@sourceAPI` name | 2.6.0 |  |
-| `SOURCE_FIELD_HTTP_BODY_INVALID` | If `@sourceField` specifies http.body, it must be a valid `JSONSelection` matching available arguments and fields | 2.6.0 |  |
-| `SOURCE_FIELD_HTTP_METHOD_INVALID` | The `@sourceField` directive must specify at most one of `http.{GET,POST,PUT,PATCH,DELETE}` | 2.6.0 |  |
-| `SOURCE_FIELD_HTTP_PATH_INVALID` | The `@sourceField` directive must specify a valid URL template for `http.{GET,POST,PUT,PATCH,DELETE}` | 2.6.0 |  |
-| `SOURCE_FIELD_NOT_ON_ROOT_OR_ENTITY_FIELD` | The `@sourceField` directive must be applied to a field of the `Query` or `Mutation` types, or of an entity type | 2.6.0 |  |
-| `SOURCE_FIELD_PROTOCOL_INVALID` | If `@sourceField` specifies a protocol, it must match the corresponding `@sourceAPI` protocol | 2.6.0 |  |
-| `SOURCE_FIELD_SELECTION_INVALID` | The `selection` argument of the `@sourceField` directive must be a valid `JSONSelection` that outputs fields of the GraphQL type | 2.6.0 |  |
-| `SOURCE_HTTP_HEADERS_INVALID` | The `http.headers` argument of `@source*` directives must specify valid HTTP headers | 2.6.0 |  |
-| `SOURCE_TYPE_API_ERROR` | The `api` argument of the `@sourceType` directive must match a valid `@sourceAPI` name | 2.6.0 |  |
-| `SOURCE_TYPE_HTTP_BODY_INVALID` | If the `@sourceType` specifies `http.body`, it must be a valid `JSONSelection` | 2.6.0 |  |
-| `SOURCE_TYPE_HTTP_METHOD_INVALID` | The `@sourceType` directive must specify exactly one of `http.GET` or `http.POST` | 2.6.0 |  |
-| `SOURCE_TYPE_HTTP_PATH_INVALID` | The `@sourceType` directive must specify a valid URL template for `http.GET` or `http.POST` | 2.6.0 |  |
-| `SOURCE_TYPE_ON_NON_OBJECT_OR_NON_ENTITY` | The `@sourceType` directive must be applied to an object or interface type that also has `@key` | 2.6.0 |  |
-| `SOURCE_TYPE_PROTOCOL_INVALID` | The `@sourceType` directive must specify the same protocol as its corresponding `@sourceAPI` | 2.6.0 |  |
+| `SOURCE_API_HTTP_BASE_URL_INVALID` | The `@sourceAPI` directive must specify a valid http.baseURL | 2.7.0 |  |
+| `SOURCE_API_NAME_INVALID` | Each `@sourceAPI` directive must take a unique and valid name as an argument | 2.7.0 |  |
+| `SOURCE_API_PROTOCOL_INVALID` | Each `@sourceAPI` directive must specify exactly one of the known protocols | 2.7.0 |  |
+| `SOURCE_FIELD_API_ERROR` | The `api` argument of the `@sourceField` directive must match a valid `@sourceAPI` name | 2.7.0 |  |
+| `SOURCE_FIELD_HTTP_BODY_INVALID` | If `@sourceField` specifies http.body, it must be a valid `JSONSelection` matching available arguments and fields | 2.7.0 |  |
+| `SOURCE_FIELD_HTTP_METHOD_INVALID` | The `@sourceField` directive must specify at most one of `http.{GET,POST,PUT,PATCH,DELETE}` | 2.7.0 |  |
+| `SOURCE_FIELD_HTTP_PATH_INVALID` | The `@sourceField` directive must specify a valid URL template for `http.{GET,POST,PUT,PATCH,DELETE}` | 2.7.0 |  |
+| `SOURCE_FIELD_NOT_ON_ROOT_OR_ENTITY_FIELD` | The `@sourceField` directive must be applied to a field of the `Query` or `Mutation` types, or of an entity type | 2.7.0 |  |
+| `SOURCE_FIELD_PROTOCOL_INVALID` | If `@sourceField` specifies a protocol, it must match the corresponding `@sourceAPI` protocol | 2.7.0 |  |
+| `SOURCE_FIELD_SELECTION_INVALID` | The `selection` argument of the `@sourceField` directive must be a valid `JSONSelection` that outputs fields of the GraphQL type | 2.7.0 |  |
+| `SOURCE_HTTP_HEADERS_INVALID` | The `http.headers` argument of `@source*` directives must specify valid HTTP headers | 2.7.0 |  |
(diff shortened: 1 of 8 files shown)


internals-js/src/specs/sourceSpec.ts
```diff
@@ -140,8 +144,271 @@ export class SourceSpecDefinition extends FeatureDefinition {
...
+
+    const errors: GraphQLError[] = [];
+    const apiNameToProtocol = new Map<string, ProtocolName>();
+
+    if (sourceAPIInSchema) {
+      sourceAPIInSchema.applications().forEach(application => {
+        const { name, ...rest } = application.arguments();
+
+        if (apiNameToProtocol.has(name)) {
+          errors.push(new GraphQLError(`${sourceAPIInSchema.name} must specify unique name`));
```
benjamn, line 167: I'm not attached to any of the wording/formatting of any of these validation error messages. Please comment to suggest improvements!

package.json
```diff
@@ -108,5 +108,8 @@
   "volta": {
     "node": "20.10.0",
     "npm": "10.2.4"
+  },
+  "dependencies": {
+    "ebnf": "^1.9.1"
```
clenfest, line 113: Does this need to be here if it exists in the internals package-json?

internals-js/src/specs/sourceSpec.ts
```diff
@@ -140,8 +144,271 @@ export class SourceSpecDefinition extends FeatureDefinition {
   sourceFieldDirective(schema: Schema) {
     return this.directive<SourceFieldDirectiveArgs>(schema, 'sourceField')!;
   }
+
+  override validateSubgraphSchema(schema: Schema): GraphQLError[] {
+    const sourceAPIInSchema = schema.directive('sourceAPI') as DirectiveDefinition<SourceAPIDirectiveArgs> | undefined;
```
clenfest, line 149: Will this work if directives are renamed? I suspect not.

internals-js/src/specs/sourceSpec.ts
```diff
@@ -140,8 +144,271 @@ export class SourceSpecDefinition extends FeatureDefinition {
   sourceFieldDirective(schema: Schema) {
     return this.directive<SourceFieldDirectiveArgs>(schema, 'sourceField')!;
   }
+
+  override validateSubgraphSchema(schema: Schema): GraphQLError[] {
```
clenfest, line 148: Make multiple functions out of this one.

internals-js/src/specs/sourceSpec.ts
clenfest: The main thing is that we want to make sure we enumerate the all composition errors in error.ts and errors.md

internals-js/src/specs/sourceSpec.ts
```diff
@@ -140,8 +144,271 @@ export class SourceSpecDefinition extends FeatureDefinition {
...
+    if (sourceAPIInSchema) {
+      sourceAPIInSchema.applications().forEach(application => {
+        const { name, ...rest } = application.arguments();
+
+        if (apiNameToProtocol.has(name)) {
+          errors.push(new GraphQLError(`${sourceAPIInSchema.name} must specify unique name`));
+        }
+
+        // Ensure name is a valid GraphQL identifier.
+        if (!/^[a-zA-Z_][0-9a-zA-Z_]*$/.test(name)) {
```
clenfest, line 171: Is there a library function for this? I hate to roll our own parsing if we don't have to.

internals-js/src/specs/sourceSpec.ts
```diff
@@ -140,8 +144,271 @@ export class SourceSpecDefinition extends FeatureDefinition {
...
+          errors.push(new GraphQLError(`${sourceAPIInSchema}(name: ${
+            JSON.stringify(name)
+          }) must be valid GraphQL identifier`));
+        }
+
+        let protocol: ProtocolName | undefined;
+        KNOWN_SOURCE_PROTOCOLS.forEach(knownProtocol => {
```
clenfest, line 178: Can you not just say `KNOWN_SOURCE_PROTOCOLS.has(protocol)` instead?

internals-js/src/specs/sourceSpec.ts
```diff
@@ -140,8 +144,271 @@ export class SourceSpecDefinition extends FeatureDefinition {
...
+            }
+            protocol = knownProtocol;
+          }
+        });
+
+        if (protocol) {
+          apiNameToProtocol.set(name, protocol);
+
+          const protocolValue = rest[protocol];
+          if (protocolValue && protocol === "http") {
```
clenfest, line 193: Make 'http' a const that exists in KNOWN_PROTOCOLS?

internals-js/src/specs/sourceSpec.ts
```diff
@@ -140,8 +144,271 @@ export class SourceSpecDefinition extends FeatureDefinition {
...
+      // validating.
+      return [];
+    }
+
+    const errors: GraphQLError[] = [];
+    const apiNameToProtocol = new Map<string, ProtocolName>();
+
+    if (sourceAPIInSchema) {
+      sourceAPIInSchema.applications().forEach(application => {
+        const { name, ...rest } = application.arguments();
```
clenfest, line 164: Is there a reason you're doing this as opposed to `const { name, http }`?

internals-js/src/specs/sourceSpec.ts
```diff
@@ -140,8 +144,271 @@ export class SourceSpecDefinition extends FeatureDefinition {
...
+                  `${sourceTypeInSchema.name} http.GET or http.POST must be valid URL path template`,
+                ));
+              }
+            }
+
+            validateHTTPHeaders(headers, errors, sourceTypeInSchema.name);
+
+            if (body) {
+              try {
+                parseJSONSelection(body);
```
clenfest, line 256: Sorry to be dumb, but how does `body` within the `HTTPSourceField` relate to `selection` on `sourceType`? We will want to validate that field as well.

internals-js/src/specs/sourceSpec.ts
```diff
@@ -140,8 +144,271 @@ export class SourceSpecDefinition extends FeatureDefinition {
...
+        const ast = application.parent.sourceAST;
+        switch (ast?.kind) {
+          case "ObjectTypeDefinition":
+          case "InterfaceTypeDefinition":
+            if (!ast.directives?.some(directive => directive.name.value === "key")) {
+              errors.push(new GraphQLError(
+                `${sourceTypeInSchema.name} must be applied to an entity type that also has a @key directive`,
+              ));
+            }
+            // TODO Validate selection is valid JSONSelection for type.
```
clenfest, line 278: So here we are going to validate that the key values match the inputs to the URLPathTemplate? Isn't that `body` rather than `selection`?

internals-js/src/specs/sourceSpec.ts
```diff
@@ -140,8 +144,271 @@ export class SourceSpecDefinition extends FeatureDefinition {
...
+            break;
+          default:
+            errors.push(new GraphQLError(
+              `${sourceTypeInSchema.name} must be applied to object or interface type`,
+            ));
+        }
+      });
+    }
+
+    if (sourceFieldInSchema) {
```
clenfest, line 288: sourceField and sourceType are pretty similar. Probably some dry principles in here.

internals-js/src/specs/sourceSpec.ts
benjamn: Both `http.body` and the `selection` arguments are `JSONSelection` strings. The `http.body` selector for `@sourceField` produces an HTTP request body from the field's arguments, whereas the `selection` argument processes JSON received from the REST API so it matches the desired GraphQL fields. In other words, the inputs are different, but the selection syntax is shared.

internals-js/src/specs/sourceSpec.ts
benjamn: The `selection` argument for `@sourceType` needs to translate from REST JSON to GraphQL fields. I'm checking that here because this code theoretically has easy access to the `ObjectTypeDefinition` or `InterfaceTypeDefinition` to compare its fields to the output shape of the `selection` string.

internals-js/src/specs/sourceSpec.ts
benjamn: They definitely are similar, but unfortunately most of the errors are slightly different, so I'm struggling to eliminate the repetition. Suggestions welcome!

internals-js/src/specs/sourceSpec.ts
benjamn: Just trying to be agnostic to eventually having more than just `http` in `KNOWN_SOURCE_PROTOCOLS`.

package.json
benjamn: I'm removing all the EBNF parsing code from this PR for now, but you're right this `ebnf` package should only need to be installed in the `internals-js` package.

internals-js/src/specs/sourceSpec.ts
benjamn: https://github.com/apollographql/federation/pull/2910/commits/5ffb16f9bc34f965189c383e5175f4880a7419c5

internals-js/src/specs/sourceSpec.ts
benjamn: https://github.com/apollographql/federation/pull/2910/commits/633e78787b2b98b03647e23d2a2c1744aecdbb2e

internals-js/src/specs/sourceSpec.ts
benjamn: https://github.com/apollographql/federation/pull/2910/commits/746fc4d5d16a50ab6486dec80cc0631921d09f12

internals-js/src/specs/sourceSpec.ts
benjamn: https://github.com/apollographql/federation/pull/2910/commits/d35bc4e21890de36d982d3a3c749126717710d17

internals-js/src/specs/sourceSpec.ts
benjamn: The basic operation here is to determine which of `KNOWN_SOURCE_PROTOCOLS` is defined as an argument to the `@sourceAPI` directive, which means intersecting `KNOWN_SOURCE_PROTOCOLS` with the directive argument names, so I think iteration is appropriate, since we don't know which protocol will be used in advance (forgetting for now that there's currently only one protocol, `http`).

internals-js/src/specs/sourceSpec.ts
clenfest: It's just style, so I'm not going to hold up the release for it, but imo it's a little cleaner to say

```typescript
const { api, http, grpc, sql } = application.arguments();
if (http) {
  assert(apiNameToProtocol.get(api) === 'http', ...);
} else if (grpc) {
  assert(apiNameToProtocol.get(api) === 'grpc', ...);
...
}
```

internals-js/src/error.ts
```diff
@@ -555,6 +555,112 @@ const INTERFACE_KEY_MISSING_IMPLEMENTATION_TYPE = makeCodeDefinition(
   { addedIn: '2.3.0' },
 )
 
+const SOURCE_API_NAME_INVALID = makeCodeDefinition(
```
clenfest, line 558: All these should be `2.7.0`

docs/source/errors.md
```diff
@@ -89,6 +89,24 @@ The following errors might be raised during composition:
 | `ROOT_SUBSCRIPTION_USED` | A subgraph's schema defines a type with the name `subscription`, while also specifying a _different_ type name as the root query object. This is not allowed. | 0.x |  |
 | `SATISFIABILITY_ERROR` | Subgraphs can be merged, but the resulting supergraph API would have queries that cannot be satisfied by those subgraphs. | 2.0.0 |  |
 | `SHAREABLE_HAS_MISMATCHED_RUNTIME_TYPES` | A shareable field return type has mismatched possible runtime types in the subgraphs in which the field is declared. As shared fields must resolve the same way in all subgraphs, this is almost surely a mistake. | 2.0.0 |  |
+| `SOURCE_API_HTTP_BASE_URL_INVALID` | The `@sourceAPI` directive must specify a valid http.baseURL | 2.6.0 |  |
```
clenfest, line 92: Same here as in error.ts. These should all be 2.7.0

internals-js/src/error.ts
benjamn: https://github.com/apollographql/federation/pull/2910/commits/0736da204b9b236c36233d9d330395b6c9a17c79

docs/source/errors.md
benjamn: https://github.com/apollographql/federation/pull/2910/commits/0736da204b9b236c36233d9d330395b6c9a17c79

clenfest (approved):

PR #2914, opened 2024-01-23
Additional `SourceSpecDefinition` validations

benjamn:
Follow-up to PR #2910.

### DIFF
--- a/internals-js/src/error.ts
+++ b/internals-js/src/error.ts
@@ -561,6 +561,12 @@ const INTERFACE_KEY_MISSING_IMPLEMENTATION_TYPE = makeCodeDefinition(
   { addedIn: '2.3.0' },
 )
 
+const SOURCE_FEDERATION_VERSION_REQUIRED = makeCodeDefinition(
+  'SOURCE_FEDERATION_VERSION_REQUIRED',
+  'Schemas using `@source{API,Type,Field}` directives must @link-import v2.7 or later of federation',
+  { addedIn: '2.7.1' },
+);
+
 const SOURCE_API_NAME_INVALID = makeCodeDefinition(
   'SOURCE_API_NAME_INVALID',
   'Each `@sourceAPI` directive must take a unique and valid name as an argument',
@@ -757,6 +763,7 @@ export const ERRORS = {
   INTERFACE_KEY_NOT_ON_IMPLEMENTATION,
   INTERFACE_KEY_MISSING_IMPLEMENTATION_TYPE,
   // Errors related to @sourceAPI, @sourceType, and/or @sourceField
+  SOURCE_FEDERATION_VERSION_REQUIRED,
   SOURCE_API_NAME_INVALID,
   SOURCE_API_PROTOCOL_INVALID,
   SOURCE_API_HTTP_BASE_URL_INVALID,
--- a/internals-js/src/specs/sourceSpec.ts
+++ b/internals-js/src/specs/sourceSpec.ts
@@ -1,4 +1,4 @@
-import { DirectiveLocation, GraphQLError, assertName } from 'graphql';
+import { DirectiveLocation, GraphQLError } from 'graphql';
 import { FeatureDefinition, FeatureDefinitions, FeatureUrl, FeatureVersion, LinkDirectiveArgs } from "./coreSpec";
 import {
   Schema,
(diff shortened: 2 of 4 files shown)


internals-js/src/error.ts
```diff
@@ -561,6 +561,12 @@ const INTERFACE_KEY_MISSING_IMPLEMENTATION_TYPE = makeCodeDefinition(
   { addedIn: '2.3.0' },
 )
 
+const SOURCE_FEDERATION_VERSION_REQUIRED = makeCodeDefinition(
+  'SOURCE_FEDERATION_VERSION_REQUIRED',
+  'Schemas using `@source{API,Type,Field}` directives must @link-import v2.7 or later of federation',
```
lennyburdette, line 566: is it more accurate/useful to say that `schemas linking to "https://specs.apollo.dev/source" must link to federation version 2.7 or later`?

internals-js/src/specs/sourceSpec.ts
```diff
@@ -172,17 +175,33 @@ export class SourceSpecDefinition extends FeatureDefinition {
...
 
+    if (result.sourceAPI || result.sourceType || result.sourceField) {
+      // Since this subgraph uses at least one of the @source{API,Type,Field}
+      // directives, it must also use v2.7 or later of federation.
+      if (!federationVersion || federationVersion.lt(this.minimumFederationVersion)) {
+        errors.push(ERRORS.SOURCE_FEDERATION_VERSION_REQUIRED.err(
+          `@source{API,Type,Field} directives require @link-importing federation ${
+            this.minimumFederationVersion
+          } or later`,
+        ));
```
o0Ignition0o, line 191: Maybe
```suggestion
          `@source{API,Type,Field} directives require federation version ${
            this.minimumFederationVersion
          } or later, found ${federationVersion}`,
        ));
``` ?

File: internals-js/src/error.ts (hunk 1 above)
Comment (benjamn) [reply internals-js/src/error.ts:566]: https://github.com/apollographql/federation/pull/2914/commits/7ced7638606529424950b47d98fbb3338066ff50

File: internals-js/src/specs/sourceSpec.ts (hunk 2 above)
Comment (benjamn) [reply internals-js/src/specs/sourceSpec.ts]: See https://github.com/apollographql/federation/pull/2914/commits/264134ed930629e09d4ca84db4d1cd88caddea89 and https://github.com/apollographql/federation/pull/2914/commits/7ced7638606529424950b47d98fbb3338066ff50

File: internals-js/src/specs/sourceSpec.ts [hunk 3]
```diff
@@ -172,17 +175,35 @@ export class SourceSpecDefinition extends FeatureDefinition {
...
         }
+        if (featureUrl && featureUrl.name === 'federation') {
+          federationVersion = featureUrl.version;
+        }
       });
 
+    if (result.sourceAPI || result.sourceType || result.sourceField) {
+      // Since this subgraph uses at least one of the @source{API,Type,Field}
+      // directives, it must also use v2.7 or later of federation.
+      if (!federationVersion || federationVersion.lt(this.minimumFederationVersion)) {
```
clenfest, line 186: I don't think this error is necessary. I think it's fine to just not validate rather than throw an error. Not sure that this could ever happen.

clenfest: All directives have an issue where if you don't specify the right version composition will say that it doesn't understand the directive. Not sure why this is a special case.

clenfest (commented):
`main` has been updated, so you want to merge there rather than `next` (which is in theory 2.8 now).

internals-js/src/specs/sourceSpec.ts
lennyburdette: because the source directives are imported from their own link instead of from the federation link. it's a little weird, but i'm ok with leaning into it because enforcing a minimum version of 2.7 should minimize the number of unhappy paths:

|  | `@link` federation < 2.7 | `@link` federation >= 2.7 |
| -- | ------ | ------ |
| composition < 2.7 | *using source is silently ignored* (🦶🏻 🔫 , but can't fix) | composition fails due to link being ahead of composition |
| composition >= 2.7 | **this scenario** which we're disallowing | happy path |

clenfest: Ok, I got it. Can you make an issue to solve this for the general case? Might be a good starter project for our new hire.

internals-js/src/specs/sourceSpec.ts
```diff
@@ -456,17 +476,21 @@ export class SourceSpecDefinition extends FeatureDefinition {
...
+          { nodes: [application.sourceAST!, fieldParent.sourceAST!] },
         ));
       } else {
         const typeGrandparent = fieldParent.parent as SchemaElement<any, any>;
-        if (typeGrandparent.sourceAST?.kind !== "ObjectTypeDefinition") {
+        if (typeGrandparent.sourceAST?.kind !== Kind.OBJECT_TYPE_DEFINITION) {
           errors.push(ERRORS.SOURCE_FIELD_NOT_ON_ROOT_OR_ENTITY_FIELD.err(
             `${sourceField} must be applied to field of object type`,
-            { nodes: application.sourceAST },
+            { nodes: [
```
clenfest, line 489: What does the error look like when you include the parent and grandparent? I'm concerned that it might be confusing. I think typically we just want to point them to the line that needs to change (i.e. just `application.sourceAST`)

benjamn: The `nodes` property seems to get attached to the `GraphQLError` object, as given, so you'd have to examine that property to notice it. That said, I can change this back to `nodes: application.sourceAST` for consistency with the other errors.

benjamn: https://github.com/apollographql/federation/pull/2914/commits/16a346b3188ffb3ca3412242dfc944dc6cc39577

clenfest: That was my only comment, so if you make those changes, you can merge it.

PR #2915, opened 2024-01-26
Directive stripped when using @link alias

bpuckhaber:
When using `@link` with an alias the directive application is removed from the composed schema. This PR adds a failing compose test which demonstrates the issue.


clenfest:
Closing as https://github.com/apollographql/federation/pull/2916 shows the proper way to rename directives locally.

PR #2917, opened 2024-01-30
docs: Use `--output` rather than redirection

abernix:
This updates the suggestion in the Getting Started guide to use `--output` rather than simple redirection.  While the previous technique redirection does work on Unix, its use on Windows resulted in incorrect file encoding in some cases.

Using `--output` results in a consistent recommendation on all operating systems and also matches documentation [elsewhere], too.

[elsewhere]: https://www.apollographql.com/docs/rover/conventions/#output-to-a-file


trevor-scheer (approved):
Thanks!

PR #2921, opened 2024-02-01
Cleanup hint doc generation script

trevor-scheer:
Hint doc generation is no longer a thing. Hints now live in the `docs` repo and are exported as a shared component.


trevor-scheer:
Expected "check hints" failure, that CI check needs to be removed after landing this.

PR #2926, opened 2024-02-01
Improve validation of `URLPathTemplate` and `JSONSelection` strings passed to `@source{Type,Field}` directives

benjamn:
Follow-up to #2910.

This work is not finished because we ultimately want to validate that the variables used by `URLPathTemplate` strings and the output shapes of `JSONSelection` strings match the available field and argument names.

This PR adds parsers for the two syntaxes and tests of those parsers, which allows us to validate the strings have the expected syntax, even if we're not validating semantics yet.

As another `TODO`, when we use `JSONSelection` for `body` parameters, it's the input shape of the selection that needs to be validated, rather than the output shape, so `getSelectionOutputShape` may not be sufficient.

### DIFF
--- a/internals-js/src/specs/__tests__/sourceSpec.test.ts
+++ b/internals-js/src/specs/__tests__/sourceSpec.test.ts
@@ -1,5 +1,6 @@
 import {
   sourceIdentity,
+  parseJSONSelection,
   parseURLPathTemplate,
   getURLPathTemplateVars,
 } from '../index';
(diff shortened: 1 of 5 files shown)


internals-js/src/specs/__tests__/sourceSpec.test.ts
```diff
@@ -1,7 +1,309 @@
...
+      if (!ast) {
+        throw new Error('Generic parse failure for ' + complexSelection);
+      }
+      expect(getSelectionOutputShape(ast)).toEqual({
+        foo: 'JSON',
+        barAlias: {
+          x: 'JSON',
+          y: 'JSON',
+          z: 'JSON',
+        },
```
benjamn, line 85: The output shape of a `JSONSelection` string is a tree of nested JSON dictionaries mirroring the structure of a GraphQL query, including scalar properties as leaf nodes. The ultimate goal is to validate this structure against object fields to ensure all fields in the schema are handled by some `JSONSelection` mapping.

These `'JSON'` leaf values could have been any placeholder value, but I wanted to emphasize we can't assume much about the types of leaf fields within `JSONSelection` output shapes, except that they must be some JSON value (including possibly `null` for nullable fields).

PR #2929, opened 2024-02-02
Implicitly upgrade federation version to the max required by any used directive

tninesling:
# Overview

The recent addition of `SourceSpecDefinition` surfaced a case where a subgraph with a `@link` to federation < 2.7 and a `@link` to the source spec (which requires federation >= 2.7) results in an unexpected error during composition. Validation for this case was introduced in https://github.com/apollographql/federation/pull/2914, raising a composition error when the versions are mismatched.

Since the existing behavior will implicitly upgrade to a later minor version of federation when two subgraphs are mismatched, it makes sense to do the same with feature requirements.

# Example

Subgraph A has a `@link` to fed version 2.5
Subgraph B has a `@link` to fed version 2.7
Subgraph C has a `@link` to fed version 2.5 and a separate `@link` to the source spec

Previous behavior:
1. Composing Subgraph A and Subgraph B silently upgrades to fed version 2.7
2. Composing Subgraph A and Subgraph C results in a composition error

New behavior:
1. Composing Subgraph A and Subgraph B silently upgrades to fed version 2.7
2. Composing Subgraph A and Subgraph C upgrades to fed version 2.7 and raises a hint explaining why

### DIFF
--- a/.changeset/forty-dolls-type.md
+++ b/.changeset/forty-dolls-type.md
@@ -1,0 +1,6 @@
+---
+"@apollo/composition": minor
+"@apollo/federation-internals": minor
+---
+
+When a linked directive requires a federation version higher than the linked federation spec, upgrade with a hint
--- a/.changeset/forty-dolls-type.md
+++ b/.changeset/forty-dolls-type.md
@@ -3,4 +3,4 @@
 "@apollo/federation-internals": minor
 ---
 
-When a linked directive requires a federation version higher than the linked federation spec, upgrade with a hint
+When a linked directive requires a federation version higher than the linked federation spec, upgrade to the implied version and issue a hint
--- a/composition-js/src/__tests__/hints.test.ts
+++ b/composition-js/src/__tests__/hints.test.ts
@@ -1180,6 +1180,24 @@ describe('when shared field has intersecting but non equal runtime types in diff
 });
 
 describe('when a directive causes an implicit federation version upgrade', () => {
+  const olderFederationSchema = gql`
+    extend schema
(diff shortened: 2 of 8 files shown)


composition-js/src/__tests__/hints.test.ts
```diff
@@ -1177,3 +1178,118 @@ describe('when shared field has intersecting but non equal runtime types in diff
...
+        name: 'old-but-not-upgraded',
+        typeDefs: olderFederationSchema,
+      },
+      {
+        name: 'upgraded',
+        typeDefs: autoUpgradedSchema,
+      }
+    ]);
+
+    expect(result).toRaiseHint(
```
clenfest, line 1250: Could you also add a check for composition success? Would be good to make that explicit.

.changeset/forty-dolls-type.md
```diff
@@ -0,0 +1,6 @@
+---
+"@apollo/composition": minor
+"@apollo/federation-internals": minor
+---
+
+When a linked directive requires a federation version higher than the linked federation spec, upgrade with a hint
```
clenfest, line 6:
```suggestion
When a linked directive requires a federation version higher than the linked federation spec, upgrade to the implied version and issue a hint
```

composition-js/src/__tests__/hints.test.ts
tninesling: Added calls to `assertCompositionSuccess` in [d831c00](https://github.com/apollographql/federation/pull/2929/commits/d831c00fdfb8f4273e758b3ac6fe3e70b40d8081)

clenfest (approved):
LGTM, make sure you squash.

PR #2933, opened 2024-02-06
Limit version of auto-upgraded subgraphs

### LINKED ISSUE #2932: auto-upgraded schemas get upgraded to latest federation version.
### Issue Description

We should restrict schemas that don't manually @link federation to a fixed version of federation so they aren't upgraded to the latest (and thus require the new version of the join spec, etc. I'm thinking that we shouldn't promote any of these auto-upgraded subgraphs to `2.6` (i.e. they wouldn't have to go past version 0.3 of the join spec.

### Link to Reproduction

n/a

### Reproduction Steps

_No response_

clenfest:
Do not auto-upgrade federation subgraphs past fed 2.4. This is so that we don't inadvertently require a router that supports the latest join spec when it's not needed.

Fixes #2932

### DIFF
--- a/internals-js/src/federation.ts
+++ b/internals-js/src/federation.ts
@@ -105,9 +105,6 @@ const federationSpec = (version?: FeatureVersion): FederationSpecDefinition => {
 // directive from v2 definitions, we will continue expanding other directives (up to v2.4) to ensure backwards compatibility.
 const autoExpandedFederationSpec = federationSpec(new FeatureVersion(2, 4));
 
-// For subgraphs that don't explicitly specify a federation version, we need a version number to link in the upgraded schema.
-const autoUpgradedFederationSpec = federationSpec(new FeatureVersion(2, 6));
-
 const latestFederationSpec = federationSpec();
 
 // We don't let user use this as a subgraph name. That allows us to use it in `query graphs` to name the source of roots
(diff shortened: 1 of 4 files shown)


internals-js/src/federation.ts
```diff
@@ -93,10 +94,21 @@ import {
...
+  if (!version) return FEDERATION_VERSIONS.latest();
+  const spec = FEDERATION_VERSIONS.find(version);
+  assert(spec, `Federation spec version ${version} is not known`);
+  return spec;
+};
+
 // Some users rely on auto-expanding fed v1 graphs with fed v2 directives. While technically we should only expand @tag
 // directive from v2 definitions, we will continue expanding other directives (up to v2.4) to ensure backwards compatibility.
-const autoExpandedFederationSpec = FEDERATION_VERSIONS.find(new FeatureVersion(2, 4))!;
+const autoExpandedFederationSpec = federationSpec(new FeatureVersion(2, 4));
```
dariuszkuc, line 106: I believe we stopped auto expanding at `2.4` so folks don't get conflicts with `@authenticated` (which seemed like pretty common name for a directive) -> since auto upgrading stuff wouldn't be able to get any of those new directives (they would have to be prefixed with `@federation__` and it is highly unlikely folks using fed v1 would try this), why don't we just stick to 2.4?

clenfest: I'm a little worried that if we've been auto-upgrading people to 2.6+, that there is more risk in going back to `2.4`, but I can't really think of a case where it would be problematic. On the other hand, I don't think there is a risk in going to 2.6 either. If they want the 2.6 specific directives, they would need to explicitly import them.

dariuszkuc: just wondering whether we need to keep track of 3 different fed versions -> latest, auto upgrade to v2.6 and auto provide definitions up to v2.4

since autoupgraded folks are on fed 1 and they cannot use post 2.4 directives without explicit declaration (and they shouldn't be using fed 2 directives in fed 1 anyway) I don't think it matters

dariuszkuc (approved):

PR #2939, opened 2024-02-20
docs: remove emojis from nav

Meschreiber:
Emojis aren't rendering properly for all viewers and aren't strictly necessary.


R4ph-t (approved):
👍 🙇

PR #2943, opened 2024-02-23
Fix message for when override conflicts with other directives

lukassailer:
When working with the override directive i received this error (names changed):
`@override cannot be used on field "myType.myField" on subgraph "mySubgraph" since "myType.myField" on "mySubgraph" is marked with directive "@requires"`
But requires was used on "myOtherSubgraph".

I changed the error message to state the correct subgraph and changed the tests to match the expected behavior.


lukassailer:
Did i miss something that causes circleci to fail?
If so please let me know.

dariuszkuc (approved):
Thanks for the fix!

PR #2951, opened 2024-03-06
Improve composition hints around types that are strictly `@external`

trevor-scheer:
Note: 1st commit highlights the issue with a test case, 2nd commit proposes the change + test update. No existing tests are affected.

Composition currently gives unhelpful hints around inconsistent value types which are only defined for the sake of referencing (all fields or type is marked `@external`).

In the example test case, the `Permissions` type is only _referenced_ in the `account` (2nd) subgraph due to the `@requires`, so a partial definition seems reasonable here. The hint suggests adding the missing field in order to make the type consistent across subgraphs.

However, if we take the actionable step recommended by that hint and add the missing field (and mark it `@external`, by necessity), we end up with a composition error instead:
> [Subgraph2] Field "Permissions.canEdit" is marked @external but is not used in any federation directive (@key, @provides, @requires) or to satisfy an interface; the field declaration has no use and should be removed (or the field should not be @external).


dariuszkuc (approved):

composition-js/src/merging/merge.ts
```diff
@@ -1081,6 +1081,11 @@ class Merger {
     return this.metadata(sourceIdx).isFieldFullyExternal(field);
   }
 
+  private typeIsFullyExternal(object: ObjectType | InterfaceType): boolean {
+    return object.hasAppliedDirective('external')
+      || object.fields().every(f => f.hasAppliedDirective('external'));
+  }
```
sachindshinde, line 1087: A few things here:
- The phrase "fully external" has a special meaning in federation, so we shouldn't use it unless we want that meaning (it means a field that is either marked `@external` or its parent is, but is not in a `@provides` field set and is not in a type extension `@key`). I'd suggest renaming it e.g. to `areAllFieldsExternal()`.
- Directly checking for an `@external` directive application won't deal with renames appropriately. To deal with that, we'll want to use `FederationMetadata.isFieldExternal()` similar to `Merger.isExternal()` above (note that function checks for parent type `@external`s appropriately).

What I'd suggest is something like

```suggestion
  private areAllFieldsExternal(sourceIdx: number, type: ObjectType | InterfaceType): boolean {
    return type.fields().every(f => this.metadata(sourceIdx).isFieldExternal(f));
  }
```

Then at the callsite, you'll want to provide `sourceIdx`.

composition-js/src/merging/merge.ts
```diff
@@ -967,7 +967,7 @@ class Merger {
     }
     for (const source of sources) {
       // As soon as we find a subgraph that has the type but not the field, we hint.
-      if (source && !source.field(field.name)) {
+      if (source && !source.field(field.name) && !this.typeIsFullyExternal(source)) {
```
sachindshinde, line 970: To provide the source index here, you'll want to change the above `for` loop into:
```
for (const [i, source] of sources.entries()) {
  // ...
}
```

sachindshinde (changes requested):
Just a few things below, but other than that this looks fine.

I think the logic you've stated in the PR description makes sense. The idea behind the value type hint is that you're more likely to run into satisfiability errors if you have partial value types in subgraphs. However, if all the fields are `@external` for a subgraph, then the type should never really be resolved in that subgraph, so you won't run into satisfiability errors from that (you can technically resolve them via `@provides`, but that's for providing more optimal paths to the query planner compared to existing ones, and it shouldn't have to resolve all value type fields for that)

sachindshinde (approved):

Issue #2892, opened 2023-12-18
An alternative to reuseFragments


samuelAndalon:
[Query Planning algorithm expands all fragments](https://github.com/apollographql/federation/blob/main/query-planner-js/src/buildPlan.ts#L3018) in the main Operation Definition as starting point, then [reuseQueryFragments](https://github.com/apollographql/federation/blob/main/query-planner-js/src/config.ts#L35) configuration option will attempt to reuse fragments from the original operation with all the complexities and edge use cases that it implies.

What if, instead of relying on the human that wrote the query, which easily could avoid using fragments we attempt to programmatically try to solve this issue?

The idea is to not touch query planning at all (the algorithm is already too complex) and once query planning is done (or when creating a FetchNode), attempt to optimize the resulting operation.

The optimization would consist of: 
1. Traverse the AST.
2. Search for `InlineFragment` nodes in `SelectionSet` nodes.
3. Create a `FragmentDefinition` of the `SelectionSet` of that `InlineFragments`.
4. Append that `FragmentDefinition` to the document AST.
5. replace the `InlineFragment` in the `SelectionSet` with a `FragmentSpread`.

This optimization is scoped for operations that are heavily relying on polymorphism. It won't need a graphQL schema at all, as everything that is needed is already in the AST.

The graphql-js implementation already provides a [DFS visit function](https://graphql.org/graphql-js/language/#visit) that we can use

PR #2958, opened 2024-03-14
Add new `generateQueryFragments` option to query planner config

trevor-scheer:
If enabled, the query planner will extract inline fragments into fragment definitions before sending queries to subgraphs. This can significantly reduce the size of the query sent to subgraphs, but may increase the time it takes to plan the query.

This is a reimplementation of https://github.com/apollographql/federation/pull/2893 that applies transformations on the selection set instead of AST.

Credit to @samuelAndalon for suggesting the new approach. Resolves https://github.com/apollographql/federation/issues/2892

### DIFF
--- a/internals-js/src/operations.ts
+++ b/internals-js/src/operations.ts
@@ -1548,35 +1548,41 @@ export class SelectionSet {
   }
 
   /**
-   * 
-   * @param namedFragments {NamedFragments}
-   * @param fragmentDefinitionsById 
-   * @returns 
+   * Takes a selection set and extracts inline fragments into named fragments,
+   * reusing generated named fragments when possible.
    */
   minimizeSelectionSet(
     namedFragments: NamedFragments = new NamedFragments(),
-    fragmentDefinitionsById: Map<number, [SelectionSet, NamedFragmentDefinition][]> = new Map(),
+    seenSelections: Map<number, [SelectionSet, NamedFragmentDefinition][]> = new Map(),
   ): [SelectionSet, NamedFragments] {
     const minimizedSelectionSet = this.lazyMap((selection) => {
       if (selection.kind === 'FragmentSelection' && selection.element.typeCondition && selection.element.appliedDirectives.length === 0 && selection.selectionSet) {
-        const id = selection.key().length + selection.selectionSet.selections().length;
-        const equivalentSelectionSetCandidates = fragmentDefinitionsById.get(id);
+        // No proper hash code, so we use a unique enough number that's cheap to
+        // compute and handle collisions as necessary.
+        const mockHashCode = selection.key().length + selection.selectionSet.selections().length;
+        const equivalentSelectionSetCandidates = seenSelections.get(mockHashCode);
         if (equivalentSelectionSetCandidates) {
+          // See if any candidates have an equivalent selection set, i.e. {x y} and {y x}.
           const match = equivalentSelectionSetCandidates.find(([candidateSet]) => candidateSet.equals(selection.selectionSet!));
           if (match) {
+            // If we found a match, we can reuse the fragment (but we still need
+            // to create a new FragmentSpread since parent types may differ).
             return new FragmentSpreadSelection(this.parentType, namedFragments, match[1], []);
           }
         }
 
-        const [minimizedSelectionSet] = selection.selectionSet.minimizeSelectionSet(namedFragments, fragmentDefinitionsById);
+        // No match, so we need to create a new fragment. First, we minimize the
+        // selection set before creating the fragment with it.
+        const [minimizedSelectionSet] = selection.selectionSet.minimizeSelectionSet(namedFragments, seenSelections);
         const fragmentDefinition = new NamedFragmentDefinition(
           this.parentType.schema(),
-          `__generated_on_${selection.element.typeCondition!.name}_${id}_${equivalentSelectionSetCandidates?.length ?? 0}`,
+          `__generated_on_${selection.element.typeCondition!.name}_${mockHashCode}_${equivalentSelectionSetCandidates?.length ?? 0}`,
           selection.element.typeCondition
         ).setSelectionSet(minimizedSelectionSet);
 
+        // Create a new "hash code" bucket or add to the existing one.
         if (!equivalentSelectionSetCandidates) {
-          fragmentDefinitionsById.set(id, [[minimizedSelectionSet, fragmentDefinition]]);
+          seenSelections.set(mockHashCode, [[minimizedSelectionSet, fragmentDefinition]]);
           namedFragments.add(fragmentDefinition);
         } else {
           equivalentSelectionSetCandidates.push([minimizedSelectionSet, fragmentDefinition]);
@@ -1585,7 +1591,7 @@ export class SelectionSet {
         return new FragmentSpreadSelection(this.parentType, namedFragments, fragmentDefinition, []);
       } else if (selection.kind === 'FieldSelection') {
         if (selection.selectionSet) {
-          selection = selection.withUpdatedSelectionSet(selection.selectionSet.minimizeSelectionSet(namedFragments, fragmentDefinitionsById)[0]);
+          selection = selection.withUpdatedSelectionSet(selection.selectionSet.minimizeSelectionSet(namedFragments, seenSelections)[0]);
         }
       }
       return selection;
--- a/internals-js/src/operations.ts
+++ b/internals-js/src/operations.ts
@@ -1553,13 +1553,13 @@ export class SelectionSet {
    */
   minimizeSelectionSet(
     namedFragments: NamedFragments = new NamedFragments(),
-    seenSelections: Map<number, [SelectionSet, NamedFragmentDefinition][]> = new Map(),
+    seenSelections: Map<string, [SelectionSet, NamedFragmentDefinition][]> = new Map(),
   ): [SelectionSet, NamedFragments] {
     const minimizedSelectionSet = this.lazyMap((selection) => {
       if (selection.kind === 'FragmentSelection' && selection.element.typeCondition && selection.element.appliedDirectives.length === 0 && selection.selectionSet) {
         // No proper hash code, so we use a unique enough number that's cheap to
         // compute and handle collisions as necessary.
-        const mockHashCode = selection.key().length + selection.selectionSet.selections().length;
+        const mockHashCode = `on${selection.element.typeCondition}` + selection.selectionSet.selections().length;
         const equivalentSelectionSetCandidates = seenSelections.get(mockHashCode);
         if (equivalentSelectionSetCandidates) {
           // See if any candidates have an equivalent selection set, i.e. {x y} and {y x}.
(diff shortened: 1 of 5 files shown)


internals-js/src/operations.ts
```diff
@@ -1535,6 +1551,32 @@ export class SelectionSet {
     this._selections = mapValues(keyedSelections);
   }
 
+  minimizeSelectionSet(namedFragments: NamedFragments, fragmentSpreadsById: Map<string, FragmentSpreadSelection>): SelectionSet {
+    return this.lazyMap((selection) => {
+      if (selection.kind === 'FragmentSelection' && selection.element.typeCondition && selection.element.appliedDirectives.length === 0 && selection.selectionSet) {
+        const id = print(selection.toSelectionNode());
```
dariuszkuc, line 1557: I think `selection.toString()` should be the same?

dariuszkuc: As discussed this also has a problem of ordering in the selection sets, i.e. given 
```graphql
... on Foo {
  foo
  bar
}
... on Foo {
  bar
  baz
}
```
they should be extracted as single fragment. Since this is printed selection set, this key can be pretty large and we should look into some better hashing approach.

internals-js/src/operations.ts
```diff
@@ -1535,6 +1551,32 @@ export class SelectionSet {
...
+        const existing = fragmentSpreadsById.get(id);
+        if (existing) return existing;
+
+        const minimizedSelectionSet = selection.selectionSet.minimizeSelectionSet(namedFragments, fragmentSpreadsById);
+        const fragmentDefinition = new NamedFragmentDefinition(
+          this.parentType.schema(),
+          `qp__${fragmentSpreadsById.size}`,
```
dariuszkuc, line 1564: I think we could generate some better names -> maybe something along the lines `selection.element.typeCondition.name + "Fragment" + <some unique identifier>`

dariuszkuc (changes requested):

internals-js/src/operations.ts
```diff
@@ -1535,6 +1551,32 @@ export class SelectionSet {
...
+          selection.element.typeCondition
+        ).setSelectionSet(minimizedSelectionSet);
+        const fragmentSpread = new FragmentSpreadSelection(this.parentType, namedFragments, fragmentDefinition, []);
```
dariuszkuc, line 1567: we cannot save `FragmentSpreadSelection` as it references its parent, instead we should cache `NamedFragmentDefinition` and recreate fragment spread selection for each reference

sachindshinde (commented):
Spoke to Trevor about the PR, had some suggestions there but will summarize them here:
- The approach we're taking where "every eligible inline fragment gets replaced with a fragment spread" is okay for now, given that (1) this is feature flagged, and (2) there are some users who just flat-out need the compression above all else, and soon. But it's a terrible choice for readability, and there's extra overhead from the unnecessary fragment indirection (for processing and validation), so I'm going to write out a design for a separate algorithm some time next week. We can take that as a follow-up.
- Using the stringified selection set doesn't quite work, since we don't care about the order of various things (selections, directives, arguments, object value fields). We could implement an actual `.hashCode()` function, but I think we're too time-constrained here for that. The suggestion I've made there is to instead use the length of the stringified selection set as a hash function (as it's unaffected by order changes). Collisions will mostly happen at low sizes, but those should be fast to compare via `.equals()` anyway, so it's fine.
  - This means changing the map of stringified selection sets to named fragment definitions to instead become a map of hash codes (numbers) to lists of pairs, where each pair consists of a `SelectionSet` object, along with the named fragment definition. Lookup involves computing the hash of the `SelectionSet`, and if an entry is found, iterating through the pairs in the list, comparing the `SelectionSet` to the ones in the pairs via `.equals()`. If a pair is equal, then the `SelectionSet` is "in" the map, and the corresponding named fragment definition is the value. If no pair is equal (or no entry was found), then the `SelectionSet` is not "in" the map, and we'd append to that list a new pair with that `SelectionSet` and the new named fragment definition (or if no entry was found, add an entry with the hash code and a new list containing just that pair).

dariuszkuc:
> The suggestion I've made there is to instead use the length of the stringified selection set as a hash function (as it's unaffected by order changes)

I think "stringifying" selection set is going to be rather slow for computing the keys. Why not just go with something simple as `selection.key() + selection.selectionSet.selections().length` instead? We are going to be comparing the selection sets on hits anyway, so first using`selection.key()` to limit to the same the `type condition + directives` and then we would just need to quickly compare the size of selection set size to achieve the same thing.

If you are going to be pointing to an array of `NamedFragmentDefinition` then we still need to account for the additional index in order to generate unique fragment name -> you could simply incorporate that index with the key and avoid nested loop.

internals-js/src/operations.ts
trevor-scheer: [159d92c](https://github.com/apollographql/federation/pull/2958/commits/159d92c535a572415738d375f82a434b8e597a06)

internals-js/src/operations.ts
```diff
@@ -1535,6 +1547,53 @@ export class SelectionSet {
...
+          const match = equivalentSelectionSetCandidates.find(([candidateSet]) => candidateSet.equals(selection.selectionSet!));
+          if (match) {
+            return new FragmentSpreadSelection(this.parentType, namedFragments, match[1], []);
+          }
+        }
+
+        const [minimizedSelectionSet] = selection.selectionSet.minimizeSelectionSet(namedFragments, fragmentDefinitionsById);
+        const fragmentDefinition = new NamedFragmentDefinition(
+          this.parentType.schema(),
+          `__generated_on_${selection.element.typeCondition!.name}_${id}_${equivalentSelectionSetCandidates?.length ?? 0}`,
```
dariuszkuc, line 1574: I believe `__` is generally reserved for internal introspection stuff so I would avoid it.

internals-js/src/operations.ts
```diff
@@ -1535,6 +1547,53 @@ export class SelectionSet {
...
+   * @param fragmentDefinitionsById 
+   * @returns 
+   */
+  minimizeSelectionSet(
+    namedFragments: NamedFragments = new NamedFragments(),
+    fragmentDefinitionsById: Map<number, [SelectionSet, NamedFragmentDefinition][]> = new Map(),
+  ): [SelectionSet, NamedFragments] {
+    const minimizedSelectionSet = this.lazyMap((selection) => {
+      if (selection.kind === 'FragmentSelection' && selection.element.typeCondition && selection.element.appliedDirectives.length === 0 && selection.selectionSet) {
+        const id = selection.key().length + selection.selectionSet.selections().length;
```
dariuszkuc, line 1562: for future: we could recursively calculate the selection set sizes to come up with more unique keys but unsure how much we would gain from it

dariuszkuc: also `selection.key().length` might not be the best identifier `... on Foo` has same length as `... on Bar`. Maybe keep it as string instead?

internals-js/src/operations.ts
trevor-scheer: It's a valid `Name` and I do want to make it clear this is internal / implementation detail. I'll do 1 underscore or if you have something particular in mind lmk.

dariuszkuc: yeah it is valid but might be confusing (and some subgraphs might be rejecting anything with `__` treating it as introspection) - single underscore is fine

internals-js/src/operations.ts
trevor-scheer: [c00429d](https://github.com/apollographql/federation/pull/2958/commits/c00429db96e3e983ef1f8c21b831e931e19a8d1f)

trevor-scheer:
## Detected SAST Vulnerabilities

dariuszkuc (approved):

sachindshinde:
> > The suggestion I've made there is to instead use the length of the stringified selection set as a hash function (as it's unaffected by order changes)
> 
> I think "stringifying" selection set is going to be rather slow for computing the keys. Why not just go with something simple as `selection.key() + selection.selectionSet.selections().length` instead? We are going to be comparing the selection sets on hits anyway, so first using`selection.key()` we will limit to the same `type condition + directives` and then we would just need to quickly compare the size of selection set size to achieve the same thing.

It's a tradeoff between hash function speed and collision rate. You can indeed use the top-level field info of the selection set for the hash function instead of the entire selection set (it's like using the prefix of a string for a string's hash function instead of the entire string). The cost is that you may encounter more collisions, which means more spurious `SelectionSet.equals()` checks (asymptotically it means the algorithm leans more quadratically than linearly, though practically it's probably fine).

In any case this is a short-term fix under a feature flag, so I'm fine with whatever hash function we choose here as long as the collision rate isn't too high.

internals-js/src/operations.ts
```diff
@@ -1535,6 +1547,59 @@ export class SelectionSet {
...
+          selection.element.typeCondition
+        ).setSelectionSet(minimizedSelectionSet);
+
+        // Create a new "hash code" bucket or add to the existing one.
+        if (!equivalentSelectionSetCandidates) {
+          seenSelections.set(mockHashCode, [[minimizedSelectionSet, fragmentDefinition]]);
+          namedFragments.add(fragmentDefinition);
+        } else {
+          equivalentSelectionSetCandidates.push([minimizedSelectionSet, fragmentDefinition]);
+        }
```
sachindshinde, line 1589: Since the comparison is against the pre-minified selection set, I think you'll want the pre-minified selection set in the pair instead of the minified selection set. (The `NamedFragmentDefinition` should still use the minified selection set though.)

sachindshinde (changes requested):

dariuszkuc:
>It's a tradeoff between hash function speed and collision rate. You can indeed use the top-level field info of the selection set for the hash function instead of the entire selection set (it's like using the prefix of a string for a string's hash function instead of the entire string). The cost is that you may encounter more collisions, which means more spurious SelectionSet.equals() checks (asymptotically it means the algorithm leans more quadratically than linearly, though practically it's probably fine).

Indeed there are way more complex hashing algos that we should look into in the future. Using the prefix just narrows down the selection sets to the same type condition and then we just check for number of selections there. I'd imagine instead doing the concatenating selection set into a string to get the final length we could simply recursively iterate over the selection sets and just aggregate the final length value (but unsure whether in practice it would be any better than recursively going through selection sets to calculate total number of selected fields).

internals-js/src/operations.ts
trevor-scheer: [e8560f8](https://github.com/apollographql/federation/pull/2958/commits/e8560f8974e4b978d4ef98f0aa8dc5f173805649)

internals-js/src/operations.ts
```diff
@@ -2413,7 +2478,7 @@ export function selectionOfElement(element: OperationElement, subSelection?: Sel
   return element.kind === 'Field' ? new FieldSelection(element, subSelection) : new InlineFragmentSelection(element, subSelection!);
 }
 
-export type Selection = FieldSelection | FragmentSelection;
+export type Selection = FieldSelection | FragmentSelection | FragmentSpreadSelection;
```
sachindshinde, line 2481: `FragmentSelection` is an abstract class implemented by `InlineFragmentSelection` and `FragmentSpreadSelection`, so I don't think you'll need to change the definition here.

sachindshinde (commented):
One minor nit to fix in a followup

internals-js/src/operations.ts
dariuszkuc: https://github.com/apollographql/federation/pull/2960

dariuszkuc:
Indeed when reusing fragments we apply a logic to ensure we have selection set > 1. For fragment generation logic we decided to keep it simple for now (we could definitely improve it in the future and provide some better optimizations, e.g. join all sibling inline fragments into one named fragment etc).

PR #2960, opened 2024-03-19
fix: remove unnecessary selection type

dariuszkuc:
`FragmentSelection` is both inline fragment and named spread selection.


trevor-scheer (approved):
Thanks! I wonder what compelled me to add this in the first place, but LGTM.

Issue #2952, opened 2024-03-06
Query planner generates invalid fetch node selection sets when reuseQueryFragments=true


tinnou:
### Issue Description

When `reuseQueryFragments=true` (the default config), in certain cases the query planner will generate a fetch that will fail GraphQL validation (Field Selection Merging - overlapping fields can be merged rule) at the subgraph GraphQL server. Specifically, this happens when a fragment from the query is reused and the query planner also queries the same field separately but the two aren't identical. The example below specifically showcases when field arguments mismatch.

## Steps to Reproduce

I built a minimum viable example [unit test](https://github.com/apollographql/federation/compare/main...tinnou:federation:query-planner-fragment-invalid-bfg-822?expand=1#diff-dfde3ac629337fd42fe177c9df7a3192483b173497ac744159ebfacf0a85f307R140) that reproduces building an incorrect plan. 

Subgraph `discoveryDgs` schema:
```graphql
     type Query {
        pinotPreQuerySearchPage: PinotUIEntity
      }
    
      union PinotUIEntity = PinotBoxShotEntityTreatment | PinotEpisodicBillboardEntityTreatment | PinotBannerWithTrailerEntityTreatment
      
      type PinotBannerWithTrailerEntityTreatment @key(fields: "unifiedEntityId") {
        unifiedEntityId: ID!
        unifiedEntity: UnifiedEntity
      }
      
      type PinotBoxShotEntityTreatment @key(fields: "unifiedEntityId") {
        unifiedEntityId: ID!
        unifiedEntity: UnifiedEntity
      }
      
      type PinotEpisodicBillboardEntityTreatment @key(fields: "unifiedEntityId") {
        unifiedEntityId: ID!
        unifiedEntity: UnifiedEntity
      }
      
      interface UnifiedEntity {
          id: ID!
      }

      type GenericContainer implements UnifiedEntity @key(fields: "id"){
        id: ID!
      }
      
      type Supplemental implements UnifiedEntity @key(fields: "id") {
        id: ID!
      }
      
      type Show implements UnifiedEntity @key(fields: "id") {
        id: ID!
      }
```

and subgraph `gusto` schema:
```graphql
      type Query {
        me: String
      }
      
      interface Video {
        videoId: Int!
        taglineMessage(uiContext: String): String
        bar: String
      }
      
      interface UnifiedEntity {
          id: ID!
      }
      
      type GenericContainer implements UnifiedEntity @key(fields: "id") {
        id: ID!
[...]

PR #2963, opened 2024-03-20
Fix a query planning bug where invalid subgraph queries are generated with `reuseQueryFragments` set true

duckki:
This PR fixes https://github.com/apollographql/federation/issues/2952.

### Summary of changes

- updated computeExpandedSelectionSetAtType function not to trim fragment's validators if the fragment's type condition is not an object type.
- This change is necessary because `FieldsInSetCanMerge` rule is more restrictive in that case.
- The untrimmed validator should avoid generating invalid queries, but it may be less efficient.

### Test plan

- The operations.test.ts confirms the changes of named fragments' validators.
- Added a new test (based on the github issue's reproducer) confirming that subgraph queries are no longer invalid.

### DIFF
--- a/.changeset/invalid-reused-fragment.md
+++ b/.changeset/invalid-reused-fragment.md
@@ -1,0 +1,7 @@
+---
+"@apollo/query-planner": patch
+"@apollo/composition": patch
+"@apollo/federation-internals": patch
+---
+
+Fix a query planning bug where invalid subgraph queries are generated with `reuseQueryFragments` set true. ([#2952](https://github.com/apollographql/federation/issues/2952))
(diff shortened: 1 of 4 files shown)


duckki:
@dariuszkuc I just rebased this PR on top of current origin/main. Please review. Thanks!

duckki:
@goto-bus-stop Please also review this one :)

- The semgrep check was failing due to the lack of github access, which I'm not sure how to fix.
- I believe Note.js v16 failure is an intermittent issue.

goto-bus-stop (commented):
It took me a little bit to grasp the issue here 😇

Test and implementation looks good 👍🏻

goto-bus-stop (approved):
Pre-approval to unblock merging after the `plan: any` is addressed 😄

duckki:
@goto-bus-stop Thanks for the review. I pushed a fix.

PR #2975, opened 2024-04-08
docs: context directives

shorgi:
Docs for `@context` and `@fromContext` directives.

### DIFF
--- a/docs/source/entities-advanced.mdx
+++ b/docs/source/entities-advanced.mdx
@@ -993,6 +993,12 @@ type Child @key(fields: "id") {
 
 When the same contextual value is set in multiple places—as is done in the example with the args of `Child.prop1` and `Child.prop2`—the `FieldValue` must resolve all types from each place into a single value that matches the parameter type.
 
+<Note>
+
+Federation makes no guarantees about which context will be used if a field is reachable via multiple contexts.
+
+</Note>
+
 ### Disambiguating contexts
 
 When a set context could be fulfilled by multiple ancestor entities in the type hierarchy, the nearest ancestor is chosen. For example, if both the parent and grandparent of a type can provide the value of a context, the parent is chosen because it's the closer ancestor.
@@ -1025,3 +1031,54 @@ type C @key(fields: "id") {
 ```
 
 In a more complex graph, a field could be reachable via multiple paths, and there could be a different field used to resolve the `prop` depending on which path was used.
+
+### Referencing fields across subgraphs
+
+The definition of context scopes can only exist in one subgraph schema. You can't define `@context` in one subgraph and use `@fromContext` in another. However, you can use contexts to share data across subgraphs using the `@external` reference.
+
+Reusing our first example, let's say we had a subgraph responsible for the User and Currency type:
+
+```graphql
+scalar CurrencyCode
+
+type Currency @shareable {
+   id: ID!
+   isoCode: CurrencyCode!
+}
+
+type User @key(fields: "id") {
+   id: ID!
+   userCurrency: Currency!
+}
+```
+
+If you want to reference those fields from another subgraph, you can use the `@external` directive to pass data across subgraph boundaries:
+
+```graphql
+scalar CurrencyCode
+
+type Currency @shareable {
+   id: ID!
+   isoCode: CurrencyCode!
+}
+
+type User @key(fields: "id") @context(name: "userContext") {
+   id: ID!
+   
+  # This is a reference to the field resolved elsewhere
+   userCurrency: Currency! @external
+   
+   # We add this field to our type here
+   lastTransaction: Transaction! 
+}
+
+type Transaction @key(fields: "id") {
+  id: ID!
+  currency: Currency!
+  amount: Int!
+  amountInUserCurrency(
+    currencyCode: CurrencyCode! 
+        @fromContext(field: "$userContext { userCurrency { isoCode } }")
+  ): Int!
+}
+```
--- a/docs/source/entities-advanced.mdx
+++ b/docs/source/entities-advanced.mdx
@@ -902,3 +902,85 @@ const resolvers = {
 A basic implementation of the `fetchProductByID` function might make a database call each time it's called. If we need to resolve `Product.price` for `N` different products, this results in `N` database calls. These calls are made in addition to the call made by the Reviews subgraph to fetch the initial list of reviews (and the `id` of each product). This is where the "N+1" problem gets its name. If not prevented, this problem can cause performance problems or even enable denial-of-service attacks.
 
 This problem is not limited to reference resolvers. In fact, it can occur with any resolver that fetches from a data store. To handle this problem, we strongly recommend using [the dataloader pattern](https://github.com/graphql/dataloader). Nearly every GraphQL server library provides a dataloader implementation, and you should use it in every resolver. This is true even for resolvers that aren't for entities and that don't return a list. These resolvers can still cause N+1 issues via [batched requests](/technotes/TN0021-graph-security/#batched-requests).
+
+## Using contexts to share data along type hierarchies
+
+Use the `@context` and `@fromContext` directives to set and get a _context_ through which types of a subgraph can share data. Contexts enable a subgraph to share data between types of a nested-type hieararchy, without overloading entity keys with extraneous fields and without breaking any separation of concerns between different subgraphs.  
+
+An entity may contain nested objects, and each nested object may have a dependency on an ancestor type and thus need access to an ancestor's field(s). 
+
+To enable a descendant to access an ancestor's field, you could add it as a `@key` field to every entity along the chain of nested types, but that can become problematic. Deeply nested types can change the `@key` fields of many entities. The added field may be irrelevant to the entities it's added to. Most importantly, this overloading of `@key` fields often breaks the separation of concerns between different subgraphs.
+
+The `@context` and `@fromContext` directives enable a subgraph to share fields without overloading `@key` fields. You can use these directives to define one or more contexts of a subgraph from which a field gets retrieved.
+
+### Using one context
+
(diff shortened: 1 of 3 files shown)


docs/source/federated-types/federated-directives.mdx
```diff
@@ -894,3 +894,46 @@ If different subgraphs use different versions of a directive's corresponding spe
 
 </tbody>
 </table>
+
+## Sharing data with contexts
```
shorgi, line 898: possibly copyedit, to reflect more saving data to reference elsewhere in type hierarchy

docs/source/federated-types/federated-directives.mdx
```diff
@@ -894,3 +894,46 @@ If different subgraphs use different versions of a directive's corresponding spe
...
+
+<MinVersion version="2.8">
+
+### `@context`
+
+</MinVersion>
+
+The `@context` directive defines a named context from which any field of the annotated type can be passed to a getter of the context (a field annotated with [`@fromContext`](#fromcontext)).
```
shorgi, line 906:
```suggestion
The `@context` directive defines a named context from which any field of the annotated type can be passed to a receiver of the context (a field annotated with [`@fromContext`](#fromcontext)). 
```

docs/source/entities-advanced.mdx
```diff
@@ -902,3 +902,118 @@ const resolvers = {
...
+
+type C @key(fields: "id") {
+  id: ID!
+  prop(
+    arg: String! @fromContext(field: "$context_1 { field }")
+  ): Int!
+} 
+```
+
+In a more complex graph, a field could be reachable via multiple paths, and there could be a different field used to resolve the `prop` depending on which path was used.
```
smyrick, line 1019:
```suggestion
In a more complex graph, a field could be reachable via multiple paths, and there could be a different field used to resolve the `prop` depending on which path was used.

### Referencing fields across subgraphs

The definition of context scopes can only exist in one subgraph schema. You can not define `@context` in one subgraph and use `@fromContext` in another. However, you can use contexts to share data across subgraphs using the `@external` reference.

Using our first example, let's say we had a subgraph responsible for the `User` and `Currency` type

```graphql title="User subgraph"
scalar CurrencyCode

type Currency @shareable {
   id: ID!
   isoCode: CurrencyCode!
}

type User @key(fields: "id") {
   id: ID!
   userCurrency: Currency!
}
```

Then if in another subgraph you want to reference those fields, you can use the `@external` directive to still pass data across subgraph boundaries.

```graphql title="Transactions subgraph"
scalar CurrencyCode

type Currency @shareable {
   id: ID!
   isoCode: CurrencyCode!
}

type User @key(fields: "id") @context(name: "userContext") {
   id: ID!
   
  # This is a reference to the field resolved elsewhere
   userCurrency: Currency! @external
   
   # We add this field to our type here
   lastTransaction: Transaction! 
}

type Transaction @key(fields: "id") {
  id: ID!
  currency: Currency!
  amount: Int!
  amountInUserCurrency(
    currencyCode: CurrencyCode! 
        @fromContext(field: "$userContext { userCurrency { isoCode } }")
  ): Int!
}
```

docs/source/federated-types/federated-directives.mdx
```diff
@@ -894,3 +894,48 @@ If different subgraphs use different versions of a directive's corresponding spe
...
+directive @fromContext(field: FieldValue!) on ARGUMENT_DEFINITION;
+```
+
+A `@fromContext` directive must be used as an argument on a field. Its field value—the `FieldValue` scalar—must contain the name of a defined context and a selection of a field from the context's type.
+
+The selection syntax for `@fromContext` used in its `FieldValue` is similar to GraphQL field selection syntax, with some added rules: 
+- The first element must be the name of a context defined by `@context` and prefixed with `$` (for example, `$myContext`). This is the only context that can be referenced by the annotated field.
+- The `@skip` and `@include` directives must not be used.
+- Top-level type conditions must not overlap with one another, so that the field can be resolved to a single value.
+- All fields referenced in the `FieldValue` must be expressed within the current subgraph. If necessary, the fields must be annotated with `@external`.
```
clenfest, line 937: This is a temporary restriction, but you need to add a bullet that says something like this:

- The argument must be nullable. This occurs validation is currently done at the subgraph level and we can't guarantee that the referenced field will not become nullable during the graph merge (i.e. if the field is nullable in one subgraph but not another.)

docs/source/entities-advanced.mdx
```diff
@@ -1025,3 +1031,54 @@ type C @key(fields: "id") {
 ```
 
 In a more complex graph, a field could be reachable via multiple paths, and there could be a different field used to resolve the `prop` depending on which path was used.
+
+### Referencing fields across subgraphs
+
+The definition of context scopes can only exist in one subgraph schema. You can't define `@context` in one subgraph and use `@fromContext` in another. However, you can use contexts to share data across subgraphs using the `@external` reference.
```
clenfest, line 1044: Rather than "You can't define `@context` in one subgraph and use `@fromContext` in another."

Maybe say:
"`@fromContext` cannot reference a `@context` defined in another subgraph"

docs/source/federated-types/federated-directives.mdx
```diff
@@ -911,8 +911,26 @@ directive @context(name: String!) on OBJECT | INTERFACE | UNION;
 
 A `@context` directive must be applied to an object, interface, or union type. A type can be annotated with one or more `@context` directives.
 
-Each `@context` must be defined with a name that's unique within the supergraph.
+Each `@context` must be defined with a name that's unique within the subgraph.
```
clenfest, line 933: The following comment makes it seem like this statement isn't true.

docs/source/federated-types/federated-directives.mdx
```diff
@@ -911,8 +911,26 @@ directive @context(name: String!) on OBJECT | INTERFACE | UNION;
 
 A `@context` directive must be applied to an object, interface, or union type. A type can be annotated with one or more `@context` directives.
 
-Each `@context` must be defined with a name that's unique within the supergraph.
+Each `@context` must be defined with a name that's unique within the subgraph. 
 
+Each `@context` can be applied to multiple places within a subgraph. For example:
+
+```graphql
+type A @key(fields: "id") @context(name: "ctxt") {
```
clenfest, line 938: Cspell doesn't like the name ctxt (which I know I suggested). Either change it to a real word or add it to cspell-dict.txt

docs/source/entities-advanced.mdx
```diff
@@ -909,3 +909,183 @@ const resolvers = {
...
+   # We add this field to our type here
+   lastTransaction: Transaction! 
+}
+
+type Transaction @key(fields: "id") {
+  id: ID!
+  currency: Currency!
+  amount: Int!
+  amountInUserCurrency(
+    currencyCode: CurrencyCode!
```
clenfest, line 1087: currencyCode must be nullable. i.e. "CurrencyCode" rather than "CurrencyCode!"

docs/source/federated-types/federated-directives.mdx
```diff
@@ -913,3 +913,68 @@ If different subgraphs use different versions of a directive's corresponding spe
...
+```
+
+A `@fromContext` directive must be used as an argument on a field. Its field value—the `FieldValue` scalar—must contain the name of a defined context and a selection of a field from the context's type.
+
+The selection syntax for `@fromContext` used in its `FieldValue` is similar to GraphQL field selection syntax, with some additional rules: 
+- The first element must be the name of a context defined by `@context` and prefixed with `$` (for example, `$myContext`). This is the only context that can be referenced by the annotated field.
+- The `@skip` and `@include` directives must not be used.
+- The second element must be a selection set that resolves to a single field.
+- Top-level type conditions must not overlap with one another, so that the field can be resolved to a single value.
+- All fields referenced in the `FieldValue` must be expressed within the current subgraph. If necessary, the fields must be annotated with `@external`.
```
Meschreiber, line 975: Maybe link to relevant section in the guide page?

docs/source/federated-types/federated-directives.mdx
```diff
@@ -913,3 +913,68 @@ If different subgraphs use different versions of a directive's corresponding spe
...
+- The argument must be nullable. Because validation is currently done at the subgraph level, the referenced field may become nullable when merging subgraphs, such as when the field is nullable in one subgraph but not in another.
```
Meschreiber, line 976: Even if this changes, better to not to hint towards it now.

```suggestion
- The argument must be nullable. Because validation is done at the subgraph level, the referenced field may become nullable when merging subgraphs, such as when the field is nullable in one subgraph but not in another.
```

docs/source/federation-versions.mdx
```diff
@@ -25,6 +25,90 @@ For a comprehensive changelog for Apollo Federation and its associated libraries
...
+</CodeColumns>
+
+<hr/>
+
+#### Directive changes
+
+<table>
+  <thead>
+    <tr>
+      <th>Topic</th>
```
Meschreiber, line 67: I [previously introduced](https://github.com/apollographql/federation/pull/3013) this so directive names weren't scrunched. Annoying to have to repeat--might just make a class for all the tables on this page.

```suggestion
      <th style={{ minWidth: 200 }}>Topic</th>
```

Meschreiber (commented):
Nice clear additions! Just some suggestions for readability.

One overall suggestion: it might be nice to include a very sparse code snippet and/or diagram in the introductory section (#using-contexts-to-share-data-along-type-hierarchies). I think that can often help readers "get it" faster.

Meschreiber (approved):
Whoops, meant to approve with my previous comments.

docs/source/entities-advanced.mdx
```diff
@@ -909,3 +909,183 @@ const resolvers = {
...
+type Child @key(fields: "id") {
+  id: ID!
+  prop1(
+    arg: String! @fromContext(field: "$context_1 { field }")
+  ): Int!
+  prop2(
+    arg: String! 
+        @fromContext(field: "$context_2 ... A { field } ... B { field } ... C { anotherField }")
+  ): Int!
+}
```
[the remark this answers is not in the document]
shorgi: good point, will resolve in a follow-up PR

clenfest (approved):

PR #2978, opened 2024-04-11
WIP: chore(@override): add e2e testing of progressive overriding

metrue:


metrue:
Confirmed with Apollo team, the progressive feature for `@override` only supported on Rust router, so it's possible to use it on federation/gateway out of box.

But if we really need to do it on the federation/gateway, we can do a little hacking in the queryplan building but it will lose the queryplan caching.  I tried this commit 1da5e434e45a7d70946d6e898cba631346592510, now you can see the CI is passed, but obviously it's a thing to try on production.

PR #2993, opened 2024-05-02
generateQueryFragments generates invalid plans with missing fragment definitions

xuorig:
This PR includes a failing test that show that `generateQueryFragments` can generate invalid GraphQL queries. It seems related to the way the hashCode is computed. Any selections of the same length and type are somehow grouped together and an additional fragment definition is not added.

Opening this in case somebody gets to the root cause fix while I investigate further.

### DIFF
--- a/.changeset/stupid-lemons-rest.md
+++ b/.changeset/stupid-lemons-rest.md
@@ -2,4 +2,25 @@
 "@apollo/query-planner": patch
 ---
 
-Fix issue with missing fragment definitions due to generateQueryFragments
+Fix issue with missing fragment definitions due to `generateQueryFragments`.
+
+An incorrect implementation detail in `generateQueryFragments` caused certain queries to be missing fragment definitions. Specifically, subsequent fragment "candidates" with the same type condition and the same length of selections as a previous fragment weren't correctly added to the list of fragments. An example of an affected query is:
+
+```graphql
+query {
+  t {
+    ... on A {
+      x
+      y
+    }
+  }
+  t2 {
+    ... on A {
+      y
+      z
+    }
+  }
+}
+```
+
+In this case, the second selection set would be converted to an inline fragment spread to subgraph fetches, but the fragment definition would be missing.
--- a/.changeset/stupid-lemons-rest.md
+++ b/.changeset/stupid-lemons-rest.md
@@ -1,0 +1,5 @@
+---
+"@apollo/query-planner": patch
+---
+
+Fix issue with missing fragment definitions due to generateQueryFragments
--- a/internals-js/src/operations.ts
+++ b/internals-js/src/operations.ts
@@ -1573,6 +1573,7 @@ export class SelectionSet {
           `_generated_${mockHashCode}_${equivalentSelectionSetCandidates?.length ?? 0}`,
           selection.element.typeCondition
         ).setSelectionSet(minimizedSelectionSet);
+        namedFragments.add(fragmentDefinition);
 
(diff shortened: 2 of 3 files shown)


query-planner-js/src/__tests__/buildPlan.test.ts
```diff
@@ -5425,6 +5426,60 @@ describe('Fragment autogeneration', () => {
...
+              ..._generated_onA2_1
+            }
+          }
+          
+          fragment _generated_onA2_0 on A {
+            x
+            y
+          }
+
+          fragment _generated_onA2_1 on A {
```
xuorig, line 5475: This fragment is missing in practice.

internals-js/src/operations.ts
```diff
@@ -1574,12 +1575,12 @@ export class SelectionSet {
...
-        if (!equivalentSelectionSetCandidates) {
-          seenSelections.set(mockHashCode, [[selection.selectionSet, fragmentDefinition]]);
-          namedFragments.add(fragmentDefinition);
-        } else {
+        if (equivalentSelectionSetCandidates) {
           equivalentSelectionSetCandidates.push([selection.selectionSet, fragmentDefinition]);
+        } else {
+            seenSelections.set(mockHashCode, [[selection.selectionSet, fragmentDefinition]]);
         }
+        namedFragments.add(fragmentDefinition);
```
xuorig, line 1583: If we get here, it's because we didn't find an exact match. In these cases, we need to create a fragment definition in all cases. Double check me here, but this does seem to fix the issue.

trevor-scheer: Yeah, agree. Small nit, but could we bump this line up to where `fragmentDefinition` is assigned? Probably makes more sense right next to it.

Appreciate the if/else restructure 👍

trevor-scheer (approved):
Can you please add a changeset by running `npx changeset` and following the prompts for a patch? A small description of how a user might be affected by this bug would be great. Thanks for the very clear PR and test case!

xuorig:
Addressed your comments @trevor-scheer 🙇 thanks for the quick review!

.changeset/stupid-lemons-rest.md
```diff
@@ -0,0 +1,5 @@
+---
+"@apollo/query-planner": patch
+---
+
+Fix issue with missing fragment definitions due to generateQueryFragments
```
trevor-scheer, line 5: I can't send this as a GH suggestion due to use of triple backtick in the comment, but would you mind adding this or something like it?

> Fix issue with missing fragment definitions due to generateQueryFragments
> 
> An incorrect impementation detail in `generateQueryFragments` caused certain queries to be missing fragment definitions. Specifically, _subsequent_ fragment "candidates" with the same type condition and the same length of selections as a previous fragment weren't correctly added to the list of fragments. An example of an affected query is:
> 
> ```graphql
> query {
>   t {
>     ... on A {
>       x
>       y
>     }
>   }
>   t2 {
>     ... on A {
>       y
>       z
>     }
>   }
> }
> ```
> 
> In this case, the second selection set would be converted to an inline fragment spread to subgraph fetches, but the fragment definition would be missing.

trevor-scheer: Ideally the changeset should have the ```graphql code fencing around the query, but that doesn't paste over from my comment.

trevor-scheer (approved):
Thank you!

dariuszkuc (approved):
Thanks for the fix!

PR #3001, opened 2024-05-13
docs: clarify nested keys

Meschreiber:
Request from SA team to surface nested field pattern on directive reference page

### DIFF
--- a/docs/source/entities-advanced.mdx
+++ b/docs/source/entities-advanced.mdx
@@ -39,19 +39,7 @@ type Organization {
 
 <Note>
 
-Although nested fields are most commonly used in compound keys, you can also use a nested field as a single `@key` field.
-For example, the previous `User` entity `@key` doesn't make much sense without including the user `id`:
-
-```graphql {1} title="Users subgraph"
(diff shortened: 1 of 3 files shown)


docs/source/entities-advanced.mdx
```diff
@@ -36,6 +37,24 @@ type Organization {
 }
 ```
 
+<Note>
+
+Although nested fields are most commonly used in compound keys, you can also use a nested field as a single `@key` field.
+For example, the previous `User` entity `@key` doesn't make much sense without including the user `id`:
```
shorgi, line 43: IIUC, this example shows why nested fields are usually part of a compound key. But from the first sentnece, the purpose of this note is to say that a nested key can be the only field of a simple key? If so, is there a practical example of a simple key with a nested field?

Meschreiber: > If so, is there a practical example of a simple key with a nested field?

I was struggling to think of one 😅, let me ask the SA team...

shorgi (approved):

PR #3017, opened 2024-05-23
various setContext bug fixes

clenfest:
Some setContext bug fixes
- needsJoinDirective() logic is incorrect. We need to make sure that we add a join field when @fromDirective exists on the arguments, not the definition
- Query plans were incorrect if type was entirely in one subgraph. selectionIsFullyLocalFromAllVertices was calling SelectionSet.canRebaseOn so we fixed to return false if the selection contained a field with a contextual argument
- For top level queries, we don't want to have "... on Query" in the rewrite path.
- Fixed up  selectionSetAsKeyRenamers() logic

### DIFF
--- a/internals-js/src/operations.ts
+++ b/internals-js/src/operations.ts
@@ -377,7 +377,9 @@ export class Field<TArgs extends {[key: string]: any} = {[key: string]: any}> ex
     const fromContextDirective = federationMetadata(parentType.schema())?.fromContextDirective();
     if (fromContextDirective && isFederationDirectiveDefinedInSchema(fromContextDirective)) {
       const fieldInParent = parentType.field(this.name);
-      if (fieldInParent && fieldInParent.arguments().some(arg => arg.appliedDirectivesOf(fromContextDirective).length > 0)) {
+      if (fieldInParent && fieldInParent.arguments()
+          .some(arg => arg.appliedDirectivesOf(fromContextDirective).length > 0 && (!this.args || this.args[arg.name] === undefined))
+        ) {
         return undefined;
       }  
     }
--- a/internals-js/src/operations.ts
+++ b/internals-js/src/operations.ts
@@ -368,10 +368,21 @@ export class Field<TArgs extends {[key: string]: any} = {[key: string]: any}> ex
     if (this.name === typenameFieldName) {
       return parentType.typenameField()?.type;
     }
-
-    return this.canRebaseOn(parentType)
+    
+    const returnType = this.canRebaseOn(parentType)
       ? parentType.field(this.name)?.type
       : undefined;
+      
+    // If the field has an argument with fromContextDirective on it. We should not rebase it.
+    const fromContextDirective = federationMetadata(parentType.schema())?.fromContextDirective();
+    if (fromContextDirective && isFederationDirectiveDefinedInSchema(fromContextDirective)) {
+      const fieldInParent = parentType.field(this.name);
+      if (fieldInParent && fieldInParent.arguments().some(arg => arg.appliedDirectivesOf(fromContextDirective).length > 0)) {
+        return undefined;
+      }  
+    }
+
+    return returnType;
   }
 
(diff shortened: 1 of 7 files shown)


sachindshinde (changes requested):
A few comments below. The last comment isn't really a change request, but more just making sure I understand `selectionSetAsKeyRenamers()`.

One other thing to note (and this can be handled in a follow-up PR), but there's a bug in `GraphPath.checkDirectPathFromPreviousSubgraphTo()`. Specifically, that function is used to check a hypothetical, mainly whether if part of the graph path could instead be successfully planned "directly" against a different starting vertex (`prevSubgraphStartingVertex`). Really what this is checking for, is that if we can easily tell that planning would work, we return back where this direct graph path would now lead instead (if we can't tell whether it would work e.g. because of a `@requires` condition, or if we know it wouldn't work, then we return `undefined`).

The gist is that there's a for loop in that function that iterates through `this.props.edgeTriggers`. If any of those triggers returns a non-null composite type for `getFieldParentType()` (which you'll have to pass through from `checkDirectPathFromPreviousSubgraphTo()`'s single callsite), and that composite type has `@context` on it in at least one subgraph (you may need to pre-compute a list of such types during query graph creation, and also it'll help to pull type-matching logic from `canSatisfyConditions()` into a new function to reduce duplication), then you'll want to return `undefined` from `checkDirectPathFromPreviousSubgraphTo()`.

query-planner-js/src/buildPlan.ts
```diff
@@ -1673,22 +1673,23 @@ type FieldToAlias = {
...
   return selectionSet.selections().map((selection: Selection): FetchDataKeyRenamer[] | undefined => {
     if (selection.kind === 'FieldSelection') {
-      // We always have at least one '..' in the relative path.
-      if (relPath[relPath.length - 1] === '..') {
-        return possibleRuntimeTypes(selectionSet.parentType).map((t) => ({
-          kind: 'KeyRenamer',
-          path: [...relPath, `... on ${t.name}`, selection.element.name],
-          renameKeyTo: alias,
-        }));
+      if (relPath[relPath.length - 1] === '..' && selectionSet.parentType.name !== 'Query') {
```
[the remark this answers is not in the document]
clenfest: Correct.

sachindshinde:
As a stopgap fix for `checkDirectPathFromPreviousSubgraphTo()`, you could also just disable the optimization when `@context` is present. The way that would work, is that whenever any subgraph uses contexts at all (which you can check by looking at whether `QueryGraph.subgraphToArgs` has at least one entry), then `checkDirectPathFromPreviousSubgraphTo()` should just early return `undefined`.

internals-js/src/operations.ts
```diff
@@ -377,7 +377,9 @@ export class Field<TArgs extends {[key: string]: any} = {[key: string]: any}> ex
     const fromContextDirective = federationMetadata(parentType.schema())?.fromContextDirective();
     if (fromContextDirective && isFederationDirectiveDefinedInSchema(fromContextDirective)) {
       const fieldInParent = parentType.field(this.name);
-      if (fieldInParent && fieldInParent.arguments().some(arg => arg.appliedDirectivesOf(fromContextDirective).length > 0)) {
+      if (fieldInParent && fieldInParent.arguments()
+          .some(arg => arg.appliedDirectivesOf(fromContextDirective).length > 0 && (!this.args || this.args[arg.name] === undefined))
```
clenfest, line 381: Actually did I get this right?

sachindshinde (approved):

internals-js/src/operations.ts
sachindshinde: Looks right to me, it's saying that if there's a contextual argument, but there are no args (or there are args, but there's no entry for the contextual argument's name), then you can't rebase. Effectively it's saying that a required argument would be missing if the rebase were performed.

PR #3024, opened 2024-06-05
update context example

clenfest:


Meschreiber (approved):

docs/source/entities-advanced.mdx
```diff
@@ -977,18 +977,21 @@ type Query {
...
 
 type B @key(fields: "id") @context(name: "context_1") @context(name: "context_2") {
   id: ID!
   field: String!
+  someField: String!
   child: Child!
 } 
 
-type C @key(fields: "id") @context(name: "context_1") {
+type C @key(fields: "id") @context(name: "context_2") {
```
dariuszkuc, line 991: I think this should be 
```suggestion
type C @key(fields: "id") @context(name: "context_1") @context(name: "context_2") {
```

dariuszkuc: *as otherwise you cannot always resolve `prop1(arg: String! @fromContext(field: "$context_1 { field }")): Int!`

clenfest: Yeah, but in that case there is all of a sudden no point in having separate contexts. I think I may need to completely rework this.

PR #3026, opened 2024-06-08
docs: fix ContextFieldValue scalar

dariuszkuc:
`FieldValue` scalar was renamed to `ContextFieldValue`.

### DIFF
--- a/docs/source/federated-types/federated-directives.mdx
+++ b/docs/source/federated-types/federated-directives.mdx
@@ -960,21 +960,21 @@ type U @key(fields: "id") {
 The `@fromContext` directive sets the context from which to receive the value of the annotated field. The context must have been defined with the `@context` directive.
 
 ```graphql
-scalar FieldValue;
+scalar ContextFieldValue;
 
-directive @fromContext(field: FieldValue!) on ARGUMENT_DEFINITION;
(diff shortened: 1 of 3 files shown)


docs/source/federated-types/federated-directives.mdx
```diff
@@ -960,21 +960,21 @@ type U @key(fields: "id") {
 The `@fromContext` directive sets the context from which to receive the value of the annotated field. The context must have been defined with the `@context` directive.
 
 ```graphql
-scalar FieldValue;
+scalar ContextFieldValue;
 
-directive @fromContext(field: FieldValue!) on ARGUMENT_DEFINITION;
+directive @fromContext(field: ContextFieldValue) on ARGUMENT_DEFINITION;
```
dariuszkuc, line 965: @clenfest was changing it to nullable was intentional? IMHO it should be non-nullable...

dariuszkuc: guessing this is because of this rule
```
The argument must be nullable. Because validation is done at the subgraph level, the referenced field may become nullable when merging subgraphs, such as when the field is nullable in one subgraph but not in another.
```

clenfest (approved):

PR #3030, opened 2024-06-11
[Perf] Filter out undefined entries earlier in needsJoinField

tinnou:
## Description
This PR filters out undefined entries from the sources array in needsJoinField yielding a ~4x performance improvement when merging a large set of subgraphs.

Without the change timing of `mergeSubgraphs` function is `6:59.349 (m:ss.mmm)`, the flamegraph below shows lots of CPU time spent in `needsJoinField`:

Screenshot 2024-06-11 at 10 28 07 AM

With the change, timing of `mergeSubgraphs` function is `1:30.084 (m:ss.mmm)`, updated flamegraph afterwards:

Screenshot 2024-06-11 at 10 30 29 AM

We might be able to write a bench suite for composition, including large graphs as it's currently relatively easy to regress.

### DIFF
--- a/composition-js/src/merging/merge.ts
+++ b/composition-js/src/merging/merge.ts
@@ -1662,7 +1662,7 @@ class Merger {
     }
     
     // if there is a @fromContext directive on one of the source's arguments, we need a join__field
-    if (sources.some((s, idx) => {
+    if (sources.filter(isDefined).some((s, idx) => {
       const fromContextDirective = this.subgraphs.values()[idx].metadata().fromContextDirective();
       if (s && isFederationDirectiveDefinedInSchema(fromContextDirective)) {
         return s.kind === 'FieldDefinition' && s.arguments().some(arg => arg.appliedDirectivesOf(fromContextDirective).length > 0);


composition-js/src/merging/merge.ts
```diff
@@ -1662,7 +1662,7 @@ class Merger {
     }
     
     // if there is a @fromContext directive on one of the source's arguments, we need a join__field
-    if (sources.some((s, idx) => {
```
clenfest, line 1665: Unfortunately, this will introduce a bug because `idx` is the subgraph index which will be incorrect if the `undefined`s are pruned from the original array.

clenfest:
Hey @tinnou , I'm going to close this PR in favor of https://github.com/apollographql/federation/pull/3031 which should accomplish the same goal if that works for you.

composition-js/src/merging/merge.ts
tinnou: ah good catch, I initially had at the beginning of the closure: 
```typescript
if (!s) {
  return false;
}
``` but changed it to use `filter` thinking it was equivalent 🤦

PR #3041, opened 2024-06-19
generate proper errors for invalid context name

clenfest:
We were triggering an assert in extractSubgraphsFromSupergraph because we weren't properly detecting an invalid name at validation time. Make sure regex matches.


dariuszkuc (approved):

trevor-scheer (approved):
Thanks for the quick fix! The underscore validation right before seems redundant now but also not harmful.

PR #3042, opened 2024-06-19
repro bug with repeated fragments spreads and generate query fragments

xuorig:
`generateQueryFragments` does't handle repeated (`... on B { ... on B ... on B } } }`) fragments well. Given the query planner can actually generate these query structures, this causes this option to be potentially dangerous, generating invalid subgraph queries.

A failing test is included.

### DIFF
--- a/internals-js/src/__tests__/operations.test.ts
+++ b/internals-js/src/__tests__/operations.test.ts
@@ -34,6 +34,72 @@ function astSSet(...selections: SelectionNode[]): SelectionSetNode {
   };
 }
 
+describe('generate query fragments', () => {
+  test('generateQueryFragments handles repeated fragment spreads', () => {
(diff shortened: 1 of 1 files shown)


internals-js/src/__tests__/operations.test.ts
```diff
@@ -34,6 +34,72 @@ function astSSet(...selections: SelectionNode[]): SelectionSetNode {
...
+                      }
+                  }
+              }
+          }
+      }
+    `);
+
+    const withGeneratedFragments = operation.generateQueryFragments();
+    expect(withGeneratedFragments.toString()).toMatchString(`
+      {
```
xuorig, line 73: Ignore this expectation, the test fails due to validation errors before this.

PR #3045, opened 2024-06-24
Make enum addValue() faster by using an Map rather than an array

clenfest:
Merging enums was previously a O(n^3) operation.

### DIFF
--- a/internals-js/src/definitions.ts
+++ b/internals-js/src/definitions.ts
@@ -32,7 +32,7 @@ import {
   isCoreSpecDirectiveApplication,
   removeAllCoreFeatures,
 } from "./specs/coreSpec";
-import { assert, mapValues, MapWithCachedArrays, removeArrayElement, TimeOrderedSet } from "./utils";
+import { assert, mapValues, MapWithCachedArrays, removeArrayElement } from "./utils";
 import {
   withDefaultValues,
   valueEquals,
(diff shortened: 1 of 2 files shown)


internals-js/src/utils.ts
```diff
@@ -479,3 +479,29 @@ export function setsEqual<T>(s1: Set<T> | null, s2: Set<T> | null): boolean {
   }
   return true;
 }
+
+// ordering is done based on order that elements were added
+export class TimeOrderedSet<V extends { name: string }> {
+  private _values: V[] = [];
+  private _valueMap: Map<string, V> = new Map();
```
benjamn, line 486: In ECMAScript/JS, `Map` and `Set` are specified to preserve insertion order when you iterate over them, so you might not need this wrapper class?

internals-js/src/definitions.ts
```diff
@@ -2319,7 +2319,7 @@ export class EnumType extends BaseNamedType<OutputTypeReferencer, EnumType> {
     }
     const existing = this.value(toAdd.name);
     if (!existing) {
-      this._values.push(toAdd);
+      this._values.add(toAdd);
```
benjamn, line 2322: An important question with `Map` and `Set` is what you want to happen when you add an element that's already in the container. By default, it won't be bumped up to the most recent / last position, but you can achieve that behavior by deleting the item before adding it (if it already exists).

However, it looks like this code only ever calls `this._values.add` when the name does not yet exist, so using a `Map` for `this._values` should work without the trick of deleting elements.

benjamn (changes requested):
I'd like to see if this can be simplified to use a `Map` with no wrapper class:
```ts
private _values = new Map<string, EnumValue>();
```

internals-js/src/utils.ts
clenfest: Interesting. I didn't realize that. I'll change it.

internals-js/src/definitions.ts
clenfest: I think the default behavior should be fine.

benjamn (approved):

PR #3046, opened 2024-06-25
Connectors

dylan-apollo:
TODO:

- [x] Merge #3048 and rebase this 
- [x] Merge #3049 and rebase this
- [x] Merge #3050 and rebase this

Summary:

- Removes support for `@source*` directives
- Adds support for `@connect` and `@source`

[CNN-217]: https://apollographql.atlassian.net/browse/CNN-217?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internals-js/src/error.ts
```diff
@@ -591,123 +591,10 @@ const INTERFACE_KEY_MISSING_IMPLEMENTATION_TYPE = makeCodeDefinition(
...
-  { addedIn: '2.7.0' },
-);
-
 const CONTEXTUAL_ARGUMENT_NOT_CONTEXTUAL_IN_ALL_SUBGRAPHS = makeCodeDefinition(
-  'CONTEXTUAL_ARGUMENT_NOT_CONTEXTUAL_IN_ALL_SUBGRAPHS',
-  'Argument on field is marked contextual in only some subgraphs',
-  { addedIn: '2.7.0' },
+    'CONTEXTUAL_ARGUMENT_NOT_CONTEXTUAL_IN_ALL_SUBGRAPHS',
+    'Argument on field is marked contextual in only some subgraphs',
+    { addedIn: '2.7.0' },
```
benjamn, line 597:
```suggestion
  'CONTEXTUAL_ARGUMENT_NOT_CONTEXTUAL_IN_ALL_SUBGRAPHS',
  'Argument on field is marked contextual in only some subgraphs',
  { addedIn: '2.7.0' },
```
indentation

internals-js/src/federation.ts
```diff
@@ -1315,28 +1297,15 @@ export class FederationMetadata {
...
-    }
-
-    const contextDirective = this.contextDirective();
-    if (isFederationDirectiveDefinedInSchema(contextDirective)) {
-      baseDirectives.push(contextDirective);
-    }
+      const contextDirective = this.contextDirective();
+      if (isFederationDirectiveDefinedInSchema(contextDirective)) {
+          baseDirectives.push(contextDirective);
+      }
```
benjamn, line 1303:
```suggestion
    const contextDirective = this.contextDirective();
    if (isFederationDirectiveDefinedInSchema(contextDirective)) {
      baseDirectives.push(contextDirective);
    }
```
Looks like 4 space tabs snuck in here

internals-js/src/federation.ts
```diff
@@ -1576,8 +1545,7 @@ export class FederationBlueprint extends SchemaBlueprint {
     for (const application of contextDirective.applications()) {
       const parent = application.parent;
       const name = application.arguments().name as string;
-      
-      const match = name.match(/^([A-Za-z]\w*)$/);
+const match = name.match(/^([A-Za-z]\w*)$/);
```
benjamn, line 1548:
```suggestion
      const match = name.match(/^([A-Za-z]\w*)$/);
```

internals-js/src/specs/connectSpec.ts
```diff
@@ -0,0 +1,183 @@
...
+  POST?: URLPathTemplate;
+  PUT?: URLPathTemplate;
+  PATCH?: URLPathTemplate;
+  DELETE?: URLPathTemplate;
+  body?: JSONSelection;
+  headers?: HTTPHeaderMapping[];
+};
+
+export const CONNECT_VERSIONS = new FeatureDefinitions<ConnectSpecDefinition>(connectIdentity)
+  .add(new ConnectSpecDefinition(new FeatureVersion(0, 1), new FeatureVersion(2, 7)));
```
benjamn, line 181: This version might need to be bumped to `FeatureVersion(2, 9)`?

composition-js/src/merging/merge.ts
```diff
@@ -2902,6 +2909,42 @@ class Merger {
     });
   }
 
+  // After merging, if we added any join__directive directives, we want to
+  // remove types imported from the original `@link` directive to avoid
+  // orphaned types. When extractSubgraphsFromSupergraph is called, it will
+  // add the types back to the subgraph.
+  //
+  // TODO: this doesn't handle renamed imports
+  private removeTypesAfterJoinDirectiveSerialization(schema: Schema) {
```
benjamn, line 2918: I believe this TODO means getting the `name` in `${namespace}__${name}` from the `link.arguments().args.import` argument rather than using `featureDefinition.allElementNames()` (since those are the original names, ignoring potential renaming).

However, I believe all the types we're removing here are the auto-imported types to support the various arguments of `@source` and `@connect`, like `connect__URLPathTemplate` or `connect_HTTPHeaderMapping`, so it's unlikely anyone would explicitly `@link`-import them, so they probably won't (ever?) be renamed, so this TODO is probably moot (or very low priority).

benjamn (commented):
Some very minor nits about rogue indentation, a `FeatureVersion` bump, and a preemptive explanation of an old TODO.

internals-js/src/federation.ts
```diff
@@ -1255,25 +1249,13 @@ export class FederationMetadata {
...
-  sourceTypeDirective(): Post20FederationDirectiveDefinition<SourceTypeDirectiveArgs> {
-    return this.getPost20FederationDirective(FederationDirectiveName.SOURCE_TYPE);
-  }
-
-  sourceFieldDirective(): Post20FederationDirectiveDefinition<SourceFieldDirectiveArgs> {
-    return this.getPost20FederationDirective(FederationDirectiveName.SOURCE_FIELD);
-  }
-
-  fromContextDirective(): Post20FederationDirectiveDefinition<{ field: string }> {
+fromContextDirective(): Post20FederationDirectiveDefinition<{ field: string }> {
```
lennyburdette, line 1252: nit: this indentation seems wrong?

dylan-apollo: Yeah... this project isn't set up to format everything with Prettier. I thought I fixed that in #3050 but that only reformats tests? I don't know enough about Prettier to try to fix this for real so, unfortunately, I gotta do more manual formatting 💀

benjamn (approved):

composition-js/src/merging/merge.ts
```diff
@@ -656,6 +658,10 @@ class Merger {
     // we want to make sure everything is ready.
     this.addMissingInterfaceObjectFieldsToImplementations();
 
+    // After converting some `@link`ed definitions to use `@join__directive`,
+    // we might have some imported scalars and input types to remove from the schema.
+    this.removeTypesAfterJoinDirectiveSerialization(this.merged);
```
sachindshinde, line 663: I'm a bit confused why we're merging subgraph types/directives that we later throw away here. (It sounds like we should update `isMergedType()` and `isMergedDirective()` to avoid merging them entirely.)

internals-js/src/specs/connectSpec.ts
```diff
@@ -0,0 +1,183 @@
...
+    }));
+
+    this.registerDirective(createDirectiveSpecification({
+      name: SOURCE,
+      locations: [DirectiveLocation.SCHEMA],
+      repeatable: true,
+      composes: false,
+    }));
+
+    this.registerType({ name: URL_PATH_TEMPLATE, checkOrAdd: () => [] });
```
sachindshinde, line 44: I'm curious why we don't use `createScalarTypeSpecification()` here (a comment would be helpful if there's a reason, at least).

internals-js/src/federation.ts
```diff
@@ -1986,13 +1951,12 @@ export function buildSubgraph(
...
-    const schema = typeof source === 'string'
-      ? buildSchema(new Source(source, name), buildOptions)
-      : buildSchemaFromAST(source, buildOptions)
+    const parsed = typeof source === 'string' ? parse(source) : source;
+    const schema =  buildSchemaFromAST(parsed, buildOptions)
     subgraph = new Subgraph(name, url, schema);
   } catch (e) {
     if (e instanceof GraphQLError && name !== FEDERATION_UNNAMED_SUBGRAPH_NAME) {
-      throw addSubgraphToError(e, name, ERRORS.INVALID_GRAPHQL);
+      throw addSubgraphToError(e, name);
```
sachindshinde, line 1959: May be useful to know what the reasons for these changes are/if they have any user-visible behavior.

sachindshinde (changes requested):

composition-js/src/merging/merge.ts
dylan-apollo: Ah, this is the other action item I forgot about. @lennyburdette who should tackle looking into this?

.changeset/pre.json
```diff
@@ -0,0 +1,14 @@
+{
+  "mode": "pre",
+  "tag": "connectors",
```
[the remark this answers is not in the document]
dylan-apollo: Updated to be 2.10 with `-alpha`

internals-js/src/specs/connectSpec.ts
lennyburdette: using createScalarTypeSpecification — i noticed that there isn't `createInputObjectTypeSpecification`, is that something i should also add and use?

lennyburdette: https://github.com/apollographql/federation/pull/3118/commits/9f34ce4c6379554019ed156b67a49cf26b5aeb82

composition-js/src/merging/merge.ts
```diff
@@ -3120,6 +3131,15 @@ class Merger {
...
     }
 
+    const linkDirective = this.linkSpec.coreDirective(this.merged);
+    for (const link of linksToPersist) {
+      dest.applyDirective(linkDirective, {
+        url: link.toString(),
+        for: link.defaultCorePurpose,
+        feature: undefined
+      });
+    }
```
sachindshinde, line 3141: Two things here:
1. I'm assuming that the `@connect` spec, when used in the supergraph schema, is being used purely as a marker spec.
2. This code will break when multiple versions of the `@connect` spec in subgraphs come about. (In general, there's a better place to put this kind of code.)

internals-js/src/specs/connectSpec.ts
```diff
@@ -0,0 +1,148 @@
...
+    connect.addArgument(SOURCE, schema.stringType());
+
+    /*
+      input HTTPHeaderMapping {
+        name: String!
+        from: String
+        value: String
+      }
+    */
+    const HTTPHeaderMapping = schema.addType(new InputObjectType(this.typeNameInSchema(schema, HTTP_HEADER_MAPPING)!));
```
sachindshinde, line 78: Ultimately we should add a `createInputObjectTypeSpecification()` function and not override `addElementsToSchema()` to properly get check-or-add functionality, but doing that is more complicated than the other `createXTypeSpecification()` functions, due to some edge cases with references.

lennyburdette (approved):

sachindshinde (approved):
There's some things below we'll want before we make a full release, but this PR is fine for `2.10` pre-releases.

Outside of this PR's code, I do see some other bugs/edge cases in  `@join__directive` code, but given that it's only used by `@connect`, this is also fine for now.

PR #3049, opened 2024-06-25
Split out satisfiability from `composeServices`

dylan-apollo:
Create a `validateSatisfiability` function separate from `composeServces`. Allow fow disabling satisfiability when running `composeServices`.

This is part of #3046, split out for easier review.

### DIFF
--- a/composition-js/src/compose.ts
+++ b/composition-js/src/compose.ts
@@ -141,7 +141,7 @@ type ValidateSubgraphsAndMergeResult = MergeResult | { errors: GraphQLError[] };
  * @param subgraphs 
  * @returns ValidateSubgraphsAndMergeResult
  */
-export function validateSubgraphsAndMerge(subgraphs: Subgraphs) : ValidateSubgraphsAndMergeResult {
+function validateSubgraphsAndMerge(subgraphs: Subgraphs) : ValidateSubgraphsAndMergeResult {
   const upgradeResult = upgradeSubgraphsIfNecessary(subgraphs);
   if (upgradeResult.errors) {
     return { errors: upgradeResult.errors };
--- a/composition-js/src/compose.ts
+++ b/composition-js/src/compose.ts
@@ -45,9 +45,95 @@ function validateCompositionOptions(options: CompositionOptions) {
   assert(!options?.allowedFieldTypeMergingSubtypingRules?.includes("list_upgrade"), "The `list_upgrade` field subtyping rule is currently not supported");
 }
 
-export function compose(subgraphs: Subgraphs, options: CompositionOptions = {}): CompositionResult {
+/**
+ * Used to compose a supergraph from subgraphs
+ * 
+ * @param subgraphs Subgraphs
+ * @param options CompositionOptions
+ * @param runSatisfiability Boolean - flag used to toggle satisfiability. Defaults to `true`
+ */
+export function compose(subgraphs: Subgraphs, options: CompositionOptions = {}, runSatisfiability = true): CompositionResult {
   validateCompositionOptions(options);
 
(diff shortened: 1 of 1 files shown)


composition-js/src/compose.ts
```diff
@@ -45,66 +45,105 @@ function validateCompositionOptions(options: CompositionOptions) {
...
-export function composeServices(services: ServiceDefinition[], options: CompositionOptions = {}): CompositionResult  {
+/**
+ * Method to validate and compose services
+ * 
+ * @param services List of Service definitions
+ * @param options CompositionOptions
+ * @param runSatisfiability Flag to toggle satisfiability check within composition. Defaults to `true`
+ * @returns CompositionResult
+ */
+export function composeServices(services: ServiceDefinition[], options: CompositionOptions = {}, runSatisfiability = true): CompositionResult  {
```
clenfest, line 99: Why don't we just add this to `CompositionOptions` rather than having an extra argument?

composition-js/src/compose.ts
```diff
@@ -45,66 +45,105 @@ function validateCompositionOptions(options: CompositionOptions) {
...
+  // We pass `null` for the `supportedFeatures` to disable the feature support validation. Validating feature support
+  // is useful when executing/handling a supergraph, but here we're just validating the supergraph we've just created,
+  // and there is no reason to error due to an unsupported feature.
+  const supergraph = supergraphSchema ? new Supergraph(supergraphSchema, null) : Supergraph.build(supergraphSdl);
+  const supergraphQueryGraph = buildSupergraphAPIQueryGraph(supergraph);
+  const federatedQueryGraph = buildFederatedQueryGraph(supergraph, false);
+  return validateGraphComposition(supergraph.schema, supergraph.subgraphNameToGraphEnumValue(), supergraphQueryGraph, federatedQueryGraph);
+}
+
+export function validateSubgraphsAndMerge(subgraphs: Subgraphs){
```
clenfest, line 136: Do you really want to export this function? I think it should be internal only, yes?

composition-js/src/compose.ts
```diff
@@ -45,66 +45,105 @@ function validateCompositionOptions(options: CompositionOptions) {
...
+} | { supergraphSdl: string, supergraphSchema?: never };
+
+/**
+ * Run satisfiability check for a supergraph
+ * 
+ * Can pass either the supergraph's Schema or SDL to validate
+ * @param args: SatisfiabilityArgs 
+ * @returns { errors? : GraphQLError[], hints? : CompositionHint[] }
+ */
+export function validateSatisfiability({ supergraphSchema, supergraphSdl} : SatisfiabilityArgs) : {
```
clenfest, line 123: Same here. Does this need to be exported?

composition-js/src/compose.ts
clenfest, line 136: Specify return type for readability.

composition-js/src/compose.ts
clenfest: Actually seems like yes from the fact that it's taking different types of parameters?

composition-js/src/compose.ts
tayrrible: addressed on 04db517c

composition-js/src/compose.ts
tayrrible: addressed on c28b45a5

composition-js/src/compose.ts
tayrrible: Yes, we are using this within the rust compiler for connectors

composition-js/src/compose.ts
tayrrible: addressed on 0357a23c

clenfest (approved):

PR #3056, opened 2024-06-28
Use a `Set` for referencers

clenfest:
When working with elements with many referencers, `addReferencer()` became very hot.
    
If you have eg. 100 referencers, each one is added one by one, and each new addition iterates the entire existing referencers list. This could cause quadratic performance.

This changes the `_referencers` values to be JS `Set`s. This maintains the same order for algorithms that rely on order.


dariuszkuc (approved):
Approved.

Follow up question though -> should we also convert other array usages to sets? `unionsWhereMember`/`allImplementations` etc?

goto-bus-stop:
It actually might be good to convert `unionsWhereMember` and `allImplementations`, too. I didn't do it because it's more work. But `allImplementations` in particular is used for the `runtimeTypesIntersect` check, which is currently O(n^2), and could be O(n) if `allImplementations` used a `Set`.

PR #3057, opened 2024-06-28
Make SchemaUpgrader faster

clenfest:
The idea is to make SchemaUpgrader faster by precomputing each type into a map that can be passed around once rather. Ideally we would be able to entirely get rid of `otherSubgraphs` as a parameter to SchemaUpgrader before this PR is ready.

Note that although this works with tests, I think there may be an issue with interfaces that isn't covered by tests, so do not merge unless someone has a chance to look at this with a critical eye. (I think the issue is that the map is built by iterating over `objectTypes` but later code iterates over `types()` which may include interfaces.


TylerBloom (approved):
LGTM

PR #3059, opened 2024-07-01
Control join directive usage from the directive definition spec

tninesling:
The primary issue with controlling the use of join directives via identity strings is that any logic relying on the presence of `@join__directive` won't work if the affected directive is imported from the federation spec.

More concretely, `@link(url: "https://specs.apollo.dev/source/v0.1", import: ["@sourceAPI"])` results in a `@join__directive` for each use of `@sourceAPI`. In contrast, `@link(url: "https://specs.apollo.dev/federation/v2.8", import: ["@sourceAPI"])` does not.

To make the second case work, we don't determine join usage by the URL used in the `@link`. Instead, each directive definition spec has a `usesJoinDirective` flag which determines this behavior. When we handle `@link` usages, composition will use a `@join__directive` to capture that link if _any_ of its imports have the `usesJoinDirective` flag set. The resulting behavior change is that the supergraph SDL will have a net-new `@join__directive` application if any of the `@source*` directives are imported from the federation spec instead of the source spec.

### DIFF
--- a/.changeset/thick-schools-divide.md
+++ b/.changeset/thick-schools-divide.md
@@ -1,0 +1,6 @@
+---
+"@apollo/composition": patch
+"@apollo/federation-internals": patch
+---
+
+When a directive composes with @join__directive, also use @join__directive for links that import that directive.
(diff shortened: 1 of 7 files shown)


tninesling:
The join directive usage is making the subsequent implementation of cost/listSize a bit too complicated. I think we can vastly simplify things by merging their arguments together in the default way instead of composing manually with the join spec. That approach doesn't need this initial PR, so I'm closing for now.

PR #3069, opened 2024-07-08
Use sparse `Map`s rather than dense `sources` arrays in `mergeType` code.

sachindshinde:
From @benjamn :

In the `mergeType` code responsible for merging subgraphs schemas into a supergraph schema, I noticed we were often looping over a `sources` array with as many elements as the number of subgraphs, inside a loop over all the fields currently being merged, leading to running time proportional to the product of the number of subgraphs and the total number of fields being merged.

This PR demonstrates that we can instead represent just the contributions from various subgraphs that are directly relevant to the current type or field being merged, so the `sources` data structure does not always remain as large as the number of subgraphs. Swapping the `type Sources<T> = Map<number, T | undefined>` type for all the array parameters can be done without altering behavior or performance (see the 2nd commit), and that foundation enables the sparse representation optimization (preserving behavior while improving performance; see the `addFieldsShallow` changes in the 3rd commit).

### DIFF
--- a/composition-js/src/merging/reporter.ts
+++ b/composition-js/src/merging/reporter.ts
@@ -1,6 +1,7 @@
 import { addSubgraphToASTNode, assert, ErrorCodeDefinition, joinStrings, MultiMap, NamedSchemaElement, printSubgraphNames, SubgraphASTNode } from '@apollo/federation-internals';
 import { ASTNode, GraphQLError } from 'graphql';
 import { CompositionHint, HintCodeDefinition } from '../hints';
+import { Sources } from './merge';
 
 export class MismatchReporter {
   pushError: (error: GraphQLError) => void;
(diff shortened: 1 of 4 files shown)


composition-js/src/merging/reporter.ts
```diff
@@ -143,7 +144,7 @@ export class MismatchReporter {
   // Not meant to be used directly: use `reportMismatchError` or `reportMismatchHint` instead.
   private reportMismatch<TMismatched extends { sourceAST?: ASTNode }>(
     supergraphElement:TMismatched | undefined,
-    subgraphElements: (TMismatched | undefined)[],
+    subgraphElements: Sources<TMismatched>,
```
sachindshinde, line 147: Note that when `includeMissingSources` is `true`, the absence of pairs in the `Map` will result in different behavior for this block down below:
```typescript
if (!subgraphElt) {
  if (includeMissingSources) {
    distributionMap.add('', this.names[i]);
  }
  continue;
}
```

PR #3074, opened 2024-07-10
Implement @cost and @listSize directives for demand control

tninesling:
# Overview

Implements two new directives for demand control, based on the [IBM Cost Specification](https://ibm.github.io/graphql-specs/cost-spec.html).

```
directive @cost(weight: Int!) on 
  | ARGUMENT_DEFINITION
  | ENUM
  | FIELD_DEFINITION
  | INPUT_FIELD_DEFINITION
  | OBJECT
  | SCALAR

directive @listSize(
  assumedSize: Int,
  slicingArguments: [String!],
  sizedFields: [String!],
  requireOneSlicingArgument: Boolean = true
) on FIELD_DEFINITION
```

The `@cost` directive allows users to specify a custom weight for fields, enums, input objects, and arguments. The weight is used in the demand control cost calculation, both for static estimates as well as actual cost calculations.

The `@listSize` directive allows users to specify expected sizes of list fields in their schema. This can be a static value set through `assumedSize` or a dynamic value using `slicingArguments` to get the value from some paging parameters.

## Differences from the spec

The main difference from the IBM spec is that we use an `Int!` for weight argument of `@cost`. This allows the parser to enforce this is parameterized with proper numeric values instead of finding out at runtime that an invalid `String!` weight was passed.

## Caveats for shared fields

When `@cost` or `@listSize` are used on a `@shareable` field with different values, the composed directive will use a merged value that takes the maximum weight or assumed size, when applicable.


internals-js/src/specs/listSizeSpec.ts
```diff
@@ -0,0 +1,40 @@
+import { DirectiveLocation } from 'graphql';
+import { createDirectiveSpecification } from '../directiveAndTypeSpecification';
```
clenfest, line 2: I'm open to other opinions, but my preference would be for both of these directives to live in the cost spec rather than having a seperate spec for each.

internals-js/src/federation.ts
```diff
@@ -1831,9 +1851,9 @@ export function setSchemaAsFed2Subgraph(schema: Schema, useLatest: boolean = fal
 
 // This is the full @link declaration as added by `asFed2SubgraphDocument`. It's here primarily for uses by tests that print and match
 // subgraph schema to avoid having to update 20+ tests every time we use a new directive or the order of import changes ...
-export const FEDERATION2_LINK_WITH_FULL_IMPORTS = '@link(url: "https://specs.apollo.dev/federation/v2.8", import: ["@key", "@requires", "@provides", "@external", "@tag", "@extends", "@shareable", "@inaccessible", "@override", "@composeDirective", "@interfaceObject", "@authenticated", "@requiresScopes", "@policy", "@sourceAPI", "@sourceType", "@sourceField", "@context", "@fromContext"])';
+export const FEDERATION2_LINK_WITH_FULL_IMPORTS = '@link(url: "https://specs.apollo.dev/federation/v2.9", import: ["@key", "@requires", "@provides", "@external", "@tag", "@extends", "@shareable", "@inaccessible", "@override", "@composeDirective", "@interfaceObject", "@authenticated", "@requiresScopes", "@policy", "@sourceAPI", "@sourceType", "@sourceField", "@context", "@fromContext"])';
```
clenfest, line 1854: Do you not want to add your directives to the FULL_IMPORTS?

internals-js/src/argumentCompositionStrategies.ts
```diff
@@ -59,4 +59,41 @@ export const ARGUMENT_COMPOSITION_STRATEGIES = {
...
+  },
+  NULLABLE_UNION: {
+    name: 'NULLABLE_UNION',
+    isTypeSupported: (_: Schema, type: InputType) => ({ valid: isListType(type) }),
+    mergeValues: (values: any[]) => {
+      if (values.every((v) => v === undefined)) {
+        return undefined;
+      }
+
+      const combined = new Set();
```
clenfest, line 88: We've talked recently about preserving ordering of directives. I know this might be a bit of a lost cause, but should we consider trying to do a better job of preserving the original ordering as best we can?

internals-js/src/specs/listSizeSpec.ts
```diff
@@ -0,0 +1,40 @@
...
+    super(new FeatureUrl(listSizeIdentity, 'listSize', version), minimumFederationVersion);
+
+    this.registerDirective(createDirectiveSpecification({
+      name: 'listSize',
+      locations: [DirectiveLocation.FIELD_DEFINITION],
+      args: [
+        { name: 'assumedSize', type: (schema) => schema.intType(), compositionStrategy: ARGUMENT_COMPOSITION_STRATEGIES.NULLABLE_MAX },
+        { name: 'slicingArguments', type: (schema) => new ListType(new NonNullType(schema.stringType())), compositionStrategy: ARGUMENT_COMPOSITION_STRATEGIES.NULLABLE_UNION },
+        { name: 'sizedFields', type: (schema) => new ListType(new NonNullType(schema.stringType())), compositionStrategy: ARGUMENT_COMPOSITION_STRATEGIES.NULLABLE_UNION },
+        { name: 'requireOneSlicingArgument', type: (schema) => schema.booleanType(), defaultValue: true, compositionStrategy: ARGUMENT_COMPOSITION_STRATEGIES.NULLABLE_AND },
```
clenfest, line 21: Since there will always be a default value, should we use just a regular AND?

composition-js/src/__tests__/compose.demandControl.test.ts
```diff
@@ -0,0 +1,552 @@
...
+          type Query {
+            sharedWithCost: Int @shareable @cost(weight: 10)
+          }
+        `)
+      };
+  
+      const result = composeServices([subgraphA, subgraphB]);
+      assertCompositionSuccess(result);
+      expect(result.hints).toMatchInlineSnapshot(`
+        Array [
```
clenfest, line 288: Wondering if we'll get requests to disable the hint.

composition-js/src/__tests__/compose.demandControl.test.ts
```diff
@@ -0,0 +1,552 @@
...
+            "message": "Directive @listSize is applied to \\"Query.sharedWithListSize\\" in multiple subgraphs with different arguments. Merging strategies used by arguments: { \\"assumedSize\\": NULLABLE_MAX, \\"slicingArguments\\": NULLABLE_UNION, \\"sizedFields\\": NULLABLE_UNION, \\"requireOneSlicingArgument\\": NULLABLE_AND }",
+            "nodes": undefined,
+          },
+        ]
+      `);
+    });
+  });
+});
+
+describe('demand control directive extraction', () => {
```
clenfest, line 353: I would put these tests into their own file extractSubgraphsFromSupergraph.demandControl.test.ts

composition-js/src/__tests__/compose.demandControl.test.ts
```diff
@@ -0,0 +1,552 @@
...
+// Used to test @listSize applications on FIELD_DEFINITION
+function fieldWithListSize(result: CompositionResult): FieldDefinition<ObjectType> | undefined  {
+  return result
+    .schema
+    ?.schemaDefinition
+    .rootType('query')
+    ?.field('fieldWithListSize');
+}
+
+describe('demand control directive composition', () => {
```
clenfest, line 205: What happens when a directive is renamed in some, but not all of the subgraphs? I honestly can't remember if this works with other directives or not.

internals-js/src/specs/listSizeSpec.ts
tninesling: I agree, although all my attempts to combine them have failed. When I've combined them into the same spec, I either get some `@listSize directive not known` error or a `duplicate inclusion of feature` error when I try to import both from the same place.

internals-js/src/argumentCompositionStrategies.ts
tninesling: The JS `Set` should preserve insertion order, so I think the ordering of arguments should be the same after this concatenation.

composition-js/src/__tests__/compose.demandControl.test.ts
tninesling: I'm not sure. I think it's important to let people know that we're aggregating costs on shared fields, as my expectation without any hint would be that the original weights would be preserved. I do not know how common it will be to try setting different costs for different subgraphs with a shared field, but I was thinking we could ship this simpler version and get some customer feedback. If customers want that use case, then we could have a future version of the cost spec use joins instead.

I tried implementing this with joins originally, but I ran into a few places where the join spec wasn't sufficient (some issues with renamed directives, not being available on certain locations, and not being fully implemented on the rust side of things).

composition-js/src/__tests__/compose.demandControl.test.ts
tninesling: The main reason I didn't do this is because I wanted to share the subgraph templates between the composition and extraction tests. The existing extraction tests are in internals-js, so importing them from here would introduce a circular dependency in those packages.

composition-js/src/__tests__/compose.demandControl.test.ts
tninesling: Just pushed a test case for this. It results in a `LINK_IMPORT_NAME_MISMATCH` error

internals-js/src/specs/listSizeSpec.ts
tninesling: It looks like the types don't quite match up in this case. I need to use `schema.booleanType()` here so the directive definition matches the spec (ie. the directive has `requireOneSlicingArgument: Boolean = true`). The argument merger doesn't do any auto-upgrading if it has a nullable type with a default value, so it has to handle `schema.booleanType()` which implies null/undefined handling. Setting the merger's require type to `new NonNullType(schema.booleanType())` results in a type mismatch error during composition.

internals-js/src/federation.ts
tninesling: Forgot to pull that one in. Just added in the latest commit. Thanks!

internals-js/src/specs/listSizeSpec.ts
tninesling: Combined to a single spec in https://github.com/apollographql/federation/pull/3074/commits/36d6c71e2c150b6aed0d0bc18362cf31609e7343

clenfest (approved):

PR #3086, opened 2024-07-16
Handle multiple composable directives from the same spec

tninesling:
# Overview

This PR targets the branch of an existing open PR. The main difference is that this one combines the new `@cost` and `@listSize` directives are combined into the same spec. This required a small update to how the supergraph schema is prepared.

When a directive spec definition is imported by a subgraph, it was previously included in the supergraph without an import argument. So a subgraph with `@link(url: "https://specs.apollo.dev/cost/v0.1", import: ["@cost"])`, is translated to the supergraph as `@link(url: "https://specs.apollo.dev/cost/v0.1")`. If multiple imports are included, the supergraph wouldn't get any of them. With this change, the supergraph now gets an equivalent `@link(url: "https://specs.apollo.dev/cost/v0.1", import: ["@cost"])`.

## Renaming

The core/link directive can have an argument called "as" to rename the directive from a spec. This would take a subgraph import `@link(url: "https://specs.apollo.dev/cost/v0.1", import: [{ name: "@cost", as: "@renamedCost" }])`, and translate it to `@link(url: "https://specs.apollo.dev/cost/v0.1", as: "renamedCost")`. With this change, the supergraph now gets an equivalent `@link(url: "https://specs.apollo.dev/cost/v0.1", import: [{ name: "@cost", as: "@renamedCost" }])`.

## Import aggregation

As we add directive spec features to the supergraph, we check for existing links with the same feature version (both the link identity and the version must match). If one is found, we append the new directive to the existing list of imports.


internals-js/src/specs/coreSpec.ts
```diff
@@ -552,6 +552,26 @@ export class CoreSpecDefinition extends FeatureDefinition {
...
 
+  applyFeatureAsLink(schema: Schema, feature: FeatureDefinition, purpose?: CorePurpose, imports?: CoreImport[]): GraphQLError[] {
+    const existing = schema.schemaDefinition.appliedDirectivesOf("link").find((link) => link.arguments().url === feature.toString());
+    if (existing) {
+      existing.remove();
+    }
+
+    const coreDirective = this.coreDirective(schema);
+    const args = {
+      url: feature.toString(),
```
clenfest, line 563: Nit, but I prefer naming the type rather than casting. That way you aren't fighting the type system.

```
    const args: LinkDirectiveArgs = {
      url: feature.toString(),
      import: (existing?.arguments().import ?? []).concat(imports?.map((i) => i.as ? { name: `@${i.name}`, as: `@${i.as}` } : `@${i.name}`)),
      feature: undefined,
    };
```

clenfest (approved):
Two minor comments, but otherwise LGTM.

PR #3087, opened 2024-07-17
Allow @interfaceObject to stand in for all runtime types

clenfest:
Rather than relying on the fact that one filteredPaths is reached, allow @interfaceObject types to intersect with all runtime types


duckki:
Wouldn't it make sense to deduplicate while computing the `filteredPaths` here?
```ts
    const filteredPaths = newSubgraphPathInfos.map((p) => p.path.path).filter((p) => isAbstractType(p.tail.type));
```

Also, I'm concerned about having duplicate paths in `newSubgraphPathInfos`, which means we do duplicate work as well.

clenfest:
> Wouldn't it make sense to deduplicate while computing the `filteredPaths` here?
> 
> ```ts
>     const filteredPaths = newSubgraphPathInfos.map((p) => p.path.path).filter((p) => isAbstractType(p.tail.type));
> ```
> 
> Also, I'm concerned about having duplicate paths in `newSubgraphPathInfos`, which means we do duplicate work as well.

I'm also concerned about why the duplicate paths exist, and I'm going to keep investigating that. I think that this is an ok compromise as it should unblock the customer and we can deliver a more comprehensive fix shortly. I agree that there is something else wrong, but I think this should be safe.

duckki (approved):
Looks good! I could reproduce the issue with the included test. And the fix makes sense.
Also, even without duplicate paths, interface object + 2 other paths with abstract types would cause this problem anyways. We can track the duplicate path issue separately.

PR #3088, opened 2024-07-17
chore(docs): rename supported subgraphs to compatible subgraphs

dariuszkuc:
Renaming page in order to avoid any confusion on what subgraph implementations are actually supported by Apollo.


Meschreiber (approved):
Makes sense to me!

PR #3092, opened 2024-07-19
Context names don't allow underscores. Change docs to be valid

clenfest:


Meschreiber (approved):
Thanks for fixing this! I think it's also worth calling out in a note. Will add that to this PR momentarily.

PR #3094, opened 2024-07-23
Don't ever add @shareable to subscription fields during schema upgrading

clenfest:
It could be valid to have two fields of the same that are not shareable if one of them is being overridden. Don't add @sharable to subscription fields and just let it be a composition error if neither are overrides. As @shareable is not allowed on a subscription field, they should never exist there in any case.

### DIFF
--- a/internals-js/src/__tests__/schemaUpgrader.test.ts
+++ b/internals-js/src/__tests__/schemaUpgrader.test.ts
@@ -361,4 +361,7 @@ test("don't add @shareable to subscriptions", () => {
 
   expect(printSchema(result.subgraphs!.get("subgraph1")!.schema!)).not.toContain('update: String! @shareable');
   expect(printSchema(result.subgraphs!.get("subgraph2")!.schema!)).not.toContain('update: String! @shareable');
+  
+  expect(result.subgraphs!.get("subgraph1")!.schema.type('Subscription')?.appliedDirectivesOf('@shareable').length).toBe(0);
+  expect(result.subgraphs!.get("subgraph2")!.schema.type('Subscription')?.appliedDirectivesOf('@shareable').length).toBe(0);
 });
--- a/internals-js/src/__tests__/schemaUpgrader.test.ts
+++ b/internals-js/src/__tests__/schemaUpgrader.test.ts
@@ -325,3 +325,40 @@ test("fully upgrades a schema with no @link directive", () => {
 }`
   );
 });
+
+test("don't add @shareable to subscriptions", () => {
+  const subgraph1 = buildSubgraph(
+    "subgraph1",
+    "",
+    `#graphql
+    type Query {
+      hello: String
+    }
+    
+    type Subscription { 
+      update: String!
+    }
+  `
+  );
+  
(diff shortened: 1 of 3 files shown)


internals-js/src/schemaUpgrader.ts
```diff
@@ -714,7 +714,7 @@ class SchemaUpgrader {
     // - to every "value type" (in the fed1 sense of non-root type and non-entity) if it is used in any other subgraphs
     // - to any (non-external) field of an entity/root-type that is not a key field and if another subgraphs resolve it (fully or partially through @provides)
     for (const type of this.schema.objectTypes()) {
-      if (type.hasAppliedDirective(keyDirective) || type.isRootType()) {
+      if (type.hasAppliedDirective(keyDirective) || (type.isRootType() && !type.isSubscriptionRootType())) {
```
tinnou, line 717: More a general question here, can @shareable be added on fields of Query or Mutation types? If yes:
- what does that mean in terms of planning? When will one subgraph be hit over another?
- the @shareable [spec/doc](https://www.apollographql.com/docs/federation/federated-schemas/sharing-types/#sharing-object-types) only specifies "value types", maybe it would be good to add Query and Mutation types to be included in that case or Subscription to be mentioned.

If not, then we could just by pass root types altogether? i.e
```
if (type.isRooType()) {
  continue;
}
if (type.hasAppliedDirective(keyDirective) {
  ...
} else {
  ...
}
```

internals-js/src/__tests__/schemaUpgrader.test.ts
```diff
@@ -325,3 +325,40 @@ test("fully upgrades a schema with no @link directive", () => {
...
+      update: String!
+    }
+  `
+  );
+  const subgraphs = new Subgraphs();
+  subgraphs.add(subgraph1);
+  subgraphs.add(subgraph2);
+  const result = upgradeSubgraphsIfNecessary(subgraphs);
+
+  expect(printSchema(result.subgraphs!.get("subgraph1")!.schema!)).not.toContain('update: String! @shareable');
```
tinnou, line 362: Adding the condition, this will go into the else block right which is:
```
else {
        const typeInOtherSubgraphs = Array.from(this.objectTypeMap.get(type.name)!.entries()).filter(([subgraphName, _]) => subgraphName !== this.subgraph.name);
        if (typeInOtherSubgraphs.length > 0 && !type.hasAppliedDirective(shareableDirective)) {
          type.applyDirective(shareableDirective);
          this.addChange(new ShareableTypeAddition(type.coordinate, typeInOtherSubgraphs.map(([s]) => s)));
        }
 }
```
Doesn't it mean in this case that @shareable will be added on the whole `Subscription` type itself? 
Hence the field will not have the @shareable and this test will seem to pass but the result will effectively be the same?

internals-js/src/schemaUpgrader.ts
clenfest: You are correct.

tinnou (approved):
Looks good thanks!

sachindshinde (approved):

PR #3096, opened 2024-07-24
Merge default array arguments for external types

tninesling:
When using an external type with a default array argument, the defaults from the different subgraphs were being compared with strict equality. Now, when we come across arrays, we use `arrayEquals` to check the equality of the individual elements in the arrays.


composition-js/src/merging/merge.ts
```diff
@@ -1794,7 +1794,9 @@ class Merger {
         if (!sameType(destArg.type!, arg.type!) && !this.isStrictSubtype(arg.type!, destArg.type!)) {
           invalidArgsTypes.add(name);
         }
-        if (destArg.defaultValue !== arg.defaultValue) {
+        const defaultIsArray = Array.isArray(destArg.defaultValue) && Array.isArray(arg.defaultValue);
```
sachindshinde, line 1797: Because default values can be any combination of arrays and objects (provided it makes sense for the type), I think this should be something closer to what's in the `mergeDefaultValue()` code (it uses `valueEquals()` after confirming both are not `undefined`).

sachindshinde (changes requested):

composition-js/src/merging/merge.ts
tninesling: Updated to use `valueEquals()` in [8d73b93](https://github.com/apollographql/federation/pull/3096/commits/8d73b937d2b9ec889854ba1db0c1292535173392)

duckki (approved):
Looks go to me. Sachin's comment has been addressed. And the test reproduces the issue and confirms that it's been fixed.

composition-js/src/merging/reporter.ts
```diff
@@ -182,7 +182,6 @@ export class MismatchReporter {
       }
     }
     const supergraphMismatch = (supergraphElement && mismatchAccessor(supergraphElement, true)) ?? '';
-    assert(distributionMap.size > 1, () => `Should not have been called for ${supergraphElement}`);
```
duckki, line 185: Maybe we should keep this `assert`? It make sense to assert that we have at least two things to report a mismatch here.

sachindshinde: Agreed with @duckki here, seems like we should keep the assert.

sachindshinde (approved):
Should be good to go after addressing @duckki 's comment above.

PR #3102, opened 2024-07-31
Add specific language to context docs that reference resolvers are necessary for contextual fields

clenfest:


docs/source/entities/use-contexts.mdx
```diff
@@ -191,3 +191,35 @@ type Transaction @key(fields: "id") {
   ): Int!
 }
 ``` 
+
+## A note on resolvers
+
+Fields that are used with the context directives must be resolvable via a [reference resolver](/federation/entities/#2-define-a-reference-resolver). Query plans that fetch contextual fields are based on the entity where the `@context` directive is set. Therefore, all subgraphs that resolve that entity must support looking up their fields based on the entity's `@keys` fields.
+
+Take the [previous example](#referencing-fields-across-subgraphs). Suppose the first subgraph has a single query to retrieve the logged in user:
```
Meschreiber, line 199: Added this anchor, as I assumed that's the example we're talking about. Could you please confirm @clenfest ?

docs/source/entities/use-contexts.mdx
```diff
@@ -191,3 +191,35 @@ type Transaction @key(fields: "id") {
...
+  me: User!
+} 
+```
+
+The following example resolver would not correctly support contextual variables:
+
+```typescript title="❌ Incorrect example" disableCopy showLineNumbers=false
+Query: {
+  me: () => ({
+    id: 1, userCurrency: { id: 431, currencyCode: 'USD'}
```
Meschreiber, line 212: Should this be `isoCode`?

```suggestion
    id: 1, userCurrency: { id: 431, isoCode: 'USD'}
```

docs/source/entities/use-contexts.mdx
```diff
@@ -191,3 +191,35 @@ type Transaction @key(fields: "id") {
...
+  })
+}
+```
+
+This is because the fetch set to subgraph 1 to retrieve the `isoCode` would be done on the entity. Therefore a section to resolve a User by keys would need to be added:
```
Meschreiber, line 217: This sentence confuses me. Does it mean something like this:

```suggestion
It's an adequate resolver because retrieving the `isoCode` relies on the retrieving the `User` entity itself. Therefore, a reference resolver for the `User` entity must be included:
```

Meschreiber: I'm also a bit confused by the examples--the incorrect example shows just a resolver for the `me` query, while the addition below is a reference resolver for `User`.  Would that `me` query resolver work as written with just the addition? Or does it have to be updated as well?

Meschreiber (commented):
Thanks for the addition @clenfest . I pushed some copyedits and have some inline clarifying questions.

Finally, it seems like this is a prerequisite and should be mentioned earlier, for example in a Note at the top of the relevant section(s).  (E.g.:

```
Note: Fields that are used with the context directives must be resolvable via a reference resolver. [Learn more](#link-to-section).
```

Does this apply to just [Referencing fields across subgraphs](https://www.apollographql.com/docs/federation/entities/use-contexts/#referencing-fields-across-subgraphs) or to all sections on the page?

PR #3103, opened 2024-08-01
add hasura to federation compatible subgraph libraries

praveenweb:
This PR adds Hasura as one of the Federation compatible subgraph solution in the docs page with support that is available as of today.


dariuszkuc:
Thanks for the PR! @praveenweb before we can merge this we need a reproducible results - i.e. all those results are auto generated based on the results from https://github.com/apollographql/apollo-federation-subgraph-compatibility. Please open a PR there by following the [subgraph maintainer implementation guide](https://github.com/apollographql/apollo-federation-subgraph-compatibility/blob/main/SUBGRAPH_GUIDE.md)

PR #3119, opened 2024-08-15
Fragment variable definitions erased in subgraph queries

### LINKED ISSUE #3112: Undefined variable reference in generated subgraph query
## Describe the bug
[This repo](https://github.com/carldunham/apollo-router-5648) includes schema and a sample query that fails validation using the new Rust validation code.

## To Reproduce
1. Setup services: clone the repo, run `make run`
3. Submit request in `query.graphql`, provide a value for "time" (can be any enum value).

## Expected behavior
The query should succeed.

## Output
```
{
  "errors": [
    {
      "message": "Variable \"$time\" is not defined.",
      "locations": [
        {
          "line": 1,
          "column": 572
        }
      ],
      "extensions": {
        "code": "GRAPHQL_VALIDATION_FAILED"
      }
    },
    {
      "message": "Variable \"$time\" is not defined.",
      "locations": [
        {
          "line": 1,
          "column": 572
        }
      ],
      "extensions": {
        "code": "GRAPHQL_VALIDATION_FAILED"
      }
    }
  ]
}
```
yet `$time` is clearly defined and used.

## Desktop (please complete the following information):
 - OS: iOS
 - Version 

## Additional context
Router version v1.52.0
Rover version 0.12.2

clenfest:
Fixes #3112

Before creating an Operation, we call `collectUsedVariables`, which will only pull values from the selection set and not from fragments. This isn't great, because a variable used in a fragment won't be collected, and it doesn't make sense to collect variables from fragments because it's before they are optimized and many will be unused.

The inelegant solution I came up with is to pass in available variables in calls to `optimize` or  `generateQueryFragments` for an operation where we can add back in the unused variables. This should be ok, because we are guaranteed that exactly one of them will get called by `toPlanNode`. Pretty sure there won't be too much overhead added because we'll only call this once per subgraph fetch.


duckki:
Rust QP went a different way by not using fragments with unexpected variables in them, instead of including added variables in the query (@goto-bus-stop). I hope Rust QP and JS QP go the same direction at least for now.

See this commit: https://github.com/apollographql/router/pull/5730/commits/8c8340d684e15969eecccadf23f9a2cae566b684

goto-bus-stop:
Yeah. I didn't realise this was the same case. Optimizing can introduce a new variable reference, but only inside a dead branch (otherwise, that fragment would not be reused). So there's three correct ways to go:
1. Removing dead branches when reusing a fragment
2. Ignoring fragments that introduce new variable dependencies
3. Counting variable references *after* optimizing

My preference was 2), because 1) is hard and 3) is adding a "dead" variable to the subgraph query. That said, a fix is a fix, and this is not common enough that it will be a big problem for mismatch testing with Rust, I think.

internals-js/src/operations.ts
```diff
@@ -1001,16 +1019,17 @@ export class Operation extends DirectiveTargetElement<Operation> {
...
-  generateQueryFragments(): Operation {
+  generateQueryFragments(allAvailableVariables: VariableDefinitions = new VariableDefinitions()): Operation {
     const [minimizedSelectionSet, fragments] = this.selectionSet.minimizeSelectionSet();
+    
     return new Operation(
       this.schema(),
       this.rootKind,
       minimizedSelectionSet,
-      this.variableDefinitions,
+      allAvailableVariables ? this.collectVariablesFromFragments(allAvailableVariables, fragments) : this.variableDefinitions,
```
sachindshinde, line 1032: Since the type of `allAvailableVariables` is `VariableDefinitions`, this ternary will always evaluate to true.

internals-js/src/operations.ts
```diff
@@ -1001,16 +1019,17 @@ export class Operation extends DirectiveTargetElement<Operation> {
       }
     }
 
-    return this.withUpdatedSelectionSetAndFragments(optimizedSelection, finalFragments ?? undefined);
+    return this.withUpdatedSelectionSetAndFragments(optimizedSelection, finalFragments ?? undefined, allAvailableVariables);
   }
 
-  generateQueryFragments(): Operation {
+  generateQueryFragments(allAvailableVariables: VariableDefinitions = new VariableDefinitions()): Operation {
```
sachindshinde, line 1025: I'm confused how fragment generation for an operation could introduce fragments with unseen variables (there's a case to be made for fragment reuse, where they can introduce fields that are never executed at runtime but introduce references, but we can't say the same for generation).

internals-js/src/operations.ts
```diff
@@ -934,25 +936,41 @@ export class Operation extends DirectiveTargetElement<Operation> {
...
       return this;
     }
-
+    
     return new Operation(
       this.schema(),
       this.rootKind,
       newSelectionSet,
-      this.variableDefinitions,
+      (allAvailableVariables && newFragments) ? this.collectVariablesFromFragments(allAvailableVariables, newFragments) : this.variableDefinitions,
```
sachindshinde, line 966: Note that this makes this method subtly incorrect, as it will omit the selection set's fragments in the variable definitions.

sachindshinde (changes requested):

internals-js/src/operations.ts
clenfest: I think you may be right about that. I had been thinking that we need to cover our bases by doing it for both paths, but I think this isn't necessary. I'll take it out.

duckki:
I don't have a strong opinion on which way we want to fix this in JS. I don't want to hold you back especially since this case is not common.

sachindshinde (approved):
I've tweaked the logic to keep variable definition ordering the same, but otherwise LGTM.

PR #3136, opened 2024-09-03
Handle edge cases with subgraph extraction logic

sachindshinde:
While reviewing recent code changes that released in `2.9.0`, I noticed a few bugs in the code. This PR contains fixes for subgraph extraction bugs.

Specifically, this PR updates subgraph extraction to use pre-existing functions/patterns instead of repeating logic, to avoid copy/pasting and bugs/divergence (the copied logic wasn't looking at spec renaming, for example).

### DIFF
--- a/internals-js/src/extractSubgraphsFromSupergraph.ts
+++ b/internals-js/src/extractSubgraphsFromSupergraph.ts
@@ -40,7 +40,7 @@ import { parseSelectionSet } from "./operations";
 import fs from 'fs';
 import path from 'path';
 import { validateStringContainsBoolean } from "./utils";
-import { CONTEXT_VERSIONS, ContextSpecDefinition, DirectiveDefinition, FeatureUrl, FederationDirectiveName, SchemaElement, errorCauses, isFederationDirectiveDefinedInSchema, printErrors } from ".";
+import { ContextSpecDefinition, CostSpecDefinition, SchemaElement, errorCauses, isFederationDirectiveDefinedInSchema, printErrors } from ".";
 
 function filteredTypes(
   supergraph: Schema,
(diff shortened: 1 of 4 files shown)


internals-js/src/supergraphs.ts
```diff
@@ -95,7 +103,27 @@ export function validateSupergraph(supergraph: Schema): [CoreFeatures, JoinSpecD
     throw ERRORS.INVALID_FEDERATION_SUPERGRAPH.err(
       `Invalid supergraph: uses unsupported join spec version ${joinFeature.url.version} (supported versions: ${JOIN_VERSIONS.versions().join(', ')})`);
   }
-  return [coreFeatures, joinSpec];
+
+  const contextFeature = coreFeatures.getByIdentity(ContextSpecDefinition.identity);
```
sachindshinde, line 107: This function seemed like the most natural place to fetch the context and core specs (if present), as it's also where we fetch the join spec (it's also the place we error if we do not know our own spec version).

internals-js/src/extractSubgraphsFromSupergraph.ts
```diff
@@ -478,78 +478,52 @@ function extractObjOrItfContent(args: ExtractArguments, info: TypeInfo<ObjectTyp
...
- * ```
- * results in a map entry of `cost -> cost`. This duals as a way to check if a directive
- * is included in the supergraph schema.
- * 
- * **Important:** This map does _not_ include directives imported from identities other
- * than `specs.apollo.dev`. This helps us avoid extracting directives to subgraphs
- * when a custom directive's name conflicts with that of a default one.
- */
-function getApolloDirectiveNames(supergraph: Schema): Record<string, string> {
-  const originalDirectiveNames: Record<string, string> = {};
```
sachindshinde, line 509: The logic in this function is already performed by `FeatureDefinition.directive()`, and there's a pre-existing pattern of adding helper methods to the spec's file to call `FeatureDefinition.directive()`. In this PR, we accordingly add `costDirective()` and `listSizeDirective()` to `costSpec.ts`.

Also, note the logic in this function:
1. Is checking identity via `identity.includes("specs.apollo.dev")` instead of strict equality.
2. Does not check for spec renaming via `as` (which is separate from import `as`).

tninesling (approved):
lgtm!

internals-js/src/extractSubgraphsFromSupergraph.ts
lrlna: @tninesling @sachindshinde to triple check, this fix was also mirrored in rust, right?

PR #3148, opened 2024-09-19
Demand control directive validations

tninesling:
Adds validations to subgraph directive applications, per the specification in https://ibm.github.io/graphql-specs/cost-spec.html#sec-Validation. These rules primarily assert that the @cost and @listSize directive arguments reference valid fields of the correct type for each application.


internals-js/src/federation.ts
```diff
@@ -1736,6 +1737,100 @@ export class FederationBlueprint extends SchemaBlueprint {
       }
     }
```
clenfest, line 1739: I know I didn't necessarily follow this for context, but we should probably make this a separate function.

internals-js/src/federation.ts
```diff
@@ -1736,6 +1737,100 @@ export class FederationBlueprint extends SchemaBlueprint {
...
+          { nodes: sourceASTs(application, parent) }
+        ));
+      }
+    }
+
+    // Validate @listSize
+    for (const application of listSizeDirective?.applications() ?? []) {
+      const parent = application.parent;
+      // @listSize can only be applied to FIELD_DEFINITION
+      if (parent instanceof FieldDefinition && parent.type) {
```
clenfest, line 1761: Do you want to just assert to get rid of the if?

internals-js/src/federation.ts
```diff
@@ -1736,6 +1737,100 @@ export class FederationBlueprint extends SchemaBlueprint {
...
+
+        // Validate assumed size, but we differ from https://ibm.github.io/graphql-specs/cost-spec.html#sec-Valid-Assumed-Size.
+        // Assumed size is used as a backup for slicing arguments in the event they are both specified. 
+        // The spec aims to rule out cases when the assumed size will never be used because there is always
+        // a slicing argument. Two applications which are compliant with that validation rule can be merged
+        // into an application which is not compliant, thus we need to handle this case gracefully at runtime regardless.
+        // We omit this check to keep the validations to those that will otherwise cause runtime failures.
+        //
+        // With all that said, assumed size should not be negative.
+        if (assumedSize && assumedSize < 0) {
```
clenfest, line 1779: Technically the logic of this is correct, but an `assumedSize` of 0 will not actually check it's size in the second part of the condition. Very much a nit, but I'd probably 
```suggestion
        if (assumedSize !== undefined && assumedSize < 0) {
```

clenfest (approved):

internals-js/src/federation.ts
tninesling: I kept this block, but moved the individual validations into separate functions.

internals-js/src/federation.ts
```diff
@@ -1055,6 +1056,115 @@ function validateShareableNotRepeatedOnSameDeclaration(
...
+  if (sizedFields.length) {
+    if (!parent.type || !isCompositeType(parent.type)) {
+      // The output type must have fields
+      errorCollector.push(ERRORS.LIST_SIZE_INVALID_SIZED_FIELD.err(
+        `Sized fields cannot be used because "${parent.type}" is not an object type`,
+        { nodes: sourceASTs(application, parent)}
+      ));
+    } else {
+      for (const sizedFieldName of sizedFields) {
+        const sizedField = parent.type.field(sizedFieldName);
```
sachindshinde, line 1149: Note that this validation is currently applied at the subgraph level, which means that if `sizedFields` references a field in another subgraph, then this validation will fail (that might be fine here, but wanted to make sure y'all were aware).

sachindshinde: (Since the merge logic unions `sizedFields` together, it shouldn't block anyone/the user can just add a `@listSize` to the other subgraph.)

tninesling: I'm not sure how if would be possible to reference something in another subgraph. Do you mean if the field's output type is @external?

internals-js/src/federation.ts
```diff
@@ -1055,6 +1056,115 @@ function validateShareableNotRepeatedOnSameDeclaration(
...
+  parent: FieldDefinition<CompositeType>,
+  errorCollector: GraphQLError[]
+) {
+  const { sizedFields = [] } = application.arguments();
+  // Validate sizedFields https://ibm.github.io/graphql-specs/cost-spec.html#sec-Valid-Sized-Fields-Target
+  if (sizedFields.length) {
+    if (!parent.type || !isCompositeType(parent.type)) {
+      // The output type must have fields
+      errorCollector.push(ERRORS.LIST_SIZE_INVALID_SIZED_FIELD.err(
+        `Sized fields cannot be used because "${parent.type}" is not an object type`,
```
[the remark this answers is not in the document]
sachindshinde: @tninesling Unions can have a selection, but their only field is `__typename`, and `__typename` doesn't return a list (so a field with `@listSize` returning a union would always fail validation regardless). Regarding error messages, you're right that people wouldn't know the phrase "composite type", but it would be fine saying e.g. "object or interface type".

internals-js/src/federation.ts
sachindshinde: @tninesling I mean if the field returned an entity type, and some other subgraph resolved a field for that entity.

sachindshinde: (I suppose the other subgraph might not have the field that returns that type, so you may not actually be able to apply `@listSize` in the other subgraph, though arguably this is just a limitation of the cost IBM spec in the context of federation, so it should be fine/could be addressed later.)

PR #3156, opened 2024-09-26
fix handling of __typename optimizations

dariuszkuc:
Fixes handling of a `__typename` selection during query planning process.

When expanding fragments we were keeping references to the same `Field`s regardless where those fragments appeared in our original selection set. This was generally fine as in most cases we would have same inline fragment selection sets across whole operation but was causing problems when we were applying another optimization by collapsing those expanded inline fragments creating a new selection set. As a result, if any single field selection (within that fragment) would perform optimization around the usage of `__typename`, ALL occurrences of that field selection would get that optimization as well. See example below

```graphql
foo {
  f1 {
    ... on Bar {
      __typename
      ...onBar # will be collapsed and sub selections will optimize __typename
    }
  }
  f2 {
    ...onBar # sub selections will get __typename optimization from above
  }
}

fragment onBar on Bar {
  b
  c
}
```

### DIFF
--- a/.changeset/shaggy-phones-know.md
+++ b/.changeset/shaggy-phones-know.md
@@ -1,0 +1,8 @@
+---
+"@apollo/query-planner": patch
+"@apollo/federation-internals": patch
+---
+
+Fixes handling of a `__typename` selection during query planning process.
+
+When expanding fragments we were keeping references to the same `Field`s regardless where those fragments appeared in our original selection set. This was generally fine as in most cases we would have same inline fragment selection sets across whole operation but was causing problems when we were applying another optimization by collapsing those expanded inline fragments creating a new selection set. As a result, if any single field selection (within that fragment) would perform optimization around the usage of `__typename`, ALL occurrences of that field selection would get that optimization as well.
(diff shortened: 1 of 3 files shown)


sachindshinde (approved):
Pushed up some changes in [this commit](https://github.com/apollographql/federation/pull/3156/commits/f168c7b4172c3e0f848208fe3b547b6bdab65fa0) to fix some things, other than that looks good to me.

To summarize:
1. In the changes to `optimizeSiblingTypenames()`, a new variable named `firstFieldIndex` was being set, that is expected to be the index of the first non-optimized-away field in `updatedSelections` (which is only populated if there are any selections to update, i.e. if a `__typename` field is found, or a field/inline fragment's selection set was recursively updated). The code before was not setting `firstFieldIndex` if `updatedSelections` was falsy, i.e. if no selections to update had been encountered by that point in the loop. I've updated `optimizeSiblingTypenames()` to always set `firstFieldIndex` when `firstFieldSelection` is set (regardless of whether `updatedSelections` actually gets populated).
2. I've moved the `Field`-copying code inside `FieldSelection.withAttachment()` into a new `Field.copy()` method that's next to the other `Field.with*()` methods. The idea here is that if someone adds an optional argument to `Field`'s constructor, it would be easier to not accidentally omit this argument copying if the methods that copy were inside `Field` (IMO we probably shouldn't be using optional arguments in those constructors to begin with, but that's not something to worry about dealing with now).
    - There's also a slight change here, in that we copy the attachments from the old field as well. This is for if we ever decide to add more attachments in the future, and shouldn't affect the case of the typename attachment (since we don't try to sibling-optimize the results of a sibling-optimized query). 
3. I've updated comments for the `attachement`->`attachment` rename.

PR #3158, opened 2024-09-27
fix: fix recursion logic in minimizeSelectionSet

dariuszkuc:
When generating fragments we were only "minimizing" subselections for fields and inline fragments that could be extracted. If inline fragment cannot be replaced with a fragment spread, we should still minimize its selection set as it potentially can be optimized as well.

### DIFF
--- a/.changeset/violet-cows-talk.md
+++ b/.changeset/violet-cows-talk.md
@@ -1,0 +1,5 @@
+---
+"@apollo/federation-internals": patch
+---
+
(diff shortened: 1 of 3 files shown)


TylerBloom (approved):
LGTM. Do you know if the Rust side needs the same fix?

dariuszkuc:
> LGTM. Do you know if the Rust side needs the same fix?

RS implementation is good -> that's how we found this.

goto-bus-stop (approved):
nice#

PR #3160, opened 2024-10-01
fix(graph-path): When eliminating suboptimal indirect paths, properly check for direct key edge

sachindshinde:
This PR fixes a bug where when removing suboptimal indirect paths, we didn't properly check for a direct key edge at the end of the potential direct path.


goto-bus-stop (approved):

goto-bus-stop:
Should this have a test? Here and in Rust?

duckki:
I confirmed this fixes the remaining mismatches from the graph in question.

PR #3162, opened 2024-10-04
fix: normalize field set selection

dariuszkuc:
`FieldSet` scalar represents a selection set without outer braces. This means that users could potentially specify some selections that could be normalized (i.e. eliminate duplicate field selections, hoist/collapse unnecessary inline fragments, etc). Previously we were using `@requires` field set selection AS-IS for edge conditions. With this change we will now normalize the `FieldSet` selections before using them as fetch node conditions.


duckki (approved):
Looks good!
I confirmed the mismatch example is also gone.

PR #3163, opened 2024-10-10
fix: ensure we are removing ALL useless groups

dariuszkuc:
When removing "useless" fetch nodes/groups we remove them in-place while still iterating over the same list. This leads to potentially skipping processing of some the children fetch nodes, as when we remove nodes we left shift all remaining children but the iterator keeps the old position unchanged effectively skipping next child.


goto-bus-stop (approved):
nice that stuff like this is like impossible to write in Rust 😁

PR #3170, opened 2024-10-17
remove vestigial connector preview errors

lennyburdette:
these errors are technically in fed 2.7 but were removed for 2.9 and beyond. we never formally documented the directives that would trigger them so they can be removed.

this file isn't generated is it?


Meschreiber (approved):
Thanks for updating!

The file isn't autogenerated afaik.

PR #3171, opened 2024-10-17
docs: fed versions

shorgi:
Fix spacing and line breaks for Fed versions page

### DIFF
--- a/docs/source/federation-versions.mdx
+++ b/docs/source/federation-versions.mdx
@@ -34,17 +34,17 @@ For a comprehensive changelog for Apollo Federation and its associated libraries
 
 <div>
 
-First release
+First release <br/><br/>
 
-**August 2024**
+**August 2024** <br/><br/>
 
 </div>
 
 <div>
 
-Minimum router version
+Minimum router version <br/><br/>
 
-**1.53.0**
+**1.53.0** <br/><br/>
 
 </div>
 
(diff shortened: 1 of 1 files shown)


Meschreiber (commented):
I wonder if there's a way to add spacing but keep things horizontal? (I feel like using an extra line break might not be necessary?)

shorgi:
> I wonder if there's a way to add spacing but keep things horizontal? (I feel like using an extra line break might not be necessary?)

I think this has been vertical all along? (Can't prove it with both old docs and wayback machine down ...)

Meschreiber:
> > I wonder if there's a way to add spacing but keep things horizontal? (I feel like using an extra line break might not be necessary?)
> 
> I think this has been vertical all along? (Can't prove it with both old docs and wayback machine down ...)

I mean the `First release: Date` vs `First release /n date

Current: 
image

This PR:
image

shorgi:
>
> I mean the `First release: Date` vs `First release /n date`
> 

ah, I meant that having the `/n` was what was there before in the old docs. (And there wasn't any period or colon after First release or Minimum router version. they were already on separate lines)

Meschreiber:
Got it. I have a preference for the horizontal version since it seems more natural/space-efficient for me, but I'm fine with whatever you decide!

Meschreiber (approved):

PR #3177, opened 2024-10-28
fix(query-graph): allow multiple best path options per subgraph

duckki:
It's necessary when advancing to a subgraph with multiple overlapping @interfaceObject implementations.


duckki:
Closing. We better add a validation rule preventing overlapping interface object declarations.

PR #3191, opened 2024-12-11
When we see that `@provides` specifies an overridden field, remove it from the field selection.

clenfest:


composition-js/src/__tests__/compose.test.ts
```diff
@@ -5290,3 +5290,37 @@ describe('@source* directives', () => {
     assertCompositionSuccess(result);
   });
 });
+
+it('errors on unused @external', () => {
```
clenfest, line 5294: This test is essentially moved from `subgraphValidation.test.ts`

composition-js/src/merging/merge.ts
```diff
@@ -2058,6 +2072,69 @@ class Merger {
       });
     }
   }
+  
+  private removeOverriddenFieldFromFieldSet <T extends FieldDefinition<ObjectType | InterfaceType> | InputFieldDefinition>(
```
dariuszkuc, line 2076: nit: I think the naming is somewhat confusing as this is specifically for `@provides` but function name doesnt mention it at all

composition-js/src/merging/merge.ts
```diff
@@ -2058,6 +2072,69 @@ class Merger {
...
+          validSelectionsIndex[validSelectionsIndex.length-1] = false;
+        }
+        return field;
+      }
+    });
+    
+    // now we should have a SelectionSet with an array of _selections the same length as `validSelectionsIndex`
+    // We should be able to filter out the ones that are invalid and we'll be left with the valid selection string
+    const selections = selectionSet.selections();
+    assert(selections.length === validSelectionsIndex.length, 'selection parsing failed');
```
dariuszkuc, line 2133: would it be possible to have some `overriddenCoordinates` that are applicable to different `@provides`? (so there would be no field set filtering required for this `@provides`)

composition-js/src/merging/merge.ts
```diff
@@ -2045,10 +2055,14 @@ class Merger {
       const external = this.isExternal(idx, source);
       const sourceMeta = this.subgraphs.values()[idx].metadata();
       const name = this.joinSpecName(idx);
+      
+      // fields in this subgraph that are no longer viable because they've been overridden 
+      const overriddenFields = this.completelyOverriddenFieldMap.get(idx);
```
sachindshinde, line 2060: Note that:
1. Since the result of overridden field removal may be to remove the `@provides` entirely, this could result in the `@join__field` becoming unnecessary. This means the computation of the "updated" `@provides` needs to be done before that, so that `needsJoinField()` can be aware of it. (Might want to put it in `FieldMergeContextProperties`.)
2. In `@join__field`, the `usedOverridden` argument is set based on whether the overridden field is "used" (i.e. is in `@key`/`@requires`/`@provides`/`@fromContext` or is an object field that implements some interface field in the subgraph). If it's used, then the field still needs to be in the overridden subgraph, but otherwise (for non-progressive overrides) it is effectively dropped from supergraph schema metadata.
    - This definition of "used" should account for the `@provides` effectively omitting (non-progressive) overridden fields. My suggestion here is that the `collectUsedFields()` function should collect two separate sets of `FieldDefinition`s: one for `@provides`, and one for the rest. Both of these should be stored in `FederationMetadata`.
    - The `FederationMetadata.isFieldUsed()` method should check both sets. You should add a new method to `FederationMetadata` called e.g. `isFieldUsedIgnoringProvides()` that only checks the non-`@provides` set.
    - In `validateOverride()`, when computing `overriddenFieldIsReferenced`, use `isFieldUsedIgnoringProvides()` instead of `isFieldUsed()` if `overrideLabel` is unset.

sachindshinde (changes requested):

composition-js/src/merging/merge.ts
```diff
@@ -3561,4 +3657,25 @@ class Merger {
     }
     return fields;
   }
+  
+  private validateAllExternalFieldsUsed(): void {
```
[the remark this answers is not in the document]
clenfest: Sorry, I should have been more explicit about this. The reason I did this is because we don't put the field back into the `@provides` when we call `extractSubgraphsFromSupergraph`, so unless we add a new field to the join spec to put the `@provides` back, we won't have a record of it. I can do that if you prefer.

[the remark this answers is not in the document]
clenfest: Thought about it some more and agree that we should add to the join spec.

PR #3192, opened 2024-12-11
docs: no router redeploy required when updating percent label

shorgi:
Clarify that no router redeploy is required to update percent label for progressive override

### DIFF
--- a/docs/source/schema-design/federated-schemas/entities/migrate-fields.mdx
+++ b/docs/source/schema-design/federated-schemas/entities/migrate-fields.mdx
@@ -286,10 +286,10 @@ A few strategies to mitigate this concern:
 
 ### Customizing progressive `@override` behavior with a feature flag service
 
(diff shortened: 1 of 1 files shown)


docs/source/schema-design/federated-schemas/entities/migrate-fields.mdx
```diff
@@ -286,10 +286,10 @@ A few strategies to mitigate this concern:
 
 ### Customizing progressive `@override` behavior with a feature flag service
 
-Out of the box, the router supports the `percent(x)` syntax for resolving labels based on a given percentage. Unfortunately, updating this number requires a subgraph publish and router redeploy.
+Out of the box, the router supports the `percent(x)` syntax for resolving labels based on a given percentage. Updating this number requires publishing the subgraph.
```
Meschreiber, line 289: Should we explicitly say that a router redeploy is not needed? Fine w/o it, too.

Meschreiber (approved):

PR #3202, opened 2025-01-16
docs: link directive

shorgi:
Clarify required `@link` directive for Fed 2

### DIFF
--- a/docs/shared/link-directive.mdx
+++ b/docs/shared/link-directive.mdx
@@ -1,4 +1,4 @@
-Subgraph schemas "opt in" to Federation v2 features by applying the `@link` directive to the `schema` type. The `import` list of this definition must include each federation-specific directive that the subgraph schema uses. In the example below, the schema uses the `@key` and `@shareable` directives:
+Subgraph schemas opt in to Federation v2 features by applying the `@link` directive to the `schema` type. You can optionally add an `import` list to this definition to include each federation-specific directive that the subgraph schema uses. In the example below, the schema uses the `@key` and `@shareable` directives:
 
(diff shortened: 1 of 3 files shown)


docs/shared/link-directive.mdx
```diff
@@ -0,0 +1,13 @@
+Subgraph schemas "opt in" to Federation v2 features by applying the `@link` directive to the `schema` type. The `import` list of this definition must include each federation-specific directive that the subgraph schema uses. In the example below, the schema uses the `@key` and `@shareable` directives:
```
Meschreiber, line 1: Not sure if `"opt in"` requires the quotes, but I'm a minimalist when it comes to characters like `"` and `(` so take that with a grain of sale.

Meschreiber (approved):
Thanks for tackling so quickly!

docs/shared/link-directive.mdx
dariuszkuc, line 1: > The `import` list of this definition must include
Imports are optional and it is perfectly fine to skip it. Side effect is that without explicit imports those directives have to be namespaced to the spec (or provided alias), i.e. `@key` becomes `@federation__key` without explicit import

shorgi: thanks @dariuszkuc, do these edits make sense:

```suggestion
Subgraph schemas opt in to Federation v2 features by applying the `@link` directive to the `schema` type. You can optionally add an `import` list to this definition to include each federation-specific directive that the subgraph schema uses. In the example below, the schema uses the `@key` and `@shareable` directives:
```

PR #3209, opened 2025-02-10
feat: onSchema will be used new hook implmenetation

kamila-brylewska-zendesk:
Hello,
We need new hook that will notify me just before new schema will be used. I created solution based on current hooks and I updated tests.


.changeset/metal-colts-begin.md
```diff
@@ -0,0 +1,5 @@
+---
+"@apollo/gateway": minor
+---
+
+feat: onSchema will be used new hook implementation
```
nogates, line 5: should be "add `onSchemaWillBeUsed` hook. Consumers can use this new hook to trigger custom logic just before a new schema is passed to the Gateway executor`

nogates (approved):

PR #3215, opened 2025-02-13
federation 2.10.0

lennyburdette:
Merging this should create a release PR that will actually cut the 2.10.0 packages.

### DIFF
--- a/composition-js/CHANGELOG.md
+++ b/composition-js/CHANGELOG.md
@@ -1,21 +1,5 @@
 # CHANGELOG for `@apollo/composition`
 
-## 2.10.0-alpha.4
-
-### Patch Changes
-
-- Updated dependencies []:
-  - @apollo/federation-internals@2.10.0-alpha.4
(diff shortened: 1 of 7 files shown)


composition-js/CHANGELOG.md
```diff
@@ -1,5 +1,21 @@
 # CHANGELOG for `@apollo/composition`
 
+## 2.10.0-alpha.4
```
dylan-apollo, line 3: Do we want to get rid of these?

lennyburdette: yeah good call, i'll remove all these manually

lennyburdette: requires a PR, this is a protected branch: https://github.com/apollographql/federation/pull/3216

dylan-apollo (approved):
LGTM (pending the other PR merging into this one). I'm not _positive_ that changeset will handle things correctly, but I don't see a way to test it either. So we'll just have to tweak if the generated PR doesn't look right.

PR #3222, opened 2025-03-03
Add support to automatically cancel query plan request if it takes longer than a pre-configured amount of time.

clenfest:


query-planner-js/src/buildPlan.ts
```diff
@@ -105,6 +105,7 @@ import { Conditions, conditionsOfSelectionSet, isConstantCondition, mergeConditi
 import { enforceQueryPlannerConfigDefaults, QueryPlannerConfig, validateQueryPlannerConfig } from "./config";
 import { generateAllPlansAndFindBest } from "./generateAllPlans";
 import { QueryPlan, ResponsePath, SequenceNode, PlanNode, ParallelNode, FetchNode, SubscriptionNode, trimSelectionNodes } from "./QueryPlan";
+import { performance } from 'perf_hooks';
```
dariuszkuc, line 108: guess this is a leftover from previous changes?

clenfest: No, this is to be able to use `performance.now()`

query-planner-js/src/buildPlan.ts
```diff
@@ -3531,10 +3539,10 @@ function computePlanInternal({
   let main: PlanNode | undefined = undefined;
   let primarySelection: MutableSelectionSet | undefined = undefined;
   let deferred: DeferredNode[] = [];
-
-  const { operation, processor } = parameters;
+  const { operation, processor, config } = parameters;
+  const deadline = performance.now() + config.maxQueryPlanningTime;
```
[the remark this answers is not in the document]
clenfest: `performance.now()` is a replacement for `Date.now` with a couple of differences:
1. It's fast and safe to use inside loops
2. It's more accurate for microsecond time resolution
3. Instead of time since the epoch, it's an accurate representation of time since some arbitrary initialization point in the past. So the time is always relatively accurate, but the number itself doesn't really mean anything.

query-planner-js/src/buildPlan.ts
```diff
@@ -445,6 +449,9 @@ class QueryPlanningTraversal<RV extends Vertex> {
   }
 
   private handleOpenBranch(selection: Selection, options: SimultaneousPathsWithLazyIndirectPaths<RV>[]) {
+    if (performance.now() > this.deadline) {
```
sachindshinde, line 452: We ended up adding deadline checks to some other parts in Rust QP (see [this commit](https://github.com/apollographql/router/pull/6840/commits/adc76b0d6d1e2c94bb35bd38c8497a65276a41cd)), can you also add those deadline checks to the corresponding parts in gateway's QP?

query-planner-js/src/config.ts
```diff
@@ -118,6 +118,11 @@ export type QueryPlannerConfig = {
    * If you aren't aware of this flag, you probably don't need it.
    */
     typeConditionedFetching?: boolean,
+    
+    /**
+     * The maximum time to wait for query planning to complete before giving up and throwing an exception. Defaults to 2 minutes
```
sachindshinde, line 123: You should mention in the docstring here that this configuration option is in milliseconds.

sachindshinde (changes requested):

PR #3226, opened 2025-03-10
[docs] Document new dry-run query plan header

smyrick:
This new header option was added but only mentioned in release notes


Meschreiber (approved):
Sorry to have missed this! Please feel free to ping in #inbox-docs if a review is lagging for something!

PR #3228, opened 2025-03-11
add connect/v0.2 and allow @connect on OBJECT

lennyburdette:
not making this conditional on spec version because that will be enforced by rust code

### DIFF
--- a/.changeset/lucky-pillows-rhyme.md
+++ b/.changeset/lucky-pillows-rhyme.md
@@ -1,0 +1,11 @@
+---
+"@apollo/composition": minor
+"@apollo/federation-internals": minor
+"apollo-federation-integration-testsuite": minor
+"@apollo/gateway": minor
+"@apollo/query-graphs": minor
+"@apollo/query-planner": minor
+"@apollo/subgraph": minor
+---
+
+Add connect spec v0.2
--- a/composition-js/src/__tests__/connectors.test.ts
+++ b/composition-js/src/__tests__/connectors.test.ts
@@ -121,6 +121,125 @@ describe("connect spec and join__directive", () => {
     }
   });
 
+  it("composes v0.2", () => {
+    const subgraphs = [
+      {
+        name: "with-connectors",
+        typeDefs: parse(`
+                    extend schema
+                    @link(
+                        url: "https://specs.apollo.dev/federation/v2.11"
+                        import: ["@key"]
+                    )
+                    @link(
+                        url: "https://specs.apollo.dev/connect/v0.2"
+                        import: ["@connect", "@source"]
+                    )
+                    @source(name: "v1", http: { baseURL: "http://v1" })
+
+                    type Query {
+                        resources: [Resource!]!
+                        @connect(source: "v1", http: { GET: "/resources" }, selection: "")
+                    }
+
+                    type Resource @key(fields: "id") @connect(source: "v1", http: { GET: "/resources" }, selection: "") {
+                        id: ID!
+                        name: String!
+                    }
+                `),
+      },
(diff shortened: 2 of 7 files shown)


.changeset/lucky-pillows-rhyme.md
```diff
@@ -0,0 +1,11 @@
...
+"@apollo/composition": minor
+"@apollo/federation-internals": minor
+"apollo-federation-integration-testsuite": minor
+"@apollo/gateway": minor
+"@apollo/query-graphs": minor
+"@apollo/query-planner": minor
+"@apollo/subgraph": minor
+---
+
+Add connect spec v0.2
```
dylan-apollo, line 11: Should we wait on the changelog entry until we know what all will be included in 0.2? I guess for changeset to bump versions it needs _something_, so maybe we leave ourselves a task to update this?

composition-js/src/__tests__/connectors.test.ts
```diff
@@ -121,6 +121,167 @@ describe("connect spec and join__directive", () => {
...
+    const result = composeServices(subgraphs);
+    expect(result.errors ?? []).toEqual([]);
+    const printed = printSchema(result.schema!);
+    expect(printed).toMatchInlineSnapshot(`
+      "schema
+        @link(url: \\"https://specs.apollo.dev/link/v1.0\\")
+        @link(url: \\"https://specs.apollo.dev/join/v0.5\\", for: EXECUTION)
+        @link(url: \\"https://specs.apollo.dev/connect/v0.2\\", for: EXECUTION)
+        @join__directive(graphs: [WITH_CONNECTORS_V0_1_], name: \\"link\\", args: {url: \\"https://specs.apollo.dev/connect/v0.1\\", import: [\\"@connect\\", \\"@source\\"]})
+        @join__directive(graphs: [WITH_CONNECTORS_V0_2_], name: \\"link\\", args: {url: \\"https://specs.apollo.dev/connect/v0.2\\", import: [\\"@connect\\", \\"@source\\"]})
```
dylan-apollo, line 187: Just so I understand, the idea is that the link shows the highest spec number so Router can double-check it's supported? And then the `join__directive`s show which spec is used for each subgraph?

We gotta make sure that when we expand and when we build up connectors for runtime that we're checking the `join__directive` for spec version and not the `link`, I get the feeling we're just checking link right now...

dylan-apollo: I see the comment in merge 👍. Let's either add a follow-up task to audit how we're parsing these or add it to the current open PR on router for `@connect` on type?

dylan-apollo (approved):
LGTM, I think my comments should just become follow-up tickets

composition-js/src/__tests__/connectors.test.ts
lennyburdette: we only put the directive in the supergraph so we can add `for: EXECUTION` and disallow the router from using a supergraph with features it doesn't understand

.changeset/lucky-pillows-rhyme.md
lennyburdette: yeah that's exactly right, i just need something for changeset to use, and we can edit the changelog for the official release. the ticket is CNN-656

composition-js/src/__tests__/connectors.test.ts
lennyburdette: to be more accurate, extract_subgraphs_from_supergraph converts the join__directive link into a normal link in the extracted subgraph. so yeah, we are using the link, but because it's in the subgraph, we know we're using the correct one

PR #3240, opened 2025-04-07
allow `@connect` as default import

lennyburdette:
according to https://specs.apollo.dev/link/v1.0/, we actually shouldn't be importing `@connect` — it's available by default when importing the spec. it can be renamed with `as:`. calling `import: ["@connect"]` should actually be importing something called `@connect__connect`, but we'll leave the current behavior.

this change applies only to join__directive serialization.

### DIFF
--- a/composition-js/src/__tests__/connectors.test.ts
+++ b/composition-js/src/__tests__/connectors.test.ts
@@ -121,6 +121,243 @@ describe("connect spec and join__directive", () => {
     }
   });
 
+  it("does not require importing @connect", () => {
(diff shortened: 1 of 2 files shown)


dylan-apollo (approved):
Pretty sure I've _done_ this before and it worked 🤔. Did something change at one point? Also, do we also need to update LSP?

lennyburdette:
it's never worked afaik

lennyburdette:
i don't think there's anything to change in the language server? maybe a snippet but i don't see one for the link

Issue #3113, opened 2024-08-12
Bug: Query Planner Fails to Build Query for Nested Entity Fields Decorated With `@requires`


ryan01010111:
### Issue Description

When querying for nested entity fields decorated with `@requires`, a gateway for a federated graph fails to build a query plan, resulting in error `"Cannot add selection of field \"X\" to selection set of parent type \"Y\""`. This is a critical issue for my org.

All `@apollo/*` packages at latest, and composition done with latest (v2.8.4) rules.

### Link to Reproduction

[...]

PR #3246, opened 2025-04-13
fix(query-planner): @requires fetch group optimization selects invalid key

ryan01010111:
### Notes
Fixes https://github.com/apollographql/federation/issues/3113.
FWIW, this fix has held up in production systems for 6+ months.

### Problem
Fetch group optimization/handling for `@requires` conditions "assumes" existence of a key in the parent group's schema. This leads to failed query planning in cases where the parent group's schema does not define the same keys for an entity as that of the child group.

### Fix Summary
- Added logic that resolves a select-able key field from a fetch group's parent, and incorporated it into the fetch group optimization flow.

### DIFF
--- a/.changeset/pink-pans-stare.md
+++ b/.changeset/pink-pans-stare.md
@@ -1,0 +1,5 @@
+---
+"@apollo/query-planner": patch
+---
+
+Fixed a bug where, when composing fetch groups, the query planner would select a key field on a parent group, without checking if the field is defined in the respective schema. This happened particularly when optimizing fetch groups that satisfy @requires conditions. With these changes, the query planner will properly resolve a selectable key from the parent group's schema.
--- a/query-planner-js/src/__tests__/buildPlan.test.ts
+++ b/query-planner-js/src/__tests__/buildPlan.test.ts
@@ -3162,6 +3162,157 @@ describe('@requires', () => {
       }
     `);
   });
+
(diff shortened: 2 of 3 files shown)


ryan01010111:
Could someone confirm requirements met for having this reviewed? @duckki @lennyburdette

duckki:
> Could someone confirm requirements met for having this reviewed? @duckki @lennyburdette

Thank you for creating this PR. The federation team will review.

sachindshinde:
Thanks for filing the PR and including a test. The issue more specifically is that `canSatisfyConditions()` guarantees that the subgraph has a locally satisfiable key, but it doesn't guarantee that any upstream parent groups will have that entire key.

This means that when `createPostRequiresGroup()` tries to fetch that key in `parent.group`, it might fail. You've updated the code in `addPostRequireInputs()`/downstream functions to look for a different key in the parent group, but that may not find one, in which case query planning will still error. I've filed a [separate PR](https://github.com/apollographql/federation/pull/3293) that effectively falls back to `group` if the locally satisfiable key can't be found in `parent.group`, and simplifies your test case a bit (there are some extra fields/recursion that aren't needed).

ryan01010111:
@sachindshinde Thank you very much for the review, and explanation! Do you happen to have an ETA on whatever release you expect the fix to go into?

PR #3249, opened 2025-04-14
WIP: Add @connect.batch.maxSize to connect spec

andrewmcgivery:
Add @connect.batch.maxSize to connect spec

### DIFF
--- a/internals-js/src/specs/connectSpec.ts
+++ b/internals-js/src/specs/connectSpec.ts
@@ -17,6 +17,7 @@ const SOURCE = "source";
 const URL_PATH_TEMPLATE = "URLPathTemplate";
 const JSON_SELECTION = "JSONSelection";
 const CONNECT_HTTP = "ConnectHTTP";
+const CONNECT_BATCH = "ConnectBatch";
 const SOURCE_HTTP = "SourceHTTP";
 const HTTP_HEADER_MAPPING = "HTTPHeaderMapping";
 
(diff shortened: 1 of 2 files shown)


internals-js/src/specs/connectSpec.ts
```diff
@@ -106,6 +107,10 @@ export class ConnectSpecDefinition extends FeatureDefinition {
       new ListType(new NonNullType(HTTPHeaderMapping));
     connect.addArgument('http', new NonNullType(ConnectHTTP));
 
+    const ConnectBatch  = schema.addType(new InputObjectType(this.typeNameInSchema(schema, CONNECT_BATCH)!));
+    ConnectBatch.addField(new InputFieldDefinition('maxSize')).type = schema.intType();
```
andrewmcgivery, line 111: I assume from a GQL perspective, this should be `maxSize`... but in rust it'll be named `max_size`? I assume this is fine?

lennyburdette (commented):
use the argument in this test https://github.com/apollographql/federation/blob/next/composition-js/src/__tests__/connectors.test.ts#L382 and this will be good to go

andrewmcgivery:
> use the argument in this test https://github.com/apollographql/federation/blob/next/composition-js/src/__tests__/connectors.test.ts#L382 and this will be good to go

Done!

lennyburdette (approved):

PR #3250, opened 2025-04-16
Add errors property to @connect and @source

andrewmcgivery:
Add errors property to @connect and @source

### DIFF
--- a/internals-js/src/specs/connectSpec.ts
+++ b/internals-js/src/specs/connectSpec.ts
@@ -27,9 +27,8 @@ const URL_PATH_TEMPLATE = 'URLPathTemplate';
 const JSON_SELECTION = 'JSONSelection';
 const CONNECT_HTTP = 'ConnectHTTP';
 const CONNECT_BATCH = 'ConnectBatch';
-const CONNECT_ERRORS = "ConnectErrors";
+const CONNECTOR_ERRORS = "ConnectorErrors";
 const SOURCE_HTTP = "SourceHTTP";
-const SOURCE_ERRORS = "SourceErrors";
 const HTTP_HEADER_MAPPING = 'HTTPHeaderMapping';
 
 export class ConnectSpecDefinition extends FeatureDefinition {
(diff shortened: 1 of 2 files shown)


internals-js/src/specs/connectSpec.ts
```diff
@@ -153,6 +155,11 @@ export class ConnectSpecDefinition extends FeatureDefinition {
     ConnectBatch.addField(new InputFieldDefinition('maxSize')).type = schema.intType();
     connect.addArgument('batch', ConnectBatch);
 
+    const ConnectErrors  = schema.addType(new InputObjectType(this.typeNameInSchema(schema, CONNECT_ERRORS)!));
```
dylan-apollo, line 158: Should we make this and the source one a single `ConnectorsErrors` type for simplicity since they're identical? Like how we share headers.

dylan-apollo (commented):
Make sure you also have an open task to update the LSP definitions of this and to write some docs about it!

internals-js/src/specs/connectSpec.ts
andrewmcgivery: 🤔 I think the definitions should be the same. The validation in Rust is slightly different (implementation is different and variables available are different) but yea, the actual fields should be identical. 

I'd have to change this in Rust too so I'll wait till you're done reviewing to actually execute this change to make sure we're still confident in it once you're done looking at both PRs 😅

internals-js/src/specs/connectSpec.ts
```diff
@@ -153,6 +155,11 @@ export class ConnectSpecDefinition extends FeatureDefinition {
...
+    ConnectErrors.addField(new InputFieldDefinition('message')).type = JSONSelection;
+    ConnectErrors.addField(new InputFieldDefinition('extensions')).type = JSONSelection;
```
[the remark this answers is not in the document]
andrewmcgivery: From the design:

> Configuring @source(error:) or @connect(error:) overrides default or fallback values for both message and extensions, even if extensions is not set.

I interpret that as saying that that both fields are optional... but yea, I agree that it's a bit of a confusing behaviour to have one of the fields just "disappear" if not set.

So, my rust implementation allows you to override just message or extensions (or both obviously). (see: https://github.com/apollographql/router/pull/7311/files#diff-1c93268ec072033d27937077ca358823a0569f2ac1919d2976059c7a625f1fd6)

internals-js/src/specs/connectSpec.ts
andrewmcgivery: Done!

dylan-apollo (approved):
LGTM

PR #3252, opened 2025-04-22
docs: review subgraph should contribute to rating

DaleSeo:
It seems a bit counterintuitive that the `Reviews` subgraph contributes to the `inStock` field of the `Product` type. I'd like to update the code example so that it contributes to the `rating` field instead. This change is also consistent with the code example in the overview section of [the document](http://apollographql.com/docs/graphos/schema-design/federated-schemas/entities/intro).


Meschreiber (approved):
Ah, good catch! Thanks as always @DaleSeo 🙇‍♀️

PR #3254, opened 2025-05-13
Add new CompositionOption to bail if total number of subgraph path infos in the stack gets above a configured value.

clenfest:

### DIFF
--- a/composition-js/src/compose.ts
+++ b/composition-js/src/compose.ts
@@ -39,6 +39,8 @@ export interface CompositionOptions {
   allowedFieldTypeMergingSubtypingRules?: SubtypingRule[];
   /// Flag to toggle if satisfiability should be performed during composition
   runSatisfiability?: boolean;
+  /// Maximum allowable number of outstanding subgraph paths to validate
+  maxValidationSubgraphPaths?: number;
 }
 
 function validateCompositionOptions(options: CompositionOptions) {
@@ -123,7 +125,7 @@ type SatisfiabilityArgs = {
  * @param args: SatisfiabilityArgs
  * @returns { errors? : GraphQLError[], hints? : CompositionHint[] }
  */
-export function validateSatisfiability({ supergraphSchema, supergraphSdl} : SatisfiabilityArgs) : {
+export function validateSatisfiability({ supergraphSchema, supergraphSdl} : SatisfiabilityArgs, options: CompositionOptions = {}) : {
   errors? : GraphQLError[],
   hints? : CompositionHint[],
 } {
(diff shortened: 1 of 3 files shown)


composition-js/src/validate.ts
```diff
@@ -720,13 +730,29 @@ class ValidationTraversal {
       subgraphNameToGraphEnumValue,
     );
   }
+  
+  pushStack(state: ValidationState) {
+    this.totalValidationSubgraphPaths += state.subgraphPathInfos.length;
```
duckki, line 735: I think that the number of `ValidationState` is correlated to`state.supergraphPath.nextEdges()`, than `subgraphPathInfos`.
Ref: https://github.com/apollographql/federation/blob/02ebb9ed09f157c4bcc00dbe27de6b57422b98cc/composition-js/src/validate.ts#L788

duckki (commented):
Look reasonable. Just a question.

composition-js/src/compose.ts
```diff
@@ -66,11 +68,15 @@ export function compose(subgraphs: Subgraphs, options: CompositionOptions = {}):
...
-      return { errors: satisfiabilityResult.errors };
+    try {
+      satisfiabilityResult = validateSatisfiability({
+        supergraphSchema: mergeResult.supergraph,
+      }, { maxValidationSubgraphPaths });
+      if (satisfiabilityResult.errors) {
+        return { errors: satisfiabilityResult.errors };
+      }
+    } catch (err) {
+      return { errors: [err] };
```
sachindshinde, line 79: Note there's an implicit unchecked cast here to `GraphQLError`; preferably we wouldn't do this (because if something were thrown that wasn't a `GraphQLError`, it could mean that callers of `compose()` could unexpectedly error).

I think the safer approach would be to either:
1. Have `pushStack()` potentially return an error, and if it does, have `ValidationTraversal.handleState()` and `ValidationTraversal.validate()` immediately return with that error.
    - We could alternatively push to `ValidationTraversal.validationErrors`, and then trigger the immediate return. (E.g., when `apollo-compiler` validation decides to stop because a limit has been exceeded, it generally doesn't return just the single limit-exceeded error.)
3. Have the `try`/`catch` check if the error is a `GraphQLError`, and if it isn't, re-throw.

composition-js/src/validate.ts
```diff
@@ -720,13 +730,29 @@ class ValidationTraversal {
       subgraphNameToGraphEnumValue,
     );
   }
+  
+  pushStack(state: ValidationState) {
```
sachindshinde, line 734: There's going to be a merge conflict with https://github.com/apollographql/federation/pull/3258 (the gist is "stack" gets renamed to "queue", and the `push()`/`pop()` methods renamed to `add()`/`poll()`).

composition-js/src/validate.ts
sachindshinde: There's more of an explanation in https://github.com/apollographql/federation/pull/3258 , but the gist is that the number of subgraph paths has a multiplicative effect that we want to watch out for, and this limit in particular is to avoid there being too many subgraph paths stored in the stack (which could cause the callee to run out of memory).

sachindshinde (changes requested):
After further investigation,

composition-js/src/validate.ts
clenfest: I'm going to let this get fixed in the other PR if / when we decide to merge it.

sachindshinde (approved):

PR #3274, opened 2025-06-11
add cacheTag directive

bnjjj:
Add support of `@cacheTag` directive.

```
directive @cacheTag(format: String) repeatable on FIELD_DEFINITION | OBJECT
```

### DIFF
--- a/composition-js/src/merging/merge.ts
+++ b/composition-js/src/merging/merge.ts
@@ -378,7 +378,7 @@ class Merger {
   private inaccessibleDirectiveInSupergraph?: DirectiveDefinition;
   private latestFedVersionUsed: FeatureVersion;
   private joinDirectiveFeatureDefinitionsByIdentity = new Map<string, FeatureDefinitions>();
-  private directiveUsingJoinDirective = new Set<string>();
+  private directivesUsingJoinDirective = new Set<string>();
   private fieldsWithFromContext: Set<string>;
   private fieldsWithOverride: Set<string>;
 
@@ -540,7 +540,7 @@ class Merger {
           compositionSpec,
         });
         if (compositionSpec.useJoinDirective) {
-          this.directiveUsingJoinDirective.add(nameInSupergraph);
+          this.directivesUsingJoinDirective.add(nameInSupergraph);
         }
       }
     }
@@ -620,6 +620,11 @@ class Merger {
     // application is an type-system element and we don't want to merge it).
     if (this.composeDirectiveManager.shouldComposeDirective({ subgraphName, directiveName: definition.name })) {
       return true;
+    }
+    if (this.directivesUsingJoinDirective.has(definition.name)) {
+      // This directive will be added as `@join__directive` by the `addJoinDirectiveDirectives`
+      // method. So, we skip the normal merging logic.
+      return false;
     }
     if (definition instanceof Directive) {
       // We have special code in `Merger.prepareSupergraph` to include the _definition_ of merged federation
@@ -2968,12 +2973,6 @@ class Merger {
   }
 
   private mergeAppliedDirective(name: string, sources: Sources<SchemaElement<any, any>>, dest: SchemaElement<any, any>) {
-    if (this.directiveUsingJoinDirective.has(name)) {
-      // This directive will be added as `@join__directive` by the `addJoinDirectiveDirectives`
-      // method. So, skip the normal merging logic here.
-      return;
-    }
-
     // TODO: we currently "only" merge together applications that have the exact same arguments (with defaults expanded however),
     // but when an argument is an input object type, we should (?) ignore those fields that will not be included in the supergraph
     // due the intersection merging of input types, otherwise the merged value may be invalid for the supergraph.
@@ -3168,7 +3167,7 @@ class Merger {
           // See if this directive is one of the directives that should use the @join__directive.
           if (
             !shouldIncludeAsJoinDirective
-            && this.directiveUsingJoinDirective.has(directive.name)
+            && this.directivesUsingJoinDirective.has(directive.name)
           ) {
             shouldIncludeAsJoinDirective = true;
             if (sourceFeature) {
--- a/internals-js/src/specs/cacheTagSpec.ts
+++ b/internals-js/src/specs/cacheTagSpec.ts
@@ -16,7 +16,7 @@ export class CacheTagSpecDefinition extends FeatureDefinition {
   public static readonly specName = CACHE_TAG;
   public static readonly identity = `https://specs.apollo.dev/${CacheTagSpecDefinition.specName}`;
 
-  constructor(version: FeatureVersion, minimumFederationVersion: FeatureVersion) {
+  constructor(version: FeatureVersion, minimumFederationVersion?: FeatureVersion) {
     super(
       new FeatureUrl(
         CacheTagSpecDefinition.identity,
@@ -36,8 +36,5 @@ export const CACHE_TAG_VERSIONS =
   new FeatureDefinitions<CacheTagSpecDefinition>(
     CacheTagSpecDefinition.identity
   ).add(
-    new CacheTagSpecDefinition(
-      new FeatureVersion(0, 1),
-      new FeatureVersion(2, 12),
-    ),
+    new CacheTagSpecDefinition(new FeatureVersion(0, 1)),
   );
--- a/internals-js/src/directiveAndTypeSpecification.ts
+++ b/internals-js/src/directiveAndTypeSpecification.ts
@@ -34,7 +34,7 @@ export type DirectiveSpecification = {
 
 export type DirectiveCompositionSpecification = {
   supergraphSpecification: (federationVersion: FeatureVersion) => FeatureDefinition,
-  useJoinDirective?: boolean,
+  useJoinDirective: boolean,
   argumentsMerger?: (schema: Schema, feature: CoreFeature) => ArgumentMerger | GraphQLError,
   staticArgumentTransform?: StaticArgumentsTransform,
 }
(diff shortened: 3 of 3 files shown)


internals-js/src/specs/cacheTagSpec.ts
```diff
@@ -0,0 +1,75 @@
...
+      ) repeatable on FIELD_DEFINITION
+        | OBJECT
+    */
+    const cacheTag = this.addDirective(schema, CACHE_KEY).addLocations(
+      DirectiveLocation.FIELD_DEFINITION,
+      DirectiveLocation.OBJECT,
+    );
+    cacheTag.repeatable = true;
+
+    cacheTag.addArgument(FORMAT, schema.stringType());
```
dariuszkuc, line 57: should this be non-null? i.e. what would be the use case for `@cacheTag` without any value?

internals-js/src/specs/cacheTagSpec.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    return [];
+  }
+
+  get defaultCorePurpose(): CorePurpose {
+    return 'EXECUTION';
+  }
+}
+
+export const CACHE_KEY_VERSIONS =
```
duckki, line 67: I hope we use `CACHE_TAG_VERSIONS` here as well, for consistency.

dariuszkuc (dismissed):
Please add directive validations to `FederationBlueprint#onValidation` ([link](https://github.com/apollographql/federation/blob/main/internals-js/src/federation.ts#L1585)).

We should check whether tag value is provided (i.e. non-null string) and also whether it is a valid selection (for dynamic values).

internals-js/src/specs/cacheTagSpec.ts
```diff
@@ -0,0 +1,75 @@
...
+  FeatureDefinition,
+  FeatureDefinitions,
+  FeatureUrl,
+  FeatureVersion,
+} from './coreSpec';
+import { Schema } from '../definitions';
+import { registerKnownFeature } from '../knownCoreFeatures';
+import { createDirectiveSpecification } from '../directiveAndTypeSpecification';
+
+export const cacheTagIdentity = 'https://specs.apollo.dev/cacheTag';
```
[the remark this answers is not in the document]
duckki: Yeah, I think we should once we finalize the implementation.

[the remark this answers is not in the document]
duckki: Wait a sec. I don't think we need a separate `cacheTag` spec at all.
- In subgraph, it will be a federation spec directive.
- In supergraph, it will be a join spec directive.
So, no need for a separate spec.

duckki (commented):
I think the best way to add `@cacheTag` was

- to make it a federation directive (no new link spec), but we need to bump up the federation version.
- to use `@join__directive` in supergraph (this was already the case).
- But, this combination was not a paved path. So, I implemented some paving for it.

I also added some subgraph validation tests (as a placeholder).
Sorry for pushing my commit without a prior discussion.

duckki (commented):
BTW, shall we rename `cachekey.test.ts` to `cachetag.test.ts`?

internals-js/src/specs/federationSpec.ts
```diff
@@ -181,6 +182,14 @@ export class FederationSpecDefinition extends FeatureDefinition {
     if (version.gte(new FeatureVersion(2, 9))) {
       this.registerSubFeature(COST_VERSIONS.find(new FeatureVersion(0, 1))!);
     }
+
+    if (version.gte(new FeatureVersion(2, 12))) {
+      this.registerDirective(createDirectiveSpecification({
+        name: FederationDirectiveName.CACHE_TAG,
+        locations: [DirectiveLocation.OBJECT, DirectiveLocation.INTERFACE, DirectiveLocation.FIELD_DEFINITION],
```
dariuszkuc, line 189: Whats the use case for applying this on interface? We always have to return a concrete object types so whats the expected behavior here? If I apply it on interface should that be inherited by implementations?

```graphql
interface Product @cacheTag(format: "product-${id}") {
  id: ID!
  title: String!
}

type Book @key(fields: "id") {
  id: ID!
  # other fields
}

type Movie @key(fields: "id") {
  id: ID!
  # other fields
}
```

composition-js/src/merging/merge.ts
```diff
@@ -418,6 +420,8 @@ class Merger {
     // Represent any applications of directives imported from these spec URLs
     // using @join__directive in the merged supergraph.
     this.joinDirectiveFeatureDefinitionsByIdentity.set(CONNECT_VERSIONS.identity, CONNECT_VERSIONS);
+    // Some federation directives are translated to @join__directive in the supergraph.
```
dariuszkuc, line 423: nit: we don't translate it, `@join__directive` is for preserving target directive information in the supergraph

internals-js/src/specs/federationSpec.ts
duckki: Exactly, the semantics will be as if all implementors have the same cache tag definition.
The motivation to allow interface was consistency with `@key`. But, I realized that it should be only allowed if there IS a `@key` on interface.

duckki: Added that limitation on the validation doc: https://docs.google.com/document/d/1oF28QJZGFbRRhGN4FBpD2Cbsrt2kecJkyoSrHzfjuUk/edit?usp=sharing

duckki:
BTW, we may not be able to merge into `next` just yet. I'm afraid there might be `next` releases between now and summit.

sachindshinde (changes requested):
Two things here:
1. There's no way to easily determine whether a supergraph schema is using `cacheTag` (you have to parse all `@join__directive` calls to know). This means gateway won't know it should reject such supergraph schemas, as it currently relies on checking for specs it knows (similar story if router wanted to check for entitlements/rate limits). The simplest way to fix this is to add a `cacheTag` spec to the supergraph schema. It doesn't necessarily need any directives in it, and can be just a marker spec; that's currently what the `connect` spec means in supergraph schemas, IIRC.
2. The implementation for `addJoinDirectiveDirectives()` has several things wrong with it. That may have been fine for connectors, but if we're going to use `@join__directive` more generally, we should fix the issues.
    - It currently relies on `schemaToImportNameToFeatureUrl`, which is computed by `computeMapFromImportNameToIdentityUrl()`. This function doesn't account for default-named directives, e.g. `@federation__cacheTag`, only imported ones. It should really be using `Schema.coreFeatures`, specifically the `CoreFeatures.sourceFeature()` function, which already accounts for this.
    - It checks `directive.name` directly against `link`, but it should really be checking against what the link's name in the subgraph schema is. 
    - It's re-parsing each `@link`'s information, when it should really just use `Schema.coreFeatures`.
    - For some reason, the code thinks it's possible `directive.name` can start with `@` (the line with `directive.name.startsWith('@')`), but it can't, and that part should be simplified.

duckki:
> Two things here:
> 
> 1. There's no way to easily determine whether a supergraph schema is using `cacheTag` (you have to parse all `@join__directive` calls to know). This means gateway won't know it should reject such supergraph schemas, as it currently relies on checking for specs it knows (similar story if router wanted to check for entitlements/rate limits). The simplest way to fix this is to add a `cacheTag` spec to the supergraph schema. It doesn't necessarily need any directives in it, and can be just a marker spec; that's currently what the `connect` spec means in supergraph schemas, IIRC.

I'd like to avoid adding a new supergraph spec link, since we have too many already. I wish there was a way to indicate the composition version in supergraph beyond the `join` spec version. Or, we could increment the join spec version just to indicate the federation version is 2.12+.

Having said that, in practice, older Router versions will fail to resolve `@cacheTag` after extracting subgraphs. Yes, it will be better to error at an earlier stage, but it's not disastrous. I'm not 100% sure if Gateway will fail or not. JS composition (via composer-tool) didn't fail. So, there is a concern.

> 2. The pre-PR/existing implementation for `addJoinDirectiveDirectives()` has several things wrong with it. That may have been fine for connectors, but if we're going to use `@join__directive` more generally, we should fix the issues.
>    
>    * It currently relies on `schemaToImportNameToFeatureUrl`, which is computed by `computeMapFromImportNameToIdentityUrl()`. This function doesn't account for default-named directives, e.g. `@federation__cacheTag`, only imported ones. It should really be using `Schema.coreFeatures`, specifically the `CoreFeatures.sourceFeature()` function, which already accounts for this.
>    * It checks `directive.name` directly against `link`, but it should really be checking against what the link's name in the subgraph schema is.
>    * It's re-parsing each `@link`'s information, when it should really just use `Schema.coreFeatures`.
>    * For some reason, the code thinks it's possible `directive.name` can start with `@` (the line with `directive.name.startsWith('@')`), but it can't, and that part should be simplified.

100% agreed. I'll address that.

internals-js/src/specs/federationSpec.ts
duckki: Thanks. Yes, I added that to the limitation doc as a validation rule.

composition-js/src/__tests__/cachetag.test.ts
```diff
@@ -0,0 +1,364 @@
...
+
+      directive @join__implements(graph: join__Graph!, interface: String!) repeatable on OBJECT | INTERFACE
+
+      directive @join__unionMember(graph: join__Graph!, member: String!) repeatable on UNION
+
+      directive @join__enumValue(graph: join__Graph!) repeatable on ENUM_VALUE
+
+      directive @join__directive(graphs: [join__Graph!], name: String!, args: join__DirectiveArguments) repeatable on SCHEMA | OBJECT | INTERFACE | FIELD_DEFINITION
+
+      directive @cacheTag(format: String!) repeatable on OBJECT | INTERFACE | FIELD_DEFINITION
```
dariuszkuc, line 57: I thought we only wanted this as a marker spec in the supergraph format and not a new directive?

duckki: No, it's a unfortunate side effect of adding the spec. I'll find a way to drop it.

PR #3293, opened 2025-07-29
fix(query-planner): Fix bug where subgraph jump key fields are fetched from the wrong group

sachindshinde:
When a `@requires` edge needs a subgraph jump, this would sometimes cause the locally satisfiable key fields needed by that jump to be fetched from the parent group instead of the current group (which apparently was done as an optimization). This can be a problem, since `canSatisfyConditions()` only really guarantees the current group has those key fields.

This PR updates the code using the parent group to now explicitly check whether the locally satisfiable key can be fetched from that group, and if not, falls back to the current group (adding it as a parent, if needed). This lets us keep the optimization and avoid unnecessarily changing query plans.

Fixes https://github.com/apollographql/federation/issues/3113 and https://github.com/apollographql/federation/issues/2987 , and derives a simpler example test from the one given in https://github.com/apollographql/federation/pull/3246 .

### DIFF
--- a/.changeset/afraid-singers-arrive.md
+++ b/.changeset/afraid-singers-arrive.md
@@ -2,4 +2,4 @@
 "@apollo/query-planner": patch
 ---
 
-Fix bug in query planning when a subgraph jump for @requires can sometimes try to fetch @key fields from a subgraph that doesn't have them.
+Fix bug in query planning where a subgraph jump for `@requires` can sometimes try to fetch `@key` fields from a subgraph that doesn't have them. This bug would previously cause query planning to error with a message that looks like "Cannot add selection of field `T.id` to selection set of parent type `T`".


.changeset/afraid-singers-arrive.md
```diff
@@ -0,0 +1,5 @@
+---
+"@apollo/query-planner": patch
+---
+
+Fix bug in query planning when a subgraph jump for @requires can sometimes try to fetch @key fields from a subgraph that doesn't have them.
```
duckki, line 5: Why don't we mention the error message here? It's a clearly recognizable symptom, even though this PR may not fix every one of them.

query-planner-js/src/buildPlan.ts
```diff
@@ -4814,13 +4814,34 @@ function createPostRequiresGroup(
...
+    // in `canSatisfyConditions()` guarantees such a locally satisfiable key
+    // exists in `group`), but it can also potentially come from `parent.group`,
+    // and previous code had (wrongfully) always assumed it could.
+    //
+    // To keep this previous optimization, we now explicitly check whether the
+    // known locally satisfiable key can be rebased in `parent.group`, and we
+    // fall back to `group` if it doesn't.
+    const keyCondition = getLocallySatisfiableKey(dependencyGraph.federatedQueryGraph, edge.head);
+    assert(keyCondition, () => `Due to @requires, validation should have required a key to be present for ${edge}`);
+    if (!keyCondition.canRebaseOn(typeAtPath(preRequireGroup.selection.parentType, requirePath.inGroup()))) {
```
duckki, line 4830: Wouldn't it make more sense to guard `if (triedToMerge)` condition up there, so it takes the `else`-branch if parent.group's key can't be used? I see some logic in the `else`-branch is a bit different. I'm not sure if that difference is ok.

duckki (commented):
It makes sense. But, I just have a question.

.changeset/afraid-singers-arrive.md
sachindshinde: That's fair, I've updated the changeset to include an example error message.

query-planner-js/src/buildPlan.ts
sachindshinde: It's a combination of a few things:
1. There is indeed currently a few differences in the logic:
    - The created group's parent relation's `path` isn't populated in the `triedToMerge` case, but is populated otherwise.
    - Since the created groups may be hoisted upstream in the `triedToMerge` case, we have to manually add a dependency to `requiresParent` then.
    - The `deferRef` isn't copied in the non-`triedToMerge` case, though this is likely just a bug. (I suspect there's a few separate bugs with `@defer` and conditions, though, and I'm pushing that to a later PR to fix.)
2. Pre-computing whether the key can be rebased takes cycles (we re-parse the key conditions, compute the type at path in `parent.group`, and then check whether the key conditions can be rebased on that type). This is currently only being done when no created groups (really unmerged groups) are present, but if we hoisted that check it would be computed more frequently.
3. We don't want to change query plans too significantly in a patch release.

duckki (approved):

PR #3301, opened 2025-09-23
fix form-data vulnerability with usage of unsafe random function

briannafugate408:
for choosing boundary. upgrade to patched versions 4.0.4, 3.0.4, 2.5.4

Co-authored-by: Chidimma O <conwuegb@users.noreply.github.com>


duckki (commented):
Looks good. I think we need a changeset entry for this security update, mentioning we are updating our form-data dependency versions due to vulnerabilities in the old versions.

PR #3305, opened 2025-09-29
fix(composition): Catch unsatisfiable  `@shareable` mutation fields

sachindshinde:
This PR fixes an issue where composition may erroneously declare a schema with a `@shareable` mutation field satisfiable when it shouldn't be.

Specifically, this PR updates composition satisfiability to ensure that all query paths under a `@shareable` mutation field can be satisfied starting at the same subgraph. (I.e., that no matter what selection set is on the `@shareable` mutation field, the query planner will not need to make multiple calls to that mutation field across subgraphs.)

This is done by partitioning subgraph paths in `ValidationState` by the initial subgraph when a shared mutation field is encountered, and keeping track of separate error stacks as well. If all these partitions experience satisfiability errors, then there's no way to satisfy the selection set containing the witness operation from each partition using a single `@shareable` mutation field call, and the user is asked to fix (at least) one of the partition's satisfiability errors.

This PR may cause composition errors for a previously successful composition if that composition was mistakenly satisfiable due to this bug. We've determined that this may not be a rare occurrence, and accordingly this fix will land in the next minor release. This PR is currently stacked on top of https://github.com/apollographql/federation/pull/3304 for tests to pass, but once that PR merges and is pulled into `next`, this PR's base will change to `next`.


composition-js/src/__tests__/validation_errors.test.ts
```diff
@@ -413,6 +413,151 @@ describe('when shared field has non-intersecting runtime types in different subg
...
+        The following supergraph API query:
+        mutation {
+          f {
+            y
+          }
+        }
+        cannot be satisfied by the subgraphs because:
+        - from subgraph \"A\":
+          - cannot find field \"F.y\".
+          - cannot move to subgraph \"B\", which has field \"F.y\", because type \"F\" has no @key defined in subgraph \"B\".
```
duckki: From the witness query, it's not clear why we need to move back to subgraph "B" here.
Maybe `mutation { f { x y }}` does have that problem.

composition-js/src/validate.ts
```diff
@@ -731,20 +1031,34 @@ class ValidationTraversal {
       subgraphNameToGraphEnumValue,
     );
   }
-  
+
   pushStack(state: ValidationState): { error?: GraphQLError } {
-    this.totalValidationSubgraphPaths += state.subgraphPathInfos.length;
+    if (state.subgraphPathInfos instanceof Array) {
```
duckki: nit: Why don't we add a method to abstract the number of paths calculation (similar to `allSubgraphPathInfos`)? That will simplify `pushStack` and `popStack` a bit.

composition-js/src/validate.ts
```diff
@@ -625,9 +804,105 @@ export class ValidationState {
     return { state: updatedState, hint };
   }
 
+  validateTransitionforSubgraphPaths(
```
duckki: nit: Shall we capitalize `for` here (and its references)?

composition-js/src/validate.ts
```diff
@@ -469,6 +479,73 @@ export class ValidationState {
...
 
+  // Returns whether the entire entire visit to this state can be skipped. If
+  // the state is partitioned, note that each individual partition must be
+  // skippable for the state to be skippable.
+  canSkipVisit(
+    subgraphNameToGraphEnumValue: Map<string, string>,
+    previousVisits: QueryGraphState<VertexVisit[]>,
+  ): boolean {
+    const vertex = this.supergraphPath.tail;
+    if (this.subgraphPathInfos instanceof Array) {
```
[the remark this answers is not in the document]
sachindshinde: Changed these to wrapper classes in the 3rd PR commit (in JS, classes have a bit more overhead compared to over languages, but if it makes the code clearer it's probably fine here).

composition-js/src/validate.ts
sachindshinde: Good catch, fixed in the 3rd PR commit.

duckki (approved):

PR #3307, opened 2025-09-30
Revert "Revert "fix(query-planner): Fix bug where subgraph jump key fields are fetched from the wrong group""

sachindshinde:
Reverts apollographql/federation#3300 (which reverted apollographql/federation#3293), as we're finishing up testing and the original PR should be fine to merge soon.


duckki (approved):
Looks good. (This was once approved and merged before.)

The results from the Rust version of this fix looked good (improved planning with no regression). So, I expect this PR will be positive as well.

PR #3310, opened 2025-10-01
fix: update connector spec to allow re-entry

dariuszkuc:
Updates connector spec to follow the same patterns as other federation spec blueprints (i.e. register types/directives in the constructor and use default logic for adding them to the schema that checks whether they need to be added or not).

### DIFF
--- a/internals-js/src/directiveAndTypeSpecification.ts
+++ b/internals-js/src/directiveAndTypeSpecification.ts
@@ -3,10 +3,10 @@ import {
   ArgumentDefinition,
   CoreFeature,
   DirectiveDefinition,
-  EnumType, InputObjectType,
+  EnumType,
   InputType,
   isCustomScalarType,
-  isEnumType, isInputObjectType,
+  isEnumType,
   isListType,
   isNonNullType,
   isObjectType,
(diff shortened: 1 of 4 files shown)


lennyburdette (approved):
thanks for doing this!

internals-js/src/directiveAndTypeSpecification.ts
```diff
@@ -251,6 +257,62 @@ export function createObjectTypeSpecification({
...
+                { nodes: existingField.sourceAST },
+            ));
+          }
+        }
+        return errors;
+      } else {
+        const createdType = schema.addType(new InputObjectType(actualName, asBuiltIn));
+        for (const { name, type, defaultValue } of expectedFields) {
+          const newField = createdType.addField(name, type);
+          if (defaultValue) {
```
sachindshinde, line 306: `defaultValue` is just a POJO in this case, so we can't rely on its JS truthiness; we can just unequivocally set `newField.defaultValue = defaultValue` instead.

internals-js/src/specs/connectSpec.ts
```diff
@@ -103,15 +89,82 @@ export class ConnectSpecDefinition extends FeatureDefinition {
...
+          name: SOURCE_HTTP,
+          inputFieldsFct: (schema, feature) => {
+            const jsonSelectionType =
+                lookupFeatureTypeInSchema<ScalarType>(JSON_SELECTION, 'ScalarType', schema, feature);
+            const httpHeaderMappingType =
+                lookupFeatureTypeInSchema<InputObjectType>(HTTP_HEADER_MAPPING, 'InputObjectType', schema, feature);
+            return [
+              {
+                name: 'baseURL',
+                type: schema.stringType()
```
sachindshinde, line 150: The field is `baseURL: String!`, so this needs a non-null wrapper.

internals-js/src/specs/connectSpec.ts
```diff
@@ -128,79 +181,149 @@ export class ConnectSpecDefinition extends FeatureDefinition {
...
+            type: (schema, feature) =>
+                lookupFeatureTypeInSchema<ScalarType>(JSON_SELECTION, 'ScalarType', schema, feature)
+          },
+          {
+            name: 'entity',
+            type: (schema) => schema.booleanType(),
+            defaultValue: false
+          },
+          {
+            name: 'errors',
```
sachindshinde, line 281: nit: The `errors` argument is ordered between `batch` and `selection` in the original code.

internals-js/src/specs/connectSpec.ts
```diff
@@ -128,79 +181,149 @@ export class ConnectSpecDefinition extends FeatureDefinition {
...
+          },
+          {
+            name: 'batch',
+            type: (schema, feature) =>
+                lookupFeatureTypeInSchema<InputObjectType>(CONNECT_BATCH, 'InputObjectType', schema, feature)
+          },
+          {
+            name: 'selection',
+            type: (schema, feature) =>
+                lookupFeatureTypeInSchema<ScalarType>(JSON_SELECTION, 'ScalarType', schema, feature)
```
sachindshinde, line 273: The argument is `selection: JSONSelection!`, so this needs a non-null wrapper.

internals-js/src/specs/connectSpec.ts
```diff
@@ -128,79 +181,149 @@ export class ConnectSpecDefinition extends FeatureDefinition {
...
-        baseURL: String!
-        headers: [HTTPHeaderMapping!]
-
-        # added in v0.2
-        path: JSONSelection
-        query: JSONSelection
-      }
+      directive @source(
+        name: String!
+        http: SourceHttp!
```
sachindshinde, line 296: This type should be `SourceHTTP!` instead of `SourceHttp!`.

sachindshinde (changes requested):
There's various changes we need to do to support spec input object types in the general case. For now though, we can do a one-off hack for connectors.

It's complicated to explain/implement, so I've filed a PR on top of this PR's branch at https://github.com/apollographql/federation/pull/3311. That PR also addresses all the comments I've made below. Once that PR merges into this one, it should be good to go.

PR #3314, opened 2025-10-17
fix: fixed demand control validations

dariuszkuc:
Validations are run against subgraphs so we need to verify `@cost`/`@listSize` from federation spec and not the supergraph cost spec.


tninesling (approved):
Changesets create the weirdest file names some times... 🍋 🦶

PR #3316, opened 2025-10-21
Merge `main` into `next` before 2.12.0 release candidate

benjamn:
Most of the commits visible below were on `main` but not yet on `next`. The merge commit at the end (https://github.com/apollographql/federation/pull/3316/commits/c8f5465f2a01bb00c0d5fe6bd38c3e2bc03bfb75) is my attempt to reconcile the branches without discarding any functionality, so it probably deserves review the most.

### DIFF
--- a/.changeset/lemon-toes-sort.md
+++ b/.changeset/lemon-toes-sort.md
@@ -1,0 +1,7 @@
+---
+"@apollo/federation-internals": patch
+---
+
+Fixed demand control validations
+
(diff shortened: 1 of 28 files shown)


sachindshinde (changes requested):
Still looking through it, but for quick/early feedback, I've noticed at least that you dropped `main` changes to these files:
`docs/source/schema-design/federated-schemas/entities/migrate-fields.mdx`
`docs/source/schema-design/federated-schemas/reference/directives.mdx`
was that intentional?

PR #3318, opened 2025-10-24
fix: handle @requires dependency on fields returned by @interfaceObject

dariuszkuc:
Depending on the merge order of the types, we could fail composition if a type that `@requires` data from an `@interfaceObject` is merged before the interface. Updated merge logic to use explicit merge order of scalars, input objects, interfaces and finally objects.


tninesling (approved):

sachindshinde (approved):
There's one fix needed for the changeset message below, but otherwise the changes should be fine.

PR #3321, opened 2025-11-04
composition updates related to merging auth directives

dariuszkuc:
Composition updates:
* reject authorization directives on interface fields
* enforce transitive authorization requirements on `@requires` and `@fromContext`
* stricter merge rules for `@requiresScopes` and `@policy` (change from `OR` to `AND` logic)
* automatically propagate authorization requirements from types to interfaces in the supergraph

### DIFF
--- a/.cspell/cspell-dict.txt
+++ b/.cspell/cspell-dict.txt
@@ -103,6 +103,7 @@ iface
 impelementation
 implemenations
 implementationt
+impls
 inacessible
 incopatibilities
 ineffectient
@@ -112,6 +113,7 @@ insteances
 instrospection
 Inteface
 Intelli
+intf
 invalidities
 isnodelike
 issuels
@@ -191,6 +193,7 @@ Renamers
 reoptimized
 repeateable
 reponse
+Reqs
 resovable
 Reviwed
 Rhai
--- a/internals-js/src/__tests__/subgraphValidation.test.ts
+++ b/internals-js/src/__tests__/subgraphValidation.test.ts
@@ -1672,15 +1672,15 @@ describe('@cacheTag', () => {
   it('applies on root field', () => {
     const doc = gql`
       extend schema
-      @link(
-        url: "https://specs.apollo.dev/federation/v2.12"
-        import: ["@cacheTag"]
-      )
+        @link(
+          url: "https://specs.apollo.dev/federation/v2.12"
+          import: ["@cacheTag"]
+        )
 
       type Query {
         f(x: Int!): String!
-        @cacheTag(format: "query-f-{$args.x}")
-        @cacheTag(format: "any-query")
+          @cacheTag(format: "query-f-{$args.x}")
+          @cacheTag(format: "any-query")
       }
     `;
     const name = 'S';
@@ -1690,10 +1690,10 @@ describe('@cacheTag', () => {
   it('applies on entity type', () => {
     const doc = gql`
       extend schema
-      @link(
-        url: "https://specs.apollo.dev/federation/v2.12"
-        import: ["@key", "@cacheTag"]
-      )
+        @link(
+          url: "https://specs.apollo.dev/federation/v2.12"
+          import: ["@key", "@cacheTag"]
+        )
 
       type P @key(fields: "id") @cacheTag(format: "p-{$.id}") {
         id: ID!
@@ -1716,10 +1716,10 @@ describe('@cacheTag', () => {
     `;
 
     expect(
-        buildForErrors(doc, {asFed2: true, includeAllImports: true}),
+      buildForErrors(doc, { asFed2: true, includeAllImports: true }),
     ).toStrictEqual(
-        // TODO
-        undefined,
+      // TODO
+      undefined,
     );
   });
 });


composition-js/src/merging/merge.ts
```diff
@@ -3606,4 +3689,344 @@ class Merger {
...
+  policies?: string[][];
+
+  satisfies(other: AuthRequirementsOnElement): boolean {
+    const authenticatedSatisfied = this.isAuthenticated || !other.isAuthenticated;
+    const scopesSatisfied = this.isSubset(this.scopes, other.scopes);
+    const policiesSatisfied = this.isSubset(this.policies, other.policies);
+    return authenticatedSatisfied && scopesSatisfied && policiesSatisfied;
+  }
+
+  isSubset(first: string[][] | undefined, second: string[][] | undefined): boolean {
```
sachindshinde: I don't think this function is quite right for use in `satisfies()`. In particular, what you're looking to check for is whether the first predicate implies the second predicate. For that to be the case with DNF-minus-NOT, it must be true that:
- For every conjunction `A` in the first predicate:
  - There must exist at least one conjunction `B` in the second predicate such that:
    - `A`'s members are a non-strict superset of `B`'s members.

E.g. `[["A", "B"], ["A", "C"]]` implies `[["A"]]`.

composition-js/src/merging/merge.ts
```diff
@@ -3606,4 +3689,344 @@ class Merger {
...
+
+  constructor(coordinate: string, directive: string) {
+    this.fieldCoordinate = coordinate;
+    this.directive = directive;
+  }
+
+  satisfies(authOnElement: AuthRequirementsOnElement | undefined): boolean {
+    if (authOnElement) {
+      const typeSatisfiesRequirements = this.type && this.type.satisfies(authOnElement);
+      const fieldSatisfiesRequirements = this.field && this.field.satisfies(authOnElement);
```
sachindshinde: You want a DNF conjunction of `type`'s and `field`'s auth requirements.

E.g. if `type` has `scopes` of `[["A"]]` and `field` has `scopes` of `[["B"]]`, then it's fine for a field in `@requires` to have `scopes` `[["A", "B"]]`.

composition-js/src/merging/merge.ts
```diff
@@ -3606,4 +3689,344 @@ class Merger {
...
+
+    if (requirements.isAuthenticated || requirements.scopes || requirements.policies) {
+      return requirements;
+    } else {
+      return;
+    }
+  }
+
+  private verifyAuthRequirementsOnSelectionSet(authRequirements: AuthRequirements, selectionSet: SelectionSet) {
+    const parentType = this.schema.type(selectionSet.parentType.name);
```
sachindshinde: Since the selection set started out being parsed on the merged schema, I'm not sure why we're refetching the parent type from the merged schema here.

composition-js/src/merging/merge.ts
```diff
@@ -3606,4 +3689,344 @@ class Merger {
...
+          for (const implType of parentType.possibleRuntimeTypes()) {
+            const requirementsOnImplType = this.authRequirementsOnElement(implType);
+            if (!authRequirements.satisfies(requirementsOnImplType)) {
+              const msg = `Field "${authRequirements.fieldCoordinate}" does not specify necessary @authenticated, @requiresScopes `
+                  + `and/or @policy auth requirements to access the transitive interface "${parentType}" data from ${authRequirements.directive} selection set.`;
+              throw ERRORS.MISSING_TRANSITIVE_AUTH_REQUIREMENTS.err(msg);
+            }
+            // re-check the implementation auth condition
+            this.verifyAuthOnFieldSelection(implType, selection, authRequirements);
+          }
```
sachindshinde: Since we've upwards-propagated auth requirements by this point, I don't think there's a reason to type explode here; it should be fine to just check the auth requirements on the interface field and the field's return type, before recursing into the field's selection set (if any).

It's particularly important not to type-explode in this case, since type explosion may lead to differing types downstream in the query due to output type covariance, and those types may have less-strict requirements than the abstract types they implement.

composition-js/src/merging/merge.ts
```diff
@@ -3606,4 +3689,344 @@ class Merger {
...
+            throw ERRORS.MISSING_TRANSITIVE_AUTH_REQUIREMENTS.err(msg);
+          }
+        }
+        this.verifyAuthRequirementsOnSelectionSet(authRequirements, selection.selectionSet);
+      }
+    }
+  }
+
+  private verifyAuthOnFieldSelection(parentType: ObjectType, selection: FieldSelection, authRequirements: AuthRequirements) {
+    const field = parentType.field(selection.element.name);
```
sachindshinde: If we don't type-explode as mentioned above, this function shouldn't have to refetch its `field` from the merged schema (it also accordingly won't need to take in `parentType`).

internals-js/src/argumentCompositionStrategies.ts
```diff
@@ -54,6 +62,104 @@ function unionValues(values: any[]): any {
...
+    result = deduplicateSubsumedValues(accumulator);
+  }
+  return result;
+}
+
+function filterNestedArrayDuplicates<T>(values: T[][][]): T[][][] {
+  const filtered: T[][][] = [];
+  const seen = new Set<string>;
+  values.forEach((value) => {
+    value.sort();
```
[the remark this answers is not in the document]
dariuszkuc: Indeed this does not sort inner most array values so given `[["c"], ["b", "a"]]` we end up with `[["b", "a"], ["c"]]` (instead of expected `[["a", "b"], ["c"]`). I'll add the extra sort

composition-js/src/merging/merge.ts
dariuszkuc: fixed in `isSubset` check

composition-js/src/merging/merge.ts
dariuszkuc: Indeed this is a leftover from some previous versions.

composition-js/src/merging/merge.ts
dariuszkuc: Yeah this was added when were not going to propagate auth reqs. This is redundant now. I'll remove it.

PR #3333, opened 2025-11-10
composition updates related to access control directives

dariuszkuc:


composition-js/src/merging/merge.ts
```diff
@@ -2954,6 +2981,182 @@ class Merger {
...
+
+    if (inaccessibleName && names.has(inaccessibleName)) {
+      this.mergeAppliedDirective(inaccessibleName, sources, dest);
+      names.delete(inaccessibleName);
+    }
+
+    if (dest instanceof InterfaceType) {
+      // we need to propagate access control from implementing types upward to the interface
+      for (const { name, nameInSupergraph } of this.accessControlDirectivesInSupergraph) {
+        if (nameInSupergraph) {
```
sachindshinde: If you changed `nameInSupergraph` to being non-`undefined` as mentioned above, you won't need this `if` check.

composition-js/src/merging/merge.ts
```diff
@@ -2954,6 +2981,182 @@ class Merger {
...
+          for (const application of accessControlDirective.applications()) {
+            const candidate = application.parent;
+            if (candidate instanceof ObjectType) {
+              // we will be propagating access control from objects up to the interfaces
+              // - record ALL implementations that define access control and implement an interface
+              const implementedInterfaces = implToIntfMap.get(candidate.name);
+              if (implementedInterfaces) {
+                for (const intf of implementedInterfaces) {
+                  // save implementation as additional source for impl
+                  const key = `${intf}_${accessControlDirective.name}`;
```
sachindshinde: nit: Slightly safer to use `intf.name` here instead of relying on `toString()`.

sachindshinde: Ah you're right it's already a string

PR #3337, opened 2025-11-11
fix: addressing feedback on updated access control logic

dariuszkuc:
Fixes:
* we now correctly sort all conditions so we can properly deduplicate entries in `dnfConjunction` 
* added extra type checks when validating usage of access control on interfaces
* fixed `AuthRequirementsOnElement.isSubset()` to verify that `@requires`/`@fromContext` field auth requirements satisfy transitive requirements
* fixed `AuthRequirements.satisfies()` logic to use DNF conjunction of auth requirements from type and field to get the final requirements to validate against
* cleaned up computation of additional access control sources
* fixed handling of renamed access control directives
* added comments around our usage of additional sources for access control
* updated logic to handle `@interfaceObject` -> object type -> other interface access control propagation chains
* `copyNonJoinAppliedDirectives()` was updated to pull additional sources from `accessControlAdditionalSources()` for auth directives, instead of just copying the interface field directives (which were already merged)
* update DNF handling logic to convert empty `[]` into `[[]]` to ensure we correctly calculate conjunctions for empty cases
* update `dnfConjunction()` to copy its inputs as it mutates the values when sorting

### DIFF
--- a/composition-js/src/merging/merge.ts
+++ b/composition-js/src/merging/merge.ts
@@ -14,6 +14,7 @@ import {
   connectIdentity,
   CONTEXT_VERSIONS,
   ContextSpecDefinition,
+  convertEmptyToTrue,
   CoreFeature,
   coreFeatureDefinitionIfKnown,
   coreIdentity,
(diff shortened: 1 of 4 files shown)


composition-js/src/merging/merge.ts
```diff
@@ -2993,22 +2989,24 @@ class Merger {
...
-          }
+        let additionalSources = this.accessControlAdditionalSources().get(`${dest.coordinate}_${name}`);
+        if (!additionalSources) {
+          additionalSources = [];
+        }
+        // WARNING: In order to propagate access control directive requirements, we are hijacking existing merge
+        // directive logic by providing "additional sources" from their interfaces/implementations. This means
+        // that their corresponding subgraphIndex will be incorrect and shouldn't be used/relied on when merging
+        // access control directives.
+        if (names.has(nameInSupergraph) || additionalSources.length) {
```
duckki: `additionalSources.length > 0` is clearer.

dariuszkuc: [878d2b6](https://github.com/apollographql/federation/pull/3337/commits/878d2b6dc7bdf9d4c7a00d5feed935c0d7edd8b9)

composition-js/src/merging/merge.ts
```diff
@@ -3123,27 +3134,31 @@ class Merger {
...
+                        // other interfaces implemented by the given type
+                        for (const intf of otherInterfaces) {
+                          if (objectName == intf) {
+                            // skip current @interfaceObject
+                            continue;
+                          }
+                          const key = `${intf}.${candidate.name}_${name}`;
+                          this.recordAdditionalSource(sourceMap, key, candidate);
+                        }
                       }
```
sachindshinde: It's possible that some of the object fields implement the same other interface fields, in which case this code will iterate through the same other interfaces more the once.

To fix this, you could:
1. While iterating through the `for` loop of `for (const impl of implementations)`, accumulate other interfaces/`implToIntfMap.get(impl)` into a `Set`.
2. After that iteration, remove the current interface object name/`objectName` from the `Set`.
3. Iterate through the interfaces of the `Set`, creating the key for the interface field and calling `this.recordAdditionalSource()`.

PR #3356, opened 2025-12-05
fix: add missing connect specs to the list of router supported features

dariuszkuc:
NOTE: We defined `DEFAULT_SUPPORTED_SUPERGRAPH_FEATURES` and `ROUTER_SUPPORTED_SUPERGRAPH_FEATURES` to distinguish between gateway and router feature sets. With completion of composition in Rust rewrite I don't think it makes much sense to keep those around (as everything is be supported by the router).

### DIFF
--- a/internals-js/src/supergraphs.ts
+++ b/internals-js/src/supergraphs.ts
@@ -1,10 +1,7 @@
 import { DocumentNode, GraphQLError } from "graphql";
 import { CoreFeatures, Schema, sourceASTs } from "./definitions";
 import {
-  CORE_VERSIONS,
   ErrCoreCheckFailed,
-  FeatureDefinition,
-  FeatureDefinitions,
   FeatureUrl,
   FeatureVersion,
 } from './specs/coreSpec';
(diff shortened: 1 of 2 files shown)


internals-js/src/supergraphs.ts
```diff
@@ -24,28 +40,19 @@ export const DEFAULT_SUPPORTED_SUPERGRAPH_FEATURES = new Set([
...
-  'https://specs.apollo.dev/join/v0.5',
-  'https://specs.apollo.dev/tag/v0.1',
-  'https://specs.apollo.dev/tag/v0.2',
-  'https://specs.apollo.dev/tag/v0.3',
-  'https://specs.apollo.dev/inaccessible/v0.1',
-  'https://specs.apollo.dev/inaccessible/v0.2',
-  'https://specs.apollo.dev/authenticated/v0.1',
-  'https://specs.apollo.dev/requiresScopes/v0.1',
-  'https://specs.apollo.dev/policy/v0.1',
-  'https://specs.apollo.dev/source/v0.1',
```
dariuszkuc: `source` spec was renamed to `connect` spec. `source` spec is not supported outside of some very old composition previews.

PR #3391, opened 2026-03-02
fix(composition): resolve @composeDirective definitions per-directive instead of per-spec

apollo-mateuswgoettems:
## Summary

Fixes a bug in `ComposeDirectiveManager` where `@composeDirective` would fail
with a false `DIRECTIVE_COMPOSITION_ERROR` when two subgraphs linked the same
custom spec but imported different subsets of its directives.

Closes FED-849

---

## Problem

`getLatestDirectiveDefinition()` resolved a directive's definition through a two-hop chain:

```
directiveName → spec identity → "latest" subgraph for that spec → schema.directive(...)
```

The "latest" subgraph was determined solely by its spec version (highest minor).
This meant that if the latest-version subgraph had not imported the specific directive
being resolved, the lookup would silently fail and produce a false error — even though
another subgraph had a perfectly valid definition for it.

**Failing scenarios:**

```graphql
# SG1 — imports @foo AND @bar, composes both
@link(url: "https://myorg.dev/myspec/v1.0", import: ["@foo", "@bar"])
@composeDirective(name: "@foo") @composeDirective(name: "@bar")

# SG2 — imports only @foo from the same spec
@link(url: "https://myorg.dev/myspec/v1.0", import: ["@foo"])
@composeDirective(name: "@foo")
```

If the iterator elected SG2 as "latest", resolving `@bar` would fail:

```
DIRECTIVE_COMPOSITION_ERROR: Core feature "https://myorg.dev/myspec" in subgraph "SG2"
does not have a directive definition for "@bar"
```

The bug was order-dependent: it succeeded only when the elected subgraph happened to be
a superset of all other subgraphs' imports. Completely disjoint imports (SG1 has `@foo`,
SG2 has `@bar`) always failed regardless of order.

---

## Fix

Added a `directiveDefinitionMap: Map<string, DirectiveDefinition>` built during `validate()`.
For each composed directive, it finds the highest-minor-version subgraph that **actually
imported** that specific directive, and stores its definition directly.
`getLatestDirectiveDefinition()` now does a single lookup on this map instead of the
indirect two-hop chain.

**Before:**
```
directiveName → spec identity → latestSubgraph (may not have the directive!) → definition
```

**After:**
```
directiveName → definition  (sourced from the subgraph that actually imported it)
```

---

## Changes

| File | Description |
|---|---|
| `composition-js/src/composeDirectiveManager.ts` | New `directiveDefinitionMap`, populated in `validate()`, `getLatestDirectiveDefinition()` simplified to a direct lookup |
| `composition-js/src/__tests__/compose.composeDirective.test.ts` | Two new regression tests covering the subset and fully-disjoint import scenarios; one existing test updated that was asserting the previous (buggy) error behavior |

---

## Testing

```bash
# Regression tests for FED-849
node_modules/.bin/jest --projects composition-js --no-coverage --testNamePattern="FED-849"

# Full composition suite
node_modules/.bin/jest --projects composition-js --no-coverage
```

All 15 test suites pass (400 tests, 0 failures).

### DIFF
--- a/.changeset/fix-compose-directive-subset-imports.md
+++ b/.changeset/fix-compose-directive-subset-imports.md
@@ -1,0 +1,29 @@
+---
+"@apollo/composition": patch
+---
+
+Fix `@composeDirective` failing with a false `DIRECTIVE_COMPOSITION_ERROR` when subgraphs import different subsets of directives from the same custom spec.
+
+**Affected scenario:** two or more subgraphs `@link` the same custom spec URL but import different directives from it:
+
+```graphql
+# subgraph A — imports @foo AND @bar, composes both
+@link(url: "https://myorg.dev/myspec/v1.0", import: ["@foo", "@bar"])
+@composeDirective(name: "@foo") @composeDirective(name: "@bar")
+
+# subgraph B — imports only @foo from the same spec
+@link(url: "https://myorg.dev/myspec/v1.0", import: ["@foo"])
+@composeDirective(name: "@foo")
+```
+
+This would produce a false error:
+```
+DIRECTIVE_COMPOSITION_ERROR: Core feature "https://myorg.dev/myspec" in subgraph "B"
+does not have a directive definition for "@foo"
+```
+
+The bug was order-dependent (failed when the subgraph elected as "latest" for the spec was not a superset of all other subgraphs' imports) and always failed for fully disjoint import sets (e.g. subgraph A imports only `@foo`, subgraph B imports only `@bar`).
+
+**Root cause:** `ComposeDirectiveManager.getLatestDirectiveDefinition()` resolved definitions through a two-hop chain — `directiveName → spec identity → "latest" subgraph for that spec → schema.directive(...)`. The "latest" subgraph was determined by spec version alone, without checking whether it had actually imported the directive being resolved.
+
+**Fix:** a `directiveDefinitionMap` (`directiveName → DirectiveDefinition`) is now built during `validate()`, sourcing each entry from the highest-minor-version subgraph that actually imported that specific directive. `getLatestDirectiveDefinition()` does a direct lookup on this map instead of the indirect chain.
--- a/.changeset/fix-compose-directive-subset-imports.md
+++ b/.changeset/fix-compose-directive-subset-imports.md
@@ -1,8 +1,8 @@
 ---
-"@apollo/composition": patch
+"@apollo/composition": minor
 ---
 
-Fix `@composeDirective` failing with a false `DIRECTIVE_COMPOSITION_ERROR` when subgraphs import different subsets of directives from the same custom spec.
+Update `@composeDirective` merge logic to allow subgraphs to import different directives from the same spec. Previously, all subgraphs had to import ALL (exactly the same) custom directives (even if they did not use them) as otherwise composition would fail with a false `DIRECTIVE_COMPOSITION_ERROR`.
 
 **Affected scenario:** two or more subgraphs `@link` the same custom spec URL but import different directives from it:
 
--- a/composition-js/src/__tests__/compose.composeDirective.test.ts
+++ b/composition-js/src/__tests__/compose.composeDirective.test.ts
@@ -156,6 +156,75 @@ describe('composing custom core directives', () => {
     expectDirectiveOnElement(schema, 'User.a', 'foo', { name: 'a' });
   });
 
+  // FED-849: composition should succeed when subgraphs import different subsets
+  // of directives from the same custom spec (e.g. SG1 imports [@foo, @bar], SG2 imports [@foo]).
+  it('two subgraphs with same spec but disjoint directive imports (FED-849)', () => {
+    // subgraphA imports both @foo and @bar from the spec and requests both to be composed
+    const subgraphA = generateSubgraph({
+      name: 'subgraphA',
+      linkText: '@link(url: "https://specs.apollo.dev/foo/v1.0", import: ["@foo", "@bar"])',
+      composeText: '@composeDirective(name: "@foo") @composeDirective(name: "@bar")',
+      directiveText: `
+        directive @foo(name: String!) on FIELD_DEFINITION
+        directive @bar(tag: String!) on FIELD_DEFINITION
+      `,
+      usage: '@foo(name: "a") @bar(tag: "t")',
+    });
+
+    // subgraphB imports only @foo from the same spec and requests only @foo to be composed
+    const subgraphB = generateSubgraph({
+      name: 'subgraphB',
+      linkText: '@link(url: "https://specs.apollo.dev/foo/v1.0", import: ["@foo"])',
+      composeText: '@composeDirective(name: "@foo")',
+      directiveText: 'directive @foo(name: String!) on FIELD_DEFINITION',
+      usage: '@foo(name: "b")',
+    });
+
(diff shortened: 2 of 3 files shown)


.changeset/fix-compose-directive-subset-imports.md
```diff
@@ -0,0 +1,29 @@
+---
+"@apollo/composition": patch
+---
+
+Fix `@composeDirective` failing with a false `DIRECTIVE_COMPOSITION_ERROR` when subgraphs import different subsets of directives from the same custom spec.
```
dariuszkuc:
```suggestion
---
"@apollo/composition": minor
---

Update `@composeDirective` merge logic to allow subgraphs to import different directives from the same spec. Previously, all subgraphs had to import ALL (exactly the same) custom directives (even if they did not use them) as otherwise composition would fail with a false `DIRECTIVE_COMPOSITION_ERROR`.
```

dariuszkuc (approved):

.changeset/fix-compose-directive-subset-imports.md
```diff
@@ -0,0 +1,29 @@
...
+```
+DIRECTIVE_COMPOSITION_ERROR: Core feature "https://myorg.dev/myspec" in subgraph "B"
+does not have a directive definition for "@foo"
+```
+
+The bug was order-dependent (failed when the subgraph elected as "latest" for the spec was not a superset of all other subgraphs' imports) and always failed for fully disjoint import sets (e.g. subgraph A imports only `@foo`, subgraph B imports only `@bar`).
+
+**Root cause:** `ComposeDirectiveManager.getLatestDirectiveDefinition()` resolved definitions through a two-hop chain — `directiveName → spec identity → "latest" subgraph for that spec → schema.directive(...)`. The "latest" subgraph was determined by spec version alone, without checking whether it had actually imported the directive being resolved.
+
+**Fix:** a `directiveDefinitionMap` (`directiveName → DirectiveDefinition`) is now built during `validate()`, sourcing each entry from the highest-minor-version subgraph that actually imported that specific directive. `getLatestDirectiveDefinition()` does a direct lookup on this map instead of the indirect chain.
```
sachindshinde, line 29: Implementation details shouldn't be in a changelog message, changelogs are tailored for package users and they don't need to read about the internal details of code changes unless it's relevant. (If they're curious, they can follow the auto-generated PR link in the changelog.)

.changeset/fix-compose-directive-subset-imports.md
```diff
@@ -0,0 +1,29 @@
...
+@composeDirective(name: "@foo")
+```
+
+This would produce a false error:
+```
+DIRECTIVE_COMPOSITION_ERROR: Core feature "https://myorg.dev/myspec" in subgraph "B"
+does not have a directive definition for "@foo"
+```
+
+The bug was order-dependent (failed when the subgraph elected as "latest" for the spec was not a superset of all other subgraphs' imports) and always failed for fully disjoint import sets (e.g. subgraph A imports only `@foo`, subgraph B imports only `@bar`).
```
sachindshinde: Currently the change is being phrased as a feature improvement, in the sense that we're relaxing a validation. There's simultaneously bugs in that validation (the order-dependence you've mentioned), but the bugfix for that validation is a separate matter (if there's a need, we can file a patch for that).

.changeset/fix-compose-directive-subset-imports.md
```diff
@@ -0,0 +1,29 @@
+---
+"@apollo/composition": minor
+---
+
+Update `@composeDirective` merge logic to allow subgraphs to import different directives from the same spec. Previously, all subgraphs had to import ALL (exactly the same) custom directives (even if they did not use them) as otherwise composition would fail with a false `DIRECTIVE_COMPOSITION_ERROR`.
```
sachindshinde, line 5: Note that this issue isn't really with federation `import`s per se; spec directives have a default name in a schema without imports, and the issue persists there too. This is really about definitions being absent in some subgraphs with the latest spec version. (This wording needs to be changed in the code comments as well.)

composition-js/src/__tests__/compose.composeDirective.test.ts
```diff
@@ -156,6 +156,115 @@ describe('composing custom core directives', () => {
     expectDirectiveOnElement(schema, 'User.a', 'foo', { name: 'a' });
   });
 
+  // FED-849: composition should succeed when subgraphs import different subsets
```
sachindshinde, line 159: Generally, we prefer to keep private ticket numbers and their discussion outside the codebase, as this is a source-available project and it makes it harder for outside readers/contributors to understand. If there's relevant information from a ticket that's suitable for public view, it's recommended to copy/condense it to a comment, or place it in PR discussion (if it ends up being relevant in the future, it can be found via `git blame`).

composition-js/src/composeDirectiveManager.ts
```diff
@@ -455,33 +488,11 @@ export class ComposeDirectiveManager {
...
-            {
-              nodes: feature.directive.sourceAST,
-            },
-          ));
-        }
-        return directive;
-      }
-    }
-    return undefined;
+    // Resolve directly from the per-directive map that was populated in validate().
```
sachindshinde: This should be a doc comment on the method/function instead a code comment. (We also don't really need to mention that `directiveDefinitionMap` was computed earlier, it's self-evident from the code.)

composition-js/src/__tests__/compose.composeDirective.test.ts
```diff
@@ -156,6 +156,115 @@ describe('composing custom core directives', () => {
...
+  // of directives from the same custom spec (e.g. SG1 imports [@foo, @bar], SG2 imports [@foo]).
+  it('two subgraphs with same spec but disjoint directive imports (FED-849)', () => {
```
sachindshinde, line 161: The directive sets aren't disjoint in this test (they are in the next one though).

sachindshinde (commented):
There's some breaking changes/bugs in the PR. The problem pointed out is legitimate, but I think we need to change the approach used.

To be specific, the problem today is that multiple subgraphs may contain the definition for a spec directive, but when looking for those definitions, the `@composeDirective` code only chooses one of those subgraphs with the latest version. If that one subgraph doesn't contain all the composed directive definitions, then composition fails. Now the rules today would ask that those definitions be defined in all such subgraphs, and that technically subverts the issue, but it's onerous. So we'd like to relax the validation.

The approach in this PR does more than relaxing it though. This PR's code:
- Changes directive definition finding to look at subgraphs with _all_ versions of the spec, not just the latest one. We don't want to do this, because it means the supergraph schema may e.g. have the user's spec at `v1.1`, but use the directive definition from `v1.0` of the user's spec, and the user's downstream code that cares about the directive (e.g. a plugin or coprocessor) may need changes to deal with these outdated directive definitions.
- Changes directive definition finding to only consider a directive definition from a subgraph when that subgraph contains a `@composeDirective` for that directive. This was not the case before this PR; the code would search all definitions from all subgraphs that had the desired spec version. Technically a definition should exist in the subgraph that declared the `@composeDirective`, but that definition may be from an older spec version, leading to the problem mentioned above.

I think the right approach here to the validation relaxation is to consider all definitions from all subgraphs with the desired spec version, not just picking an arbitrary subgraph with the desired spec version. I've filed a PR for this and updated the changeset message/tests in this PR https://github.com/apollographql/federation/pull/3422.

There's also a few general review comments below.

PR #3392, opened 2026-03-02
fix: relax @interfaceObject validation for Fed 1 subgraphs

apollo-mateuswgoettems:
### What this PR does

Previously, any use of `@interfaceObject` in a Fed 2 subgraph would cause an
`INTERFACE_OBJECT_USAGE_ERROR` if **any** Fed 1 subgraph was present in the
composition — regardless of whether the types actually conflicted.

This was overly restrictive. The real problem is narrower: `@interfaceObject`
on type `T` in a Fed 2 subgraph requires other subgraphs to be able to resolve
`__typename` via an interface `@key` on `T`. Fed 1 subgraphs silently drop
`@key` from interfaces during upgrade (the schema upgrader removes them), so
they **cannot** fulfill that requirement for `T` — but they are perfectly fine
for any other types.

This PR replaces the blanket check with a **per-type** cross-check:

- An error is only raised when a Fed 2 subgraph uses `@interfaceObject` on
  type `T` **and** a Fed 1 subgraph has `@key` on an interface also named `T`.
- Fed 2 subgraphs using `@interfaceObject` on a type that has no corresponding
  interface `@key` in any Fed 1 subgraph compose successfully.

### Changes

**`internals-js/src/schemaUpgrader.ts`**
- `SchemaUpgrader`: added `interfaceKeyTypes: Set<string>` field to track
  interface types whose `@key` was removed during upgrade; the set is returned
  from `upgrade()` on success.
- `upgradeSubgraphsIfNecessary`: replaced `subgraphsUsingInterfaceObject` with
  two `SetMultiMap<string, string>` instances (`fed2InterfaceObjectTypesToSubgraphs`
  and `fed1InterfaceKeyTypesToSubgraphs`) that are cross-checked per type name
  after all subgraphs are processed.

**`internals-js/src/__tests__/schemaUpgrader.test.ts`**
- Updated the existing `@interfaceObject` rejection test with the new, more
  precise error message.
- Added a new test confirming that composition succeeds when the same type name
  does **not** have a `@key` on a Fed 1 interface.

### DIFF
--- a/internals-js/src/__tests__/schemaUpgrader.test.ts
+++ b/internals-js/src/__tests__/schemaUpgrader.test.ts
@@ -235,12 +235,11 @@ test('remove tag on external field if found on definition', () => {
   ).toStrictEqual(['@tag(name: "a tag")']);
 });
 
-test('reject @interfaceObject usage if not all subgraphs are fed2', () => {
-  // Note that this test both validates the rejection of fed1 subgraph when @interfaceObject is used somewhere, but also
-  // illustrate why we do so: fed1 schema can use @key on interface for backward compatibility, but it is ignored and
-  // the schema upgrader removes them. Given that actual support for @key on interfaces is necesarry to make @interfaceObject
-  // work, it would be really confusing to not reject the example below right away, since it "looks" like it the @key on
-  // the interface in the 2nd subgraph should work, but it actually won't.
+test('reject @interfaceObject usage when a fed1 subgraph has @key on the same interface type', () => {
+  // This test validates that when a fed2 subgraph uses @interfaceObject on a type, and a fed1 subgraph
+  // has @key on an interface of the same name, we produce an error. Fed1 subgraphs ignore @key on
+  // interfaces entirely (the schema upgrader removes them), so they cannot resolve type names via an
+  // interface @key, which is required for @interfaceObject to work correctly.
 
   const s1 = `
     extend schema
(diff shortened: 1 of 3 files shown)


internals-js/src/__tests__/schemaUpgrader.test.ts
```diff
@@ -235,12 +235,11 @@ test('remove tag on external field if found on definition', () => {
...
 
-test('reject @interfaceObject usage if not all subgraphs are fed2', () => {
-  // Note that this test both validates the rejection of fed1 subgraph when @interfaceObject is used somewhere, but also
-  // illustrate why we do so: fed1 schema can use @key on interface for backward compatibility, but it is ignored and
-  // the schema upgrader removes them. Given that actual support for @key on interfaces is necesarry to make @interfaceObject
-  // work, it would be really confusing to not reject the example below right away, since it "looks" like it the @key on
-  // the interface in the 2nd subgraph should work, but it actually won't.
+test('reject @interfaceObject usage when a fed1 subgraph has @key on the same interface type', () => {
+  // This test validates that when a fed2 subgraph uses @interfaceObject on a type, and a fed1 subgraph
+  // has @key on an interface of the same name, we produce an error. Fed1 subgraphs ignore @key on
```
sachindshinde: > Fed1 subgraphs ignore @key on interfaces entirely (the schema upgrader removes them), so they cannot resolve type names via an interface @key, which is required for @interfaceObject to work correctly.

Similarly, I would rephrase this to

> @key on an interface in a Fed 1 subgraph does not indicate it can resolve type names for that interface via `_entities`, which is required for @interfaceObject to work correctly.

sachindshinde (changes requested):

sachindshinde (approved):

PR #3416, opened 2026-03-18
chore: Use node 24 and npm trusted publishing, and update cspell (backport of #3411 for 2.10)

sachindshinde:
This PR is a backport of #3411 for the `version-2.10` branch. Specifically, this PR:
- Adds an ignore rule to cspell for GHSA IDs.
- Backports the node orb version bump and lint CircleCI jobs from `main` (previously using `node/run`).
- Bumps `actions/checkout` to v4 and adds OIDC/changesets permissions in the release GitHub action.
- Bumps node to v24 for the repo (which is bundled with npm v11), updating the lint CircleCI jobs and release GitHub action as well.
  - This is because npm trusted publishing requires at least npm v11.5.1.
- Bumps the npm constraint in `engines` in `package.json` to `<12` (to allow running npm v11).
- Runs the test CircleCI jobs for node v22, v24, and latest.
  - The job was previously installing npm v9 because node v14's bundled npm version was too low to support the `package-lock.json`, but it was doing this for all node versions. We've now changed this to only happen for node v14.
- Stops passing `NPM_TOKEN` to `changesets/action` in the release GitHub action.
  - Note that npm versions that support trusted publishing will [ignore locally configured tokens when OIDC environment variables are present](https://github.com/npm/cli/blob/8afa3bd21461c0984caf1bcc2e486c4881bda516/lib/utils/oidc.js#L141) (which they are for GH actions, since we've set that up), so `NPM_TOKEN` would be ignored anyway. But the main reason for this change is to stop `changesets/action` from printing a misleading log message saying `NPM_TOKEN` has been placed into `.npmrc` (which, while true, will be ignored by `npm publish`).
  - Note that we still need to keep the `NPM_TOKEN` GitHub secret around to run any tag-changing scripts, as `npm dist-tag` sadly doesn't support OIDC yet (see https://github.com/npm/cli/issues/8547 for discussion). This means we'll need to manually rotate the token every 90 days.
- Uses `npm ci` instead of `npm i` in the release GitHub action.
  - As a requirement for releasing (or filing a release PR), the `package-lock.json` should be aligned with the `package.json`.
- Remove the `federation-rs` step in the release GitHub action, as this is no longer necessary in the release process. 
  - We also stop setting `FEDERATION_VERSION`, since it's no longer used.
- Sets the tag for npm publishes to the branch name for `version-*` branches.
  - Previously, it was the default of `latest`, which was causing backport releases to mistakenly change the `latest` tag (used when someone `npm i`s without a version number).
  - We also tightened the `version-*` branch pattern down to `version-[0-9]+.[0-9]+` (since it's getting passed around in bash now).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
-  node: circleci/node@5.1.0
+  node: circleci/node@7.2.0
 
 jobs:
   # Unfortunately cimg/node doesn't tag its images with major only, you have to specify a minor version.
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,7 +20,11 @@ jobs:
           node-version: << parameters.node-version >>
       # node v14 defaults to npm 6, which is too old for our package-lock.json
       # should be able to remove this step when we drop node v14
-      - run: npm install -g npm@9
+      - when:
+          condition:
+            equal: [ "14", << parameters.node-version >> ]
+          steps:
+            - run: npm install -g npm@9
       - node/install-packages
       - run:
           name: Run tests
@@ -61,6 +65,9 @@ workflows:
                 - "16"
                 - "18"
                 - "20"
+                - "22"
+                - "24"
+                - "latest"
       - Lint:
           name: Lint - << matrix.script >>
           matrix:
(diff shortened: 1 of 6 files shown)


duckki (approved):
Looks good!

PR #3420, opened 2026-03-19
Merge `main` into `next`

sachindshinde:
Merge `main` into `next`


sachindshinde:
There's only one real commit on `next` (#3289) and a bunch of merge commits. To simplify history, we're just going to rebase that PR's commit on `main` and push that as `next`.

PR #3422, opened 2026-03-26
fix(composition): Address post-merge feedback for #3391

sachindshinde:
This PR:
- Address post-merge feedback given for [#3391](https://github.com/apollographql/federation/pull/3391) (specifically, [this review](https://github.com/apollographql/federation/pull/3391#pullrequestreview-4002938809)).
- Fixes an unrelated bug in `sourceFeature()`, where it would return the wrong name-in-spec sometimes.
  - Note this bug only occurs when both a `@composeDirective` spec (or the connect spec) is aliased and a directive with the same name as the spec is used (which should be rare).


apollo-mateuswgoettems (approved):

dariuszkuc (approved):
looks good

question: do we want to allow scenarios where older specs define more directives than new one? this would make it easier to evolve* but at the same time limiting to latest spec version is safer as it ensures that all those custom directives are still valid

*if they do need to bump the spec folks will have to import all of them in a single subgraph until migration is over (so following current behavior)

sachindshinde:
@dariuszkuc 
When folks have to upgrade, they'll indeed need to have the first subgraph that upgrades define all the spec directives in-use by the supergraph if they don't want to break composition. That's not quite the current behavior though, since current composition arbitrarily picks a subgraph (so e.g. the second subgraph to upgrade could cause an error today, whereas after these PR changes it would not). After the upgrade is done, subgraphs can shift back to just defining what they use. I'm not sure if there's a nicer way to do it tbh; we can't really place older directive definitions into the supergraph schema since it may break downstream user logic.

PR #3423, opened 2026-04-01
Support requestless connectors

andrewmcgivery:
…s connectors

### DIFF
--- a/internals-js/src/specs/connectSpec.ts
+++ b/internals-js/src/specs/connectSpec.ts
@@ -396,13 +396,6 @@ export const CONNECT_VERSIONS = new FeatureDefinitions<ConnectSpecDefinition>(
       new FeatureVersion(2, 13),
     ),
     { preview: true },
-  )
-  .add(
-    new ConnectSpecDefinition(
-      new FeatureVersion(0, 5),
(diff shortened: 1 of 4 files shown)


internals-js/src/specs/connectSpec.ts
```diff
@@ -249,7 +249,7 @@ export class ConnectSpecDefinition extends FeatureDefinition {
       directive @connect(
         source: String
         id: String
-        http: ConnectHTTP!
+        http: ConnectHTTP
```
dariuszkuc, line 252: this is changing the arg type so probably should bump the spec? 0.4 is in preview so maybe it is fine?

dariuszkuc:
This PR should be targeting `next` instead of `main` (we'll also need same change in the Rust federation crate).

benjamn (approved):
Approved pending switching the target branch to `next`, as @dariuszkuc suggested.

dariuszkuc:
Do we need a changeset?

internals-js/src/specs/connectSpec.ts
benjamn: Yes, I believe we are ok with changes like this during the v0.4 preview. Working code will continue to work.

PR #3424, opened 2026-04-08
fix: printHumanReadableList now correctly checks for max length

dariuszkuc:
Fixed reduce logic when calculating the max number of elements to include in the message. Since there is no short-circuit logic in reduce, we need to pass the length that exceeded the cutoff for subsequent evaluations (as otherwise check may fail on a subgraph with a very long name BUT succeed on a subsequent subgraph name that is shorter and as a result we would include too many items in the list).

### DIFF
--- a/internals-js/src/__tests__/utils.test.ts
+++ b/internals-js/src/__tests__/utils.test.ts
@@ -13,7 +13,7 @@ it('subgraph names are truncated correctly', () => {
   expect(printedNames).toEqual(
     'subgraphs "some-graphql-service", "another-graphql-service", "third-graphql-service-name", ...',
   );
-})
+});
 
 describe('OrderedMap', () => {
   it('updating value works', () => {
--- a/internals-js/src/__tests__/utils.test.ts
+++ b/internals-js/src/__tests__/utils.test.ts
@@ -1,4 +1,19 @@
 import { OrderedMap } from '../utils';
+import { printSubgraphNames } from '../federation';
+
+it('subgraph names are truncated correctly', () => {
+  const printedNames = printSubgraphNames([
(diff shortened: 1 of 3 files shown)


duckki:
🤔 Arguably, JS logic is not a bad design choice, since we can fit more names within the limit by picking shorter names over longer names.

dariuszkuc:
We don't pick shorter names at the end - if we have a long name that doesn't fit followed by a shorter name that fits within the limit we increment the last index. This means we incorrectly pick up the additional long name that makes the string longer than the limit.

internals-js/src/__tests__/utils.test.ts
```diff
@@ -1,4 +1,19 @@
...
+  const printedNames = printSubgraphNames([
+    'some-graphql-service',
+    'another-graphql-service',
+    'third-graphql-service-name',
+    // total length 112 which is greater than 100 limit
+    'some-very-long-graphql-service-name',
+    'short-service-name',
+  ]);
+  expect(printedNames).toEqual(
+    'subgraphs "some-graphql-service", "another-graphql-service", "third-graphql-service-name", ...',
```
dariuszkuc, line 14: prior to this fix the behavior would be
- check max length if including `some-very-long-graphql-service-name` -> this would exceed the max length so we would pass current length and continue (as reduce does not have short-circuit mechanism)
- check max length if including next element `short-service-name` -> new length would be within the limit so we would increase the index of subgraph names to include
- final list of subgraphs would include the "very-long" name that exceeds the length limit

duckki (approved):
I see we don't pick a subset of names to fit the space. We just pick the range of `[0, x]`.
The fix makes sense.

PR #3430, opened 2026-04-27
fix(composition): Add missing `@link` validations

sachindshinde:
This PR adds `@link` validations to detect invalid alias names and alias/import conflicts. To maintain backwards compatibility, we do allow some exceptions, and we also auto-generate supergraph spec aliases when conflicts are detected.


composition-js/src/merging/merge.ts
```diff
@@ -887,6 +910,50 @@ class Merger {
...
+    // alias. The gist is we use a trie to determine a GraphQL name that isn't a
+    // prefix of any existing aliases (this prefix also doesn't use "_" outside
+    // the first character, so it should be safe for aliases). We then add an
+    // incrementing index to it, to account for core features that have the same
+    // spec name. Finally, we need to add some separator to avoid the case where
+    // the spec's name starts with a number and blends into the index (note spec
+    // names don't have to be valid GraphQL names, just name-in-schemas do). The
+    // separator also can't end in "_", since the spec name may begin with "_".
+    // We've arbitrarily chosen the separator "_composed" here.
+    const computePrefix = (): string => {
```
dariuszkuc: If the intent is to ensure we have a unique alias prefix - why do all this work instead of just aliasing the conflicts with `composed<increasing_number>_` prefix (or something along those lines (i.e. `@composed1_foo` for `@foo`). Since this is very rare I'd keep the logic simple and don't even check for `_` (so we could end up with `@composed2__bar` for `@_bar`). It's just optics anyway

sachindshinde: The prefix most of the time should be `_0`, `_1`, etc., which seems short and readable. The trie BFS algorithm itself isn't too bad (less than 50 lines, seems an LLM can one-shot it when I try it). IMO we should also get used to trie BFSs to compute unique prefixes. We have a few different places in code that just assume a given prefix is unique, and it may lead to undesirable behavior at runtime.

composition-js/src/merging/merge.ts
```diff
@@ -367,6 +369,18 @@ interface MergedDirectiveInfo {
   staticArgumentTransform?: StaticArgumentsTransform;
 }
 
+type SupergraphInfoByIdentity = Map<
```
dariuszkuc, line 372: I know this is just extracted out of `validateAndMaybeAddSpecs` but I'm just wondering whether we should be capturing the types as well? e.g. our `@requiresScopes` and `@policy` supergraph specs do define custom scalars

While we only allow folks to `@composeDirective`s they could potentially define the same overlapping scalars?

dariuszkuc (approved):

composition-js/src/merging/merge.ts
sachindshinde: > whether we should be capturing the types as well

`validateAndMaybeAddSpecs()` is for Apollo specs, and for them, subgraph spec type definitions are validated via `checkOrAdd()` but then basically ignored (we don't enforce that their names be the same across subgraphs like we do for Apollo spec directives that compose). The spec type names in the supergraph schema are usually the default namespaced names found in the `FeatureDefinition` for the spec (the type/directive definitions also come from the `FeatureDefinition` ).

> While we only allow folks to `@composeDirective`s they could potentially define the same overlapping scalars?

The code for `@composeDirective` ignores user-defined spec types (the supergraph `@link` won't contain imports for them, but the type names aren't changed if they're namespaced either). Those types are treated like any other subgraph type, merged and placed into the supergraph schema (which may or may not compose successfully, depending on any differences). `@composeDirective` also doesn't do any significant validation of the user-defined directive definitions of composed directives, so they could e.g. reference non-built-in types outside their spec. That could be user-defined types (which get merged into the supergraph schema), though it could also be types defined in other specs, which could cause a composition error since they may not appear in the supergraph schema (e.g. the `FieldSet` or `_Any` scalar) or they may appear under a different name (e.g. Apollo supergraph spec types).

PR #3439, opened 2026-05-26
Pass scope version as a separate argument in telemetry instantiation

### LINKED ISSUE #3436: gateway-js: pass instrumentation scope version to getMeter and getTracer
## Problem

In [`gateway-js/src/utilities/opentelemetry.ts`](https://github.com/apollographql/federation/blob/main/gateway-js/src/utilities/opentelemetry.ts), the instrumentation scope version is not correctly set on either the meter or the tracer:

**Meter** (line 191) — version is omitted entirely:
```ts
const meter = meterProvider.getMeter("apollo/gateway-js")
```

**Tracer** (line 76) — version is concatenated into the name instead of passed as the second argument:
```ts
export const tracer = opentelemetry.trace.getTracer(\`\${name}/\${version}\`);
```

## Expected behavior

The [OTel JS API](https://open-telemetry.github.io/opentelemetry-js/interfaces/_opentelemetry_api.TracerProvider.html#getTracer) accepts `name` and `version` as separate arguments:

```ts
getMeter(name: string, version?: string, options?: MeterOptions): Meter
getTracer(name: string, version?: string, options?: TracerOptions): Tracer
```

Both calls should be updated to pass the version separately:

```ts
const meter = meterProvider.getMeter("apollo/gateway-js", version)
export const tracer = opentelemetry.trace.getTracer(name, version)
```

## Impact

Without a version on the meter, consumers (including Apollo's telemetry ingress) cannot distinguish metrics by gateway-js version. The tracer scope name currently looks like `apollo/gateway/2.x.y` rather than the expected `apollo/gateway`, which if we at some point wanted to use these traces in Apollo insights, it would need additional parsing.

SharkBaitDLS:
Fixes #3436.

### DIFF
--- a/gateway-js/src/utilities/opentelemetry.ts
+++ b/gateway-js/src/utilities/opentelemetry.ts
@@ -73,7 +73,7 @@ export enum OpenTelemetryAttributeNames {
 }
 
 const { name, version } = require('../../package.json');
-export const tracer = opentelemetry.trace.getTracer(`${name}/${version}`);
+export const tracer = opentelemetry.trace.getTracer(name, version);
 
 const APOLLO_OTEL_ENDPOINT = "https://usage-reporting.api.apollographql.com";
 const EXPORT_INTERVAL_SECONDS = 60 * 60; // one hour
@@ -188,7 +188,7 @@ export function createDataCollectionMeterProvider(metricExporter: PushMetricExpo
   const osAttrs = osDetectorSync.detect().attributes
   const osType = osAttrs[ATTR_OS_TYPE]
   const hostArch = hostAttrs[ATTR_HOST_ARCH]
-  const meter = meterProvider.getMeter("apollo/gateway-js")
+  const meter = meterProvider.getMeter(name, version)
 
   const uptimeGauge = meter.createObservableGauge(METRIC_PROCESS_UPTIME, {
     "description": "The uptime of the JS federation gateway running"
--- a/.changeset/fix-gateway-otel-scope-version.md
+++ b/.changeset/fix-gateway-otel-scope-version.md
@@ -1,0 +1,5 @@
+---
+"@apollo/gateway": patch
+---
+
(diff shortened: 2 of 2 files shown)


gateway-js/src/utilities/opentelemetry.ts
```diff
@@ -188,7 +188,7 @@ export function createDataCollectionMeterProvider(metricExporter: PushMetricExpo
   const osAttrs = osDetectorSync.detect().attributes
   const osType = osAttrs[ATTR_OS_TYPE]
   const hostArch = hostAttrs[ATTR_HOST_ARCH]
-  const meter = meterProvider.getMeter("apollo/gateway-js")
+  const meter = meterProvider.getMeter(name, version)
```
SharkBaitDLS, line 191: This will change the name back to `@apollo/gateway` instead of `apollo/gateway-js`. @timbotnik will that require changes on the ingestion side?

dariuszkuc: instead of `name` shouldn't you pass `apollo/gateway-js` (here and at line 76)?

SharkBaitDLS: Yeah I'm not sure based on the original issue Tim wrote up whether the desire was to have this name actually change (he mentions the desired scope is `@apollo/gateway`), or just to swap out the APIs.

timbotnik: From what I can tell based on our existing conventions we should use what was here before `apollo/gateway-js` 👍
- https://expert-giggle-8wv59kk.pages.github.io/schema/registry/constellation/0.0.4/METADATA/
- https://expert-giggle-8wv59kk.pages.github.io/schema/registry/operator/1.2.0/METADATA/

The router uses `apollo-router` but it's the only outlier 🙄

timbotnik: I'm not sure where the `@` symbols crept in on that ticket but you can ignore.

dariuszkuc (approved):

PR #3445, opened 2026-05-28
fix(composition): pass subgraph name instead of field name to `isMergedDirective`

dariuszkuc:
`validateAndFilterExternal` was passing `source.name` (the field name) instead of `this.names[i]` (the subgraph name) to `isMergedDirective`. This caused `shouldComposeDirective` to miss `@composeDirective`-composed directives on external fields, since it looked up the directive under a nonexistent subgraph name.


duckki (approved):
👍

benjamn (approved):

PR #3450, opened 2026-06-22
Filter cloud resource detection to only the data we process

SharkBaitDLS:


gateway-js/src/utilities/opentelemetry.ts
```diff
@@ -151,10 +151,30 @@ export function recordExceptions(
...
+// Non-identifying cloud resource attributes that we process.
+const ALLOWED_CLOUD_RESOURCE_ATTRIBUTES: Set<string> = new Set([
+  ATTR_CLOUD_PROVIDER,
+  ATTR_CLOUD_PLATFORM,
+]);
+
+function filterToAllowedCloudAttributes(resource: Resource): Resource {
+  return new Resource(
+    {},
+    Promise.resolve(resource.waitForAsyncAttributes?.()).then(() => {
```
briannafugate408, line 163: Should we drop the optional chaining and require a minimum OTel SDK version that has `waitForAsyncAttributes` or add a fallback like:

`Promise.resolve(resource.waitForAsyncAttributes ? resource.waitForAsyncAttributes() : undefined)` and document the version requirement?

SharkBaitDLS: I'm not quite sure I follow, the optional chaining should be functionally equivalent to your suggested fallback. If you prefer the verbose solution for readability I can definitely change to that though.

I took this code straight from how OTel handles this case defensively:

https://github.com/open-telemetry/opentelemetry-js/blob/v1.30.1/packages/opentelemetry-resources/src/detect-resources.ts#L73

briannafugate408: Makes sense. If the method doesn't exist, there are no async attributes, so reading synchronously sounds safe. Thanks for the OTel source reference, that clears it up. LGTM!

briannafugate408 (approved):

PR #3454, opened 2026-07-14
docs fix: entity key field docs and differing keys

mabuyo:
We received user feedback that the example snippet was confusing, given the content in the docs section on [Differing keys](https://www.apollographql.com/docs/graphos/schema-design/federated-schemas/entities/define-keys.md#differing-keys-across-subgraphs). 

This fixes the schema example and edits the explanation on differing keys.

### DIFF
--- a/docs/source/schema-design/federated-schemas/entities/intro.mdx
+++ b/docs/source/schema-design/federated-schemas/entities/intro.mdx
@@ -18,7 +18,7 @@ In Apollo Federation, federated data objects are represented as _entities_. Enti
 
 Entities are defined in subgraph schemas. Each subgraph can contribute different fields to an entity it defines and is responsible for _resolving_ it—returning only the fields that it contributes. This enables subgraphs to adhere to the separation of concerns principle.
 
-An _entity type_ is an object type that has been [defined as an entity](#defining-an-entity). Because an entity is keyed, an entity type's definition must have a `@key` directive. For example, this `Product` entity's fields are defined and resolved across two subgraphs:
+An _entity type_ is an object type that has been [defined as an entity](#defining-an-entity). Because an entity is keyed, an entity type's definition must have a `@key` directive. For example, this `Product` entity's fields are defined and resolved across two subgraphs. Both subgraphs use `upc` as a `@key` field so the router can merge their fields into a single `Product` type:
 
 <CodeColumns>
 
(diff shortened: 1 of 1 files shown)


docs/source/schema-design/federated-schemas/entities/intro.mdx
```diff
@@ -31,8 +31,8 @@ type Product @key(fields: "upc") {
 ```
 
 ```graphql title="Reviews subgraph"
-type Product @key(fields: "productUpc") {
-  productUpc: ID!
```
mabuyo, line 34: In the original example, I get a composition error so the code snippet is definitely wrong

```
  SATISFIABILITY_ERROR: The following supergraph API query:
    {
      products {
        productUpc
      }
    }
    cannot be satisfied by the subgraphs because:
    - from subgraph "products":
      - cannot find field "Product.productUpc".
      - cannot move to subgraph "reviews" using @key(fields: "productUpc") of "Product", the key field(s) cannot be resolved from subgraph "products".
    SATISFIABILITY_ERROR: The following supergraph API query:
    {
      products {
        rating
      }
    }
    cannot be satisfied by the subgraphs because:
    - from subgraph "products":
      - cannot find field "Product.rating".
      - cannot move to subgraph "reviews" using @key(fields: "productUpc") of "Product", the key field(s) cannot be resolved from subgraph "products".
```

dariuszkuc: Indeed original example is invalid as there needs to be a way to jump between subgraphs.

* `@key` is used to indicate this object type is an entity
* `@key` defines exit points (or jump points) from a subgraph so given a key `upc` I can use `upc` to jump to other subgraph
  * if we specify `resolvable: false` then it means we cannot jump out and can only jump into this subgraph
* `@key` can differ between subgraphs as long as there are some overlapping key fields declared across subgraphs
  * as long as there is a matching field on the target type it doesn't have to be explicitly marked as `@key` to be used for jumping into the graph, i.e. subgraph A can declare `@key = upc` and subgraph B can declare `@key = productUpc` as long as there is a way to reach from A to B and/or B to A (thats the satisfiability check) - so in order to make this valid either subgraph A would need to define `productUpc` field (doesn't have to be marked as key) OR subgraph B needs to add `upc` field (doesn't need to be a key)

dariuszkuc (approved):
Its an intro to entities so yeah we dont need to introduce any of those more advanced concepts here.

docs/source/schema-design/federated-schemas/entities/intro.mdx
mabuyo: Thanks for the clarification! On that last point, [this other docs page](https://www.apollographql.com/docs/graphos/schema-design/federated-schemas/entities/define-keys) does explain it correctly.

PR #3349, opened 2025-11-13
Initial support for federation version 2.13 and connect/v0.4

benjamn:
This PR is important for releasing previews builds where `@link(url: "https://specs.apollo.dev/connect/v0.4")` can be enabled, similar to how we tested `connect/v0.3` before it was released.

I suspect this functionality will end up targeting federation version 2.13.0 (the next minor version), hence the `minor` version bump in the changeset. To support that, I also introduced `new FederationSpecDefinition(new FeatureVersion(2, 13))`, so we can begin releasing previews of v2.13.

PR #3455, opened 2026-07-17
Initial support for connect/v0.5

benjamn:
## Summary

Adds initial support for `connect/v0.5` so we can release preview builds where `@link(url: "https://specs.apollo.dev/connect/v0.5")` can be enabled, similar to how we tested `connect/v0.3` and `connect/v0.4` before they were released.

- Registers `connect/v0.5` in `CONNECT_VERSIONS` as a **preview** version (`{ preview: true }`), so it does not leak into the supergraph `@link` automatically — only subgraphs that explicitly `@link` to v0.5 get it (`latestNonPreview()` continues to drive automatic selection). `connect/v0.4` remains preview as well.
- Maps `connect/v0.5` to a minimum federation version of `2.14`. Unlike the `connect/v0.4` work (#3349), this needs **no new federation spec version** — `2.14` already exists.
- Adds `https://specs.apollo.dev/connect/v0.5` to `ROUTER_SUPPORTED_SUPERGRAPH_FEATURES`.

The `@connect`/`@source` directive surface for v0.5 is currently **identical to v0.4** — this PR just stakes out the version to unblock further development (both composition-side and router-side). Version-specific `@connect`/`@source` arguments will be added behind the v0.5 version gate (`version.gte(new FeatureVersion(0, 5))`) as they are finalized, the same pattern `federationSpec.ts` uses for per-version directive args.

> **Note:** this PR also merges the current `main` into `next` (merge-forward), then adds the v0.5 changes on top, per the request to do it in a single PR.

## Test plan

- [x] New test: "uses v0.5 in supergraph @link when a subgraph explicitly links to v0.5"
- [x] Strengthened the neighboring v0.3/v0.4 selection tests to assert `connect/v0.5` does not leak
- [x] `npm run compile` + full `connectors.test.ts` suite passes (14 tests)

A changeset will be added in a follow-up commit so it can link back to this PR.

### DIFF
--- a/.changeset/connect-v05-initial.md
+++ b/.changeset/connect-v05-initial.md
@@ -1,0 +1,7 @@
+---
+"@apollo/composition": minor
+"@apollo/federation-internals": minor
+"@apollo/query-graphs": minor
+---
+
+Add initial support for `connect/v0.5` (preview) to unblock further development ([#3455](https://github.com/apollographql/federation/pull/3455)). The directive surface is currently identical to `connect/v0.4`; version-specific `@connect`/`@source` arguments will be added behind the v0.5 version gate as they are finalized.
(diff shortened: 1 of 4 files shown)


briannafugate408 (approved):

dariuszkuc:
This needs to be re-implemented in the Rust version

PR #3456, opened 2026-07-17
fix(composition): propagate directives from @interfaceObject fields to @external implementations

dariuszkuc:
When an implementation re-declares a field as `@external` (e.g. to reference it in `@requires`), the field's only resolvable definition lives on the abstracting `@interfaceObject`. Directives like `@tag` applied there were not being propagated to the implementation's copy in the supergraph.

During `add_interface_object_fields`, detect implementation fields where every `@join__field` is `external: true` and the field is provided by an `@interfaceObject`, then copy applicable directives onto the implementation field.

[FED-1074]: https://apollographql.atlassian.net/browse/FED-1074?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/composition-js/src/merging/merge.ts
+++ b/composition-js/src/merging/merge.ts
@@ -1377,10 +1377,23 @@ class Merger {
     // For each merged object types, we check if we're missing a field from one of the implemented interface.
     // If we do, then we look if one of the subgraph provides that field as a (non-external) interface object
     // type, and if that's the case, we add the field to the object.
+    const joinFieldDirective = this.joinSpec.fieldDirective(this.merged);
     for (const type of this.merged.objectTypes()) {
       for (const implementedItf of type.interfaces()) {
         for (const itfField of implementedItf.fields()) {
-          if (type.field(itfField.name)) {
+          const existingField = type.field(itfField.name);
+          if (existingField) {
+            // An implementation may locally declare a field as `@external` purely to reference it in
+            // a `@requires`, in which case it already exists on the object so it won't be propagated
+            // from the interface object. But that field's only "real" definition is the abstracting
+            // `@interfaceObject` field, so directives that only apply there (e.g. `@tag`) still need
+            // to be propagated onto it.
+            const joinFieldApplications = existingField.appliedDirectivesOf(joinFieldDirective);
+            const isFullyExternal = joinFieldApplications.length > 0
+              && joinFieldApplications.every((d) => d.arguments().external === true);
+            if (isFullyExternal && this.isFieldProvidedByAnInterfaceObject(itfField.name, implementedItf.name)) {
+              this.copyNonJoinAppliedDirectives(itfField, existingField);
+            }
             continue;
           }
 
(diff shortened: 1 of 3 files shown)


composition-js/src/merging/merge.ts
```diff
@@ -1980,12 +1993,14 @@ class Merger {
     // Note that we don't truly merge externals: we don't want, for instance, a field that is non-nullable everywhere to appear nullable in the
     // supergraph just because someone fat-fingered the type in an external definition. But after merging the non-external definitions, we
     // validate the external ones are consistent.
-    this.mergeDescription(withoutExternal, dest);
-    this.recordFieldAppliedDirectivesToMerge(withoutExternal, dest);
-    this.addArgumentsShallow(withoutExternal, dest);
-    for (const destArg of dest.arguments()) {
-      const subgraphArgs = mapSources(withoutExternal, f => f?.argument(destArg.name));
-      this.mergeArgument(subgraphArgs, destArg);
+    if (someSources(withoutExternal, isDefined)) {
```
dariuszkuc, line 1996: We should only merge below IF we have non external sources. Without this guard we end up merging access control directives twice -> once during normal merge field logic and then again when propagating `@interfaceObject` field directives.

tninesling (approved):

composition-js/src/merging/merge.ts
```diff
@@ -1377,10 +1377,23 @@ class Merger {
...
+            // An implementation may locally declare a field as `@external` purely to reference it in
+            // a `@requires`, in which case it already exists on the object so it won't be propagated
+            // from the interface object. But that field's only "real" definition is the abstracting
+            // `@interfaceObject` field, so directives that only apply there (e.g. `@tag`) still need
+            // to be propagated onto it.
+            const joinFieldApplications = existingField.appliedDirectivesOf(joinFieldDirective);
+            const isFullyExternal = joinFieldApplications.length > 0
+              && joinFieldApplications.every((d) => d.arguments().external === true);
+            if (isFullyExternal && this.isFieldProvidedByAnInterfaceObject(itfField.name, implementedItf.name)) {
+              this.copyNonJoinAppliedDirectives(itfField, existingField);
```
sachindshinde, line 1395: Note there's a bug here, in that `copyNonJoinAppliedDirectives()` assumes it's only called once per destination element in the supergraph schema. (The code below calls it once, because it ultimately adds the missing implementing field once, and then later iterations will skip that field since it will exist afterward.) If you try to call `copyNonJoinAppliedDirectives()` more than once, you're effectively skipping merge logic that should have ran. For non-repeatable federation directives, they'll be applied multiple times, resulting in a composition error (e.g. similar to the issue you noticed below where `recordFieldAppliedDirectivesToMerge()` will apply access control directives during `mergeField()`, but so will `copyNonJoinAppliedDirectives()`).

This can happen if you have interface objects of different names that provide the same object field (somewhat related to [this issue](https://github.com/apollographql/router/pull/9386#issuecomment-4443845039)). Since `recordFieldAppliedDirectivesToMerge()` will cause merge behavior for access control directives while `copyNonJoinAppliedDirectives()` does not, this also means the changes made in this PR to `mergeField()` will cause a regression where composition will fail when it previously succeeded, shown in [this branch](https://github.com/apollographql/federation/tree/sachin/interface-object-access-control-regression).

PR #3459, opened 2026-07-21
fix(composition): propagate directives from @interfaceObject fields to @external implementations (backport 2.12)

dariuszkuc:
When an implementation re-declares a field as `@external` (e.g. to reference it in `@requires`), the field's only resolvable definition lives on the abstracting `@interfaceObject`. Directives like `@tag` applied there were not being propagated to the implementation's copy in the supergraph.

During `add_interface_object_fields`, detect implementation fields where every `@join__field` is `external: true` and the field is provided by an `@interfaceObject`, then copy applicable directives onto the implementation field.

[FED-1074]: https://apollographql.atlassian.net/browse/FED-1074?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/giant-teachers-divide.md
+++ b/.changeset/giant-teachers-divide.md
@@ -1,0 +1,12 @@
+---
+"@apollo/composition": patch
+---
+
+Propagate directives from @interfaceObject fields to @external implementations
+
+When an implementation re-declares a field as `@external` (e.g. to reference it in `@requires`), the field's only 
+resolvable definition lives on the abstracting `@interfaceObject`. Directives like `@tag` applied there were not 
+being propagated to the implementation's copy in the supergraph.
+
+During `add_interface_object_fields`, detect implementation fields where every `@join__field` is `external: true` and 
+the field is provided by an `@interfaceObject`, then copy applicable directives onto the implementation field.
--- a/composition-js/src/__tests__/compose.test.ts
+++ b/composition-js/src/__tests__/compose.test.ts
@@ -4214,6 +4214,53 @@ describe('composition', () => {
       result = composeAsFed2Subgraphs([subgraph2, subgraph1]);
       assertCompositionSuccess(result);
     })
+
+    it('propagates @tag from @interfaceObject field to locally @external implementation field', () => {
+      const subgraphA = {
+        typeDefs: gql`
+          type Query {
+            product: Product
+          }
+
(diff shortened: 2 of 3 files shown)


briannafugate408:
A few questions from review:

**1. `mergeField` guard interaction with access-control propagation**

The `someSources(withoutExternal, isDefined)` guard changes behavior beyond the `@tag` case: `recordFieldAppliedDirectivesToMerge` also does access-control directive propagation (`@requiresScopes`/`@authenticated`/`@policy`) via `accessControlAdditionalSources`, independent of `withoutExternal`. Before this change that ran unconditionally for these "fully-external, real-def-on-interfaceObject" fields; now it's skipped, and `copyNonJoinAppliedDirectives` (which has its own access-control block) becomes the sole path for them.

Was that swap intentional — to avoid double-applying access-control directives on these fields now that `copyNonJoinAppliedDirectives` also handles them? If so, could we get a test confirming an access-control directive on the `@interfaceObject` field still ends up on the `@external` implementation field exactly once?

**2. Multiple interfaces declaring the same field name**

If `type` implements two interfaces that both declare a field with this name and both qualify (fully-external + interfaceObject-provided), `copyNonJoinAppliedDirectives` runs once per interface against the same `existingField`. Since `@tag` is repeatable, could that append a duplicate `@tag` application? Might be worth a test or a guard to copy at most once per field.

**3. Negative-case coverage**

Is there a test for the case where only *some* `@join__field` applications on the implementation field are `external: true` (i.e., a genuinely shared/non-external-everywhere field)? Want to confirm `isFullyExternal` doesn't fire there.

dcwalter (approved):

dariuszkuc:
> 1. mergeField guard interaction with access-control propagation

Yes this was intentional.

See https://github.com/apollographql/federation/pull/3456#discussion_r3605342518

> 2. Multiple interfaces declaring the same field name

This is a limitation of a GraphQL spec as it does not specify any inheritance rules. If an interface defines some directives they have to be re-declared on the implementing type. 

We pick the first encountered interface object field. Our merge order is deterministic given inputs but if the inputs change (i.e. subgraphs get renamed) we could end up with different merge order. This means that it is possible to end up with different directive information depending on which interface object gets picked.

